PR #5, opened 2018-10-30
updating with external github links

triskadecaepyon:
The matching ghpages fixes for PR #3


triskadecaepyon:
This seems to be an effect of the Sphinx versions and odd environment behaviors.  Will retry with a new env and a docker container of sphinx and related dependencies.

triskadecaepyon:
Retrying with another method.  Will open a new PR.

Issue #6, opened 2018-10-31
Native pandas dataframe handling


triskadecaepyon:
Most algorithms in daal4py take numpy arrays, but this is an extra step that most data scientists prefer not to have to do.

It would be easier to be able to put in a direct Dataframe instead of a NumPy array into the models, as we normally just call '.values' anyways.

It would go from this:
```
target = np.array(dataset.columnname.values, ndmin=2).T
d4p_lm = d4p.linear_regression_training(interceptFlag=True)
lm_trained = d4p_lm.compute(feature_dataset.values, target)
```

to this:
```
d4p_lm = d4p.linear_regression_training(interceptFlag=True)
lm_trained = d4p_lm.compute(feature_dataset, target_df_or_series)
```

Note that this would 1) simplify the calls if using a single Y vector to not have the above numpy array cast, and 2) would allow us to keep internal information (such as column names and dropped feature frames) within the model.  Both of these would be good additions to daal4py.

fschlimb:
Down with 29f5c75c446466cb79bf774ee35b704b32c062ef

PR #21, opened 2018-12-04
Tc/pandas

fschlimb:
This provides direct support pandas DataFrames. DataFrames can now be used as input, similar to numpy arrays and file-names.

@oleksandr-pavlyk 
@triskadecaepyon I made a few edits to the docs so that DataFrames are mentioned whenever numpy arrays are
@Alexander-Makaryev I changed the test and example so that we test with numpy arrays and pandas data-frames

### DIFF
--- a/src/daal4py.cpp
+++ b/src/daal4py.cpp
@@ -28,34 +28,12 @@
 // ************************************************************************************
 
 #define is_array(a)            ((a) && PyArray_Check(a))
-#define array_is_behaved(a)    (PyArray_ISCARRAY_RO((PyArrayObject*)a))
+#define array_type(a)          PyArray_TYPE((PyArrayObject*)a)
+#define array_is_behaved(a)    (PyArray_ISCARRAY_RO((PyArrayObject*)a) && array_type(a)<NPY_OBJECT)
 #define array_is_native(a)     (PyArray_ISNOTSWAPPED((PyArrayObject*)a))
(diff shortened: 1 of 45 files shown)


src/daal4py.cpp
```diff
@@ -244,75 +222,97 @@ extern PyObject * make_nda(daal::data_management::KeyValueDataCollectionPtr * di
...
+                        std::cerr << "Found wrong dimensionality (" << PyArray_NDIM(ary) << ") of array in list when constructing SOA table (must be 1d)";
+                        delete soatbl;
+                        soatbl = NULL;
+                        break;
+                    }
+
+#define SETARRAY_(_T) {daal::services::SharedPtr< _T > _tmp(reinterpret_cast< _T * >(PyArray_DATA(ary)), NumpyDeleter(ary)); soatbl->setArray(_tmp, i);}
+                    SET_NPY_FEATURE(PyArray_DESCR(ary)->type, SETARRAY_, throw std::invalid_argument("Found unsupported array type"));
+#undef SETARRAY_
+
```
fschlimb: @RuslanIsrafilov is it enough to create a SOANumericTable with size and then just set the column pointers? Or do we need something else?
Note: the columns pointers are shared-pointers. daal4py will hold a ref until the numpy array gets GC'ed/deleted.

RuslanIsrafilov: It is enough

fschlimb:
Fixes #6

PR #27, opened 2018-12-17
adding more batch algos

fschlimb:
@oleksandr-pavlyk This PR adds all missing algorithms in Batch mode, except quality metrics and boosting.
Boosting and quality metrics will be dealt with in separate PRs.

@Alexander-Makaryev: the tests were refactored to mostly require a simple config only.

Parameter/Input variables are now derived from extracting set'ers in DAAL headers, no longer require a matching get.

@triskadecaepyon Another PR will handle new docu generation.


oleksandr-pavlyk:
@fschlimb I compiled, and checked `quatiles`, `sortign`, `em_gmm_init` and `em_gmm` as well as `stump_training` for dense batch modes.

PR #30, opened 2019-01-09
Create CODE_OF_CONDUCT.md

fschlimb:


fschlimb:
@rscohn2 @anton-malakhov

anton-malakhov:
Looks like a good idea! Do we want to apply it to other repos as well?

fschlimb:
I suggest we do. Similar we should come up with a CONTRIBUTING.md. It's requires more work than this, though.

PR #31, opened 2019-01-09
fixing communication bug in tree-reduce, bumping build number

fschlimb:
@Alexander-Makaryev
@dzagorny 
@oleksandr-pavlyk


Alexander-Makaryev:
ok

fschlimb:
@dmitrygladkov

fschlimb:
Not sure I understand your question.
We are forming a reduction tree. Any process can send only one partial result (it's own or a merged one). The upper (initially partial) half sends these. In the next level of the tree the senders have no more to do. Without this fix, processes would send their partial result multiple times but without a matching receive and so lead to dead-locks.

PR #38, opened 2019-01-23
adding sklearn testing for cron job

fschlimb:
@oleksandr-pavlyk we need to add the light-tests and I will need to define the cron job
 
Will need to investigate how to report back to github a success/failure indicator


oleksandr-pavlyk:
Would output of this run be saved in run logs?

The run log is probably where scikit-learn would like refer customer to.

PR #35, opened 2019-01-10
Feature/sklearn patches

oleksandr-pavlyk:
This is a continuation of #15 which got accidentally merged.

PR #41, opened 2019-01-28
Feature/sklearn patches

fschlimb:
Continuation of #35


fschlimb:
The branch CI will not succeed, we'll accept if only the PR CI in Travis is green.

PR #60, opened 2019-02-07
Modelbuilder

fschlimb:
@Alexander-Makaryev @oleksandr-pavlyk 

The modelbuilder's method names are all lower case with '_' instead of camel-case to align with the rest of daal4py.

Also, I made parent and position optional keyword args, users can then add root nodes more easily.


fschlimb:
We need to protect the code against earlier versions of DAAL.
Most of the code could be easily auto-generated.

PR #64, opened 2019-02-21
 added PCA normalization param

Alexander-Makaryev:
implementation of https://github.com/IntelPython/daal4py/issues/61

### DIFF
--- a/examples/pca_batch.py
+++ b/examples/pca_batch.py
@@ -33,8 +33,10 @@ except:
 def main(readcsv=read_csv, method='svdDense'):
     infile = "./data/batch/pca_normalized.csv"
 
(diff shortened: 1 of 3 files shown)


examples/pca_batch.py
```diff
@@ -33,8 +33,10 @@
 def main(readcsv=read_csv, method='svdDense'):
     infile = "./data/batch/pca_normalized.csv"
 
+    # This is an optional argument z score can be configured differently
```
fschlimb: `# 'normalization' is an optional parameter to PCA; we use z-score which could be configured differently`

fschlimb:
@oleksandr-pavlyk

PR #67, opened 2019-02-27
Feature/optional args

Alexander-Makaryev:
implementation of https://github.com/IntelPython/daal4py/issues/65


Alexander-Makaryev:
@fschlimb Please give me advice what other inputs from ignore list must be added as optional args

fschlimb:
@Alexander-Makaryev We definitely want PCA:correlation. From the top of my head I can exclude the ones in Kmeans-init and the ones with 'Optional' in the name (optimization solvers).

I suggest you identify which ones in the list are actually input arguments.

generator/wrappers.py
```diff
@@ -149,14 +134,65 @@ def wrap_algo(algo, ver):
...
+        'weights': True,
+    },
+    'algorithms::ridge_regression::training': {
+        'weights': True,
+    },
+    'algorithms::stump::classification::training': {
+        'weights': True,
+    },
+    'algorithms::stump::regression::training': {
+        'weights': True,
```
fschlimb: Could you make these one-liners?

examples/optional_arg_kdtree_knn_cls_batch.py
```diff
@@ -0,0 +1,68 @@
+#*******************************************************************************
```
fschlimb: I suggest renaming the files to someting like kdtree_knn_weights_batch.py

tests/test_examples.py
```diff
@@ -52,6 +52,11 @@ class Base():
     We also use generic functions to test these, they get added later.
     """
 
+    def test_optional_arg_kdtree_knn_cls_batch(self):
+        import optional_arg_kdtree_knn_cls_batch as ex
+        (_, predict_result, test_labels) = self.call(ex)
+        self.assertTrue(np.count_nonzero(test_labels != predict_result.prediction) < 170)
```
fschlimb: can you move this assertion into the example?

PR #75, opened 2019-04-01
Test sklearn master

fschlimb:


fschlimb:
The latest version triggers sklearn tests on its tip of master from github. The idea is to have a permanent branch just as a placeholder so we can get a separate badge.

@oleksandr-pavlyk for some reason we get an error, saying daal4py is not available. Do you have some time today to investigate further?

fschlimb:
@oleksandr-pavlyk this branch does not trigger on CircleCi, it's triggered by cron-job. Adjust cron time minutes (in config.yaml) to 2-3 minutes in future and it will trigger after pushing.

fschlimb:
We need 2 PR to cleanup the dummy branch.
I will merge now and open another PR with the necessary cleanup.

PR #79, opened 2019-04-03
Add compilers to conda recipe

ehsantn:


fschlimb:
This fails on OsX.
I suspect something goes wrong when linking. The missing symbol _daal_del_ls_ptr is defined in libdaal_thread. It is listed at link-time but not as a 'Needed DSO' (like in older succeeding builds).
The symbol has only one leading '_' on Linux, but two ('__') in the message, not sure if that's normal.
@averbukh @dr-pain any idea? Have you seen something like this before?

ehsantn:
I moved the compilers to the build section in case it matters. But I see the same error on my local mac when I run the tests after build.

fschlimb:
@ehsantn Could you rebase to latest master and force-push your branch? I'd like to see if this builds now with latest changes to configs.

napetrov:
Already implemented

PR #82, opened 2019-04-09
avoid re-building if not needed, better deps, better docs

fschlimb:
@Alexander-Makaryev 
* Avoiding re-generating sources unless generator code has changed (or genereated files are missing).
* adjusted meta data

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,18 +27,11 @@ from os.path import join as jp
 from distutils.sysconfig import get_config_vars
 from Cython.Build import cythonize
 from Cython.Distutils import build_ext
+import glob
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -27,19 +27,12 @@
 from distutils.sysconfig import get_config_vars
 from Cython.Build import cythonize
 from Cython.Distutils import build_ext
+import glob
 
 import numpy as np
 
 npyver = int(np.__version__.split('.')[1])
```
Alexander-Makaryev: We don't use it anywhere else. Does it must be removed too?

fschlimb: removed

PR #89, opened 2019-05-06
fixing mem leak

fschlimb:
@Alexander-Makaryev 
`get_dat_or_file` was allocating memory for the data_or_file wrapper which was never freed.
This was "hiding" a ref-count issue for SOA tables (e.g. when reading from pandas)

### DIFF
--- a/generator/format.py
+++ b/generator/format.py
@@ -63,7 +63,7 @@ def flat(typ):
 def cy_callext(arg, typ_cy, typ_cyext, s2e=None):
     '''Where needed decorate argument with conversion when calling C++ from cython'''
     if 'data_or_file' in typ_cy:
-        return 'mk_data_or_file({0})'.format(arg)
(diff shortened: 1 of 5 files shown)


oleksandr-pavlyk:
I ran the following script:

```python
import pandas, daal4py, psutil, numpy as np
d = {'col1': [x**2 for x in range(10**6)], 'col2': np.random.randn(10**6), 'col3': np.random.randn(10**6).astype(np.single) }
df = pandas.DataFrame(d)


def test(it=10000):
    mem_bef = psutil.virtual_memory().used
    for _ in range(it):
        res = daal4py.low_order_moments(fptype='double').compute(df)
    del res
    mem_aft = psutil.virtual_memory().used
    return mem_aft - mem_bef

sum([ test(1000) for _ in range(100) ])

# master 
# Out[3]: 400609280

# this PR
# Out[3]: -9678848
```

PR #99, opened 2019-05-27
set 2020.1 ver

Alexander-Makaryev:

PR #101, opened 2019-05-28
Bump sklearn version

oleksandr-pavlyk:
Bumped up version of sklearn patches are known to be compatible with.

Updated version of Intel(R) DAAL for which certain examples (stump, boosting) apply.

Updated version of the packed in meta.yaml

Because DAAL is known to be non-deterministic in the constructed model for decision forest regression, exclude `check_fit_idempotent`  for `RandomForestRegressor` in  test_estimators.py

DAAL team has been notified.

@fschlimb @Alexander-Makaryev

### DIFF
--- a/examples/run_examples.py
+++ b/examples/run_examples.py
@@ -41,11 +41,11 @@ else:
 req_version = defaultdict(lambda:(2019,0))
 req_version['decision_forest_classification_batch.py'] = (2019,1)
 req_version['decision_forest_regression_batch.py'] = (2019,1)
-req_version['adaboost_batch.py'] = (2019,4)
-req_version['brownboost_batch.py'] = (2019,4)
-req_version['logitboost_batch.py'] = (2019,4)
-req_version['stump_classification_batch.py'] = (2019,4)
-req_version['stump_regression_batch.py'] = (2019,4)
+req_version['adaboost_batch.py'] = (2019,5)
+req_version['brownboost_batch.py'] = (2019,5)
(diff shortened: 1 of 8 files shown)


examples/run_examples.py
```diff
@@ -41,11 +41,11 @@
 req_version = defaultdict(lambda:(2019,0))
 req_version['decision_forest_classification_batch.py'] = (2019,1)
 req_version['decision_forest_regression_batch.py'] = (2019,1)
-req_version['adaboost_batch.py'] = (2019,4)
-req_version['brownboost_batch.py'] = (2019,4)
-req_version['logitboost_batch.py'] = (2019,4)
-req_version['stump_classification_batch.py'] = (2019,4)
-req_version['stump_regression_batch.py'] = (2019,4)
```
fschlimb: These version checks are done in a separate PR #99. The correct version would be 2020.1

oleksandr-pavlyk: So would you rather I merge #99 into this branch, and we merge these changes together, or I drop changes to these lines breaking CI and wait until #99 is merged ?

fschlimb: I suggest the latter, mergine #99 and then yours should work just fine even if you rebase your's onto #99 now

oleksandr-pavlyk: I have cherry-picked 2 commits changing this versions from the branch behind #99.

The change to `normalization_zscore` in #99 to use `doScale` is better in my PR, since it allows daal4py to build and work with older DAAL versions as well

PR #93, opened 2019-05-14
added specialisation for parallelPlusDense in kmeans

Alexander-Makaryev:
Implementation of https://github.com/IntelPython/daal4py/issues/77

PR #106, opened 2019-06-05
Try conda forge circleci

oleksandr-pavlyk:
Changed `set_up_env.sh` and used default channel alongside conda-forge for conda-build.

The 13-15 test failures we saw in test suite of unpatched scikit-learn, see #93, are gone.

There is a single test failure `test_omp_cv`, which I suspect is just flaky. Need to discuss this with @ogrisel or @jeremiedbb.

The patched run also fails ``sklearn/tests/test_pipeline.py::test_pipeline_memory``, with DAAL optimized SVM using more than expected.

@fschlimb @Alexander-Makaryev

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ jobs:
             . ~/miniconda/etc/profile.d/conda.sh
             conda activate bld
             mkdir -p /tmp/out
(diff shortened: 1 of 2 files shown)


fschlimb:
👍

PR #107, opened 2019-06-07
initial abstraction layer for communication

fschlimb:


fschlimb:
Removed hard runtime dependence on MPI
* MPI stuff is now in python module which gets loaded/imported at runtime
* conda recipe adds no runtime dependence to MPI
* you can run non-distributed daal4py without any problem
* when asking for distributed and MPI is not available, an exception is thrown like `ImportError: libmpi.so.12: cannot open shared object file: No such file or directory`
* as a POC you can set D4P_TRANSCEIVER env var; it will try to load that module and bail out if not available: `ModuleNotFoundError: No module named 'no_transceiver'`

@tomashek @triskadecaepyon @oleksandr-pavlyk @shssf

fschlimb:
Lin/OSX python2.7 fail because of new module initialization. Maybe it's time stop supporting python 2.

src/mpi/mpi_transceiver.h
```diff
@@ -0,0 +1,49 @@
...
+public:
+    mpi_transceiver() : transceiver_impl() {}
+
+    virtual void init();
+
+    virtual void fini();
+    
+    virtual size_t nMembers();
+
+    virtual size_t me();
```
shssf: it would be better to rename `me` to `id`

src/mpi/mpi_transceiver.h
```diff
@@ -0,0 +1,49 @@
...
+class mpi_transceiver : public transceiver_impl
+{
+public:
+    mpi_transceiver() : transceiver_impl() {}
+
+    virtual void init();
+
+    virtual void fini();
+    
+    virtual size_t nMembers();
```
shssf: It would be better to rename `nMembers` to `size`

fschlimb: size of what? This is not MPI!

src/mpi/mpi_transceiver.h
fschlimb: Not sure I agree. id of what?
Any other opinion?

> It is not clear why so complicated inheritance schema used.

What is unclear? I hoped the comments clarify the intent. I see no other way to provide the same functionality with a clean separation of concerns. Suggestions?

fschlimb:
OK, I'll merge now. We can always suggest changes with new PRs. This is an internal API only, not exposed to users.
At some point someone might want to provide an alternative communication layer, like TCP or OFI. This will be a good time to add the missing implementations in transciever_impl.

PR #108, opened 2019-06-10
feature/lasso regression and coordinate descent

Alexander-Makaryev:


examples/lasso_regression_batch.py
```diff
@@ -0,0 +1,64 @@
...
+    predict_algo = d4p.lasso_regression_prediction()
+    # read test data (with same #features)
+    pdata = readcsv(testfile, range(10))
+    ptdata = readcsv(testfile, range(10,11))
+    # now predict using the model from the training above
+    predict_result = predict_algo.compute(pdata, train_result.model)
+
+    # The prediction result provides prediction
+    assert predict_result.prediction.shape == (pdata.shape[0], dep_data.shape[1])
+
```
fschlimb: Please add checking for correct prediction (similar to other examples where we expect a certain accuracy).

generator/wrappers.py
```diff
@@ -121,6 +123,8 @@ def wrap_algo(algo, ver):
     'algorithms::em_gmm::init': ['seed',], # deprecated
     'algorithms::pca': ['covariance'], # parameter defined multiple times with different types
     'algorithms::kdtree_knn_classification': ['seed',], # deprecated
+    'algorithms::lasso_regression::training': ['optionalArgument', 'gramMatrix'], # internal stuff, optional input
+    'algorithms::lasso_regression::prediction': ['algorithms::linear_model::interceptFlag',], # parameter
```
fschlimb: why is this ignored?

generator/wrappers.py
```diff
@@ -395,56 +400,6 @@ def wrap_algo(algo, ver):
...
-            'specs': [
-                {
-                    'template_decl': ['fptype'],
-                    'expl': OrderedDict([('method', 'daal::algorithms::ridge_regression::prediction::defaultDense')]),
-                },
-            ],
-        },
-    },
-}
-
```
fschlimb: How did you check this is still producing the correct code?

fschlimb: Oh I see, it's not used anywhere...Nevermind.

generator/wrappers.py
Alexander-Makaryev: optionalArgument is internal parameter for kernel calls which is not accessible for end users.
gramMatrix is optional input that is not ready for using now, ignoring is requested by DAAL team. It will be moved to optional args.
interceptFlag is common interface parameter which is not provided in prediction algo/stage directly, it is a part of model from training stage.

examples/lasso_regression_batch.py
fschlimb: I assume the expected accuracy has been verified with the DAAL team.

PR #110, opened 2019-06-13
Consolidate all sklearn related documentation at a single URL

ogrisel:
Here is a proposal to gather all the relevant information about scikit-learn integration on a single page that can be referenced from a single URL.

Otherwise it's hard to discover the patching instruction and API usage code snippet that were on the home page from within the sklearn.html page.


fschlimb:
👍 
@oleksandr-pavlyk ?

PR #111, opened 2019-06-13
Make sklearn patching verbose by default

ogrisel:
I think it's important to let the user know when sklearn is patched or not.

I think patching could be made silent if some environment variable is explicitly defined but by default otherwise users might be surprised that they are using a class from daal4py while they are expecting only code from the sklearn package.

### DIFF
--- a/daal4py/sklearn/monkeypatch/dispatcher.py
+++ b/daal4py/sklearn/monkeypatch/dispatcher.py
@@ -99,7 +99,7 @@ def enable(name=None, verbose=True):
         for key in _mapping:
             do_patch(key)
     if verbose and sys.stderr is not None:
-        sys.stderr.write("Intel DAAL solvers for sklearn enabled: "
+        sys.stderr.write("Intel(R) DAAL solvers for sklearn enabled: "
                          "https://intelpython.github.io/daal4py/sklearn.html\n")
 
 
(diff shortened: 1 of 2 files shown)


ogrisel:
A single URL with all relevant information as done in #110 is therefore beneficial.

daal4py/sklearn/monkeypatch/dispatcher.py
```diff
@@ -97,6 +98,9 @@ def enable(name=None):
     else:
         for key in _mapping:
             do_patch(key)
+    if verbose and sys.stderr is not None:
+        sys.stderr.write("Intel DAAL solvers for sklearn enabled: "
+                         "https://intelpython.github.io/daal4py/sklearn.html\n")
```
fschlimb: We probably need a message that accounts for selective patching, e.g. if only some algorithms have been patched. Might be confusing otherwise.

ogrisel: I wanted to do that initially (e.g. one line for each class / function patched) but I was it afraid it would maybe be too verbose. The information about the class that are patched is available by clicking on the link.

@oleksandr-pavlyk what's your opinion?

fschlimb: @ogrisel I meant to somehow indicate which of all the algos were patched if name!=None. The message is ok if name==None.

oleksandr-pavlyk:
Travis CI PR job fails because of GATWAY timeout due to use of older version of conda build. The newer version uses their new archival method.

I think this is safe to merge.

oleksandr-pavlyk:
A fix for the error we see in Travis jobs is under review: https://github.com/conda/conda-build/pull/3585

PR #113, opened 2019-06-20
dbscan example, default weights, required versions, tests

Alexander-Makaryev:
Executing of example takes too many time. We should double check why it is so and maybe change test data or parameters of algorithm.


examples/dbscan_batch.py
```diff
@@ -0,0 +1,61 @@
...
+except:
+    # fall back to numpy loadtxt
+    read_csv = lambda f, c, t=np.float64: np.loadtxt(f, usecols=c, delimiter=',', ndmin=2)
+
+
+def main(readcsv=read_csv, method='defaultDense'):
+    infile = "./data/batch/dbscan_dense.csv"
+    epsilon = 0.02;
+    minObservations = 180;
+
```
fschlimb: no ';'s please

tests/test_examples.py
```diff
@@ -157,6 +157,7 @@ def test_svm_batch(self):
     ('stump_regression_batch', None, None, (2020, 1)),
     ('svm_multiclass_batch', 'svm_multiclass_batch.csv', lambda r: r[0].prediction),
     ('univariate_outlier_batch', 'univariate_outlier_batch.csv', lambda r: r[1].weights),
+    ('dbscan_batch', None, None, (2020, 0)),
```
fschlimb: This is not sufficient. The built-in test needs to check more, like the actual assignments.

Issue #50, opened 2019-02-04
Add computation of leftSingularMatrix to distributed SVD


fschlimb:
- add an extra step (Distributed<step3>) to corresponding entry in ```wrappers.py:has_dist```
- Most work will be adding the new pattern
  - use something like ```map-reduce-star-plus``` if same can be used for QR(#49)
  - otherwise follow ```dist_custom``` pattern
- add using leftSingularMatrix to svd examples

Alexander-Makaryev:
implemented

Issue #49, opened 2019-02-04
Add distributed mode to QR


fschlimb:
- add all 3 steps to new entry in ```wrappers.py:has_dist```
- Most work will be adding the new pattern
  - use something like ```map-reduce-star-plus``` if same can be used for SVC
  - otherwise follow ```dist_custom``` pattern
- add distributed example

Alexander-Makaryev:
implemented

PR #115, opened 2019-06-24
Feature/distributed svd qr

Alexander-Makaryev:
Implementation of #50 and #49.


Alexander-Makaryev:
@fschlimb It was easier to create new branch and then cherry-pick necessary commits. Previous PR is closed.

src/daal4py_dist.h
```diff
@@ -71,10 +71,29 @@ struct PartialIOManagerSingle : public PartialIOManager< A, O >
     }
 };
 
+// Input/Output manager for intermediate steps with both: result and partial result
+// abstracts from input/output types
+// also defines how to get results and finalize
```
fschlimb: The comment doesn't seem right.

PR #117, opened 2019-07-01
Feature/optional algo setup args

Alexander-Makaryev:


examples/lasso_regression_batch.py
```diff
@@ -39,21 +39,21 @@ def main(readcsv=read_csv, method='defaultDense'):
 
     # Read data. Let's have 10 independent, and 1 dependent variable (for each observation)
     indep_data = readcsv(infile, range(10))
-    dep_data   = readcsv(infile, range(10,11))
+    dep_data   = readcsv(infile, range(10,12))
```
fschlimb: Pls fix the above comment if you want to use 2 dep vars

PR #118, opened 2019-07-01
adding cfg-knob to let distributed steps keep state, used for kmeans_…

fschlimb:
…init::step3

@Alexander-Makaryev Can you please add kmeans-init parallel++ stuff from your old branch? Also, please add your tests comparing distributed vs batch implementations.

### DIFF
--- a/examples/kmeans_spmd.py
+++ b/examples/kmeans_spmd.py
@@ -23,11 +23,7 @@
 import daal4py as d4p
 from numpy import loadtxt, allclose
 
-if __name__ == "__main__":
-
-    # Initialize SPMD mode
-    d4p.daalinit()
-
+def main():
     infile = "./data/distributed/kmeans_dense.csv"
     nClusters = 10
     maxIter = 25
@@ -60,8 +56,17 @@ if __name__ == "__main__":
     algo = d4p.kmeans(nClusters, 0, assignFlag=True) # maxIt=0; not distributed, we compute on local data only!
     assignments = algo.compute(data, result.centroids).assignments
 
-    print("\nFirst 10 cluster assignments:\n", assignments[0:10])
-    print("\nFirst 10 dimensions of centroids:\n", result.centroids[:,0:10])
-    print("\nObjective function value:\n", result.objectiveFunction)
-    print('All looks good!')
+    return (assignments, result)
+
+
+if __name__ == "__main__":
+    # Initialize SPMD mode
+    d4p.daalinit()
+    (assignments, result) = main()
+    # result is available on all processes - but we print only on root
+    if d4p.my_procid() == 0:
+        print("\nFirst 10 cluster assignments:\n", assignments[0:10])
+        print("\nFirst 10 dimensions of centroids:\n", result.centroids[:,0:10])
+        print("\nObjective function value:\n", result.objectiveFunction)
+        print('All looks good!')
     d4p.daalfini()
--- a/examples/kmeans_spmd.py
+++ b/examples/kmeans_spmd.py
@@ -23,13 +23,13 @@
 import daal4py as d4p
 from numpy import loadtxt, allclose
 
-def main():
+def main(method='plusPlusDense'):
     infile = "./data/distributed/kmeans_dense.csv"
     nClusters = 10
     maxIter = 25
 
     # configure a kmeans-init
-    init_algo = d4p.kmeans_init(nClusters, method="plusPlusDense", distributed=True)
+    init_algo = d4p.kmeans_init(nClusters, method=method, distributed=True)
     # Load the data
     data = loadtxt(infile, delimiter=',')
     # now slice the data, it would have been better to read only what we need, of course...
--- a/src/dist_kmeans_init.h
+++ b/src/dist_kmeans_init.h
@@ -198,6 +198,122 @@ namespace dist_custom {
         }
     };
 
+    // specialize dist_custom for kemans_init<parallelPlusDense>
+    template< typename fptype >
+    class dist_custom< kmeans_init_manager< fptype, daal::algorithms::kmeans::init::parallelPlusDense > >
+    {
+    public:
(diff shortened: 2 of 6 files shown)


Alexander-Makaryev:
@fschlimb please look into the last commit https://github.com/IntelPython/daal4py/pull/118/commits/641035462c2309c97b8c24ffbc1ee0d0ce92a7f6

src/transceiver.h
```diff
@@ -351,26 +351,33 @@ std::vector<daal::services::SharedPtr<T> > transceiver::gather(const daal::servi
...
-            // This is inefficient, we need to write our own DatArchive to avoid extra copy
-            daal::data_management::OutputDataArchive out_arch(reinterpret_cast<daal::byte*>(buff+offset), sizes[i]);
-            all[i] = daal::services::staticPointerCast<T>(out_arch.getAsSharedPtr());
-            offset += sizes[i];
+            if(sizes[i] > 0) {
+                // This is inefficient, we need to write our own DatArchive to avoid extra copy
+                daal::data_management::OutputDataArchive out_arch(reinterpret_cast<daal::byte*>(buff+offset), sizes[i]);
+                all.push_back(daal::services::staticPointerCast<T>(out_arch.getAsSharedPtr()));
+                offset += sizes[i];
+            }
```
fschlimb: This should have a 'else' and store NULL in the sharedptr for empty archives. Generally the expectation is that we find result for process i at position i. This also means you can keep the resize. Moreover, callers must be aware of the possibility to get an empty shared-ptr.

src/dist_kmeans_init.h
```diff
@@ -284,7 +284,7 @@ namespace dist_custom {
                         step4OutMerged->addNumericTable(step14OutMaster[i]);
                     }
                 }
-                tcvr->bcast(step4OutMerged, data_rank);
+                tcvr->bcast(step4OutMerged, 0);
```
fschlimb: Can we remove all use of data_rank?

src/dist_kmeans_init.h
```diff
@@ -198,6 +198,122 @@ namespace dist_custom {
...
+                } else {
+                    step14Out = daal::data_management::NumericTablePtr();
+                }
+
+                // we need to gather all exist results on root, merge them into one table and then share it with all non-roots
+                auto step14OutMaster = tcvr->gather(step14Out);
+                daal::data_management::RowMergedNumericTablePtr step4OutMerged(new daal::data_management::RowMergedNumericTable());
+                if(rank == 0)
+                {
+                    for (int i = 0; i < step14OutMaster.size(); i++)
```
fschlimb: See other comment, gather should return result of rank/process i at position i. This code must handle possible NULL pointers.

src/dist_kmeans_init.h
```diff
@@ -198,6 +198,122 @@ namespace dist_custom {
         }
     };
 
+    // specialize dist_custom for kemans_init<parallelPlusDense>
```
fschlimb: pls add a description how parallel++ differs from ++ method (or add a full description like for ++).

examples/kmeans_spmd.py
```diff
@@ -23,11 +23,7 @@
 import daal4py as d4p
 from numpy import loadtxt, allclose
 
-if __name__ == "__main__":
-
-    # Initialize SPMD mode
-    d4p.daalinit()
-
+def main():
```
fschlimb: pls allow configuring for other init-methods

tests/spmd_test_examples.py
```diff
@@ -18,14 +18,23 @@ def test_qr_spmd(self):
...
-            batch_result = d4p.kmeans(nClusters=nClusters, maxIterations=maxIter, assignFlag=True).compute(data, batch_init_res.centroids)
-            self.assertTrue(np.allclose(result.centroids, batch_result.centroids))
+            spmd_init_res = d4p.kmeans_init(nClusters=nClusters, method="plusPlusDense", distributed=True).compute(spmd_data)
+
+            self.assertTrue(np.allclose(batch_init_res.centroids, spmd_init_res.centroids))
+
+            batch_res = d4p.kmeans(nClusters=nClusters, maxIterations=maxIter).compute(data, batch_init_res.centroids)
+            spmd_res = d4p.kmeans(nClusters=nClusters, maxIterations=maxIter, distributed=True).compute(spmd_data, spmd_init_res.centroids)
+
+            self.assertTrue(np.allclose(batch_res.centroids, batch_res.centroids))
```
fschlimb: pls add the same for parallel++ and deterministic-dense (see other comment: add parameter to example)

src/dist_kmeans_init.h
```diff
@@ -204,6 +204,18 @@ namespace dist_custom {
...
+            then we repeat nRounds times in the loop:
+                step2 computes data for step3 on each rank
+                then we gather results from step2 and execute step3 on root
+                step3 provides inputs for step4 on each rank
+                on each rank if there is data for step4 then we run it
+                results of step4 are input for step2 of next iteration
+            after the loop we execute step2 one more time with data from last iteration of loop on each rank
+            and in step5 we select the initial centroids
+        */
         static typename Algo::iomb_type::result_type
```
fschlimb: What about
```
/*
            step1 provides initial input for step2, inside the loop step4 produces the input for step2.
            We have to keep input for step2 because it will also be used as input for final step5.
            Now we iterate/loop for nRounds:
                - step2 computes on each rank
                - we gather results from all step2's
                - gathered data from step2 is input to step3, executed on root only
                - output of step3 is scattered to all ranks
                - ranks which received non-empty output from step3 will execute step4 on its data
                - results of step4 are input for step2 of next iteration
            After the loop
                - we execute step2 one more time with data from last iteration of loop on each rank
                - and finally select the initial centroids in step5 on root
            The resulting centroids are broadcasted to all processes.
*/
```

PR #126, opened 2019-08-07
Sklearn 0.21.3 update

oleksandr-pavlyk:
@Alexander-Makaryev @fschlimb 

Updated dispatcher to be compatible with sklearn==0.21.3, also removed SVM test failures from deselected.yaml

Updated message printed to stderr when patches are activated to use full product name, rather than the abbreviation

### DIFF
--- a/daal4py/sklearn/monkeypatch/dispatcher.py
+++ b/daal4py/sklearn/monkeypatch/dispatcher.py
@@ -99,7 +99,7 @@ def enable(name=None, verbose=True):
         for key in _mapping:
             do_patch(key)
     if verbose and sys.stderr is not None:
-        sys.stderr.write("Intel(R) DAAL solvers for sklearn enabled: "
+        sys.stderr.write("Intel(R) Data Analytics Acceleration Library (Intel DAAL) solvers for sklearn enabled: "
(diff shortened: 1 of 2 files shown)


daal4py/sklearn/monkeypatch/dispatcher.py
```diff
@@ -99,7 +99,7 @@ def enable(name=None, verbose=True):
         for key in _mapping:
             do_patch(key)
     if verbose and sys.stderr is not None:
-        sys.stderr.write("Intel(R) DAAL solvers for sklearn enabled: "
+        sys.stderr.write("Intel(R) Data Analytics Acceleration Library (Intel DAAL) solvers for sklearn enabled: "
```
Alexander-Makaryev: Intel(R) Data Analytics Acceleration Library (Intel(R) DAAL)
looks like the second '(R)' sign is necessary according
http://prod-namesdb.intel.com/search.asp with search string 'DAAL':
The approved usage is: Intel® Data Analytics Acceleration Library (Intel® DAAL)

oleksandr-pavlyk: Done!

PR #128, opened 2019-08-20
Fix tests against master

oleksandr-pavlyk:
Updated SVM patch to work with master


oleksandr-pavlyk:
@Alexander-Makaryev @fschlimb

PR #131, opened 2019-08-21
fix lasso example with sparse matrix

Alexander-Makaryev:


examples/lasso_regression_batch.py
```diff
@@ -53,6 +53,11 @@ def main(readcsv=read_csv, method='defaultDense'):
 
     # The prediction result provides prediction
     assert predict_result.prediction.shape == (pdata.shape[0], dep_data.shape[1])
+
+    # the example is used in tests with the scipy.sparse matrix
+    # we use this trick until subtracting a sparse matrix is not supported
+    if hasattr(ptdata, 'toarray'):
+        ptdata = ptdata.toarray()
     assert np.square(predict_result.prediction - np.array(ptdata)).mean() < 2.2
```
oleksandr-pavlyk: better `np.array(ptdata, copy=False)` or `np.asarray(ptdata)`.

PR #132, opened 2019-08-30
Fixes of test failures against master

oleksandr-pavlyk:
These were due to changes in default keyword values.

`multi_class` and `solver` keywords of `LogisticRegression` changed defaults
from 'warn' to 'auto' and 'lbfgs' respectively.


oleksandr-pavlyk:
@fschlimb @Alexander-Makaryev

PR #133, opened 2019-09-09
added dummy examples for model builders

Alexander-Makaryev:


Alexander-Makaryev:
@oleksandr-pavlyk please look one more time at names of files. Are they fine?

PR #134, opened 2019-09-26
Adaboost algorithm and support of changes in DAAL

Alexander-Makaryev:


generator/wrappers.py
```diff
@@ -131,6 +131,7 @@ def wrap_algo(algo, ver):
     'algorithms::kdtree_knn_classification': ['seed',], # deprecated
     'algorithms::lasso_regression::training': ['optionalArgument'], # internal stuff
     'algorithms::lasso_regression::prediction': ['algorithms::linear_model::interceptFlag',], # parameter
+    'algorithms::multi_class_classifier': ['algorithms::multi_class_classifier::getTwoClassClassifierModels',] # unsupported return type ModelPtr*
```
oleksandr-pavlyk: Will this continue working? https://github.com/IntelPython/daal4py/blob/master/daal4py/sklearn/svm/svm.py#L229

fschlimb: yes

PR #135, opened 2019-09-30
KMeans sklearn API class change

oleksandr-pavlyk:
If initial clusters are provided, or non-random, override n_init keyword,
and perform only 1 iteration.

@Alexander-Makaryev


oleksandr-pavlyk:
Unsure why 'commit' workflow failed. Test suites were run against both patched and unpatched code-base, and passed. 

Commit workflow failed because shell command 

```
export SKL=`grep -e '\d* passed.* second' ~/skl.out`
```

not longer selects the status of the run, which now reads as 

```
= 12066 passed, 66 skipped, 196 deselected, 3 xfailed, 1890 warnings in 618.38s (0:10:18) =
```

I will change the pattern to ``grep -E "=(\s\d*\w*,?)+ in .*\s=" ./d4p.out`` in a moment.

PR #139, opened 2019-10-28
Support MPI_LIBNAME env.

oleksandr-pavlyk:
When compiling daal4py with Cray MPI, the library name is not 'mpi',
so an edit to setup.py is needed to compile DAAL4PY.

Now one can just set MPI_LIBNAME to the intended library name.
If the MPI shared library is `libmpi_gnu44.so`, we set `MPI_LIBNAME=mpi_gnu44`

@Alexander-Makaryev @fschlimb


oleksandr-pavlyk:
ping.

PR #141, opened 2019-11-04
Use DAAL in SVC only if break_ties=False

bibikar:
Running `sklearn.svm.SVC.fit()` with daal4py 2019.5 with scikit-learn 0.21.x was not using DAAL acceleration. We were checking that `getattr(self, 'break_ties', False)` is `True`, but the wrapper for `predict` was checking the opposite, so neither fit nor predict were running with DAAL backend.

### DIFF
--- a/daal4py/sklearn/svm/svm.py
+++ b/daal4py/sklearn/svm/svm.py
@@ -426,7 +426,7 @@ def fit(self, X, y, sample_weight=None):
         # see comment on the other call to np.iinfo in this file
         seed = rnd.randint(np.iinfo('i').max)
 
-        if ( not sparse and not self.probability and getattr(self, 'break_ties', False) and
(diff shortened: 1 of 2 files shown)


oleksandr-pavlyk:
This is good. The failure of Py_PI is due to reintroduction of SVC optimization breaking tests for which exceptions were removed.

Change is needed to `deselected_tests.yaml`. In

```
  # See: https://github.com/scikit-learn/scikit-learn/issues/12738
  - svm/tests/test_svm.py::test_sample_weights <0.21.3
  - svm/tests/test_svm.py::test_precomputed    <0.21.3
  - svm/tests/test_sparse.py::test_sparse      <0.21.3
  - svm/tests/test_sparse.py::test_svc_iris    <0.21.3
  - ensemble/tests/test_bagging.py::test_sparse_classification <0.21.3
```

Qualifies `<0.21.3` need to be removed.

PR #142, opened 2019-11-05
BUG: mpi_transciever_impl should be always initialized

oleksandr-pavlyk:
@Alexander-Makaryev @fschlimb 

In the scenario where `import mpi4py` was executed prior to a
call to daal4py.daalinit(), and MPI has already been initialized
the mpi_transciever_impl internal state was not being initialized
causing a crash in the following example:

```
import numpy as np
from mpi4py import MPI
import os, os.path
import daal4py as d4p

d4p.daalinit()

fn = os.path.join('data', 'X_' + str( d4p.my_procid() ) + '.csv')
X = np.loadtxt(fn)
assert X.size > 0

svd_distr = d4p.svd(
    fptype='double',
    distributed=True,
    )
 res = svd_distr.compute(X)

d4p.daalfini()
``

### DIFF
--- a/src/mpi/mpi_transceiver.cpp
+++ b/src/mpi/mpi_transceiver.cpp
@@ -20,13 +20,13 @@
 
 void mpi_transceiver::init()
 {
-    int is_initialized;
-    MPI_Initialized(&is_initialized);
+    int is_mpi_initialized = 0;
+    MPI_Initialized(&is_mpi_initialized);
     // protect against double-init
-    if(!is_initialized) {
+    if(!is_mpi_initialized) {
         MPI_Init(NULL, NULL);
-        transceiver_impl::init();
     }
+    transceiver_impl::init();	
 }
 
 void mpi_transceiver::fini()
--- a/src/transceiver.h
+++ b/src/transceiver.h
@@ -153,16 +153,20 @@ class transceiver_impl : public transceiver_iface
 public:
     transceiver_impl()
         : m_me(-1),
-          m_nMembers(0)
+	  m_nMembers(0),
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by fschlimb** [comment]
Good catch.
Still, I don't think this will behave exactly like I wanted it to behave.
The call to parent::init() should be protected so that it gets called only once per instance even if called multiple times.
We probably need a second guard, a member variable that checks for prior object initialization. The guard and its implementation should live in transceiver_impl.

**Comment by oleksandr-pavlyk** [comment]
Fixed that, by introducing a private member `m_initialized` to `transceiver_impl` and a method to read it. 

Notice however, that `transceiver` class itself seems to have `m_inited` private variable, but it appears unused. Should be the removed?

File: src/mpi/mpi_transceiver.cpp [hunk 1]
```diff
@@ -20,13 +20,14 @@
...
-    MPI_Initialized(&is_initialized);
+    int is_mpi_initialized = 0;
+    MPI_Initialized(&is_mpi_initialized);
     // protect against double-init
-    if(!is_initialized) {
+    if(!is_mpi_initialized) {
         MPI_Init(NULL, NULL);
-        transceiver_impl::init();
     }
+    if (!transceiver_impl::is_initialized())
```
fschlimb: sorry, I was unclear, I would do the check in transceiver_impl and always do the call here. It's a bit cleaner and less error prone for other concrete implementations (to come).

oleksandr-pavlyk: Ok, I will make the change.

PR #143, opened 2019-11-21
Update recipe

oleksandr-pavlyk:
The intent is to use conda's compiler jinja variable to
take care of DSO warnings emitted by conda-build

@Alexander-Makaryev @fschlimb


oleksandr-pavlyk:
Merging. `build_pypi` is failing because of the patches and upgrade of scikit-learn.

PR #144, opened 2019-12-09
removing py27 builds

oleksandr-pavlyk:


oleksandr-pavlyk:
To be merged in early 2020.

PR #148, opened 2019-12-16
_make_nt converts FC ndarrays to SOA

oleksandr-pavlyk:
@fschlimb @Alexander-Makaryev

Correctness check:

```ipython
(psk022) [15:21:28 fxsatlin03 tmp]$ ipython
Python 3.7.3 | packaged by conda-forge | (default, Dec  6 2019, 08:54:18)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.10.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import daal4py, numpy as np, daal4py.sklearn

In [2]: X = np.random.randn(5, 100).T

In [3]: y = np.random.randn(100)

In [4]: daal4py.sklearn.linear_model.Ridge().fit(X, y).coef_
Out[4]: array([-0.0370632 , -0.01074325, -0.10366734,  0.20030913, -0.0705896 ])

In [5]: daal4py.sklearn.linear_model.Ridge().fit(np.ascontiguousarray(X), y).coef_
Out[5]: array([-0.0370632 , -0.01074325, -0.10366734,  0.20030913, -0.0705896 ])

In [6]: quit
```

This closes the performance gap between using pandas DataFrame (underlying f-contiguous layout)
and the actual F-contiguous ndarray input.

### DIFF
--- a/src/daal4py.cpp
+++ b/src/daal4py.cpp
@@ -293,10 +293,9 @@ daal::data_management::NumericTablePtr make_nt(PyObject * obj)
             } else {
 		if(array_is_behaved_F(ary) && (PyArray_NDIM(ary) == 2)) {
 		    int _axes = 0;
-		    npy_intp i = 0;
 		    npy_intp N = PyArray_DIM(ary, 1); // number of columns
 		    npy_intp column_len = PyArray_DIM(ary, 0);
-		    int array_type = PyArray_TYPE(ary);
+		    int ary_numtype = PyArray_TYPE(ary);
 		    /*
 		     * Input is 2D F-contiguous array: represent it as SOA numeric table
 		     */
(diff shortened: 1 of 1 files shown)


src/daal4py.cpp
```diff
@@ -290,7 +291,44 @@ daal::data_management::NumericTablePtr make_nt(PyObject * obj)
...
+		    PyArrayIterObject * it = (PyArrayIterObject *) PyArray_IterAllButAxis(obj, &_axes);
+		    if (it == NULL) {
+			Py_XDECREF(it);
+			throw std::runtime_error("Creating DAAL SOA table from F-contigous NumPy array failed: iterator could not be created");
+		    }
+
+		    soatbl = new daal::data_management::SOANumericTable(N, column_len);
+		    
+		    while (PyArray_ITER_NOTDONE(it)) {
+			PyArrayObject *slice = (PyArrayObject *) PyArray_SimpleNewFromData(1, &column_len, array_type, (void *)PyArray_ITER_DATA(it));
```
fschlimb: Why do we need a `slice` temporary? It seems to be used for ref-counting only and the current code keeps a double-indirection. Can't we use `ary`  directly for `NumpyDeleter` (and of course keep the `Py_INCREF`)?

src/daal4py.cpp
```diff
@@ -290,7 +291,44 @@ daal::data_management::NumericTablePtr make_nt(PyObject * obj)
...
+		    // iterate over columns
+		    PyArrayIterObject * it = (PyArrayIterObject *) PyArray_IterAllButAxis(obj, &_axes);
+		    if (it == NULL) {
+			Py_XDECREF(it);
+			throw std::runtime_error("Creating DAAL SOA table from F-contigous NumPy array failed: iterator could not be created");
+		    }
+
+		    soatbl = new daal::data_management::SOANumericTable(N, column_len);
+		    
+		    while (PyArray_ITER_NOTDONE(it)) {
```
fschlimb: `for(npy_intp i = 0; PyArray_ITER_NOTDONE(it); ++i, ++it)`

src/daal4py.cpp
oleksandr-pavlyk: I did this initially, but `NumpyDeleter` has an assertion checking that the pointer being deleted coincides with the data pointer of the array, which it would not.

oleksandr-pavlyk:
I ran a script:

```python
import psutil
import os
import numpy as np
import daal4py
import gc

X = np.load('./input_data/Xf.npy')
Y = np.load('./input_data/Yf.npy')

Y = Y.reshape((-1,1))

_ridgeParam = np.asarray(1.0, dtype=np.float64).reshape((1,-1))
n_iter_inner = 10

proc = psutil.Process(os.getpid())
rss0 = proc.memory_info().rss

while(True):
    for _ in range(n_iter_inner):
        ridge_alg = daal4py.ridge_regression_training(
            fptype='double',
            method='defaultDense',
            interceptFlag=True,
            ridgeParameters=_ridgeParam
        )
        ridge_res = ridge_alg.compute(X, Y)

    del ridge_res
    del ridge_alg
    rss1 = proc.memory_info().rss
    print( (rss1 - rss0) )
    gc.collect()
```

and while the `rss` was always increasing, the %MEM of the process is `top` stays constant. 

Intel(R) Inspector did not indicate any memory leaks in the daal4py code.  So merging.

PR #151, opened 2020-01-06
Use Python 3.7.3 on OSX, instead of 3.7 to work around an issue of Py…

oleksandr-pavlyk:
…Init__daal4py symbol being absent from the produced native extension binary

@fschlimb @AndresGuzman-Ballen @Alexander-Makaryev


oleksandr-pavlyk:
This works around https://github.com/IntelPython/daal4py/issues/152

oleksandr-pavlyk:
The latest change closes #152

oleksandr-pavlyk:
The commit workflow has failed due do 10m without any output which did not happen in prior runs. 

I therefore am going to merge the PR.

PR #161, opened 2020-02-03
Impi

Alexander-Makaryev:


.circleci/config.yml
```diff
@@ -20,7 +20,7 @@ jobs:
             . ~/miniconda/etc/profile.d/conda.sh
             conda activate bld
             mkdir -p /tmp/out
-            conda build --python=37 --override-channels -c conda-forge -c defaults --output-folder /tmp/out --no-test conda-recipe
+            conda build --python=37 --override-channels -c intel -c conda-forge -c defaults --output-folder /tmp/out --no-test conda-recipe
```
oleksandr-pavlyk: It's best if we do not do this, and leave conda-forge to be first in the list.

oleksandr-pavlyk: Can we use `-c conda-forge -c intel -c defaults`?

PR #164, opened 2020-02-06
Update Decision forest example for probabilities calculation

masdevas:


Alexander-Makaryev:
@masdevas We should use condition by version in purpose of back compatibility like it is done in https://github.com/IntelPython/daal4py/blob/master/examples/gradient_boosted_classification_batch.py

PR #167, opened 2020-02-12
Fix/numpy version

Alexander-Makaryev:
The try to defeat the warning:
No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.11
WARNING:conda_build.metadata:No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.11


Alexander-Makaryev:
@oleksandr-pavlyk 
conda build config was added but warning still occurs.

oleksandr-pavlyk:
The message goes away with using `conda build -c conda-forge -c intel $PYVER --numpy=1.17 ./conda-recipe`.

napetrov:
--numpy variable is now passed

PR #168, opened 2020-02-13
estimators_ attribues is now a property

oleksandr-pavlyk:
Conversion of DAAL model to list of Scikit-Learn estimators may be 
time consuming, and not needed for predict, unless out-of-bag score
is requested.

Moving the conversion to property allows to compute them on demand.
The result is cached, and reused afterwards.

@Alexander-Makaryev 
@ShvetsKS


ShvetsKS:
could you also please fix 
https://github.com/IntelPython/daal4py/pull/168/files#diff-ba842fd2454b0591b24fdb09b73005faR172

Seems we have to provide True for classification

oleksandr-pavlyk:
> Seems we have to provide True for classification

Done.

PR #172, opened 2020-02-18
using of create function instead call of the constructor

Alexander-Makaryev:
Used method ``NumericTable::create`` instead of constructor in preparation for future changes.

### DIFF
--- a/src/daal4py.cpp
+++ b/src/daal4py.cpp
@@ -232,19 +232,20 @@ extern PyObject * make_nda(daal::data_management::KeyValueDataCollectionPtr * di
 }
 
 template<typename T>
-static daal::data_management::NumericTable * _make_hnt(PyObject * nda)
+static daal::data_management::NumericTablePtr _make_hnt(PyObject * nda)
 {
-    daal::data_management::NumericTable * ptr = NULL;
+    daal::data_management::NumericTablePtr ptr;
     PyArrayObject * array = (PyArrayObject*)nda;
 
     assert(is_array(nda) && array_is_behaved(array));
 
     if(array_numdims(array) == 2) {
         // we provide the SharedPtr with a deleter which decrements the pyref
-        ptr = new daal::data_management::HomogenNumericTable<T>(daal::services::SharedPtr<T>((T*)array_data(array),
-                                                                                             NumpyDeleter(array)),
-                                                                (size_t)array_size(array,1),
-                                                                (size_t)array_size(array,0));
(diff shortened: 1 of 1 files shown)


fschlimb:
There are matching delete calls missing. The new code lets new'ed pointers get lost (usually under error-conditions).

Alexander-Makaryev:
@KalyanovD @prutskov please review

src/daal4py.cpp
```diff
@@ -232,19 +232,20 @@ extern PyObject * make_nda(daal::data_management::KeyValueDataCollectionPtr * di
...
 
     assert(is_array(nda) && array_is_behaved(array));
 
     if(array_numdims(array) == 2) {
         // we provide the SharedPtr with a deleter which decrements the pyref
-        ptr = new daal::data_management::HomogenNumericTable<T>(daal::services::SharedPtr<T>((T*)array_data(array),
-                                                                                             NumpyDeleter(array)),
-                                                                (size_t)array_size(array,1),
-                                                                (size_t)array_size(array,0));
+        ptr = daal::services::dynamicPointerCast<daal::data_management::NumericTable>(
```
prutskov: What is the reason of using `dynamicPointerCast`? You can initialize as follows `ptr=HomogenNumericTable<T>::create(...)`

src/daal4py.cpp
```diff
@@ -330,16 +331,16 @@ daal::data_management::NumericTablePtr make_nt(PyObject * obj)
...
-		    daal::data_management::SOANumericTable * soatbl = NULL;
+		    daal::data_management::SOANumericTablePtr soatbl;
 
 		    // iterate over columns
 		    PyArrayIterObject * it = (PyArrayIterObject *) PyArray_IterAllButAxis(obj, &_axes);
 		    if (it == NULL) {
-			Py_XDECREF(it);
-			throw std::runtime_error("Creating DAAL SOA table from F-contigous NumPy array failed: iterator could not be created");
+                Py_XDECREF(it);
+                throw std::runtime_error("Creating DAAL SOA table from F-contigous NumPy array failed: iterator could not be created");
```
KalyanovD: Wrong alignment

Alexander-Makaryev: Looks like we should do refactoring of the alignment in whole file. But if we will do it right now then it will be very hard to look into changes related to goal of this PR. It should be done later in separate PR.

src/daal4py.cpp
Alexander-Makaryev: Thanks a lot, fixed in next commit.

PR #173, opened 2020-02-19
enable build with oneDAL in CI

napetrov:
In this change we introduce following changes
- Enable CI with oneDAL + DPC++ build
- update req_version check to support complex versioning
- Introduce req_device check for examples
- Remove device selection failback to default device - fixes https://github.com/IntelPython/daal4py/issues/171 
- Update sycl examples to support execution if GPU device not available

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,4 +1,37 @@
 jobs:
+- job: DPCPP
+  pool:
+    vmImage: 'ubuntu-18.04'
+  steps:
+  - script: |
+      wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+      sudo apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+      rm GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+      echo "deb https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
+      sudo add-apt-repository "deb https://apt.repos.intel.com/oneapi all main"
+      sudo apt-get update
+      sudo apt-get install              \
+          intel-oneapi-common-vars      \
+          intel-oneapi-common-licensing \
+          intel-oneapi-tbb-devel        \
+          intel-oneapi-daal-devel       \
+          intel-oneapi-dpcpp-compiler   \
+          intel-oneapi-dev-utilities    \
+          intel-oneapi-libdpstd-devel   \
+          cmake
+      sudo bash -c 'echo libintelocl.so > /etc/OpenCL/vendors/intel-cpu.icd'
+      sudo mv /opt/intel/inteloneapi/compiler/latest/linux/lib/oclfpga /opt/intel/inteloneapi/compiler/latest/linux/lib/oclfpga_
+    displayName: 'apt-get'
+  - script: conda create -q -y -n CB python=3.7 conda-build conda-verify
+    displayName: Create Anaconda environment
+  - script: |
+      export DPCPP_VAR=/opt/intel/inteloneapi/compiler/latest/env/vars.sh
+      export DAALROOT=/opt/intel/inteloneapi/daal/latest
+      . /usr/share/miniconda/etc/profile.d/conda.sh
+      conda activate CB
+      conda build --override-channels -c conda-forge -c intel conda-recipe
+    displayName: conda build
+
 - job: Windows
   pool:
     vmImage: 'vs2017-win2016'
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -43,8 +43,6 @@ test:
         - scikit-learn
         - mpich # [osx]
         - impi_rt # [not osx]
-    imports:
-        - daal4py
     source_files:
         - examples
         - tests
--- a/conda-recipe/run_test.sh
+++ b/conda-recipe/run_test.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# if dpc++ vars path is specified
+if [ ! -z "${DPCPP_VAR}" ]; then
+    source ${DPCPP_VAR}
+fi
+
+# if DAALROOT is specified
+if [ ! -z "${DAALROOT}" ]; then
+    conda remove daal --force -y
+    source ${DAALROOT}/env/vars.sh
+fi
--- a/examples/run_examples.py
+++ b/examples/run_examples.py
@@ -99,6 +99,7 @@ def get_exe_cmd(ex, nodist, nostream):
             return None
         if not check_device(req_device["sycl/" + os.path.basename(ex)], availabe_devices):
             return None
+            
     if os.path.dirname(ex).endswith("examples"):
         if not check_version(req_version[os.path.basename(ex)], daal_version):
             return None
--- a/examples/run_examples.py
+++ b/examples/run_examples.py
@@ -44,9 +44,22 @@ try:
 except:
     sycl_available = False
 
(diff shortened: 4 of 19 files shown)


Alexander-Makaryev:
@fschlimb  please review, I worked on these changes so my review will be ambiguous

azure-pipelines.yml
```diff
@@ -1,4 +1,37 @@
...
+      sudo mv /opt/intel/inteloneapi/compiler/latest/linux/lib/oclfpga /opt/intel/inteloneapi/compiler/latest/linux/lib/oclfpga_
+    displayName: 'apt-get'
+  - script: conda create -q -y -n CB python=3.7 conda-build conda-verify
+    displayName: Create Anaconda environment
+  - script: |
+      export DPCPP_VAR=/opt/intel/inteloneapi/compiler/latest/env/vars.sh
+      export DAALROOT=/opt/intel/inteloneapi/daal/latest
+      . /usr/share/miniconda/etc/profile.d/conda.sh
+      conda activate CB
+      conda build --override-channels -c conda-forge -c intel conda-recipe
```
fschlimb: you want to prefer the conda-forge channel?

conda-recipe/run_test.sh
```diff
@@ -0,0 +1,12 @@
...
+# if dpc++ vars path is specified
+if [ ! -z "${DPCPP_VAR}" ]; then
+    source ${DPCPP_VAR}
+fi
+
+# if DAALROOT is specified
+if [ ! -z "${DAALROOT}" ]; then
+    conda remove daal --force -y
+    source ${DAALROOT}/env/vars.sh
+fi
```
fschlimb: What is this script supposed to do?

examples/run_examples.py
```diff
@@ -55,13 +56,19 @@
 req_version['saga_batch.py'] = (2019,3)
 req_version['dbscan_batch.py'] = (2019,5)
 req_version['lasso_regression_batch.py'] = (2019,5)
-req_version['elastic_net_batch.py'] = (2021,5)
+req_version['elastic_net_batch.py'] = (2022,0) # (2021,5) is relevant but there is an issue with versions number
+req_version['sycl/bf_knn_classification_batch.py'] = (2022,0) # there is an issue with versions number
+req_version['sycl/gradient_boosted_regression_batch.py'] = (2022,0) # there is an issue with versions number
```
fschlimb: what's the issue with the version numbers?

examples/run_examples.py
```diff
@@ -55,13 +56,19 @@
...
-    if req_version[os.path.basename(ex)] > daal_version:
-        return None
+    if os.path.dirname(ex).endswith("sycl"):
+        if not sycl_available:
+            return None
+        if req_version["sycl/" + os.path.basename(ex)] > daal_version:
+            return None
+    if os.path.dirname(ex).endswith("examples"):
+        if req_version[os.path.basename(ex)] > daal_version:
+            return None
```
fschlimb: why is this not checked for dirs within examples, like sycl?

azure-pipelines.yml
Alexander-Makaryev: yes, it is our considered decision

conda-recipe/run_test.sh
Alexander-Makaryev: This script runs at testing stage like build.sh runs at building stage. It enables DPC++ compiler runtime (paths) and replace daal from conda environment (package) with daal from DAALROOT. (if these env vars were specified and were used during the build stage)

examples/run_examples.py
Alexander-Makaryev: Our checks of versions suppose that algorithm or some behavior is available in DAAL of some version and above. But now this rule is not working properly. Another reason is that it is possible to get versions like (2021,103). We decided to discuss and rework this separately.

examples/run_examples.py
Alexander-Makaryev: There is the similar check for sycl:
if req_version["sycl/" + os.path.basename(ex)] > daal_version
I had to split checks on sycl and "classic" examples because they have the same names but different paths. Possible that it is not the best solution but I don't want to change all at the same time.

conda-recipe/run_test.sh
fschlimb: I don't get it. This script only sets env-vars. How are the tests run?

Alexander-Makaryev: Tests run as usual by commands in the test section in the meta.yaml file.

napetrov:
So i assume we are good to go and merge this to repo?

examples/run_examples.py
napetrov: We've discuss versions briefly. 
1. New features are introduced in both version and for example it can be enabled in 2020.1 and in 2021.Beta Update 5. so check logic should be updated.
2. It's interesting that we see 2021.105 versioning as it supposed to be 2021.1.0.5 and most importantly now we are at beta stage and this should be reflected in logic as  this is is true condition 2021.0 > 2021.105B

napetrov:
@fschlimb  @Alexander-Makaryev So i guess as this is not disruptive change it make sense to merge it? 

As for version problem - it's not just an examples problem for what particular version to souse, but overall way how daal4py obtains DAL version - for instance information about Beta or Product state is not collected. Also DAL use 3 parts in versioning while daal4py use just 2. 
@SmirnovEgorRu

fschlimb:
A quick glance at the changes didn't expose anything bad to me. @Alexander-Makaryev is the better person to give green light.

napetrov:
Thanks @fschlimb , so @Alexander-Makaryev  could you please merge this assuming we get review and approval from Frank and both of us are also ok with change.

examples/run_examples.py
napetrov: This have been fixed

conda-recipe/meta.yaml
```diff
@@ -43,8 +43,6 @@ test:
         - scikit-learn
         - mpich # [osx]
         - impi_rt # [not osx]
-    imports:
-        - daal4py
```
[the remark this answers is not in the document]
napetrov: What this would change? per my understanding current setup works in same way?

src/oneapi/oneapi.h
```diff
@@ -42,20 +42,13 @@ class PySyclExecutionContext
...
-        catch (cl::sycl::runtime_error const &e)
+    {  
+        if(dev == "gpu") m_ctxt = new daal::services::SyclExecutionContext(cl::sycl::queue(cl::sycl::gpu_selector()));
+        else if(dev == "cpu") m_ctxt = new daal::services::SyclExecutionContext(cl::sycl::queue(cl::sycl::cpu_selector()));
+        else if(dev == "host") m_ctxt = new daal::services::SyclExecutionContext(cl::sycl::queue(cl::sycl::host_selector()));
+        else
         {
-            std::cout << "Device \'" << dev << "\' was requested but runtime error occurs: " << e.what() << "\nFalling back to default device.\n";
+            std::cout << "Unknown device \'" << dev << "\' was specified.\nFalling back to default device.\n";
             m_ctxt = new daal::services::SyclExecutionContext(cl::sycl::queue(cl::sycl::default_selector()));
```
Alexander-Makaryev, line 52: You closed the issue https://github.com/IntelPython/daal4py/issues/171 but changes are not made.

conda-recipe/meta.yaml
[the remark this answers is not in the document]
Alexander-Makaryev: No, current setup worked (before merge of this PR) with check of importing before running other tests.
https://github.com/IntelPython/daal4py/blob/7334aeb6cf7cc53c25aaceb2f62c5041a6c4969a/conda-recipe/meta.yaml#L46
It can be significant if you will get fail in tests with MPI because the errors from MPI are often not informative so we will not understand that we simple failing during import.

PR #174, opened 2020-02-26
Fix performance degradation for Pandas input

PetrovKP:


generator/wrapper_gen.py
```diff
@@ -175,7 +175,7 @@ def my_procid():
 
 def _get_data(x):
     if isinstance(x, pdDataFrame):
-        x = [x.loc[:,i].values for i in x]
+        x = x.values
```
fschlimb: In most cases this will make a copy if the DF is not contiguous. This basically disables the SOA path.

Alexander-Makaryev: Here we should understand the memory layout of DataFrame.
We need to determine two significant for us cases appropriate to Homogen and SOA Numeric Tables and try to to except unnecessary copying of data.
Also we should pay attention to data types of columns. They can be as mix of ints and floats with various sizes but "aligned".
This is my current representation of it:
```
def _get_data(x):
    if isinstance(x, pdDataFrame):
        # get first row of Dataframe
        firstRow = df.iloc[0].values
        # try to detect c-contiguous representation of DF in memory
        if (firstRow.flags["C_CONTIGUOUS"] == True
            and firstRow.flags["OWNDATA"] == False):
            return x.values
        # np.ascontiguousarray will create a copy only when it is necessary
        # and will return the same memory otherwise
        x = [np.ascontiguousarray(x.iloc[:,i].values) for i in range(len(x.columns))]
    elif isinstance(x, pdSeries):
        x = [x.values]
    return x
```
This is a poor try to understand is the memory layout of DataFrame the c-contiguous array or not, it should be modified.

fschlimb: This looks better, but is not what this PR does :)

If a series is not contiguous, I would expect that all series of the DF are non-contiguous. Is this a fair assumption to make? If all series' are non-contiguous, you probably better just use x.values.

You also have to make the elif case contiguous.

Another solution would be to implement a NT which supports non-contiguous SOA. Basically a variant of daal4py's NpyTable. I'd rather go that route than implicitly copying. We could have a way to show performance warnings and let the user copy to contiguous explicitly.

Alexander-Makaryev:
@oleksandr-pavlyk This is PR with Pandas related changes

PetrovKP:
@fschlimb , @Alexander-Makaryev Thanks for the comments and observations!

The problem is the strong performance degradation sklearn & daal4py algorithms on pandas compared to "F" numpy format. After a little research, it turned out that the main hot spot is a function `_get_data`.  On the example of linear regression time *68 ms* on "F" numpy and *182 ms* on Pandas. A primitive solution would be to take the array directly. But you're right, in the case of heterogeneous data, we lose the feature of the SOA table. So, I propose a small dispatcher for homogeneous and heterogeneous cases. Also, I don’t understand why to use explicit copy (`iloc`) instead of getting an array (`values`). This is faster and should solve the problem of @Alexander-Makaryev .

|   |  `[x.loc[:,i].values for i in x]`  | `[series.values for _, series in x.items()]`  |  `x.values`  | 
|---|---|---|---|
| (140000, 1000)   |  117 ms  |  4.3 ms  | 0.1 ms  |
| (10000, 10000)   |  750.89 ms | 19.32 ms | 0.1 ms |

If I miss any cases or details, please tell me, I will fix it. But this degradation must be corrected in any case. This is a strong affect performance.

fschlimb:
Interesting. What are these numbers measuring? Only the conversion? I am surprised by the huge performance difference between the first and the second. I had seem similar performance advantages of converting the entire df into a contiguous array (by making a copy).

Still, we should try to avoid making copies, we can always tell the user to make a copy explicitly. We want this to work with huge data-sets. That's why we should not return df.value unless df itself is contiguous.

As the first and the second both may return a copy or a view, the second indeed seems preferable.

Have you also tried `[series.to_numpy() for _, series in x.items()]` (which is really what we want)?

generator/wrapper_gen.py
```diff
@@ -175,7 +175,10 @@ def my_procid():
 
 def _get_data(x):
     if isinstance(x, pdDataFrame):
-        x = [x.loc[:,i].values for i in x]
+        if x._is_homogeneous_type:
```
fschlimb: This seems be deprecated with pandas 1.1.0: https://github.com/pandas-dev/pandas/pull/30712
We should not rely on a deprecated and undocumented feature.

PetrovKP:
@fschlimb 

> What are these numbers measuring? Only the conversion? 

Yes, only them.

> That's why we should not return df.value unless df itself is contiguous.

I tried to see out in what cases the pandas has non-contiguous data. And in the process of work a pandas , I did not see where this case could arise. If you know such a case, can you share it?

> Have you also tried [series.to_numpy() for _, series in x.items()] (which is really what we want)?

There is no serious difference between them. But getting numpy array looks preferable.

> Don't use _is_homogeneous_type. This seems be deprecated with pandas 1.1.0. We should not rely on a deprecated and undocumented feature.

This removed Indexes and Series, but not for the DataFrames. But you're probably right about the undeclared function. I will try to create a issue in the pandas repository, because without this functionality we lose performance (~0.7-1.0 ms).

generator/wrapper_gen.py
```diff
@@ -175,12 +175,13 @@ def my_procid():
 
 def _get_data(x):
     if isinstance(x, pdDataFrame):
-        if x._is_homogeneous_type:
-            x = x.values
+        x_dtypes = x.dtypes
+        if np.all(x_dtypes == x_dtypes[0]):
+            x = x.to_numpy()
```
fschlimb: We need to check contiguity; homogeneity is only a required but not sufficient attribute.

fschlimb:
@Alexander-Makaryev can you show us the example of non-contiguous series that you mentioned last week?

Alexander-Makaryev:
We had a discussion with Kirill, we shared our minds, and I continue investigation.
While I investigated non-contiguous data I met such behavior:
```
import numpy as np
import pandas as pd
from pandas import DataFrame as pdDataFrame

row = 2
col = 2

def _get_data(x):
    if isinstance(x, pdDataFrame):
        x_dtypes = x.dtypes
        if np.all(x_dtypes == x_dtypes[0]):
            x = x.to_numpy()
        else:
            x = [xi.to_numpy() for _, xi in x.items()]
    elif isinstance(x, pdSeries):
        x = [x.to_numpy()]
    return x

na1 = np.random.randn(row, col)
na2 = np.random.randn(row, col)
df1 = pd.DataFrame(na1)
df2 = pd.DataFrame(np.asfortranarray(na2))
df3 = pd.concat([df2, df1], axis=1)
data = _get_data(df3)
```
```
  File "numpy2pandas_float64.py", line 63, in <module>
    data = _get_data(df3)
  File "numpy2pandas_float64.py", line 11, in _get_data
    if np.all(x_dtypes == x_dtypes[0]):
  File "/localdisk/work/amakarye/miniconda3/envs/d4p_dev/lib/python3.7/site-packages/pandas/core/ops/common.py", line 64, in new_method
    return method(self, other)
  File "/localdisk/work/amakarye/miniconda3/envs/d4p_dev/lib/python3.7/site-packages/pandas/core/ops/__init__.py", line 521, in wrapper
    raise ValueError("Can only compare identically-labeled Series objects")
ValueError: Can only compare identically-labeled Series objects
```

Alexander-Makaryev:
@fschlimb 
> @Alexander-Makaryev can you show us the example of non-contiguous series that you mentioned last week?

Everything below is related to the latest version of code from Kirill:
```
def _get_data(x):
    if isinstance(x, pdDataFrame):
        x_dtypes = x.dtypes
        if np.all(x_dtypes == x_dtypes[0]):
            x = x.to_numpy()
        else:
            x = [xi.to_numpy() for _, xi in x.items()]
    elif isinstance(x, pdSeries):
        x = [x.to_numpy()]
    return x
```
At this moment I can get columns with strides only when I am using c-contiguous array. This case is covered by condition with checking of homogeneity.
When we understand that data is homogeneus then we return c-contigous or f-contigous array that have a specialized processing.

I didn't find the good description in official documentation but according to this article:
https://www.dataquest.io/blog/pandas-big-data/

Pandas stores float values for all columns as one contiguous block of memory.
So if we will try to create a mix of various float columns then it will be copied to one contiguous memory block at the moment of creating of DataFrame in any case.

We still have a case when data contain several columns in floats and several in ints:
We have to copy integers in any case, because DAAL takes only floats or doubles as input.
If float data is in fortran-style then it should be cheaper to copy the rest integers into float representation and provide SOA Table to algorithm.
But if we have float data in c-style then we have to copy it too. In this case possible that will be better to create c-contiguous (Homogen, not SOA) array with all values because usually we have better performance with them.
On another hand, when user chose DataFrames to store the data, he expects that it will be columnar format.
It is only my vision, condition for making decision may be more complicated. (for example based on amount of data to be copied in each case)

generator/wrapper_gen.py
```diff
@@ -175,11 +175,11 @@ def my_procid():
 
 def _get_data(x):
     if isinstance(x, pdDataFrame):
-        x_dtypes = x.dtypes
+        x_dtypes = x.dtypes.values
         if np.all(x_dtypes == x_dtypes[0]):
             x = x.to_numpy()
         else:
-            x = [xi.to_numpy() for _, xi in x.items()]
+            x = [np.asfortranarray(xi.values) for _, xi in x.items()]
```
fschlimb: Back to values? Fortran?

PetrovKP: We have difficulties with this case. if you take on separate columns the data may be non-contiguous. So far the only way that has turned out is to use the function `np.asfortranarray`. I’m still looking at what can be done so as not to degrade and handle the whole case correctly. If there are ideas and suggestions - I will be very grateful.

fschlimb: What about np.ascontiguousarray(xi.to_numpy)`?

PetrovKP: With `to_numpy` we have the result is slower a bit in this case ~2-3 ms.

fschlimb: The user might be be able to avoid/change the non-homogen data-layout without doing a copy. Implicitly copying the DF with then make it impossible to process such data. 
If we support fragmented arrays without copies it might be slow but we would not lose support for larger datasets. The user can always decide to create a contiguous array if there is enough memory for small cases.

In other words, I suggest always preferring in-place operation if possible. Not making copies is the only reason why we actually provide support for DFs in addition to numpy arrays.

oleksandr-pavlyk:
The `make_nt` needs to be prepared to handle non-unit strided columns that may come from dataframe. 

It would be ideal if we could recognize the dataframe whose columns are views of a F- or C- contiguous numpy array.

@Alexander-Makaryev 's suggestion to use `iloc` rather than `loc` is good.

generator/wrapper_gen.py
PetrovKP: I wrote some example that covers all interesting cases with a data frame. Ok, for simple case we have numpy array and we created data frame from array.

## 1 case. homogen & сontiguous data 
```python
fptype = np.float32
row = 10**5
col = 10**3
na1 = np.random.randn(row, col)
X_df = pd.DataFrame(na1, dtype=fptype)
```
Let's see what it is
```python
ordata = X_df._data.blocks[0].get_values()

print("1. Create DF. Data n blocks {}. Block[0]:  OWNDATA: {:2}; C_CONT: {:2}; F_CONT: {:2}; strides ({} {})".
    format(len(X_df._data.blocks), ordata.flags['OWNDATA'], ordata.flags['C_CONTIGUOUS'],
            ordata.flags['F_CONTIGUOUS'], ordata.strides[0],
            ordata.strides[1]))
```
Output:
```
1. Create DF. Data n blocks 1. Block[0]:  OWNDATA:  0; C_CONT:  0; F_CONT:  1; strides (4 4000)
```

1 case inside does not owner the data itself, uses the interface through the view. The data inside is displayed in fortran format. In this case, if we get values for DataFrame. We will get a view of this data.

```python
vdf = X_df.values
print("1.1. Get values. OWNDATA: {:2}; C_CONT: {:2}; F_CONT: {:2}; strides ({} {})"
    .format(vdf.flags['OWNDATA'], vdf.flags['C_CONTIGUOUS'],
            vdf.flags['F_CONTIGUOUS'], vdf.strides[0],
            vdf.strides[1]))
```
Output:
```
1.1. Get values.  OWNDATA:  0; C_CONT:  1; F_CONT:  0; strides (4000 4)
```
There is no copy. On the daal4pu side, will call `_make_hnt` for creating `HomogenTable`. This is the fastest and most free way to work with data! If it is created from a fortran numpy format, it will be created `SOATable`. 

## 2 case. homogen & non-contiguous data 

```python
X_df = X_df.iloc[2 : row - 1, 2 : col - 2]
```
What is going on inside:
```python
ordata = X_df._data.blocks[0].get_values()

print("2. Get non-contiguous DF. Data n blocks {}. Block[0]: OWNDATA: {:2}; C_CONT: {:2}; F_CONT: {:2}; strides ({} {})".
    format(len(X_df._data.blocks), ordata.flags['OWNDATA'], ordata.flags['C_CONTIGUOUS'],
            ordata.flags['F_CONTIGUOUS'], ordata.strides[0],
            ordata.strides[1]))
```
Output:
```
2. Get non-contiguous DF. Data n blocks 1. Block[0]: OWNDATA:  0; C_CONT:  0; F_CONT:  0; strides (4 4000)
```
We still do not own the data. It can be seen that the data is now not contiguous. What happens if in this case we take the data?

```python
vdf = X_df.values
print("2.1. Get values. OWNDATA: {:2}; C_CONT: {:2}; F_CONT: {:2}; strides ({} {})"
    .format(vdf.flags['OWNDATA'], vdf.flags['C_CONTIGUOUS'],
            vdf.flags['F_CONTIGUOUS'], vdf.strides[0],
            vdf.strides[1]))
```
Output:
```
2.1. Get values.  OWNDATA:  0; C_CONT:  0; F_CONT:  0; strides (4000 4)
```
Data in this case will not be owner. Those. In this case, **copying does not occur**. Moreover, data not-contiguous is preserved. So in this case it will be created `NpyNumericTable<NpyNonContigHandler>`. Which will lead to the correct result.  **Please note that now this case does not work correctly!**

## 3 case. heterogeneous & contiguous data 

```python
X_df = X_df.astype({X_df.columns[1]: 'float64', X_df.columns[6]: 'float64'})
X_df[X_df.columns[1]] = X_df[X_df.columns[5]] * 3
```

```python
ordata = X_df._data.blocks[0].get_values()
print("3. Make geterogenious DF. Data n blocks {}. Block[0]: OWNDATA: {:2}; C_CONT: {:2}; F_CONT: {:2}; strides ({} {})". format(len(X_df._data.blocks), ordata.flags['OWNDATA'], ordata.flags['C_CONTIGUOUS'],
            ordata.flags['F_CONTIGUOUS'], ordata.strides[0],
            ordata.strides[1]))
```
Output:
```
3. Make geterogenious DF. Data n blocks 3. Block[0]: OWNDATA:  1; C_CONT:  1; F_CONT:  0; strides (399988 4)
```
Somewhere inside there was a copy and now the data frame owns the data. With this data contiguous we have got. In this case, the valid list creation code `[xi.values for _, xi in x.items()]`. You can leave it. But there is potential to improve this. In daal4pu we created `SOA Table`

## 4 case. heterogeneous & not-contiguous data 

I could not reproduce this case. But in theory, it should work as case 3, only after converting it to fortran format. But I'm not sure if this case is possible at all?

## Summary

> The user might not be able to avoid/change the non-homogen data-layout without doing a copy. Implicitly copying the DF will then make it impossible to process such data.

In no case will data be copied. The pandas provides us is always a view of data. 1 case is the most popular in our cases and benchmarks. therefore, we would not want to have an overhead in this place. I agree that the rest cases are an important. Some do not work now. My current changes should solve these problems.

fschlimb: Have you verified with looking at addresses if case 3 is really true? With the data you show I cannot verify the claim that there is no copy made.
I might just missing your point since I think, `X_df._data.blocks[0].get_values()`  should be `X_df[X_df.columns[1]]`.

PetrovKP: @fschlimb Yes, I checked.

Yes, I checked. When the data is homogeneous, they are stored in a single manager block. `X_df._data.blocks[0].get_values()` This is just an internal function of direct access to the memory of this block. `len(X_df._data.blocks)` - the number of such blocks. When we do a column addition or type change, new blocks will added. But at certain points, they are merged by type.
```python
X_df._data.blocks[0].get_values().data
```
Output:
```
<memory at 0x7f47c61928b8>
```
For get values:
```python
X_df.values.data
```
Output:
```
<memory at 0x7f47c61928b8>
```
Moreover, if copying would occur, we would get an contiguous array as a result. In this case, we get **not-contiguous**.

PetrovKP:
@Alexander-Makaryev , @fschlimb @oleksandr-pavlyk 

I checked the correctness of working with non-contiguous memory on example dpscan.

1. case numpy
```python
data = data[rpp*d4p.my_procid():rpp*d4p.my_procid()+rpp,:]
```
2. case pandas
```python
data = pd.DataFrame(data[rpp*d4p.my_procid():rpp*d4p.my_procid()+rpp,:])
```
3. case heterogeneous pandas
```python
data = pd.DataFrame(data[rpp*d4p.my_procid():rpp*d4p.my_procid()+rpp,:])
data = data.astype({data.columns[1]: 'float32'})
```

All cases **work correctly*** and print one result. The current implementation does not work correctly in cases 2 and 3.

```
Results on node with id =  0  :

First 10 cluster assignments:
 [[0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]]
Number of clusters:
 [[6]]

Results on node with id =  1  :

First 10 cluster assignments:
 [[3]
 [3]
 [3]
 [3]
 [3]
 [3]
 [3]
 [3]
 [3]
 [3]]
Number of clusters:
 [[6]]

Results on node with id =  2  :

First 10 cluster assignments:
 [[0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]]
Number of clusters:
 [[6]]

Results on node with id =  3  :

First 10 cluster assignments:
 [[3]
 [3]
 [3]
 [3]
 [3]
 [3]
 [3]
 [3]
 [3]
 [3]]
Number of clusters:
 [[6]]
```

Alexander-Makaryev:
@averbukh
That is PR where a lot of info about pandas memory layout.

oleksandr-pavlyk:
Please add the following test file to the test suite, to ensure that whatever solution is worked out passes the test:

```python
# file test_df_support.py
import numpy as np
import pandas as pd
import daal4py as d4p


def verify_on_dbscan(X):
    alg1 = d4p.dbscan(epsilon=2.0, minObservations=5, fptype='double')
    res1 = alg1.compute(X)
    Xc = np.ascontiguousarray(X)
    alg2 = d4p.dbscan(epsilon=2.0, minObservations=5, fptype='double')
    res2 = alg2.compute(Xc)
    assert np.array_equal(res1.assignments, res2.assignments)
    assert len(np.unique(res1.assignments)) > 2

def test1():
    """
    Dataframe from C-contiguous array
    """
    X = np.random.randn(13024, 16)
    df = pd.DataFrame(X)
    verify_on_dbscan(df)

def test2():
    """
    Dataframe from F-contiguous array
    """
    X = np.asfortranarray(np.random.randn(13024, 16))
    df = pd.DataFrame(X)
    verify_on_dbscan(df)

def test3():
    """
    Dataframe from non-contiguous array, case 1
    """
    X = np.random.randn(13024, 5, 16)
    df = pd.DataFrame(X[:, 2, :])
    verify_on_dbscan(df)


def test4():
    """
    Dataframe from non-contiguous array, case 2
    """
    X = np.random.randn(13024*3, 16)
    df = pd.DataFrame(X[1::3, :])
    verify_on_dbscan(df)
```

I am running it with `pytest test_df_support.py -ra`.

I am currently getting:

```
===================================================================== short test summary info ======================================================================
FAILED test_df_support.py::test1 - assert False
FAILED test_df_support.py::test3 - assert False
FAILED test_df_support.py::test4 - assert False
3 failed, 1 passed in 1.60s
```

PetrovKP:
> Please add the following test file to the test suite, to ensure that whatever solution is worked out passes the test

I added. All tests pass successfully
```
==================== 4 passed in 1.33s =========================
```

oleksandr-pavlyk:
Please also add another test where dataframe is created from a non-homogeneous dtype NumPy array. I expect this test to pass too (it passes in master). 



```
def test5():
    """
    Dataframe from multi-dtype array
    """
    X = np.random.randn(13024, 16)
    dt = []
    for i in range(8):
        dt += [('f' + str(2*i), np.float64),
               ('f' + str(2*i+1), np.float32)]
    X2 = np.empty((X.shape[0],), dtype=np.dtype(dt))
    for i, (n, _) in enumerate(dt):
        X2[n][:] = X[:, i]

    df = pd.DataFrame(X2)
    for i, (n, _) in enumerate(dt):
        assert np.allclose(df[n].values, X[:, i])
    verify_on_dbscan(df)
```

PetrovKP:
@oleksandr-pavlyk 
Added by. All tests successfully passed.

PetrovKP:
@fschlimb @Alexander-Makaryev @oleksandr-pavlyk  Are there any concerns about merge this request to the master?

fschlimb:
Have you verified that the use of `to_numpy()` does not make a copy for any of the examples?
Is it possible to add/extend test to check that no copies were made?

PetrovKP:
@fschlimb Yes. I checked it. I added a test

generator/wrapper_gen.py
```diff
@@ -175,9 +175,13 @@ def my_procid():
...
     if isinstance(x, pdDataFrame):
-        x = [x.loc[:,i].values for i in x]
+        x_dtypes = x.dtypes.values
+        if np.all(x_dtypes == x_dtypes[0]):
+            x = x.to_numpy()
+        else:
+            x = [xi.to_numpy() for _, xi in x.items()]
     elif isinstance(x, pdSeries):
-        x = [x.values]
+        x = [x.to_numpy()]
```
oleksandr-pavlyk: I do not see how this solves the problem. 

```
X = np.random.randn(10,3)
ps = pd.Series(X[:, 0])
```

and now

```
In [57]: ps.to_numpy().flags
Out[57]:
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

```

and we hit the trouble-some branch, which expects C-contiguous array, but gets a non-contiguous one.

The fix must implement handling of non-contiguous arrays in `make_nt`.

generator/wrapper_gen.py
```diff
@@ -181,7 +181,7 @@ def _get_data(x):
         else:
             x = [xi.to_numpy() for _, xi in x.items()]
     elif isinstance(x, pdSeries):
-        x = [x.to_numpy()]
+        x = x.to_numpy()
```
fschlimb: Hm, does this return a 1d or 2d array?

generator/wrapper_gen.py
fschlimb: This was exactly my suspicion. 
A short-term fix would make a copy. A proper solution would be able to handle SOA of non-contiguous arrays. We currently only support homogen non-contiguous data.

fschlimb: And yes, this will most likely degrade performance over just making a copy. See above argument on pros and cons of making copies vs operating in-place.

generator/wrapper_gen.py
oleksandr-pavlyk: 1d array:

```
In [60]: ps = pd.Series(X[:, 0])

In [61]: X[:,0].shape
Out[61]: (10,)

In [62]: ps.to_numpy().shape
Out[62]: (10,)
```

generator/wrapper_gen.py
PetrovKP: @fschlimb @oleksandr-pavlyk  Thank you for your comment! Fixed, the current behavior will go to the `make_nt` in the case non-contiguous arrays.

generator/wrapper_gen.py
PetrovKP: Thanks, fixed! Also added tests for this case.

generator/wrapper_gen.py
```diff
@@ -175,9 +175,13 @@ def my_procid():
...
     if isinstance(x, pdDataFrame):
-        x = [x.loc[:,i].values for i in x]
+        x_dtypes = x.dtypes.values
+        if np.all(x_dtypes == x_dtypes[0]):
+            x = x.to_numpy()
+        else:
+            x = [xi.to_numpy() for _, xi in x.items()]
     elif isinstance(x, pdSeries):
-        x = [x.values]
+        x = x.to_numpy().reshape(-1, 1)
```
oleksandr-pavlyk: :+1:

generator/wrapper_gen.py
```diff
@@ -175,9 +175,13 @@ def my_procid():
 
 def _get_data(x):
     if isinstance(x, pdDataFrame):
-        x = [x.loc[:,i].values for i in x]
+        x_dtypes = x.dtypes.values
+        if np.all(x_dtypes == x_dtypes[0]):
+            x = x.to_numpy()
+        else:
+            x = [xi.to_numpy() for _, xi in x.items()]
```
oleksandr-pavlyk: I would still use `np.ascontiguousarray(xi.to_numpy())` just for safety sake. 

It won't make a copy if the array is already contiguous, but will safe-guard the code against possible future changes in pandas.

PetrovKP: For some reason, there is an overhead for calling this function. But it seems that DataFrames do not allow such a case. I added three tests and the results are correct. Here is one example:

```python
        X = np.random.randn(13024*3, 16)
        df = pd.DataFrame(X[1::3, 1:])
        k = 2
        dir_dtypes = {col: 'float64' if i % 2 == 0 else 'float32' for i, col in enumerate(df.columns)}
        df = df.astype(dir_dtypes)
        ps = pd.Series(X[1::3, 0])
        df['newcol'] = ps
        self.verify_on_dbscan(df)
```

fschlimb: I strongly recommend introducing a (cheap) check which raises an error if we don't meet what we expect. Changes that introduce a copy should not be able to pass unnoticed.

PetrovKP: Add check in function `_get_data`?

fschlimb: Instead of `np.ascontiguousarray(xi.to_numpy())` I suggest adding the check and then using just `xi.to_numpy()`.

oleksandr-pavlyk: The `np.ascontiguous` is about 10 times slower than checking flags:

```
In [37]: x = np.random.randn(10**5, 50)

In [38]: %timeit y = np.ascontiguousarray(x)
814 ns ± 7.84 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [39]: %timeit y2 = x
17.8 ns ± 0.062 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [43]: %timeit x.flags['C']
76.9 ns ± 2.06 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Raising an error is not very user friendly, but if we decide on it, it is most performant to do in `make_nt` where we can do it in C.

fschlimb: Emitting a mutable warning (that a copy was made) would also be acceptable to me.

PetrovKP: OK. I added the following check to the code. Look please.

```c++

    else if (PyList_Check(obj) && PyList_Size(obj) > 0)
        { // a list of arrays for SOA?
               ....

                for (auto i = 0; i < N; i++)
                {
                    PyArrayObject * ary = (PyArrayObject *)PyList_GetItem(obj, i);
                   ....

                    if (!(array_is_behaved(ary) || array_is_behaved_F(ary)))
                    {
                        std::cerr << "Can't get the data correctly without a copy. Errors are possible.\n";
                    }

```

p.s. Sorry for the big changes in cpp. They were made automatically by the clang-format.

oleksandr-pavlyk: The `*ary`  should be a 1D array, and so `array_is_behaves(ary)` should be sufficient.

When an automatic tool makes changes like that, it is best if automatic changes are separated from meaningful changes into a separate commit.

This makes it easy to opt-out of these changes if needed.

PetrovKP: @oleksandr-pavlyk Yes. You're right. I fixed.

src/daal4py.cpp
```diff
@@ -378,6 +378,11 @@ daal::data_management::NumericTablePtr make_nt(PyObject * obj)
                         break;
                     }
 
+                    if (!array_is_behaved(ary))
+                    {
+                        std::cerr << "Can't get the data correctly without a copy.\n";
+                    }
+
```
fschlimb: This check must either raise an exception (e.g. break) or print a warning *and* auto-convert. It's not a good idea to proceed if we know it will most likely produce a seg-fault.

Can't we do this in the generated cython code?

fschlimb: The message could be more discriptive since we want to tell the user that he can work-around this (by making a contiguous copy).
Maybe "Cannot operate on column x because it is non-contiguous. Please make it contiguous before passing it to daal4py."
Or "Warning: implicitly created a copy of column x. Make it contiguous before passing it to daal4py to avoid potential negative performance implications."
If a warning and implicit conversion, the user needs to be able to switch the message of (not the conversion).

PR #181, opened 2020-03-05
GBT Scikit-Learn-compatible estimators

Alexander-Makaryev:


Alexander-Makaryev:
We can't use _validate_data function yet. It was added only at 29 Feb and is not available in current releases of Scikit-Learn.
https://github.com/scikit-learn/scikit-learn/commit/d205638475ca542dc46862652e3bb0be663a8eac#diff-f392a01b355b04b8c90c84175d118fdc

Alexander-Makaryev:
For some reason function check_is_fitted is not working properly with daal_models, so we check it separately.

daal4py/sklearn/ensemble/GBTDAAL.py
```diff
@@ -0,0 +1,306 @@
...
+from sklearn import preprocessing
+from sklearn.utils.multiclass import check_classification_targets
+from sklearn.utils import check_random_state
+import daal4py as d4p
+from ..utils import getFPType
+
+from sklearn import __version__ as sklearn_version
+from distutils.version import LooseVersion
+
+daal_version = tuple(map(int, (d4p.__daal_link_version__[0:4], d4p.__daal_link_version__[4:8])))
```
oleksandr-pavlyk: Perhaps we should use `d4py.__daal_run_version__` instead?

Alexander-Makaryev: Yes, you are right. Thanks!

PR #185, opened 2020-03-13
Feature/sklearn assert all finite

Alexander-Makaryev:


src/daal4py.cpp
```diff
@@ -641,3 +641,22 @@ size_t c_my_procid()
 #endif
 }
 } // extern "C"
+
+extern "C" {
```
fschlimb: Could be removed.

PR #189, opened 2020-03-21
added generating of get_result function

Alexander-Makaryev:
Getting of result without explicit call of compute method is necessary when algorithm is passed as parameter and used inside another algorithm.

### DIFF
--- a/generator/wrapper_gen.py
+++ b/generator/wrapper_gen.py
@@ -654,6 +654,9 @@ struct {{algo}}__iface__ : public {{prnt}}
     virtual {{result_map.class_type}} * compute({{input_args|fmt('{}', 'decl_cpp', sep=',\n')|indent(indent)}},
 {{' '*indent}}bool setup_only = false)
         {assert(false); return NULL;}
+{% if add_get_result %}
+    virtual {{result_map.class_type}} * get_result() {assert(false); return NULL;}
+{% endif %}
 {% if streaming.name %}
     virtual {{result_map.class_type}} * finalize() {assert(false); return NULL;}
 {% endif %}
@@ -753,6 +756,7 @@ private:
     }
 
 {% endif %}
+
 {% if step_specs is defined and distributed.name %}
     // Distributed computing
 public:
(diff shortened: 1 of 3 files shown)


fschlimb:
@Alexander-Makaryev Could you give an example?

Alexander-Makaryev:
@agorshk Please update your PR and provide the link to the place with using of get_result

agorshk:
An example of usage this feature might be found in PR https://github.com/IntelPython/daal4py/pull/184 in file _coordinate_descent_0_22.py

fschlimb:
Such a call conflicts with the general design of the API. I'd rather provide ways to call compute so that it returns the result of an indirect use. Or better yet, the caller should have the indirect result in its own result. 
👎

oleksandr-pavlyk:
@fschlimb I am reading your feedback as a signal that `get_result` should not become a documented method of any algorithm.

However we have pressing need to access the number of iterations performed by an iterative solver used by other ML solvers such as `LogisticRegression`, `LASSO` or `ElasticNet` to be able to optimize `sklearn.linear_model.ElasticNet`. 

Would be acceptable to demote `get_result` to a private method, say `__get_result__` to be removed once the proper implementation is in place?

generator/wrapper_gen.py
```diff
@@ -872,40 +872,34 @@ def __cinit__(self,
     # compute simply forwards to the C++ de-templatized manager__iface__::compute
     def _compute(self,
                  {{input_args|fmt('{}', 'decl_dflt_cy', sep=',\n')|indent(17)}},
-                 setup=False):
+                 setup=False{% if add_get_result %}, get_precomputed_result=False{% endif %}):
```
fschlimb: This breaks the HPAT/SDC support.

generator/wrapper_gen.py
```diff
@@ -872,40 +872,34 @@ def __cinit__(self,
...
+        {{input_args|fmt('{}', 'get_obj', sep='\n')|indent(8)}}
         res.c_ptr = deref(algo).compute({{input_args|fmt('{}', 'arg_cyext', sep=',\n')|indent(40)}},
                                         setup)
         return res
 
     # compute simply forwards to the C++ de-templatized manager__iface__::compute
     def compute(self,
                 {{input_args|fmt('{}', 'decl_dflt_cy', sep=',\n')|indent(16)}},
-                setup=False):
+                setup=False{% if add_get_result %}, get_precomputed_result=False{% endif %}):
```
fschlimb: see above

fschlimb:
Yes, I think it would be acceptable to temporarily do one of two things
1. add a private function as @oleksandr-pavlyk suggested
2. treat all `None` input to `compute` as `getResult`

I don't think adding a new argument to compute is a good idea.

PR #192, opened 2020-04-02
Modify parser for code after comments

Pahandrovich:


generator/parse.py
```diff
@@ -117,10 +117,11 @@ def enum(**enums):
 class comment_parser(object):
     """parse documentation in comments"""
     def parse(self, l, ctxt):
-        assert not re.match(r'.*\*/(.+)', l), "Found the code after closed comment in the same line"
```
Alexander-Makaryev, line 120: I think we can leave the assert after deleting of comments here too. It will helps us to detect that these changes really work an we will not miss something in future.

Pahandrovich:
@fschlimb , can you review the code, please?

Pahandrovich:
Thank you! The same error in Circle СI is in the master

PR #196, opened 2020-04-08
check of linked DAAL version during assert_all_finite monkeypatching

Alexander-Makaryev:


oleksandr-pavlyk:
The commit is not green, because CI builds the master with the PR merged into it.

PR #200, opened 2020-04-10
Add verbose mode in scikit-learn algorithms

Pahandrovich:
Verbose mod is turned on by a variable DAAL4PY_SKLEARN_VERBOSE
For kmeans print why we going to sklearn


Pahandrovich:
@oleksandr-pavlyk, can you look at the implementation of this verbose mode?
Support for this will be added for all algorithms.
The main thing is to inform the user that the algorithm uses scikit-learn instead of a daal.

oleksandr-pavlyk:
I would suggest to use existing Python mechanism, https://docs.python.org/3/howto/logging.html, instead of using print. 

The level of logging can be controlled via python  CL option.

Pahandrovich:
If we will controlled log level in CL option, then user must implement handler of log level?

daal4py/sklearn/cluster/_k_means_0_21.py
```diff
@@ -201,18 +201,31 @@ def fit(self, X, y=None, sample_weight=None):
...
         if daal_ready and sample_weight is not None:
             sample_weight = np.asarray(sample_weight)
             daal_ready = (sample_weight.shape[0] == X_len) and (
                          np.allclose(sample_weight, np.ones_like(sample_weight)))
+            if verbose() and not daal_ready:
+                text = text + "sample_weight.shape[0] != X_len or not (np.allclose(sample_weight, np.ones_like(sample_weight)))\n"
 
     if not daal_ready:
+        if verbose():
+            print("kmeans: go_to_sklearn")
```
SmirnovEgorRu, line 227: Wording should be improved completely :)

SmirnovEgorRu:
@Pahandrovich, no - the changes on library level only, not user.
But in this case users will have additional mechanism to disable/enable them and other actions.

Pahandrovich:
Now the verbose mod is turned on by the environment variable IDP_SKLEARN_VERBOSE=INFO
User may use command 
IDP_SKLEARN_VERBOSE=INFO python -m daal4py sample.py

daal4py/sklearn/cluster/_k_means_0_21.py
```diff
@@ -213,6 +214,7 @@ def fit(self, X, y=None, sample_weight=None):
                          np.allclose(sample_weight, np.ones_like(sample_weight)))
 
     if not daal_ready:
+        logging.info("kmeans:fit of sklearn")
```
SmirnovEgorRu, line 217: I think we need to improve wording. I offer to use this format:
1) sklearn.cluster.KMeans.fit: using original Scikit-learn...
2) sklearn.cluster.KMeans.fit: using Intel(R) DAAL...

but technically - LGTM too

Vika-F: Agree with Egor. I also vote for creating a set of strings somewhere in common part of daal4py. Maybe in utils.py. The set of strings will look like:
    fit_method_uses_daal = “uses Intel® Data Analytics Acceleration Library solver”
    fit_method_uses_sklearn = “uses original Scikit-learn solver”
After that we can use those strings to generate log messages.
This will prevent us from the possible error when one log message has one wording an the another message - another wording.
It would also help to change the wording faster when needed,

SmirnovEgorRu: Agree.
Only one moment - I prefer to use ```Intel® DAAL``` instead of ```Intel® Data Analytics Acceleration Library```. Just because it's shorter.

Pahandrovich:
@PetrovKP @Alexander-Makaryev @SmirnovEgorRu, Can you review this PR?

daal4py/sklearn/linear_model/_ridge_0_22.py
```diff
@@ -107,15 +108,18 @@ def fit(self, X, y, sample_weight=None):
...
     else:
         self.n_iter_ = None
         res = _daal4py_fit(self, X, y)
         if res is None:
+            logging.info("sklearn.linear_model.Ridge.fit: " + fit_method_uses_sklearn)
             if hasattr(self, 'daal_model_'):
                 del self.daal_model_
             return super(Ridge, self).fit(X, y, sample_weight=sample_weight)
         else:
+            logging.info("sklearn.linear_model.Ridge.fit: " + fit_method_uses_daal)
```
PetrovKP, line 122: I see that in some cases we print information after DAAL called, and in some cases  - before. I understand why in some cases we can print information only after. But I think it would be better if we did it the same way - everywhere. For example, after a call DAAL. What do you think about it?

SmirnovEgorRu: I think in case of when we go into daal4py, but can't do smth during the running - we need to have 2 prints:
```
"sklearn.linear_model.Ridge.fit: uses Intel® DAAL solver")
"sklearn.linear_model.Ridge.fit: uses original Scikit-learn solver, because the task can't be solved with Intel® DAAL"
```

Pahandrovich: I always print log before call daal4py/sklearn function, but this cases is an exception. Here we are forced to print after training daal4py in the condition.

PetrovKP: It much better 👍

daal4py/__init__.py
```diff
@@ -7,3 +7,14 @@
         raise ImportError(s + '\n\nActivating your conda environment or sourcing mpivars.[c]sh/psxevars.[c]sh may solve the issue.\n')
     else:
         raise
+
+import logging
+import warnings
+import os
+l = os.environ.get("IDP_SKLEARN_VERBOSE")
```
PetrovKP, line 14: minnor: `l` is not good name 😃

daal4py/sklearn/cluster/_k_means_0_23.py
```diff
@@ -270,8 +273,10 @@ def predict(self, X, sample_weight=None):
     daal_ready = sample_weight is None and hasattr(X, '__array__') # or sp.isspmatrix_csr(X)
 
     if daal_ready:
+        logging.info("sklearn.cluster.KMeans.predict: " + fit_method_uses_daal)
```
SmirnovEgorRu, line 276: Minor, but you use "**fit**_method_uses_daal" variable for predict method.

PR #201, opened 2020-04-10
Implementation of decision tree classifier

oleksandr-pavlyk:
Added decision tree classifier.


oleksandr-pavlyk:
All local tests passed, including test against 0.22.2 scikit-learn (patched test vs. unpatched test).

PR #203, opened 2020-04-16
Added dbscan sycl example; reduced size of dbscan dataset

amgrigoriev:


Alexander-Makaryev:
@amgrigoriev Why my previous PR is not used? https://github.com/IntelPython/daal4py/pull/202
We talked about it and I sent the link yesterday.
In any case, I think that comparing of results of execution (with different devices too) in example is helpful. We can expand it like here https://github.com/IntelPython/daal4py/blob/feature/sycl-dbscan/examples/sycl/dbscan_batch.py#L78.
Examples are part of our tests, so it will check that sycl version works and provides appropriate results.

Alexander-Makaryev:
I have an issue with execution time even on this small dataset
```
from daal4py.oneapi import sycl_context
import daal4py as d4p
import numpy as np

X = np.array([[1, 2], [2, 2], [2, 3],
              [8, 7], [8, 8], [25, 80]])
with sycl_context('gpu'):
    clustering = d4p.dbscan(epsilon=3, minObservations=2,
                            resultsToCompute="computeCoreIndices",
                            memorySavingMode=1).compute(X)
    print("clustering.labels_")
    print(clustering.labels_)
    print("clustering")
    print(clustering)
```
Looks like it hangs somewhere inside.

examples/sycl/dbscan_batch.py
```diff
@@ -0,0 +1,93 @@
...
+    except:
+        pass
+    return data
+
+
+
+# Common code for both CPU and GPU computations
+def compute(data, minObservations, epsilon):
+    # configure dbscan main object: we also request the indices and observations of cluster cores
+    algo = d4p.dbscan(minObservations=minObservations, epsilon=epsilon, resultsToCompute='computeCoreIndices|computeCoreObservations', memorySavingMode=1)
```
Alexander-Makaryev, line 55: I suggest to use True/False instead integers to specify value of memorySavingMode. It should work from Python side and can be replaced in example.

examples/sycl/dbscan_batch.py
```diff
@@ -0,0 +1,93 @@
...
+    
+    # Load the data
+    data = readcsv(infile, range(2))
+
+    result_classic = compute(data, minObservations, epsilon)
+
+    data = to_numpy(data)
+
+    # It is possible to specify to make the computations on GPU
+    with sycl_context('gpu'):
```
napetrov, line 73: This should be slightly updated with new examples structure or this will fail on system without gpu

PR #211, opened 2020-05-05
add check on decision_function_shape function

oleksandr-pavlyk:
Adding this check should paint sklearn-master build green.

### DIFF
--- a/daal4py/sklearn/svm/_svm_0_23.py
+++ b/daal4py/sklearn/svm/_svm_0_23.py
@@ -389,6 +389,13 @@ def fit(self, X, y, sample_weight=None):
             raise TypeError("Sparse precomputed kernels are not supported.")
         self._sparse = sparse and not callable(self.kernel)
 
+        if hasattr(self, 'decision_function_shape'):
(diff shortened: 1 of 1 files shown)


daal4py/sklearn/svm/_svm_0_23.py
```diff
@@ -389,6 +389,13 @@ def fit(self, X, y, sample_weight=None):
             raise TypeError("Sparse precomputed kernels are not supported.")
         self._sparse = sparse and not callable(self.kernel)
 
+        if hasattr(self, 'decision_function_shape'):
+            if self.decision_function_shape not in ('ovr', 'ovo'):
+                raise ValueError(
+                    f"decision_function_shape must be either 'ovr' or 'ovo', "
+                    f"got {self.decision_function_shape}."
```
Alexander-Makaryev, line 396: Maybe add quotes like in previous line: ``f"got '{self.decision_function_shape}'."``

oleksandr-pavlyk: copied verbatim from upstream.

PR #212, opened 2020-05-06
Feature/sycl sklearn

Alexander-Makaryev:
I am not sure that it is a right place or representation for such examples but let's start with it. Please suggest your vision how it should look like.


daal4py/sklearn/cluster/_dbscan_0_21.py
```diff
@@ -43,10 +43,12 @@ def _daal_dbscan(X, eps=0.5, min_samples=5, sample_weight=None):
     XX = make2d(X)
 
     fpt = getFPType(XX)
+    memorySavingMode = True if is_in_sycl_ctxt() else False
```
amgrigoriev, line 46: It might be risky to set memorySaavingMode to false on CPU in all cases.  How to process huge datasets?

daal4py/sklearn/cluster/_dbscan_0_22.py
```diff
@@ -43,10 +43,12 @@ def _daal_dbscan(X, eps=0.5, min_samples=5, sample_weight=None):
     XX = make2d(X)
 
     fpt = getFPType(XX)
+    memorySavingMode = True if is_in_sycl_ctxt() else False
```
amgrigoriev, line 46: See above

daal4py/sklearn/cluster/_dbscan_0_21.py
Alexander-Makaryev: We don't support memorySaavingMode param in Sklearn interface at this moment. But we can extend it in future. 'False' value is default for DAAL itself. You can create an issue/task about it.

daal4py/sklearn/cluster/_dbscan_0_22.py
Alexander-Makaryev: See above

napetrov:
How we will be testing this changes? so far our conformance run is made without DPC++ part. 
so we also need at least some kind of valuation

daal4py/sklearn/_utils.py
```diff
@@ -36,3 +36,10 @@ def make2d(X):
 method_uses_daal = "uses Intel® DAAL solver"
 method_uses_sklearn = "uses original Scikit-learn solver"
 method_uses_sklearn_arter_daal = "uses original Scikit-learn solver, because the task was not solved with Intel® DAAL"
+
+def is_in_sycl_ctxt():
+    try:
+        from daal4py.oneapi import is_in_sycl_ctxt as is_in_ctx
+        return is_in_ctx()
+    except:
```
oleksandr-pavlyk, line 44: It is best to be explicit here and use `except ModuleNotFoundError:`

daal4py/sklearn/cluster/_dbscan_0_21.py
oleksandr-pavlyk: more explicit `memorySavingMode = bool(is_in_sycl_ctxt())`.

daal4py/sklearn/cluster/_dbscan_0_22.py
oleksandr-pavlyk: `memorySavingMode = bool(is_in_sycl_ctxt())`.

src/daal4py.cpp
```diff
@@ -134,7 +134,9 @@ static PyObject * _make_nda_from_homogen(daal::data_management::NumericTablePtr
 
 #ifdef _DPCPP_
 #include "oneapi/oneapi_api.h"
-static int _1api_imp = import__oneapi();
+// Disable returning of sycl buffer from algorithms
+// static int _1api_imp = import__oneapi();
+static int _1api_imp = -1;
```
oleksandr-pavlyk, line 139: I would stick to canonical name such as `__oneAPI_imp = -1`.

src/oneapi/oneapi.pyx
```diff
@@ -45,25 +45,30 @@ cdef class sycl_execution_context:
         del self.c_ptr
 
 
-from contextlib import contextmanager
+# thread-local storage
+from threading import local as threading_local
+_tls = threading_local()
+if getattr(_tls, 'initialized', None) is None:
```
oleksandr-pavlyk, line 51: This is executed in the main thread, and I am not at all certain that the TLS class initialized here will end up being initialized in each additional thread.

It should be possible to find that out experimentally, but I would rather err on the side of cautious and add the logic to `def is_in_cycl_ctxt():` to check that `_tls` has `initialized` attribute, and initialize if it does not.

src/oneapi/oneapi.pyx
```diff
@@ -45,25 +45,30 @@ cdef class sycl_execution_context:
...
 
+from contextlib import contextmanager
 
 @contextmanager
 def sycl_context(dev='default'):
-    global _in_sycl_ctxt
     # Code to acquire resource
     ctxt = sycl_execution_context(dev)
-    _in_sycl_ctxt = True
+    _tls._in_sycl_ctxt = True
```
oleksandr-pavlyk, line 65: Same thing here. `_tls` may not have `initialized` variable set. 

I think you should write and use a setter routine which checks whether `_tls` has been initialized and initializes it if needed.

src/oneapi/oneapi.pyx
```diff
@@ -45,25 +45,30 @@ cdef class sycl_execution_context:
...
     try:
         yield ctxt
     finally:
         # Code to release resource
         del ctxt
-        _in_sycl_ctxt = False
+        _tls._in_sycl_ctxt = False
```
oleksandr-pavlyk, line 71: same, use setter.

src/oneapi/oneapi.pyx
```diff
@@ -102,7 +107,7 @@ cdef class sycl_buffer:
         return PyCapsule_New(todaalnt(<void*>self.sycl_buffer, self.typ, self.shape), NULL, NULL)
 
 cdef api object make_py_from_sycltable(void * ptr, int typ, int d1, int d2):
-    if not _in_sycl_ctxt:
+    if not _tls._in_sycl_ctxt:
```
oleksandr-pavlyk, line 110: use getter.

daal4py/sklearn/cluster/_dbscan_0_21.py
Alexander-Makaryev: I think that current representation is better because it is easier to read.
"memorySavingMode is True because we are in sycl context" but not "memorySavingMode is equal to is_in_sycl_context()".
Imagine any non-bool parameter that should be dependent on the context.
As example: method = 'specialSyclMethod' if is_in_sycl_ctxt() else 'defaultDense'

src/oneapi/oneapi.pyx
```diff
@@ -48,27 +48,41 @@ cdef class sycl_execution_context:
 # thread-local storage
 from threading import local as threading_local
 _tls = threading_local()
-if getattr(_tls, 'initialized', None) is None:
+
+def _is_tls_initialized():
+    return (not getattr(_tls, 'initialized', None) is None) and (_tls.initialized == True)
```
oleksandr-pavlyk, line 53: It is a valid Python, but `getattr(_tls, 'initialized', None) is not None` reads better.

daal4py/sklearn/cluster/_dbscan_0_22.py
Alexander-Makaryev: See above

daal4py/sklearn/_utils.py
Alexander-Makaryev: done

src/daal4py.cpp
Alexander-Makaryev: done

src/oneapi/oneapi.pyx
Alexander-Makaryev: done

src/oneapi/oneapi.pyx
Alexander-Makaryev: done

src/oneapi/oneapi.pyx
Alexander-Makaryev: done

src/oneapi/oneapi.pyx
Alexander-Makaryev: done

src/oneapi/oneapi.pyx
Alexander-Makaryev: done

Alexander-Makaryev:
@napetrov Thanks, you are right! Explicit test command is added in the meta.yaml.

examples/sycl/sklearn_sycl.py
```diff
@@ -0,0 +1,136 @@
...
+if __name__ == "__main__":
+    examples = [
+                k_means_init_x,
+                k_means_random,
+                linear_regression,
+                logistic_regression_lbfgs,
+                logistic_regression_newton,
+                dbscan,
+               ]
+    if gpu_available:
```
napetrov, line 131: It worth to add cpu context here so we will be validating that it's also possible to execute with CPU. and this will be validated in CI.

Alexander-Makaryev: done

PR #213, opened 2020-05-06
workaround for L0 on Beta06

napetrov:
In this change we adding check for daal version and change behavior for L0 load

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,7 @@ DAAL_DEFAULT_TYPE = 'double'
 
 
 #Level Zero workaround for oneDAL Beta06
-from generator.parse import parse_version
+from generator.parse import parse_version
 
 header_path = daal_root + '/include/services/library_version_info.h'
 
--- a/setup.py
+++ b/setup.py
@@ -98,6 +98,18 @@ else:
     DPCPP_LIBS = []
 DAAL_DEFAULT_TYPE = 'double'
 
+
+#Level Zero workaround for oneDAL Beta06
+from generator.parse import parse_version
+
+header_path = daal_root + '/include/services/library_version_info.h'
+
+
+with open(header_path) as header:
+    v = parse_version(header)
+    dal_build_version = (int(v[0]), int(v[2]))
+    print('Found DAAL version {}.{}'.format(*dal_build_version))
+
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -98,6 +98,18 @@
     DPCPP_LIBS = []
 DAAL_DEFAULT_TYPE = 'double'
 
+
+#Level Zero workaround for oneDAL Beta06
+from generator.parse import parse_version
+
+header_path = daal_root + '/include/services/library_version_info.h'
```
Alexander-Makaryev, line 105: It is better to use appropriate functions like os.path.join(daal_root, 'include', 'services', 'library_version_info.h')

setup.py
```diff
@@ -98,6 +98,18 @@
...
+
+
+with open(header_path) as header:
+    v = parse_version(header)
+    dal_build_version = (int(v[0]), int(v[2]))
+    print('Found DAAL version {}.{}'.format(*dal_build_version))
```
Alexander-Makaryev, line 111: We already have such print here
https://github.com/IntelPython/daal4py/blob/master/generator/gen_daal4py.py#L183
In that place it shows that all headers are read successfully.
Two same prints in the log can make misunderstanding.
Maybe we should change the message or remove it at all.

setup.py
```diff
@@ -138,7 +150,7 @@ def getpyexts():
     if IS_WIN:
         libraries_plat = ['daal_core_dll']
     else:
-        if dpcpp:
+        if dal_build_version == (2021,106):
```
Alexander-Makaryev, line 153: I suggest to move this behavior to another place like it is done here (branch, is not merged yet):
https://github.com/IntelPython/daal4py/blob/feature/sycl-sklearn/setup.py#L95
these changes are in this commit
https://github.com/IntelPython/daal4py/commit/8e6e84270ce10a2449b42595f5706816759b3d80

setup.py
napetrov: done

setup.py
napetrov: removed

setup.py
napetrov: done

PR #215, opened 2020-05-13
Train test split optimizations

Alexsandruss:
Optimizations for [train_test_split function from scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html).
[oneDAL pull request](https://github.com/oneapi-src/oneDAL/pull/583)
- Optimizations support numpy ndarray, pandas DataFrame and Series with float and int types.
- Introduces new RNGs from mkl_random for shuffling


Alexander-Makaryev:
I suggest to use the same paths as in original Scikit-Learn:
https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/model_selection/_split.py (with underscore)
instead
https://github.com/Alexsandruss/daal4py/blob/train_test_split_opt/daal4py/sklearn/model_selection/split.py

daal4py/sklearn/model_selection/split.py
```diff
@@ -0,0 +1,201 @@
+from sklearn.utils import indexable, _safe_indexing
```
PetrovKP, line 1: Did split.py from model_selection have different implementation for different version sklearn (0.22, 0.23, etc) ?

PetrovKP: I think you need to make similar to other algorithms

daal4py/sklearn/model_selection/split.py
```diff
@@ -0,0 +1,201 @@
...
+        else:
+            fallback = True
+
+        dtypes = get_dtypes(arr)
+        if dtypes is None:
+            fallback = True
+        else:
+            for i, dtype in enumerate(dtypes):
+                if dtype in [np.int32, 'int32']:
+                    dtypes[i] = 0
```
PetrovKP, line 107: I don't like naming for type: 
```c++
0 - int
1 - float
etc.
```
It looks strange and not justified. Why can not you use a string, enums or something similar?

daal4py/sklearn/model_selection/split.py
```diff
@@ -0,0 +1,201 @@
...
+                dtypes = dtypes * n_cols
+            dtypes = np.array(dtypes, dtype=np.int32, ndmin=2)
+
+            if origin_type == 'numpy':
+                order = 'C' if arr.flags['C_CONTIGUOUS'] == True else 'F'
+
+                train_arr = np.empty(shape=(n_train, n_cols), dtype=arr.dtype, order=order)
+                test_arr = np.empty(shape=(n_test, n_cols), dtype=arr.dtype, order=order)
+
+                d4p.daal_train_test_split(arr.reshape((arr.shape[0], n_cols), order='A'), train_arr, test_arr, train, test, dtypes)
```
PetrovKP, line 148: Why you use reshape? I don't understand why do you create 5 branches for detect input types? `data_or_file` can in inside defined need types for NumericTable.

daal4py/sklearn/model_selection/split.py
```diff
@@ -0,0 +1,201 @@
...
+    for arr in arrays:
+        fallback = False
+        if isinstance(arr, np.ndarray):
+            origin_type = "numpy"
+        elif pandas_is_imported:
+            if isinstance(arr, pd.core.frame.DataFrame):
+                origin_type = "pandas DataFrame"
+            elif isinstance(arr, pd.core.series.Series):
+                origin_type = "pandas Series"
+            else:
```
PetrovKP, line 96: I don't see branch for `CSR` format

daal4py/sklearn/model_selection/split.py
```diff
@@ -0,0 +1,201 @@
...
+
+                train_arr, test_arr = pd.DataFrame(train_arr), pd.DataFrame(test_arr)
+
+            elif origin_type == 'pandas DataFrame with multiple dtypes':
+                train_arr, test_arr = {}, {}
+                for i, col in enumerate(list(arr.columns)):
+                    train_arr[col] = np.empty(shape=(n_train,), dtype=arr.dtypes[col])
+                    test_arr[col] = np.empty(shape=(n_test,), dtype=arr.dtypes[col])
+                train_arr = pd.DataFrame(train_arr)
+                test_arr = pd.DataFrame(test_arr)
```
PetrovKP, line 171: Does Sklearn have similar logic?

src/daal4py.cpp
```diff
@@ -662,3 +662,17 @@ bool c_assert_all_finite(const data_or_file & t, bool allowNaN, char dtype)
...
+void c_train_test_split(data_or_file & orig, data_or_file & train, data_or_file & test,
+                        data_or_file & train_idx, data_or_file & test_idx)
+{
+#if INTEL_DAAL_VERSION >= 20200002
+    auto origTable = get_table(orig);
+    auto trainTable = get_table(train);
+    auto testTable = get_table(test);
+    auto trainIdxTable = get_table(train_idx);
+    auto testIdxTable = get_table(test_idx);
+    daal::data_management::internal::trainTestSplit<int>(*origTable, *trainTable, *testTable, *trainIdxTable, *testIdxTable);
```
PetrovKP, line 675: Why template `int`, not `float` and `double`?

daal4py/sklearn/monkeypatch/dispatcher.py
```diff
@@ -88,6 +91,8 @@
 if daal_run_version >= (2020, 1) and daal_link_version >= (2020, 1):
     _mapping['fin_check'] = [[(validation, '_assert_all_finite', _daal_assert_all_finite), None]]
 
+if daal_run_version >= (2020, 2) and daal_link_version >= (2020, 2):
```
PetrovKP, line 94: Will It be  this changes in oneAPI releases? Do you need to check it?

daal4py/sklearn/model_selection/split.py
```diff
@@ -0,0 +1,149 @@
...
+        test = np.arange(n_train, n_train + n_test)
+    else:
+        if stratify is not None:
+            cv = StratifiedShuffleSplit(test_size=n_test,
+                                        train_size=n_train,
+                                        random_state=random_state)
+            train, test = next(cv.split(X=arrays[0], y=stratify))
+        else:
+            if mkl_random_is_imported and (isinstance(random_state, int) or random_state is None):
+                random_state = mkl_random.RandomState(random_state, rng)
```
PetrovKP, line 74: Does MKL use for stock sklearn for generate indices?

daal4py/sklearn/model_selection/split.py
```diff
@@ -0,0 +1,149 @@
...
+            else:
+                cv = ShuffleSplit(test_size=n_test,
+                                  train_size=n_train,
+                                  random_state=random_state)
+                train, test = next(cv.split(X=arrays[0], y=stratify))
+
+    res = []
+    for arr in arrays:
+        fallback = False
+        if not isinstance(arr, np.ndarray):
```
PetrovKP, line 86: Maybe we can make a check for `numpy` and `pandas` input in a common place? I think it would be good for @owerbat  to have a check too. What do you think?

PetrovKP:
@Pahandrovich , Can you check this PR?

daal4py/sklearn/model_selection/split.py
Alexsandruss: Types naming was deleted

daal4py/sklearn/model_selection/split.py
```diff
@@ -0,0 +1,201 @@
+from sklearn.utils import indexable, _safe_indexing
+from sklearn.utils.validation import _num_samples
+from sklearn.model_selection import StratifiedShuffleSplit, ShuffleSplit, train_test_split
+from sklearn.model_selection._split import _validate_shuffle_split
+import daal4py as d4p
+import numpy as np
+
+from timeit import default_timer as timer
```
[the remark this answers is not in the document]
Alexsandruss: Deleted

src/daal4py.h
```diff
@@ -53,6 +53,10 @@ using daal::services::LibraryVersionInfo;
 #include "data_management/data/internal/finiteness_checker.h"
 #endif
 
+#if INTEL_DAAL_VERSION >= 20200002
```
Alexander-Makaryev, line 56: We can avoid failings with DPCPP if we will pay attention for beta (2021...) versions too.

Alexander-Makaryev:
@Alexsandruss 
Merge changes from master branch to fix failed scikit-learn tests.

daal4py/sklearn/model_selection/split.py
```diff
@@ -0,0 +1,149 @@
...
+            train = train.reshape((n_train,))
+            test = test.reshape((n_test,))
+            res.append(_safe_indexing(arr, train))
+            res.append(_safe_indexing(arr, test))
+        else:
+            train = train.reshape(n_train, 1)
+            test = test.reshape(n_test, 1)
+
+            if len(arr.shape) == 2:
+                one_dim = False
```
owerbat, line 111: ```one_dim``` is never used. So, I think it should be removed

daal4py/sklearn/model_selection/split.py
Alexsandruss: sklearn uses specific indexing function which is based mostly on iloc method of pd.DataFrame, for example.

daal4py/sklearn/model_selection/split.py
Alexsandruss: reshaping is mostly free operation in this case and allows to pass data without problem (data_or_file doesn't consume 1d input)

daal4py/sklearn/model_selection/split.py
Alexsandruss: CSR isn't prioritized, so it can be added later

src/daal4py.cpp
Alexsandruss: int is IndexType

daal4py/sklearn/model_selection/split.py
Alexsandruss: deleted, thanks for reminder

daal4py/sklearn/model_selection/split.py
PetrovKP: Exception? Callback?

daal4py/sklearn/monkeypatch/dispatcher.py
Alexsandruss: Yes, oneDAL master has version (2021, 8). I checked that optimization is used.

daal4py/sklearn/model_selection/split.py
Alexsandruss: Split has stable API and implementation in last versions.

daal4py/sklearn/model_selection/split.py
Alexsandruss: I don't understand question about exception and callback exactly. If input is CSR, _daal_train_test_split falls back to default implementation.

daal4py/sklearn/model_selection/split.py
Alexsandruss: stock scikit-learn uses default numpy.random

daal4py/sklearn/monkeypatch/dispatcher.py
PetrovKP: But in version oneDAL < beta08 we will have a failed. because 2021.*07 > 2020.*02

Alexsandruss: I added more complex and working version getting. It should work for both beta and mnt.

src/daal4py.h
Alexsandruss: I added more complex and working version getting. It should work for both beta and mnt.

Alexsandruss:
@PetrovKP @Alexander-Makaryev @Pahandrovich , please, review last changes. I run scikit-learn 0.23.1 tests, looks like there are no problems with train_test_split, but I was forced to fallback to default numpy RNG due to some failed tests with predefined results.

PR #219, opened 2020-05-15
fix for python3.5 building

KalyanovD:
After https://github.com/IntelPython/daal4py/pull/169 was noticed that building with python 3.5 was broken. This PR fixes it.

### DIFF
--- a/generator/wrapper_gen.py
+++ b/generator/wrapper_gen.py
@@ -1153,7 +1153,7 @@ class wrapper_gen(object):
         """
         return code for initing
         """
-        cpp = "#ifndef DAAL4PY_CPP_INC_\n#define DAAL4PY_CPP_INC_\n#include <daal4py_dist.h>\n\ntypedef daal::data_management::NumericTablePtr NumericTablePtr;"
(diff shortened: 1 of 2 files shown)


Alexander-Makaryev:
@KalyanovD Could you please also check that build works fine with/without NO_DIST option

Alexander-Makaryev:
@KalyanovD 
We already discussed it but I will save it here.
Previously we deleted definitions because we got 'multiple definition' error with some compilers from VS on Windows. Please look at these rows:
https://github.com/IntelPython/daal4py/blob/a6f2efd2a5a87b242769c6ca24704932697eedd7/src/daal4py.h#L220
https://github.com/IntelPython/daal4py/blob/a6f2efd2a5a87b242769c6ca24704932697eedd7/generator/wrapper_gen.py#L93
I suggest to remove the definition through ``data_management_NumericTablePtr`` and return one of deleted. (One of them is returned in first commit of this PR)

KalyanovD:
@Alexander-Makaryev, you were right, with returning one of removed definitions daal4py is started to build locally on Linux with Python 3.5 and 3.7 in batch and distributed modes(NO_DIST=1 or 0 for 3.7).

Vyacheslav-Smirnov:
I am ok with changes

PR #220, opened 2020-05-17
fix dbscan example/test

Alexander-Makaryev:


examples/sycl/dbscan_batch.py
```diff
@@ -79,13 +81,19 @@ def main(readcsv=read_csv, method='defaultDense'):
...
             assert np.allclose(result_classic.assignments, result_gpu.assignments)
+            assert np.allclose(result_classic.coreIndices, result_gpu.coreIndices)
+            assert np.allclose(result_classic.coreObservations, result_gpu.coreObservations)
 
     with sycl_context('cpu'):
-        sycl_data2 = sycl_buffer(data)
-        result_cpu = compute(sycl_data2, minObservations, epsilon) 
+        sycl_data = sycl_buffer(data)
+        result_cpu = compute(sycl_data, minObservations, epsilon)
+        assert np.allclose(result_classic.nClusters, result_cpu.nClusters)
```
amgrigoriev, line 93: At the moment, "classic" and "context('cpu') are using the same kernel. Only difference is SyclHomogenTable instead of HomogenTable.

PR #223, opened 2020-05-18
Update using interfaces kmeans for patch sklearn

PetrovKP:
Using choose parameters `computeAssignments`, `computeCentroids` and `computeAssignments` for efficiently calculate an algorithm without overhead.


PetrovKP:
@agorshk, @Alexander-Makaryev , @oleksandr-pavlyk  Look, please. I made it only for the 0.23 file so that it would be easier to review, before the merge I will add 0.22 and 0.21 for these changes.

daal4py/sklearn/cluster/_k_means_0_23.py
```diff
@@ -33,7 +33,7 @@
 from sklearn.cluster import KMeans as KMeans_original
 
 import daal4py
-from .._utils import getFPType, method_uses_sklearn, method_uses_daal
+from .._utils import getFPType, method_uses_sklearn, method_uses_daal, daal_check_version
```
agorshk, line 36: Do we need update interfaces for 0.21 and 0.22 Sklearn version? Does it correct work if we take the newest version of daal and old version d4p (do we need check version for dp4?)

PetrovKP:
> Am I right, that appropriate Sklearn tests were passed (it's need manual check due to the old version comes to CI)?

Of course. So far this has been verified in manual mode.

> Do we need update interfaces for 0.21 and 0.22 Sklearn version?

Of course. This will be done after the review to make it easier for review.

> Does it correct work if we take the newest version of daal and old version d4p (do we need check version for dp4?)

I checked different version DAAL. I need to check with different versions of daal4py. I think it should work.

agorshk:
> @agorshk
> 
> > Am I right, that appropriate Sklearn tests were passed (it's need manual check due to the old version comes to CI)?
> 
> Of course. So far this has been verified in manual mode.
> 
> > Do we need update interfaces for 0.21 and 0.22 Sklearn version?
> 
> Of course. This will be done after the review to make it easier for review.
> 
> > Does it correct work if we take the newest version of daal and old version d4p (do we need check version for dp4?)
> 
> I checked different version DAAL. I need to check with different versions of daal4py. I think it should work.

Ok, that's all from my side, I have no other objections

PR #224, opened 2020-05-21
Fix tests fails with new Scikit-Learn 0.23.1

Alexander-Makaryev:
For now I made changes only in branch of code for latest version of Scikit-Learn.
But I think that our work with Logistic Regression is too complicated. We have the branching on init stage in
https://github.com/IntelPython/daal4py/blob/master/daal4py/sklearn/linear_model/logistic_path.py
like
```
if LooseVersion(sklearn_version) >= LooseVersion("0.22"):
    from ._logistic_path_0_22 import *
else:
    from ._logistic_path_0_21 import *
```
and in files ``_logistic_path_*.py`` themselves.
Please help me clarify that it is really necessary and advice how to simplify it.


oleksandr-pavlyk:
> But I think that our work with Logistic Regression is too complicated.

The 0.22 file can probably be tidied up, ridding it of dead code intended to support legacy behavior.

PR #226, opened 2020-05-22
Setting random forest classifier to beta09 version

napetrov:
This problems is not fixed in Beta07 version, and not fixed within in0development Beta08 version

### DIFF
--- a/tests/test_estimators.py
+++ b/tests/test_estimators.py
@@ -77,7 +77,7 @@ class Test(unittest.TestCase):
     def test_KNeighborsClassifier(self):
         check_estimator(KNeighborsClassifier)
 
-    @unittest.skipUnless(check_version(((2019,0),(2021, 107)), daal_version), "not supported in this library version")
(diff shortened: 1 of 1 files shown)


Alexander-Makaryev:
RF Classifier is a Scikit-Learn-compatible estimator but not a drop-in replacement.
It si based on current version of DAAL and works now. Why we want to exclude it from testing? Do we have changes in DAAL that will broke this work?
I suggest to remove version limitation. We will replace it with a drop-in replacement implementation when all will be ready for it on DAAL side.

Alexander-Makaryev:
@napetrov 
We discussed it with @KalyanovD and @PivovarA.
The fix is already done and should be merged soon
https://github.com/oneapi-src/oneDAL/pull/603
Changes are targeted to existing version in limitations.
Suggestion is to leave this PR and close it when issue will be resolved in DAAL. (In this case we will not forget about it)
At the same time version limitation should be removed after merge of the fix. It can be done in this or separate PR.

PR #229, opened 2020-05-28
removing tbb

Alexander-Makaryev:


azure-pipelines.yml
```diff
@@ -14,7 +14,6 @@ jobs:
       sudo apt-get install              \
           intel-oneapi-common-vars      \
           intel-oneapi-common-licensing \
-          intel-oneapi-tbb-devel        \
```
napetrov, line 17: I guess we don't need TBB-devel, but we need tbb for runtime to actually execute CPU threaded stuff?

Alexander-Makaryev: Yes, tbb is runtime dependency of threaded DAAL.

PR #234, opened 2020-06-02
Removing pandas conversion in Linear Regression

owerbat:
List of changes:
* sklearn's ```check_array(), check_X_y(), and validate_data()``` have been modified to exclude ```pandas.DataFrame``` to ```numpy.array``` conversion
* new ```_daal_check_array(), _daal_check_X_y(), and _daal_validate_data()``` have been used in Linear Regresion
* ```getFPType()``` has been modified to work with ```pandas.DataFrame```
* a couple of tests have been added

### DIFF
--- a/daal4py/sklearn/linear_model/_linear_0_21.py
+++ b/daal4py/sklearn/linear_model/_linear_0_21.py
@@ -33,7 +33,8 @@ except ImportError:
 
 import daal4py
 from .._utils import (make2d, getFPType, method_uses_sklearn, \
-                    method_uses_daal, method_uses_sklearn_arter_daal)
+                    method_uses_daal, method_uses_sklearn_arter_daal,
+                    is_DataFrame, get_dtype)
 import logging
 
 def _daal4py_fit(self, X, y_):
@@ -119,12 +120,7 @@ def fit(self, X, y, sample_weight=None):
     X, y = _daal_check_X_y(X, y, accept_sparse=['csr', 'csc', 'coo'],
                            y_numeric=True, multi_output=True)
 
-    try:
-        from pandas import DataFrame
-        from pandas.core.dtypes.cast import find_common_type
-        dtype = find_common_type(X.dtypes) if isinstance(X, DataFrame) else X.dtype
-    except ImportError:
-        dtype = X.dtype
+    dtype = get_dtype(X, is_DataFrame(X))
 
     self.sample_weight_ = sample_weight
     self.fit_shape_good_for_daal_ = bool(X.shape[0] > X.shape[1] + int(self.fit_intercept))
@@ -187,20 +183,10 @@ def predict(self, X):
     C : array, shape = (n_samples,)
         Returns predicted values.
     """
-    try:
-        from pandas import DataFrame
-        is_DataFrame = isinstance(X, DataFrame)
-    except ImportError:
-        is_DataFrame = False
-
-    X = np.asarray(X) if not sp.issparse(X) and not is_DataFrame else X
+    is_df = is_DataFrame(X)
+    X = np.asarray(X) if not sp.issparse(X) and not is_df else X
     good_shape_for_daal = True if X.ndim <= 1 else True if X.shape[0] > X.shape[1] else False
-
-    try:
-        from pandas.core.dtypes.cast import find_common_type
-        dtype = find_common_type(X.dtypes) if is_DataFrame else X.dtype
-    except ImportError:
-        dtype = X.dtype
+    dtype = get_dtype(X, is_df)
 
     if (sp.issparse(X) or
             not hasattr(self, 'daal_model_') or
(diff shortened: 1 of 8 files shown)


daal4py/sklearn/linear_model/_linear_0_21.py
```diff
@@ -115,14 +116,22 @@ def fit(self, X, y, sample_weight=None):
     """
 
     n_jobs_ = self.n_jobs
-    X, y = check_X_y(X, y, accept_sparse=['csr', 'csc', 'coo'],
-                     y_numeric=True, multi_output=True)
+    X, y = _daal_check_X_y(X, y, accept_sparse=['csr', 'csc', 'coo'],
+                           y_numeric=True, multi_output=True)
+
+    try:
```
PetrovKP, line 122: I think it will be copied to many places. Need to move in a common place

daal4py/sklearn/linear_model/_linear_0_21.py
```diff
@@ -19,8 +19,9 @@
 from scipy import sparse as sp
```
PetrovKP, line 19: For review you just one file to make a change. For make it easier to leave comments and you did not have to forward each change to each file - 0.21, 0.22, 0.23. At the final stage, you simply copy the changes and carefully discard them into the backward files.

daal4py/sklearn/linear_model/tests/test_linear.py
```diff
@@ -0,0 +1,80 @@
+import pytest
```
PetrovKP, line 1: You need to use copyright.

daal4py/sklearn/linear_model/tests/test_linear.py
```diff
@@ -0,0 +1,80 @@
...
+def test_linear_array_vs_dataframe_heterogen():
+    pd = pytest.importorskip('pandas')
+
+    x_train, y_train = make_dataset(100, 20)
+    x_test, _ = make_dataset(100, 20, random_state=1)
+
+    df_x_train, df_y_train = make_dataset(100, 20, pd.DataFrame)
+    df_x_test, _ = make_dataset(100, 20, pd.DataFrame, random_state=1)
+
+    dir_dtypes = {col: 'float64' if i % 2 == 0 else 'float32' for i, col in enumerate(df_x_train.columns)}
```
PetrovKP, line 68: Good, I think to add of tests for combine type: `int`, `float`, `double`. And one test is little.  May be move this in  function: `make_dataset`

daal4py/sklearn/utils/base.py
```diff
@@ -0,0 +1,57 @@
+from .validation import _daal_check_array, _daal_check_X_y
```
PetrovKP, line 1: Copyright

daal4py/sklearn/utils/validation.py
```diff
@@ -62,3 +69,431 @@ def _daal_assert_all_finite(X, allow_nan=False, msg_dtype=None):
...
+    context = " by %s" % estimator_name if estimator is not None else ""
+
+    array_orig = array
+
+    try:
+        from pandas import DataFrame, option_context
+        if isinstance(array, DataFrame) and force_all_finite != 'allow-nan':
+            if force_all_finite:
+                with option_context('mode.use_inf_as_na', True):
+                    if array.isna().any().any():
```
PetrovKP, line 181: We have such functionality in DAAL. Can we use this for check? @Alexsandruss What do you think?

daal4py/sklearn/utils/validation.py
```diff
@@ -62,3 +69,431 @@ def _daal_assert_all_finite(X, allow_nan=False, msg_dtype=None):
...
+            estimator_name = estimator
+        else:
+            estimator_name = estimator.__class__.__name__
+    else:
+        estimator_name = "Estimator"
+    context = " by %s" % estimator_name if estimator is not None else ""
+
+    array_orig = array
+
+    try:
```
PetrovKP, line 176: Maybe, move in other internal function for pandas check?

daal4py/sklearn/linear_model/_linear_0_22.py
```diff
@@ -203,20 +212,32 @@ def predict(self, X):
     C : array, shape = (n_samples,)
         Returns predicted values.
     """
-    X = np.asarray(X) if not sp.issparse(X) else X
+    try:
```
PetrovKP, line 215: Make common function in utils:
```python
def is_data_frame(x):
 ...
```

daal4py/sklearn/utils/validation.py
owerbat: Now in ```_daal_assert_all_finite(X)``` any ```X``` is converted to ```numpy.array``` and it slows down everything for pandas case. So, we need to change it to use DAAL functionality

PetrovKP: Yes, it looks like you need to change. Because the implementation in DAAL can work with heterogeneous data? Am I right, @Alexsandruss ?

Alexsandruss: Yes, _daal_assert_all_finite should handle pandas w/o conversions like _daal_train_test_split, which is WIP.

PR #235, opened 2020-06-03
initial support of dppy + example

Alexander-Makaryev:
This is the initial implementation. PR is created in purpose to share it and start the conversation.

### DIFF
--- a/generator/wrapper_gen.py
+++ b/generator/wrapper_gen.py
@@ -221,6 +221,26 @@ cdef extern from "daal4py.h":
 
 def daal_train_test_split(orig, train, test, train_idx, test_idx):
     c_train_test_split(data_or_file(<PyObject*>orig), data_or_file(<PyObject*>train), data_or_file(<PyObject*>test), data_or_file(<PyObject*>train_idx), data_or_file(<PyObject*>test_idx))
+
+import sys
+def _execute_with_context(func):
+    def exec_func(*args, **keyArgs):
+        # we check is DPPY imported or not
+        # possible we should check are we in defined context or not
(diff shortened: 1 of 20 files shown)


generator/wrapper_gen.py
```diff
@@ -894,7 +894,22 @@ def compute(self,
...
+            from dppy import runtime as rt
+            from _oneapi import set_queue_to_daal_context, release_queue
+
+            ctx = rt.get_current_context()
+            q = ctx.get_sycl_queue()
+            set_queue_to_daal_context(q)
+
+            res = self._compute({{input_args|fmt('{}', 'name', sep=', ')}}, False)
+
+            release_queue(q)
```
Alexander-Makaryev, line 909: @michael-smirnov Attempt to work with dppy context fails with segmentation fault. Looks like that we should not release the queue after computation or we do it incorrect at this moment. Please clarify that we are not trying to destroy the original dppy queue.

Alexander-Makaryev:
@diptorupd Take a look, the "user experience" with DPPY will be useful)

generator/wrapper_gen.py
michael-smirnov: @Alexander-Makaryev thank you! That is not an issue with the wrong destruction of queue, but it is a known issue caused by the undefined destruction of static environment objects of oneDAL and SYCL. We are keeping a queue inside oneDAL even when dppy is trying to release SYCL context - that is the root cause of that seqfault. 
Now I emulate a "stateless" approach when oneDAL forgets a queue after each operation. It solves the problem. In the future, we will implement a stateless approach natively in new oneDAL interfaces.

PR #238, opened 2020-06-09
Add support thunder SVM method and weights for IDP sklearn

PetrovKP:


PetrovKP:
@agorshk, @Alexander-Makaryev  @oleksandr-pavlyk  Please, look this changes for SVM

oleksandr-pavlyk:
@PetrovKP Would it make sense to also make `_daal4py_kf` support CSR input as we discussed some time back? Or perhaps it should be done in a separate PR.

PetrovKP:
> @PetrovKP Would it make sense to also make `_daal4py_kf` support CSR input as we discussed some time back? Or perhaps it should be done in a separate PR.

Yes. It will be in another PR. And most likely in the next release. We need to spend some time passing all the tests.

PR #240, opened 2020-06-17
Added information in documentation about verbose

Pahandrovich:


doc/sklearn.rst
```diff
@@ -52,6 +54,40 @@ on all added and monkey-patched classes, which are discovered by means of
 introspection. This assures scikit-learn API compatibility of all
 `daal4py.sklearn` classes.
 
+.. _sklearn_verbose:
+
+scikit-learn verbose
+----------------
+
+In order to find out which implementation of the algorithm is currently used,
```
outoftardis, line 62: you can just use "to" instead of "in order to" (it's recommended by the style guide)

doc/sklearn.rst
```diff
@@ -52,6 +54,40 @@ on all added and monkey-patched classes, which are discovered by means of
...
+export the environment variable.
```
outoftardis, line 63: I'd add the name of the variable here:

```suggestion
export the ``IDP_SKLEARN_VERBOSE`` environment variable.
```

shouldn't it be "set" instead of "export"?

doc/sklearn.rst
```diff
@@ -52,6 +54,40 @@ on all added and monkey-patched classes, which are discovered by means of
...
+
+On Linux and Mac OS::
+
+    export IDP_SKLEARN_VERBOSE=INFO
+
+On Windows::
+
+    set IDP_SKLEARN_VERBOSE=INFO
+
+For example, on Linux you may use the following command::
```
outoftardis, line 73: what's that an example of exactly? how's is that different from the command above? I think you need to add a bit more context here (e.g. something like "An example below shows how to ... ")

doc/sklearn.rst
```diff
@@ -52,6 +54,40 @@ on all added and monkey-patched classes, which are discovered by means of
...
+
+    IDP_SKLEARN_VERBOSE=INFO python -m daal4py my_application.py
+
+During calls using Intel-optimized scikit-learn you will receive additional print statements,
+indicating which implementation being called.
```
outoftardis, line 78:
```suggestion
During the calls that use Intel-optimized scikit-learn, you will receive additional print statements
that indicate which implementation is being called.
```

doc/sklearn.rst
```diff
@@ -52,6 +54,40 @@ on all added and monkey-patched classes, which are discovered by means of
...
+These print statements are only available for :ref:`scikit-learn algorithms that the daal4py patches <sklearn_algorithms>`.
```
outoftardis, line 79:
```suggestion
These print statements are only available for :ref:`scikit-learn algorithms that daal4py patches <sklearn_algorithms>`.
```

```suggestion
These print statements are only available for :ref:`scikit-learn algorithms with daal4py patches <sklearn_algorithms>`.
```

doc/sklearn.rst
```diff
@@ -52,6 +54,40 @@ on all added and monkey-patched classes, which are discovered by means of
...
+
+For example, for DBSCAN::
+
+    INFO: sklearn.cluster.DBSCAN.fit: uses Intel® DAAL solver
+
+or::
+
+    INFO: sklearn.cluster.DBSCAN.fit: uses original Scikit-learn solver
```
outoftardis, line 87: I'd remove this "or"

```suggestion
For example, for DBSCAN you get one of these print statements depending on which implementation is used::

    INFO: sklearn.cluster.DBSCAN.fit: uses Intel® DAAL solver

::

    INFO: sklearn.cluster.DBSCAN.fit: uses original Scikit-learn solver
```

doc/sklearn.rst
Pahandrovich: I agree, this example duplicates command above. I removed this.

Pahandrovich:
@outoftardis, thanks for comments. I fixed them.

PR #243, opened 2020-06-24
Optimization of train_test_split indices shuffling

Alexsandruss:
Optimization of indices shuffling for scikit-learn train_test_split function.

### DIFF
--- a/src/daal4py.cpp
+++ b/src/daal4py.cpp
@@ -678,3 +678,13 @@ void c_train_test_split(data_or_file & orig, data_or_file & train, data_or_file 
 #else
 #endif
 }
+
+void c_generate_shuffled_indices(data_or_file & idx, data_or_file & random_state)
+{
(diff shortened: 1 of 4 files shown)


src/daal4py.cpp
```diff
@@ -678,3 +678,13 @@ void c_train_test_split(data_or_file & orig, data_or_file & train, data_or_file
 #else
 #endif
 }
+
+void c_generate_shuffled_indices(data_or_file & idx, data_or_file & random_state)
```
PetrovKP, line 682: don't you want to merge into function `c_train_test_split`?

Alexsandruss:
Tests of sklearn 0.21/0.22/0.23 didn't reveal problems of optimizations

src/daal4py.cpp
Alexsandruss: In some cases optimizations of `c_generate_shuffled_indices` or `c_train_test_split` may not be used. So, to use at least one of them, functions should be separate.

PR #245, opened 2020-07-10
Swithcing to new dppl name

napetrov:


napetrov:
> 
> 
> examples/sycl/svm_batch.py for some reason marked like totally changed in diff... end of lines?
> LGTM

yes. end of lines

PR #248, opened 2020-07-15
renaming dal lib

napetrov:
in this change we rename daal_sycl to onedal_sycl starting from Beta08

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,11 +103,14 @@ with open(header_path) as header:
 
 if dpcpp:
     DPCPP_CFLAGS = ['-D_DPCPP_']
-    DPCPP_LIBS = ['OpenCL', 'sycl', 'daal_sycl']
+    DPCPP_LIBS = ['OpenCL', 'sycl', 'onedal_sycl']
     if IS_LIN:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -103,11 +103,14 @@
 
 if dpcpp:
     DPCPP_CFLAGS = ['-D_DPCPP_']
-    DPCPP_LIBS = ['OpenCL', 'sycl', 'daal_sycl']
+    DPCPP_LIBS = ['OpenCL', 'sycl', 'onedal_sycl']
     if IS_LIN:
         DPCPP_LIBDIRS = [jp(dpcpp_root, 'linux', 'lib')]
     elif IS_WIN:
         DPCPP_LIBDIRS = [jp(dpcpp_root, 'windows', 'lib')]
+    if dal_build_version == (2021,7):
```
Alexander-Makaryev, line 111: maybe
dal_build_version  <= (2021,7)
?

napetrov: I tough about this - but not sure how  well  this work. 

in general all previous beta versions are obsolete and this code will be removed. together with previous versions

napetrov: so taking in mind time constraints - i would go with safe option )

PR #256, opened 2020-07-31
Paths of oneapi package changed for azure CI

michael-smirnov:


conda-recipe/run_test.sh
```diff
@@ -28,3 +28,9 @@ if [ ! -z "${DAALROOT}" ]; then
     conda remove daal --force -y
     source ${DAALROOT}/env/vars.sh
 fi
+
+# if TBBROOT is specified
+if [ ! -z "${TBBROOT}" ]; then
+    conda remove tbb --force -y
```
PetrovKP, line 34: tbb in conda isn't it installed by default from oneapi toolkit?

examples/run_examples.py
```diff
@@ -22,6 +22,7 @@
 
 from daal4py import __daal_link_version__ as dv, __has_dist__
 daal_version = tuple(map(int, (dv[0:4], dv[4:8])))
+print('DAAL version:', daal_version)
```
PetrovKP, line 25: remember delete

conda-recipe/run_test.sh
napetrov: There is no conda for oneapi for now. And then this will happen - first it will be required to split build process for 2020 and 2021 version as they have separate channels and separate packages

examples/run_examples.py
napetrov: I think this is legit print to show DAAL version used for examples run and i think it worth having it in longterm

PR #250, opened 2020-07-20
Support probability, decision function and CSR input format for SVC

RomaA2000:
Support probability, decision function and CSR input format for SVC algorithm

PR #262, opened 2020-08-07
Support probability and decision function for SVM

PetrovKP:


PetrovKP:
Move to https://github.com/IntelPython/daal4py/pull/250

PetrovKP:
Move to #250

PR #271, opened 2020-08-25
Gbt modelbuilder

RukhovichIV:
Added a model builder for GBT


ShvetsKS:
@Alexander-Makaryev could you review this PR please?

Alexander-Makaryev:
LGTM)
please add some examples/tests

src/modelbuilder.pyx
```diff
@@ -0,0 +1,104 @@
...
+        Get built model
+
+        :rtype: gbt_classification_model
+        '''
+        cdef gbt_classification_model res = gbt_classification_model.__new__(gbt_classification_model)
+        res.c_ptr = get_gbt_classification_modelbuilder_Model(self.c_ptr)
+        return res
+
+
+def model_builder(nFeatures, nIterations, nClasses = 2):
```
PetrovKP, line 95: Don't forget that there are builders for other algorithms, such as regression. So, this name is not a good one in my understanding.

RukhovichIV:
@Alexander-Makaryev ,
> LGTM)
> please add some examples/tests

We're working on it. It'll be a part of another PR (#273 ) which is not ready yet.
Could you explicitly approve this PR?

src/modelbuilder.pyx
RukhovichIV: Thank you. I've changed it to `gbt_clf_model_builder()` and also added a similar function for regression. Please look at this one more time

ShvetsKS (approved):

Alexander-Makaryev (approved):
LGTM again
Looks like the changes for regression are the same as previous

src/modelbuilder.h
```diff
@@ -0,0 +1,45 @@
+/*******************************************************************************
+* Copyright 2014-2019 Intel Corporation
```
PetrovKP, line 2:
```suggestion
* Copyright 2020 Intel Corporation
```

src/modelbuilder.pyx
```diff
@@ -0,0 +1,180 @@
+#*******************************************************************************
+# Copyright 2014-2019 Intel Corporation
```
PetrovKP, line 2:
```suggestion
# Copyright 2020 Intel Corporation
```

generator/gen_daal4py.py
```diff
@@ -942,11 +942,18 @@ def gen_daal4py(daalroot, outdir, version, warn_all=False, no_dist=False, no_str
     with open(jp(outdir, 'daal4py_cpp.cpp'), 'w') as f:
         f.write(cpp_cpp)
 
-
     with open(jp('src', 'gettree.pyx'), 'r') as f:
         pyx_gettree = f.read()
 
+    pyx_modelbuilder = ''
+    if 'algorithms::gbt::classification' in iface.namespace_dict and \
```
PetrovKP, line 949: I would transfer namespaces to dictionaries. See examples [here](https://github.com/IntelPython/daal4py/blob/master/generator/wrappers.py). This would make it easy to expand new builders without changing the generator.

owerbat (approved):

src/modelbuilder.pyx
```diff
@@ -0,0 +1,180 @@
...
+        '''
+        Create certain tree in the gradient boosted trees classification model for certain class
+
+        :param size_t nNodes: number of nodes in created tree
+        :param size_t classLabel: label of class for which tree is created. classLabel bellows interval from 0 to (nClasses - 1)
+        :rtype: tree identifier
+        '''
+        return self.c_ptr.createTree(nNodes, classLabel)
+
+    def add_leaf(self, c_gbt_clf_TreeId treeId, double response, c_gbt_clf_NodeId parentId=c_gbt_clf_noParent, size_t position=0):
```
PetrovKP, line 69: For variable names, you use `CamelCase`. For method names, you use `snake_case`. Why?

RukhovichIV: I thought that this is a right way, because the model builder for DF, which I used as an example was written like that. it'd be better to use `snake_case` everywhere I guess?

PetrovKP: I see file https://github.com/IntelPython/daal4py/blob/master/src/gettree.pyx for example.  For me, I need to make out in the same style. It is difficult to review without an example - because it is not clear how it will be used and how it will look. Can we merge after added examples and tests?

RukhovichIV: Examples will be ready soon. The point is that the examples also need examples for themselves which are not ready yet. Of course we can merge that in any order, but there's no much time remains before the code freeze. I'll try to push the first version of another PR till tomorrow's daily, so you could have a look on it. Will that suit you?

RukhovichIV: I've changed all name styles to `snake_case`. It all looks same now

src/modelbuilder.pyx
```diff
@@ -0,0 +1,180 @@
...
+    def create_tree(self, size_t n_nodes):
+        '''
+        Create certain tree in the gradient boosted trees regression model
+
+        :param size_t n_nodes: number of nodes in created tree
+        :rtype: tree identifier
+        '''
+        return self.c_ptr.createTree(n_nodes)
+
+    def add_leaf(self, c_gbt_reg_TreeId tree_id, double response, c_gbt_reg_NodeId parent_id=c_gbt_reg_noParent, size_t position=0):
```
PetrovKP, line 126:
```suggestion
    def add_leaf(self, c_gbt_reg_tree_id tree_id, double response, c_gbt_reg_node_id parent_id=c_gbt_reg_no_parent, size_t position=0):
```
?

RukhovichIV: done

PetrovKP (approved):

RukhovichIV:
@PetrovKP , could you merge this PR, as I don't have appropriate rights?

PR #278, opened 2020-09-02
Modification of K-Means Stopping Criteria

owerbat:
Stopping criteria in K-Means modified to be consistent with stock Scikit-learn. Also one Scikit-learn test has been added to deselected because of instability of the algorithm.

### DIFF
--- a/daal4py/sklearn/cluster/_k_means_0_23.py
+++ b/daal4py/sklearn/cluster/_k_means_0_23.py
@@ -123,7 +123,6 @@ def _daal4py_compute_starting_centroids(X, X_fptype, nClusters, cluster_centers_
 def _daal4py_kmeans_compatibility(nClusters, maxIterations, fptype = "double",
     method = "lloydDense", accuracyThreshold = 0.0, resultsToEvaluate = "computeCentroids"):
     kmeans_algo = None
-    accuracyThreshold = 1e-6 if accuracyThreshold == 0.0 else accuracyThreshold
     if daal_check_version((2020, 2), (2021, 107)):
         kmeans_algo = daal4py.kmeans(nClusters = nClusters,
             maxIterations= maxIterations,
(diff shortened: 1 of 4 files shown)


masdevas:
I think that we should use different "default epsilon" for float and double data types because of different precision. Also, if PR changes our behaviour under sklearn interfaces, @SmirnovEgorRu should check this request.

SmirnovEgorRu:
I suppose we need to change stopping criteria in DAAL itself, not to have a dirty hack on Scikit-learn patching side.

deselected_tests.yaml
```diff
@@ -72,3 +72,7 @@ deselected_tests:
   # We have difference (Max absolute difference: 1.97215226e-31) in computing of log_proba with np.log(y_proba)
   # Looks like using IDP NumPy is the cause for the failure due to use of more aggressive compiler optimization when compile NumPy UFunc loops
   - neural_network/tests/test_mlp.py::test_predict_proba_multilabel
+
+  # Different number of iterations because of instability of kmeans
```
PetrovKP, line 76: This is a bug, can we fix it?

daal4py/sklearn/cluster/_k_means_0_23.py
```diff
@@ -123,11 +123,13 @@ def is_string(s, target_str):
 def _daal4py_kmeans_compatibility(nClusters, maxIterations, fptype = "double",
     method = "lloydDense", accuracyThreshold = 0.0, resultsToEvaluate = "computeCentroids"):
     kmeans_algo = None
+    accuracyThreshold = 1e-6 if accuracyThreshold == 0.0 else accuracyThreshold
```
PetrovKP, line 126: We used a tolerance equal to 0. This logic should be left.

masdevas (approved):

PetrovKP (approved):

PR #286, opened 2020-09-16
Input Shape Check for kNN Predict

owerbat:


daal4py/sklearn/neighbors/knn.py
```diff
@@ -209,3 +213,60 @@ def predict(self, X):
         else:
             logging.info("sklearn.neighbors.KNeighborsClassifier.predict: " + method_uses_sklearn)
             return super(KNeighborsClassifier, self).predict(X)
+
+    def predict_proba(self, X):
```
Alexsandruss, line 217: Why this method is needed? It's optimized through self.kneighbors

daal4py/sklearn/neighbors/_base.py
```diff
@@ -0,0 +1,63 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ******************************************************************************/
+
+
+import numpy as np
+
+
+def _get_weights(dist, weights):
```
Alexsandruss, line 22: Why this function is needed if predict_proba will not be patched?

daal4py/sklearn/neighbors/knn.py
```diff
@@ -161,6 +162,9 @@ class KNeighborsClassifier(BaseKNeighborsClassifier, KNeighborsMixin):
     def predict(self, X):
         X = check_array(X, accept_sparse='csr')
 
+        if hasattr(self, 'n_features_in_') and X.shape[1] != self.n_features_in_:
```
Alexsandruss, line 165: Does 'n_features_in_' always exist and have correct value in this class?

daal4py/sklearn/neighbors/knn.py
```diff
@@ -209,3 +213,60 @@ def predict(self, X):
...
+        -------
+        p : ndarray of shape (n_queries, n_classes), or a list of n_outputs
+            of such arrays if n_outputs > 1.
+            The class probabilities of the input samples. Classes are ordered
+            by lexicographic order.
+        """
+        X = check_array(X, accept_sparse='csr')
+
+        if hasattr(self, 'n_features_in_') and X.shape[1] != self.n_features_in_:
+            raise ValueError('Input data shape {} is inconsistent with the trained model'.format(X.shape))
```
Alexsandruss, line 236: Try to move to kneighbors method

owerbat: Thanks, it should work

daal4py/sklearn/neighbors/knn.py
owerbat: As I saw in tests it doesn't always exist. That's why I'm using `getattr`

daal4py/sklearn/neighbors/knn.py
owerbat: Removed

daal4py/sklearn/neighbors/_base.py
owerbat: Removed

daal4py/sklearn/neighbors/knn.py
```diff
@@ -32,6 +32,11 @@
 
 class KNeighborsMixin(BaseKNeighborsMixin):
     def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
+        n_features = getattr(self, 'n_features_in_', None)
+        shape = getattr(X, 'shape', None)
+        if n_features and shape and shape[1] != n_features:
```
Alexsandruss, line 37: len(shape) might be 1 if 1d array is passed and error will be raised while trying to get shape[1]

daal4py/sklearn/neighbors/knn.py
```diff
@@ -161,6 +166,9 @@ class KNeighborsClassifier(BaseKNeighborsClassifier, KNeighborsMixin):
     def predict(self, X):
         X = check_array(X, accept_sparse='csr')
 
+        # if hasattr(self, 'n_features_in_') and X.shape[1] != self.n_features_in_:
+        #     raise ValueError('Input data shape {} is inconsistent with the trained model'.format(X.shape))
+
```
Alexsandruss, line 171: Delete or uncomment

daal4py/sklearn/neighbors/knn.py
owerbat: Additional check was added

Alexsandruss:
predict method of classifier should have same check.

Alexsandruss (approved):

PR #287, opened 2020-09-16
Add new deselected test for df

Pahandrovich:


PetrovKP:
It is necessary to add a comment that this test will be renamed in future versions.

agorshk (approved):

PR #288, opened 2020-09-17
Fit Method for kNN

owerbat:


daal4py/sklearn/neighbors/knn.py
```diff
@@ -30,6 +30,22 @@
 import logging
 
 
+def configure_algorithm(classifier, fptype, params):
```
PetrovKP, line 33: I see no reason return two algorithms here. I would write separately the functions:
```python
def training_algorithm(classifier, fptype, params):
...
```
and
```python
def prediction_algorithm(classifier, fptype, params):
...
```

daal4py/sklearn/neighbors/knn.py
```diff
@@ -163,6 +166,58 @@ def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
...
+                classes, self._y[:, k] = np.unique(y[:, k], return_inverse=True)
+                self.classes_.append(classes)
+
+            self.classes_ = self.classes_[0]
+            self._y = self._y.ravel()
+            self.n_samples_fit_ = X.shape[0]
+            self.n_features_in_ = X.shape[1]
+            self.effective_metric_ = 'euclidean'
+            self._fit_method = self.algorithm
+            self._fit_X = X
```
PetrovKP, line 201: Why store a training matrix?

daal4py/sklearn/neighbors/knn.py
```diff
@@ -163,6 +166,58 @@ def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
...
+            self.outputs_2d_ = False
+
+            self.classes_ = []
+            self._y = np.empty(y.shape, dtype=np.int)
+            for k in range(self._y.shape[1]):
+                classes, self._y[:, k] = np.unique(y[:, k], return_inverse=True)
+                self.classes_.append(classes)
+
+            self.classes_ = self.classes_[0]
+            self._y = self._y.ravel()
```
PetrovKP, line 196: Why store a training labels?

daal4py/sklearn/neighbors/knn.py
```diff
@@ -30,6 +30,22 @@
 import logging
 
 
+def configure_algorithm(classifier, fptype, params):
+    if classifier.algorithm == 'brute':
+        train_alg = d4p.bf_knn_classification_training
+        predict_alg = d4p.bf_knn_classification_prediction
+        # Brute force method always computes in doubles due to precision need
+        compute_fptype = 'double'
```
SmirnovEgorRu, line 38: Please explain.
cc @Alexsandruss

daal4py/sklearn/neighbors/knn.py
owerbat: Because it's using in `kneighbors()`

daal4py/sklearn/neighbors/knn.py
owerbat: Because it's using in stock `predict_proba()`

daal4py/sklearn/neighbors/knn.py
owerbat: Fixed

daal4py/sklearn/neighbors/knn.py
Alexsandruss: sklearn use same strategy - computation for brute-force always uses float64

daal4py/sklearn/neighbors/knn.py
```diff
@@ -163,6 +202,65 @@ def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
 
 
 class KNeighborsClassifier(BaseKNeighborsClassifier, KNeighborsMixin):
+    def fit(self, X, y):
+        X_incorrect_type = isinstance(X, NeighborsBase) or isinstance(X, BallTree) or isinstance(X, KDTree)
```
Alexsandruss, line 206: Minor: apply recommendation of CodeFactor
`Consider merging these isinstance calls to isinstance(X, (BallTree, KDTree, NeighborsBase)) (consider-merging-isinstance)`

daal4py/sklearn/neighbors/knn.py
```diff
@@ -163,6 +202,65 @@ def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
...
+
+        if not X_incorrect_type:
+            X, y = self._validate_data(X, y, accept_sparse="csr", multi_output=True)
```
Alexsandruss, line 209: Verify that `_validate_data` is not hotspot when pandas DataFrame/Series is passed. It might be better to handle these formats in another way.

daal4py/sklearn/neighbors/knn.py
owerbat: Fixed

daal4py/sklearn/neighbors/knn.py
owerbat: I checked it. `_validete_data` is about 1-3% from fitting time. So, I think we don't need to change something

PetrovKP (approved):

PR #290, opened 2020-09-18
[Draft] kNN 'auto' method tuning

SmirnovEgorRu:
@PetrovKP @Alexsandruss @owerbat 

I see 3 issues:
1) Performance was validated on C++ code. To be applicable on python also we should remove training of kd-tree from prediction stage.
2) 'auto' behavior in the patching is unclear for me. It looks like it's selected on original Scikit-learn side in fit and **my changes should be moved to our patching of fit method**.
3) I really don't like explicit setting 'double' fptype for brute-force. What strong reasons do we have for this? What is performance overheads from this?

Also, It's interesting to check performance Scikit-learn's 'ball_tree' against our 'brute_force' for # features > 13.

image
k=5, 20k samples for prediction, fptype=double, NT=SOA. The graphic shows only predict time, no fit.

### DIFF
--- a/daal4py/sklearn/neighbors/knn.py
+++ b/daal4py/sklearn/neighbors/knn.py
@@ -81,12 +81,25 @@ class KNeighborsMixin(BaseKNeighborsMixin):
         except ValueError:
             fptype = None
 
-        if daal_check_version((2020, 3)) and self._fit_method in ['brute', 'kd_tree'] \
+        if daal_check_version((2020, 3)) and self._fit_method in ['brute', 'kd_tree', 'auto'] \
         and (self.effective_metric_ == 'minkowski' and self.p == 2 or self.effective_metric_ == 'euclidean') \
         and fptype is not None and not sp.issparse(X):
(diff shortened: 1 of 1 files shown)


SmirnovEgorRu:
Generally speaking:
if # features >= 13 - brute-force is always better
if # features < 8 - kd-tree is always better
for [8, 12] range it depends on # of training samples

owerbat:
I've removed training from prediction stage in #288. So, I'm going to apply your changes there

SmirnovEgorRu:
@owerbat, good, thank you. Please check that 'auto' will work as expected (don't check the performance, but just method selection).

PetrovKP:
Your results are similar to the guys from sklearn, [the picture is the same](https://github.com/scikit-learn/scikit-learn/issues/8213). Only they still had a parameter in the account: the number of neighbors. As a result of the parameters, I do not really see the point of doing my own - since in sklearn 0.24 they did it in a more convenient way. I would recommend doing the same. See [here](https://github.com/scikit-learn/scikit-learn/pull/17148/files#diff-4ff1a647d9d4ef33358643ec58c914c7R437).

PR #301, opened 2020-09-24
Fix static analysis tools

Pahandrovich:


napetrov:
New  code introduce 3 code factor issues - suggest to fix them now . Overall looks good

daal4py/sklearn/decomposition/_pca_0_23.py
```diff
@@ -486,7 +486,7 @@ def transform(self, X):
             if self.mean_ is not None:
                 X = X - self.mean_
                 X_transformed = np.dot(X, self.components_.T)
```
PetrovKP, line 488: This is a mistake. It should be like this:
```python
        if self.mean_ is not None:
            X = X - self.mean_
        X_transformed = np.dot(X, self.components_.T)
        if self.whiten:
            X_transformed /= np.sqrt(self.explained_variance_)
        return X_transformed
```
Now there is a chance that you will return None. This is bad.

generator/format.py
```diff
@@ -150,7 +150,7 @@ def __init__(self, name, typ, const, dflt, inpt, algo, doc):
                     pd = (pydefaults[typ] if dflt == True else dflt).rsplit('::', 1)[-1]
                     sphinx_default = '"{}"'.format(pd) if typ == 'std::string' else '{}'.format(pd)
                     pydefault = ' = {}'.format(sphinx_default)
-                    cppdefault = ' = {}'.format(cppdefaults[typ] if dflt == True else dflt) if dflt != None else ''
+                    cppdefault = ' = {}'.format(cppdefaults[typ] if dflt == True else dflt)
```
PetrovKP, line 153:
```suggestion
                    cppdefault = ' = {}'.format(cppdefaults[typ] if dflt else dflt)
```

generator/gen_daal4py.py
```diff
@@ -679,7 +679,7 @@ def prepare_hlwrapper(self, ns, mode, func, no_dist, no_stream):
 
             for p in all_params:
                 pns, tmp = splitns(p)
-                if not tmp.startswith('_') and not ignored(pns, tmp):
+                if tmp != None and not tmp.startswith('_') and not ignored(pns, tmp):
```
PetrovKP, line 682:
```suggestion
                if tmp is not None and not tmp.startswith('_') and not ignored(pns, tmp):
```

PetrovKP (changes requested):

PetrovKP (approved):

napetrov (approved):

PR #306, opened 2020-09-28
Interoperability with dpctl

michael-smirnov:
Fix of the **Data parallel control** module integration (formerly known as DPPY, DPPL). All the code that depends on dpctl is moved to a separate module to avoid unwanted dependency in the big daal4py.

### DIFF
--- a/conda-recipe/bld.bat
+++ b/conda-recipe/bld.bat
@@ -1,8 +1,26 @@
 rem set NO_DIST=1
 
+set PATH=%PATH%;%PREFIX%\Library\bin\libfabric
+
+rem if dpc++ vars path is specified
+IF DEFINED DPCPPROOT (
+    echo "=============DPCPP==============="
+    call "%DPCPPROOT%\env\vars.bat"
+    set "CC=dpcpp.exe"
+    set "CXX=dpcpp.exe"
+)
+
+rem if DAALROOT not exists then provide PREFIX
+rem otherwise use library directly from the path provided
+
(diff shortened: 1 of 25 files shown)


PetrovKP (approved):
LGTM

setup.py
```diff
@@ -43,6 +43,8 @@
 mpi_root = None if no_dist else os.environ['MPIROOT']
 dpcpp = True if 'DPCPPROOT' in os.environ else False
 dpcpp_root = None if not dpcpp else os.environ['DPCPPROOT']
+dpctl = True if dpcpp and 'DPCTLROOT' in os.environ else False
```
napetrov, line 46: What if we will use dpctl from conda and not from dpctlroot?

napetrov:
Overall looks ok - and we might merge in this configuration.  The only question that is not clear - how we will work in case of dpctl use from conda

conda-recipe/bld.bat
```diff
@@ -1,8 +1,26 @@
...
+    set "CXX=dpcpp.exe"
+)
+
+rem if DAALROOT not exists then provide PREFIX
+rem otherwise use library directly from the path provided
+
+IF NOT DEFINED DAALROOT ( 
+    set DAALROOT=%PREFIX% 
+) ELSE (
+    call conda remove daal daal-include --force -y
```
napetrov, line 19: should we also remove daal-devel?

napetrov (approved):

PR #313, opened 2020-10-02
Fixed modelbuilder bug, which occured while parsing decision stumps

RukhovichIV:


RukhovichIV:
@PetrovKP, can you please merge this biggest PR on GitHub?)
A few days ago we discovered, that LightGBM doesn't enumerate tree node if the tree is a stump (the root is a leaf), but I believed it is, so this led to incorrect behaviour of our convertor. 
It's very strange, that we haven't met a single stump so far, despite we checked it on all our datasets (but not with the right parameters)

PetrovKP (approved):

PR #331, opened 2020-10-09
Dpctl import fix in examples

michael-smirnov:


napetrov (approved):
Good - assuming CI will be working

PR #334, opened 2020-10-12
Brute Force kNN example

Alexsandruss:


Alexsandruss:
Fails due to absence of BF kNN implementation in 2020.2 version of DAAL which is used in CI. Merge after 2020.3 release.

PetrovKP:
@Alexsandruss you can add a versioning check - then it automatically turns on - as the release appears.

see examples:
https://github.com/IntelPython/daal4py/blob/master/examples/run_examples.py#L104

PetrovKP:
@Mergifyio rebase

PetrovKP (approved):

PR #339, opened 2020-10-15
NearestNeighbors Patching

owerbat:

### DIFF
--- a/daal4py/sklearn/monkeypatch/dispatcher.py
+++ b/daal4py/sklearn/monkeypatch/dispatcher.py
@@ -74,18 +74,18 @@ from daal4py import __version__ as daal4py_version
 
 from daal4py.sklearn._utils import daal_check_version
 _mapping = {
-    'pca':           [[(decomposition_module, 'PCA', PCA_daal4py), None]],
-    'kmeans':        [[(cluster_module, 'KMeans', KMeans_daal4py), None]],
-    'distances':     [[(pairwise, 'pairwise_distances', daal_pairwise_distances), None]],
-    'linear':        [[(linear_model_module, 'LinearRegression', LinearRegression_daal4py), None]],
-    'ridge':         [[(linear_model_module, 'Ridge', Ridge_daal4py), None]],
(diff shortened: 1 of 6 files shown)


daal4py/sklearn/monkeypatch/dispatcher.py
```diff
@@ -81,7 +83,9 @@
     'lasso':         [[(linear_model_module, 'Lasso', Lasso_daal4py), None]],
     'svm':           [[(svm_module, 'SVC', SVC_daal4py), None]],
     'logistic':      [[(logistic_module, _patched_log_reg_path_func_name, daal_optimized_logistic_path), None]],
-    'knn_clsf':      [[(neighbors_module, 'KNeighborsClassifier', KNeighborsClassifier_daal4py), None]]
+    'knn_clsf':      [[(neighbors_module, 'KNeighborsClassifier', KNeighborsClassifier_daal4py), None]],
+    'nrst_nbrs':     [[(neighbors_module, 'NearestNeighbors', NearestNeighbors_daal4py), None]],
```
PetrovKP, line 87: These are the names that the user controls when choosing what he wants to patch. I would recommend giving full names.
```python
'nearest_neighbors':     [[(neighbors_module, 'NearestNeighbors', NearestNeighbors_daal4py), None]],
```

owerbat: Fixed

daal4py/sklearn/neighbors/_base.py
```diff
@@ -341,6 +341,9 @@ def _fit(self, X, y=None):
 
         weights = getattr(self, 'weights', 'uniform')
 
+        if self.algorithm == 'ball_tree' and (self.metric == 'minkowski' and self.p == 2 or self.metric == 'euclidean'):
+            self.algorithm = 'auto'
```
PetrovKP, line 345: If will not path for DAAL - we break stock sklearn

PetrovKP (approved):

Issue #316, opened 2020-10-04
Code quality improvements - adress  findings from www.codefactor.io


napetrov:
Please create PRs for addressing code quality issues reported by codefactor scans

https://www.codefactor.io/repository/github/intelpython/daal4py

Some problems are simple fixes and i would expect that they can be fixed across entire repo and not one by one
Other might be super complex and i don't think can be fixed without serious refactoring - cod complexity

PR #341, opened 2020-10-15
Resolve minor codefactor.io issues

cnguyen-uk:
# Description
This pull request resolves a few code quality issues found in codefactor.io, with the intention of making progress towards issue #316.

- Use `enumerate()` instead of `range()` and `len()`
- Remove unnecessary "else" after "return"

### DIFF
--- a/daal4py/sklearn/neighbors/_base.py
+++ b/daal4py/sklearn/neighbors/_base.py
@@ -208,33 +208,32 @@ def daal4py_kneighbors(estimator, X=None, n_neighbors=None, return_distance=True
 
     if not query_is_train:
         return results
+    # If the query data is the same as the indexed data, we would like
+    # to ignore the first nearest neighbor of every sample, i.e
+    # the sample itself.
+    if return_distance:
+        neigh_dist, neigh_ind = results
     else:
-        # If the query data is the same as the indexed data, we would like
-        # to ignore the first nearest neighbor of every sample, i.e
-        # the sample itself.
-        if return_distance:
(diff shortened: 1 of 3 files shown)


cnguyen-uk:
Those checks don't look too good - some pretty simple commits so now sure why there are so many check failures.

napetrov:
@cnguyen-uk  - build failing across all platforms 

File "/usr/share/miniconda/envs/CB/conda-bld/daal4py_1602760673071/work/generator/gen_daal4py.py", line 554, in <genexpr>
    if not any(tmp_input_args[i].name.endswith(x) for x in ordered):
TypeError: list indices must be integers or slices, not tuple

cnguyen-uk:
@napetrov Thanks for that, it was helpful. The fix that codefactor.io suggested wasn't a good one because the code following on wouldn't have been happy with it. To really fix the issue, we would need to use `enumerate()` instead of `range(len())`, as suggested by codefactor.io, but we would also need to rewrite the code after it. I'll leave that for now.

I've now submitted two new commits which undo the first set of changes in my PR's description. So really, this PR is just for one commit.

cnguyen-uk:
@napetrov Hold off from approving this - will attempt a small change which may solve the two issues.

cnguyen-uk:
@napetrov Changes made - good for a review once the tests (apart from Travis CI) pass.

napetrov:
@cnguyen-uk  Thanks. actually Travis was fixed, so it should pass here as well

napetrov (approved):

PR #344, opened 2020-10-16
Enable DPC++ build on windows

napetrov:
Enable DPC++ build on windows

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,3 +1,9 @@
+variables:
+  - name: WINDOWS_BASEKIT_URL
+    value: https://registrationcenter-download.intel.com/akdlm/irc_nas/16958/w_BaseKit_b_2021.1.9.2200.exe
+  - name: WINDOWS_DPCPP_COMPONENTS
+    value: intel.oneapi.win.dpcpp-compiler
+
 jobs:
 - job: DPCPP
   pool:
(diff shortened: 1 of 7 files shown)


azure-pipelines.yml
```diff
@@ -63,18 +69,21 @@ jobs:
...
     displayName: Add conda to PATH
-
   - script: conda create -q -y -n CB python=3.7 conda-build conda-verify
     displayName: Create Anaconda environment
-
   - script: |
       call activate CB
+      set "DPCPPROOT=C:\Program Files (x86)\Intel\oneAPI\compiler\latest"
+      set "DAALROOT=C:\Program Files (x86)\Intel\oneAPI\daal\latest"
+      set "TBBROOT=C:\Program Files (x86)\Intel\oneAPI\tbb\latest"
```
PetrovKP, line 87: I would suggest separating build from testing in order to catch problems with the compiler RT. Because the user doesn't need runtime to run the cpu part.

napetrov: I think we should do both in ideal case. make build + testing with RT . and then do testing without RT.

tomashek: Why are you setting/assuming certain system paths here and not using the dal-devel and tbb-devel conda package paths in the python environment?

napetrov: Because there is no 2021 version of dal-devel or tbb in conda

- https://anaconda.org/intel/daal-devel/labels

setup.py
```diff
@@ -110,6 +110,12 @@
         DPCPP_LIBDIRS = [jp(dpcpp_root, 'linux', 'lib')]
     elif IS_WIN:
         DPCPP_LIBDIRS = [jp(dpcpp_root, 'windows', 'lib')]
+        _project_compiler = "dpcpp-cl"   # "clang-cl"
```
Alexander-Makaryev, line 113: @napetrov Looks like `dpcpp-cl` is not exists in beta10 any more, use dpcpp `instead`.

napetrov: switched to dpcpp

PetrovKP (approved):

PR #350, opened 2020-10-22
Improving  CI build and test

napetrov:
# Description

-
-
-

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -76,10 +76,10 @@ jobs:
 
   - script: |
       call activate CB
-      conda build --override-channels -c conda-forge -c intel --numpy=1.17 --output-folder=. conda-recipe
+      conda build --override-channels -c intel -c conda-forge --numpy=1.17 --output-folder=. conda-recipe
     displayName: conda build
   - task: PublishPipelineArtifact@0
     inputs:
       artifactName: 'conda build dir'
-      targetPath: 'C:\Miniconda\envs\CB\conda-bld'
+      targetPath: '.'
     condition: always()
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -40,7 +40,7 @@ jobs:
       sudo chown -R $USER $CONDA
       conda config --set always_yes yes --set changeps1 no
       conda update -q conda
-      conda create -n CB python=3.7 daal daal-include cython mpich jinja2 numpy clang-tools scikit-learn lightgbm xgboost pandas -c intel -c conda-forge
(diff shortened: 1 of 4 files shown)


conda-recipe/meta.yaml
```diff
@@ -43,6 +43,7 @@ test:
         - pandas
         - scipy
         - scikit-learn
+        - threadpoolctl
```
PetrovKP, line 46: It should be like a temporary solution. because in runtime we shouldn't require this dependency from the user.

azure-pipelines.yml
```diff
@@ -76,5 +76,10 @@ jobs:
...
   - script: |
       call activate CB
-      conda build --override-channels -c conda-forge -c intel --numpy=1.17 --output-folder=. conda-recipe
+      conda build --override-channels -c intel -c conda-forge --numpy=1.17 --output-folder=. conda-recipe
     displayName: conda build
+  - task: PublishPipelineArtifact@0
+    inputs:
+      artifactName: 'conda build dir'
+      targetPath: '.'
+    condition: always()
```
PetrovKP, line 85: \n

azure-pipelines.yml
```diff
@@ -76,5 +76,10 @@ jobs:
 
   - script: |
       call activate CB
-      conda build --override-channels -c conda-forge -c intel --numpy=1.17 --output-folder=. conda-recipe
+      conda build --override-channels -c intel -c conda-forge --numpy=1.17 --output-folder=. conda-recipe
```
PetrovKP, line 79: it will be necessary to update it everywhere as well.

conda-recipe/run_test.bat
```diff
@@ -0,0 +1,25 @@
+ 
+set DAAL4PY_VERSION=%PKG_VERSION%
+set MPIROOT=%PREFIX%\Library
+
+IF DEFINED DPCPPROOT (
```
PetrovKP, line 5: Are we sure that the DPCPP compiler is required as runtime always?

conda-recipe/meta.yaml
napetrov: yes. But this is not a runtime but test env. this will have no impact on users

conda-recipe/run_test.bat
napetrov: it's not required - but in case we have dpcpproot passed - it will be used

PetrovKP: and if it is not? for example, when testing a cpu part where there is no runtime.

conda-recipe/meta.yaml
PetrovKP: ok

azure-pipelines.yml
PetrovKP: I'll do it later then.

PetrovKP (approved):

PR #354, opened 2020-10-23
Support of different library names

Alexsandruss:
# Description
 - Add branching for different library names: libdaal*.so and libonedal*.so.
 - Add support of major version in Win library.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,10 +55,22 @@ def get_lib_suffix():
                 return 'daal'
         return None
 
+    def walk_libdir():
+        global lib_dir
+
+        for _, _, libs in os.walk(lib_dir):
+            for lib in libs:
+                if 'onedal_core' in lib:
+                    return 'onedal'
+                if 'daal_core' in lib:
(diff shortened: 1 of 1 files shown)


napetrov:
Looks good  - if this works with new changes in DAL

napetrov (approved):

setup.py
```diff
@@ -28,9 +28,54 @@
...
 
+try:
+    from ctypes.utils import find_library
+except ImportError:
+    from ctypes.util import find_library
+
+
+def get_lib_suffix():
+    def walk_ld_library_path():
+        ld_library_path = os.environ.get('LD_LIBRARY_PATH', None)
```
PetrovKP, line 41: I think it not work for windows. If the DAAL is at the conda, then it will not be in this variable. Right?

Alexsandruss: `lib_dir` search is added to cover this

PetrovKP (approved):

PR #357, opened 2020-10-24
Update to correct oneDAL naming

napetrov:
# Description
Cleaning up oneDAL naming in code

https://github.com/oneapi-src/oneDAL/pull/1195

### DIFF
--- a/daal4py/sklearn/_utils.py
+++ b/daal4py/sklearn/_utils.py
@@ -71,11 +71,11 @@ def make2d(X):
 
 def get_patch_message(s):
     if s == "daal":
-        message = "uses optimized solver"
+        message = "uses Intel(R) oneAPI Data Analytics Library solver"
     elif s == "sklearn":
         message = "uses original Scikit-learn solver"
     elif s == "sklearn_after_daal":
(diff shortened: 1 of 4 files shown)


daal4py/sklearn/_utils.py
```diff
@@ -71,11 +71,11 @@ def make2d(X):
 
 def get_patch_message(s):
     if s == "daal":
-        message = "uses Intel® DAAL solver"
+        message = "uses optimized solver"
```
SmirnovEgorRu: My prefer to specify daal4py/oneDAL/Intel® oneDAL here: "uses daal4py solver" / "uses oneDAL solver" . Because it's unclear what "optimized solver" is.

Also, if we change the message we need to change daal4py documentation with description of verbosity: https://intelpython.github.io/daal4py/sklearn.html#scikit-learn-verbose

PetrovKP (approved):

daal4py/sklearn/_utils.py
SmirnovEgorRu: I'm oaky with new "Intel(R) oneAPI Data Analytics Library" variant, thanks. But need to update docs anyway :)

PR #368, opened 2020-10-30
Fix n_jobs problem in LogisticRegression

OnlyDeniko:


PetrovKP (approved):

YannDubs:
why is n_jobs not supported / had to be fixed ? It would be good to raise a warning for users to be aware of it also ...

YannDubs:
@PetrovKP @OnlyDeniko 
Note that the docs say that all parameters can be used: https://intel.github.io/scikit-learn-intelex/algorithms.html they should clarify that `n_jobs` cannot

PR #390, opened 2020-11-13
Reduce time for weights calculation

masdevas:
Drop extra work on weights calculation in cases when classes weights list contains ones.


SmirnovEgorRu:
Overall - good. But, please, provide perf gains info and fix CI fails.

PetrovKP:
Do other algorithms have such a problem? If not, maybe it is worth copying the solution from there or adding this part to the general one to support all algorithms?

Pahandrovich (approved):
LGTM

PetrovKP (approved):

PR #435, opened 2020-12-22
Update scikit-learn version for support with daal4py

PetrovKP:


azure-pipelines.yml
```diff
@@ -59,7 +59,9 @@ jobs:
       sudo chown -R $USER $CONDA
       conda config --set always_yes yes --set changeps1 no
       conda update -q conda
-      conda create -n CB python=3.7 daal daal-include cython mpich jinja2 numpy clang-tools scikit-learn lightgbm xgboost pandas pytest -c conda-forge -c intel
+      conda create -n CB python=3.7
+      conda install -n CB -q cython scipy pytest pandas pyyaml joblib numpydoc jinja2 numpy clang-tools lightgbm scikit-learn mpich
```
owerbat, line 63: Is it ok that we used packages from conda-forge before and now we use packages from default channel?

PetrovKP: Yes, it doesn't matter which channel we download packages from, except for daal.

owerbat (approved):

PR #436, opened 2020-12-23
Changing SVD to Covariance in PCA

owerbat:


deselected_tests.yaml
```diff
@@ -145,3 +145,5 @@ deselected_tests:
   - inspection/tests/test_partial_dependence.py::test_partial_dependence_dataframe[features-string-None-estimator-brute]
   - ensemble/tests/test_bagging.py::test_estimators_samples_deterministic
 
+  # Eigenvalues in the last component aren't equal because of different implementations of PCA
```
PetrovKP, line 148: write here for what reason the test failed. diverge accuracy?

daal4py/sklearn/decomposition/_pca.py
```diff
@@ -193,8 +193,12 @@ def _fit(self, X):
                 self._fit_svd_solver = 'full'
 
         if self._fit_svd_solver == 'full':
-            logging.info("sklearn.decomposition.PCA.fit: " + get_patch_message("daal"))
-            return self._fit_full(X, n_components)
+            if X.shape[1]/X.shape[0] < 2:
```
PetrovKP, line 196:
```suggestion
            if X.shape[1] / X.shape[0] < 2:
```
and fix CodeFactor comment

daal4py/sklearn/decomposition/_pca.py
```diff
@@ -85,15 +85,15 @@ def _fit_full_daal4py(self, X, n_components):
             daal_n_components = n_components
 
         fpType = getFPType(X)
-        centering_algo = daal4py.normalization_zscore(
-            fptype=fpType, doScale=False)
+
+        covariance_algo = daal4py.covariance(fptype=fpType, outputMatrixType='covarianceMatrix')
```
PetrovKP, line 89: I thought normalization was happening on the side of the PCA? And what is called is covariance, not correlation by default. This is not true?

owerbat: Yes, SVD PCA uses normalization, but covariance/correlation PCA hasn't got normalization as a parameter.
And yes, this is covariance. I use it to be consistent with stock scikit-learn. But in C++ correlation is still default method

deselected_tests.yaml
owerbat: Some values in `PCA.components_` (in the last component) are not equal. Also, results are not stable

PetrovKP: How much on average?

owerbat: 0.6 on average for absolute error in this test

PetrovKP:
Attach private CI with changes in oneDAL & daal4py for PCA

PetrovKP:
You have all tests for PCA. You need to add check version of daal. Right?

deselected_tests.yaml
PetrovKP: Add this info to comments

owerbat:
> You have all failed tests for PCA. You need to add check version of daal. Right?

I think it fails because PCA in oneDAL is not updated yet. It still computes correlation, not covariance

SmirnovEgorRu:
Need to think about threshold when select 'randomized' and 'full' in case of 'auto'
```py
elif n_components >= 1 and n_components < .8 * min(X.shape):
        self._fit_svd_solver = 'randomized'
```
Probably 0.8 can be reduced significantly

daal4py/sklearn/decomposition/_pca.py
```diff
@@ -17,6 +17,7 @@
 
 import numpy as np
 import numbers
+from numpy.lib.function_base import cov
```
PetrovKP, line 20: do you use this function?

PetrovKP (approved):

PR #453, opened 2021-01-13
Log reg fit was patched

masdevas:
Log reg fit was patched


daal4py/sklearn/linear_model/_logistic_path_0_22.py
```diff
@@ -1012,29 +1009,176 @@ def daal4py_predict(self, X, resultsToEvaluate):
     raise ValueError('resultsToEvaluate must be in [computeClassLabels, \
                      computeClassProbabilities, computeClassLogProbabilities]')
 
+def daal4py_fit(self, X, y, sample_weight=None):
```
PetrovKP, line 1012: does it make sense to rename the file?

PetrovKP, line 1012: not critical, but it is better to name it with the prefix daal4py, what will call directly daal4py alforithms

daal4py/sklearn/linear_model/_logistic_path_0_22.py
```diff
@@ -1012,29 +1009,176 @@ def daal4py_predict(self, X, resultsToEvaluate):
...
+    self : object
+    """
+    solver = _check_solver(self.solver, self.penalty, self.dual)
+    daal_ready = solver in ['lbfgs', 'newton-cg'] and not sparse.issparse(X)
+    daal_ready = daal_ready and sample_weight is None and self.class_weight is None
+
+    if daal_ready:
+        logging.info("sklearn.linear_model.LogisticRegression.fit: " + get_patch_message("daal"))
+    else:
+        logging.info("sklearn.linear_model.LogisticRegression.fit: " + get_patch_message("sklearn"))
```
PetrovKP, line 1039: in this place we can call the base class (original function) and not copy a lot of code next. right?

daal4py/sklearn/linear_model/_logistic_path_0_22.py
```diff
@@ -1012,29 +1009,176 @@ def daal4py_predict(self, X, resultsToEvaluate):
...
+
```
PetrovKP, line 1040: I think we are unnecessarily copying so much code. Leave code branches that are only needed for call of daal4py

daal4py/sklearn/linear_model/_logistic_path_0_22.py
```diff
@@ -21,12 +21,12 @@
 import numbers
```
PetrovKP, line 21: 1200+ lines! I think it makes sense to split the file. Good practice

daal4py/sklearn/monkeypatch/dispatcher.py
```diff
@@ -96,7 +88,9 @@ def _get_map_of_algorithms():
         'fin_check':                [[(validation, '_assert_all_finite', _daal_assert_all_finite), None]],
         'tsne':                     [[(manifold_module, 'TSNE', TSNE_daal4py), None]],
     }
-    if daal_check_version((2021,'P', 100)):
+    if LooseVersion(sklearn_version) < LooseVersion("0.22"):
+        mapping['logistic'] = [[(logistic_module, _patched_log_reg_path_func_name, daal_optimized_logistic_path), None]]
+    else:
         mapping['log_reg'] = [[(linear_model_module, 'LogisticRegression', LogisticRegression_daal4py), None]]
```
PetrovKP, line 94:
```suggestion
        mapping['logistic'] = [[(linear_model_module, 'LogisticRegression', LogisticRegression_daal4py), None]]
```

daal4py/sklearn/monkeypatch/dispatcher.py
```diff
@@ -96,7 +88,9 @@ def _get_map_of_algorithms():
         'fin_check':                [[(validation, '_assert_all_finite', _daal_assert_all_finite), None]],
         'tsne':                     [[(manifold_module, 'TSNE', TSNE_daal4py), None]],
     }
-    if daal_check_version((2021,'P', 100)):
+    if LooseVersion(sklearn_version) < LooseVersion("0.22"):
```
PetrovKP, line 91: Does it make sense to support version 0.21? It's very hard?

daal4py/sklearn/utils/fixes.py
```diff
@@ -0,0 +1,19 @@
+from functools import update_wrapper
```
PetrovKP, line 1: Copyright + I don't really understand why you need this file and functions? Why not take these functions from Scikit-learn?

PetrovKP (commented):
You have a lot of falls in public CI. You should green public CI

PetrovKP:
@Pahandrovich , Please see this PR for security issues.

daal4py/sklearn/linear_model/_logistic_path_0_22.py
masdevas: daal4py_fit is called only directly from LogisticRegression class

daal4py/sklearn/monkeypatch/dispatcher.py
masdevas: I don't think that performance improve for old versions is critical for us

PetrovKP: If merging the files 0.21 and 0.22 is not difficult, I would do it.

daal4py/sklearn/linear_model/logistic_loss.py
```diff
@@ -88,7 +88,7 @@ def _daal4py_loss_and_grad(beta, objF_instance, X, y, n):
     beta_ = make2d(beta)
     res = objF_instance.compute(X, y, beta_)
     gr = res.gradientIdx.ravel()
-    gr = gr * n # force copy
+    gr = gr.astype(np.float64) * n # force copy
```
PetrovKP, line 91: if gr is double type - will copying take place here?

daal4py/sklearn/linear_model/_logistic_0_22.py
```diff
@@ -1012,29 +895,132 @@ def daal4py_predict(self, X, resultsToEvaluate):
...
+    """
+    solver = _check_solver(self.solver, self.penalty, self.dual)
+    daal_ready = solver in ['lbfgs', 'newton-cg'] and not sparse.issparse(X)
+    daal_ready = daal_ready and sample_weight is None and self.class_weight is None
+
+    if daal_ready:
+        logging.info("sklearn.linear_model.LogisticRegression.fit: " + get_patch_message("daal"))
+    else:
+        logging.info("sklearn.linear_model.LogisticRegression.fit: " + get_patch_message("sklearn"))
+        return LogisticRegression_original.fit(self, X, y, sample_weight=sample_weight)
```
PetrovKP, line 926: from 918-926 correctly (see function name) transfer to method fit

PR #455, opened 2021-01-14
Fix supported types of KNN predict_proba

OnlyDeniko:


Alexsandruss (approved):

daal4py/sklearn/neighbors/_base.py
```diff
@@ -366,6 +366,8 @@ def stock_fit(self, X, y):
 class KNeighborsMixin(BaseKNeighborsMixin):
     def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
         daal_model = getattr(self, '_daal_model', None)
+        if not X is None:
```
PetrovKP, line 369: See [here](https://stackoverflow.com/questions/2710940/python-if-x-is-not-none-or-if-not-x-is-none)

```suggestion
        if X is not None:
```

PetrovKP (approved):

PR #464, opened 2021-01-21
Fix memory leak in Numeric Table creation

Alexsandruss:
# Description
Fix memory leak in SOA Numeric Table creation from Pandas DataFrame with Fortran data order by handling columns iterator and its slices correctly.

### DIFF
--- a/src/daal4py.cpp
+++ b/src/daal4py.cpp
@@ -434,7 +434,7 @@ daal::data_management::NumericTablePtr make_nt(PyObject * obj)
                     daal::data_management::SOANumericTablePtr soatbl;
 
                     // iterate over columns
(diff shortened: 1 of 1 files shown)


SmirnovEgorRu:
@PetrovKP, do we want to add some tests to track such issues?

PetrovKP:
> @PetrovKP, do we want to add some tests to track such issues?

I don't know how easy it will be to do it. @Alexsandruss, what do you think?

Alexsandruss:
Easy, if it's acceptable to add dependency to memory profiler like [pympler](https://github.com/pympler/pympler).

PetrovKP (approved):

PR #476, opened 2021-01-26
Continue of fix unitest issue

OnlyDeniko:
Previous fix (https://github.com/IntelPython/daal4py/pull/470) is not enough to fix problem


conda-recipe/run_test.sh
```diff
@@ -18,7 +18,18 @@
...
-python -c "import daal4py"
-mpirun -n 4 python -m unittest discover -v -s tests -p spmd*.py
-pytest --pyargs daal4py/sklearn/
-python tests/run_tests.py
+ok=0
+
+if ! python -c "import daal4py"; then
+    ok=1
+fi
+if ! mpirun -n 4 python -m unittest discover -v -s tests -p spmd*.py; then
```
PetrovKP, line 26: daal CI have build without MPI support. So, we need to skip MPI check in this case. For example:
```python
if ! $NO_DISTR; then
...
fi
```

conda-recipe/run_test.sh
```diff
@@ -18,7 +18,18 @@
 # limitations under the License.
 #*******************************************************************************
 
-python -c "import daal4py"
-mpirun -n 4 python -m unittest discover -v -s tests -p spmd*.py
-pytest --pyargs daal4py/sklearn/
-python tests/run_tests.py
+ok=0
+
+if ! python -c "import daal4py"; then
```
PetrovKP, line 23: For check of last return code exist `$?`
```suggestion
python -c "import daal4py"
ok += $?
```

tests/run_tests.py
```diff
@@ -21,4 +21,4 @@
     test_suite = unittest.defaultTestLoader.discover('tests', 'test*.py')
     test_runner = unittest.TextTestRunner()
     result = test_runner.run(test_suite)
-    sys.exit(not result.wasSuccessful())
+    sys.exit(len(result.errors) != 0 or not result.wasSuccessful())
```
PetrovKP, line 24: Do we now need a file after discovered a problem?

conda-recipe/run_test.sh
OnlyDeniko: What if exit code is not equal 0 or 1?

tests/run_tests.py
OnlyDeniko: No

PetrovKP (approved):

PR #477, opened 2021-01-27
Update pipline for docs

PetrovKP:


PetrovKP:
\azr help

PetrovKP:
After the build of daal4py, the documentation was generated correctly
[daal4py-documentation.zip](https://github.com/IntelPython/daal4py/files/5886739/daal4py-documentation.zip)

.ci/pipeline/docs.yml
```diff
@@ -16,26 +16,47 @@ pr:
     - doc
     - examples
 
+variables:
+  - name: python.version
+    value: '3.7'
+  - name: daal4py.version
+    value: '2021.2.0'
```
napetrov, line 23: I think we will have problems here and will forgot to update version here.

.ci/pipeline/docs.yml
```diff
@@ -16,26 +16,47 @@ pr:
     - doc
     - examples
 
+variables:
+  - name: python.version
+    value: '3.7'
```
napetrov, line 21: we use 3.8 as default in other places - should we use 3.7 here?

.ci/pipeline/docs.yml
```diff
@@ -16,26 +16,47 @@ pr:
...
-  inputs:
-    artifactName: 'daal4py-documentation'
-    targetPath: '$(Build.ArtifactStagingDirectory)/html'
+  - task: UsePythonVersion@0
+    displayName: 'Use Python $(python.version)'
+    inputs:
+      versionSpec: '$(python.version)'
+  - script: |
+      conda update -y -q conda
+      conda create -q -y -n CB -c intel python=$(python.version) daal-devel
```
napetrov, line 34: Suggest to change this to dal-devel package name

.ci/pipeline/docs.yml
PetrovKP: For the conda now we use python37 everywhere, due to the fact that the daal is the freshest in the Intel channel now.

.ci/pipeline/docs.yml
```diff
@@ -16,26 +16,47 @@ pr:
...
+      pip install -r requirements-dev.txt
+      pip install -r requirements-doc.txt
+    displayName: 'Install requirements'
+  - script: |
+      . /usr/share/miniconda/etc/profile.d/conda.sh
+      conda activate CB
+      export DALROOT=$CONDA_PREFIX
+      export PKG_VERSION=$(daal4py.version)
+      export NO_DIST=1
+      python setup.py install --single-version-externally-managed --record=record.txt
```
napetrov, line 45: Just as alternative option - call docbuild after regular daal4py build instead of building daal4py separately?

.ci/pipeline/docs.yml
PetrovKP: agree

.ci/pipeline/docs.yml
PetrovKP: Can we set up a trigger for a sub-task? Or always run the rules?

napetrov: yes - we can setup triggers

PetrovKP: docs build and attach for 5-10 sec. I think - we don't need a trigger :)

napetrov (approved):

PR #482, opened 2021-01-30
Check available of host and cpu devices for sycl examples

PetrovKP:
Host and Cpu devices can be not available in python package and there is a failed: 

```
python daal4py\examples\sycl\sklearn_sycl.py
********************************************************************************
device context: host
  File "daal4py\examples\sycl\sklearn_sycl.py", line 179, in <module>
    with get_context(device):
  File "C:\Users\kpetrov\Miniconda3\envs\py36\lib\contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "src\oneapi\oneapi.pyx", line 78, in sycl_context
  File "src\oneapi\oneapi.pyx", line 42, in _oneapi.sycl_execution_context.__cinit__
RuntimeError: No device of requested type available. Please check https://software.intel.com/en-us/articles/intel-oneapi-dpcpp-compiler-system-requirements-beta -1 (CL_DEVICE_NOT_FOUND)
```
These changes do a check for the availability of devices and does not run examples if there is no package

### DIFF
--- a/examples/run_examples.py
+++ b/examples/run_examples.py
@@ -62,20 +62,20 @@ availabe_devices = []
 try:
     from daal4py.oneapi import sycl_context
     sycl_extention_available = True
-except:
+except RuntimeError:
     sycl_extention_available = False
 
 if sycl_extention_available:
     try:
         with sycl_context('gpu'):
             availabe_devices.append("gpu")
-    except:
+    except RuntimeError:
         pass
     availabe_devices.append("host")
     try:
         with sycl_context('cpu'):
             availabe_devices.append("cpu")
-    except:
+    except RuntimeError:
         pass
 
 
--- a/examples/run_examples.py
+++ b/examples/run_examples.py
@@ -14,176 +14,207 @@
 # limitations under the License.
 #===============================================================================
 
-# daal4py Scikit-Learn examples for GPU
-# run like this:
-#    python -m daal4py ./sklearn_sycl.py
+import os
+import struct
+import subprocess
+import sys
 
-import numpy as np
+from daal4py import __has_dist__
+from daal4py.sklearn._utils import get_daal_version
 
-from sklearn.cluster import KMeans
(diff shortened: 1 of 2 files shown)


examples/run_examples.py
```diff
@@ -62,18 +62,25 @@
...
             availabe_devices.append("gpu")
-    except:
-        gpu_available = False
-    availabe_devices.append("host")
-    availabe_devices.append("cpu")
+    except RuntimeError:
+        pass
+    try:
+        with sycl_context('host'):
+            availabe_devices.append("host")
```
michael-smirnov, line 76: This means nothing since `req_device` contains only one requirement for one example to run only on gpu.
The suggestion is to add [here](https://github.com/IntelPython/daal4py/blob/f4b79297a4661d4a32e7cbe87933bafbdd5012b4/examples/run_examples.py#L132) generating requirements on host and cpu devices for all sycl examples. In that case, you would not need to change sklearn example at all.

michael-smirnov:
General question: how proper is it not to raise any error or warning if no host device is available?

PetrovKP:
@michael-smirnov 
> General question: how proper is it not to raise any error or warning if no host device is available?

I think the behavior should be aligned in the case when the GPU is not found.

napetrov:
> 
> 
> @michael-smirnov
> 
> > General question: how proper is it not to raise any error or warning if no host device is available?
> 
> I think the behavior should be aligned in the case when the GPU is not found.
I don't think that for cpu/host we should have same path as for GPU. There are might be cases there you don't have GPU, but you always having CPU. 

The problem is that in this setup in case of some major problems with DPC++ environment we will have green examples in both customer systems and our testing. 

Might be we should consider separation of sycl and normal examples?  In case you are running sycl examples we can expect that they are failing without dpc++ env..

michael-smirnov:
I would say at least the host device should be present always with DPC++ RT.

PR #502, opened 2021-02-04
Update dpctl interop

vlad-nazarov:


michael-smirnov (approved):
I think that the next step should be removing the dependency on dpctl from the C++-Cython level.

PetrovKP:
@michael-smirnov , @vlad-nazarov 
Can we add a dpctl to build in the public CI? For checked and testing of work for only build

PR #503, opened 2021-02-04
Unnecessary Check Removing

owerbat:


OnlyDeniko:
Will these changes affect support for types such as int(8,16,32,64), uint(8,16,32,64)?

owerbat:
@OnlyDeniko @PetrovKP I don't think so because there is another one `check_array` in `_fit()` method which has dtype check

PetrovKP (approved):

PR #509, opened 2021-02-08
CI for coverity

PetrovKP:


azure-pipelines.yml
```diff
@@ -0,0 +1,26 @@
+# Starter pipeline
```
napetrov, line 1: Just in case - i suggest not to add this pipeline to the root of repo.

PetrovKP: It is test version

PetrovKP:
Approve! Thanks you

.ci/pipeline/coverity.yml
```diff
@@ -0,0 +1,27 @@
...
+  COVERITY_TOOL_HOME: $(Agent.BuildDirectory)/cov-analysis
+  DESCRIPTION: Nightly
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- script: |
+    cd $(Agent.BuildDirectory)
+    wget https://scan.coverity.com/download/linux64 --post-data "token=$(COVERITY_AUTHKEY)&project=daal4py" -O coverity_tool.tgz
```
napetrov, line 11: Coverity recommendation to get this cashed instead of downloading each time from their server. 
For example you can use this as an example - https://github.com/oneapi-src/oneDAL/blob/master/.ci/pipeline/ci.yml#L83

.ci/pipeline/coverity.yml
```diff
@@ -0,0 +1,27 @@
+variables:
+  COVERITY_TOOL_HOME: $(Agent.BuildDirectory)/cov-analysis
+  DESCRIPTION: Nightly
```
napetrov, line 3: I would put some additional info here  as for now we will have all scan with same description


image

napetrov (approved):
Overall we can enable this

.ci/pipeline/coverity.yml
KalyanovD: But it has different upload date and proposed to used only for nightly runs for a while.

.ci/pipeline/coverity.yml
KalyanovD: We will enable caching at the nearest time, thanks for information and example. For now it is more preferable to enable continious scan.

PR #512, opened 2021-02-09
Upgrade of run-to-run tests

OnlyDeniko:


SmirnovEgorRu:
- log_proba has some issues, e.g. nan when prob == 0. It was the reason why I didn't add this.
- I offer to add both binary and multiclass classification data sets. Because at least LogReg and SVC have different code paths for them.

OnlyDeniko:
> * log_proba has some issues, e.g. nan when prob == 0. It was the reason why I didn't add this.
> * I offer to add both binary and multiclass classification data sets. Because at least LogReg and SVC have different code paths for them.

We have already had both data sets.

PR #517, opened 2021-02-12
output sycl and devices availability status to logs

napetrov:
Output sycl and devices availability

### DIFF
--- a/examples/run_examples.py
+++ b/examples/run_examples.py
@@ -62,7 +62,7 @@ availabe_devices = []
 try:
     from daal4py.oneapi import sycl_context
     sycl_extention_available = True
-except:
+except ModuleNotFoundError:
     sycl_extention_available = False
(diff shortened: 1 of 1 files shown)


examples/run_examples.py
```diff
@@ -62,18 +62,34 @@
...
-    except:
+    except RuntimeError:
         gpu_available = False
     availabe_devices.append("host")
     availabe_devices.append("cpu")
+    #validate that host and cpu devices avaialbe for logging reasons. Examples and
+    #vaidaton logic assumes that host and cpu devices are always available
+    try:
+        with sycl_context('cpu'):
+            cpu_available = True
```
michael-smirnov, line 82: Why we cannot use `availabe_devices` list to detect host and device availability?

napetrov: for now avaialbe devices for cpu/host get assignment regardless actual support

PetrovKP (approved):

PR #536, opened 2021-02-25
change modin conversion condition

PivovarA:
Add new modin to numpy conversion condition
```python
import sys
if "modin" in sys.modules:
    # do manipulation with modin
```

### DIFF
--- a/generator/wrapper_gen.py
+++ b/generator/wrapper_gen.py
@@ -83,11 +83,6 @@ except ImportError:
     class pdSeries:
         pass
 
-if "modin" in sys.modules:
-    from modin import pandas as mdpd
-    mdDataFrame = mdpd.DataFrame
-    mdSeries = mdpd.Series
-
 npc.import_array()
 
 hpat_spec = []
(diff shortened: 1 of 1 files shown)


SmirnovEgorRu:
Please check 2 scenarios:
1) Modin is installed, but not imported in application - no import of modin and ray processes.
2) Modin is installed and imported - daal4py works with md.DataFrame properly.

PivovarA:
> Please check 2 scenarios:
> 
> 1. Modin is installed, but not imported in application - daal4py doesn't import modin and doesn't create ray processes.
> 2. Modin is installed and imported - daal4py works with md.DataFrame properly.

I can do this, but do we really need a **modin_installed** check ? Because if modin has been imported already then it is installed too. 
Also I'm afraid that we need ```try import modin``` for  **modin_installed** check.

SmirnovEgorRu:
@PivovarA, I missed you :)
Just to clarify - I don't have concerns to the code. I just want to test the scenarios above manually.

generator/wrapper_gen.py
```diff
@@ -81,15 +83,10 @@ class pdDataFrame:
...
-    mdSeries = pandas.Series
-except ImportError:
-    class mdDataFrame:
-        pass
-    class mdSeries:
-        pass
+if "modin" in sys.modules:
+    from modin import pandas as mdpd
+    mdDataFrame = mdpd.DataFrame
+    mdSeries = mdpd.Series
```
SmirnovEgorRu, line 89: Are they local or global variables? Are they alive after leaving "if" statement?

PetrovKP: I agree with Egor, there is no need for a global search

PetrovKP:
+ pleas, fix PEP8:

./generator/wrapper_gen.py:192:1: W293 blank line contains whitespace

generator/wrapper_gen.py
PivovarA: Agree. I have replaced it already

PivovarA:
> @PivovarA, I missed you :)
> Just to clarify - I don't have concerns to the code. I just want to test the scenarios above manually from user's perspective.

I tested modin support on kmeans example. 
If modin is not installed, everything works fine. CI is showing it now. 
If the modin is installed but not imported, then import modin inside daal4py will not happen

PetrovKP (approved):

SmirnovEgorRu (approved):

PR #539, opened 2021-02-26
define versions for different pythons

napetrov:
# Description
Use pip ability to specify pythons 

https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers


napetrov:
@PetrovKP  @Pahandrovich  - let's merge this assuming it will pass testing?

PetrovKP (approved):

PR #566, opened 2021-03-17
Appropriate error handling for LightGBM model converter

RukhovichIV:
Added an explicit error messages for cases when LightGBM model has categorical features since oneDAL model builder doesn't support such an interface

### DIFF
--- a/src/gbt_convertors.pyx
+++ b/src/gbt_convertors.pyx
@@ -70,9 +70,13 @@ def get_gbt_model_from_lightgbm(model: Any) -> Any:
             continue
 
         # add root
+        feat_val = struct["threshold"]
+        if isinstance(feat_val, str):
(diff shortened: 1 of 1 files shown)


src/gbt_convertors.pyx
```diff
@@ -70,9 +70,13 @@ def get_gbt_model_from_lightgbm(model: Any) -> Any:
             continue
 
         # add root
+        feat_val = struct["threshold"]
```
PetrovKP, line 73: minnor: struct bad name, github even emphasizes as a keyword

PetrovKP (approved):
LGTM

src/gbt_convertors.pyx
RukhovichIV: renamed to `tree_struct` :D

PR #581, opened 2021-03-24
Fix recipe testing

OnlyDeniko:


Pahandrovich (approved):

OnlyDeniko:
http://intel-ci.intel.com/eb906e15-9376-f15f-a746-8cdcd4b723a1
http://intel-ci.intel.com/eb9073d5-f0a9-f1ff-9cad-8cdcd4b723a1

examples/sycl/log_reg_binary_dense_batch.py
```diff
@@ -135,7 +135,7 @@ def cpu_context():
     assert result_classic.prediction.shape == (predict_data.shape[0],
                                                train_labels.shape[1])
 
-    assert np.allclose(result_classic.prediction, result_cpu.prediction)
+    assert np.allclose(result_classic.prediction, result_cpu.prediction, atol=1e-5)
```
michael-smirnov, line 138: It is very strange why we have differences between "classic" and "host" results. Can you please comment on what caused such a difference? Is it due to run2run instability?

OnlyDeniko: Yes, it is due to run2run instability in LogisticRegression

PetrovKP: We will fix this later, but for now we want green private CI

michael-smirnov: I think it's worth placing a comment or TODO here on why the results are not exactly the same - to avoid a possible confusion of users.

PetrovKP (approved):

michael-smirnov (approved):

PR #582, opened 2021-03-24
Fix import sklearnex problem

Pahandrovich:

### DIFF
--- a/sklearnex/dispatcher.py
+++ b/sklearnex/dispatcher.py
@@ -15,12 +15,9 @@
 # limitations under the License.
 #===============================================================================
 
-from daal4py.sklearn import patch_sklearn as patch_sklearn_orig
-from daal4py.sklearn import unpatch_sklearn as unpatch_sklearn_orig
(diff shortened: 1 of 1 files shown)


.ci/pipeline/ci.yml
```diff
@@ -43,6 +43,8 @@ jobs:
       . /usr/share/miniconda/etc/profile.d/conda.sh
       conda activate CB
       ./conda-recipe/run_test.sh
+      pip uninstall -y scipy scikit-learn
+      python -c "import daal4py.sklearn"
```
napetrov, line 47: shouldn't this be in script instead of direct specification in pipeline ?

PetrovKP (approved):

PetrovKP:
@Mergifyio backport rls/2021.2.1-rls

.ci/pipeline/ci.yml
Pahandrovich: Now I remove this lines, because daal4py.sklearn assumes the presence of scikit-learn module.

PR #654, opened 2021-05-18
Fix memory test due to wrong leak in Logistic Regression

OnlyDeniko:


sklearnex/tests/test_memory_usage.py
```diff
@@ -79,28 +79,30 @@ def dataframe_f(x, y):
 ]
 
 
-def gen_clsf_data():
+def gen_clsf_data(estimator_name):
```
Alexsandruss, line 82: It's better to count algorithm extra memory usage outside of data generation function

sklearnex/tests/test_memory_usage.py
```diff
@@ -79,28 +79,30 @@ def dataframe_f(x, y):
...
     data, label = make_classification(
         n_samples=2000, n_features=50, random_state=777)
-    return data, label, \
-        data.size * data.dtype.itemsize + label.size * label.dtype.itemsize
+    data_size = data.size * data.dtype.itemsize + label.size * label.dtype.itemsize
+    if estimator_name == 'SVC':
+        data_size += 2048 * label.size * label.dtype.itemsize
```
Alexsandruss, line 87: Why 2048? Isn't SVC used memory depending on some parameter?

sklearnex/tests/test_memory_usage.py
```diff
@@ -79,28 +79,30 @@ def dataframe_f(x, y):
...
     for train_index, test_index in kf.split(x):
         if isinstance(x, np.ndarray):
             x_train, x_test = x[train_index], x[test_index]
-            y_train, y_test = y[train_index], y[test_index]
+            y_train = y[train_index]
         elif isinstance(x, pd.core.frame.DataFrame):
             x_train, x_test = x.iloc[train_index], x.iloc[test_index]
-            y_train, y_test = y.iloc[train_index], y.iloc[test_index]
+            y_train = y.iloc[train_index]
+        mem_before, _ = tracemalloc.get_traced_memory()
```
Alexsandruss, line 105: Memory tracing should be outside of `kf.split` cycle. For every loop, memory leak grows and becomes more visible. Previous cases, when full input data leaked, were easy to find, but future leaks might be more sneaky so this cycle is used to find them

Alexsandruss (changes requested):

sklearnex/tests/test_memory_usage.py
```diff
@@ -109,6 +110,7 @@ def _kfold_function_template(estimator, data_transform_function):
             alg.transform(x_test)
         elif hasattr(alg, 'kneighbors'):
             alg.kneighbors(x_test)
+        gc.collect()
     del alg, x_train, x_test, y_train, y_test
     mem_after, _ = tracemalloc.get_traced_memory()
```
Alexsandruss, line 115:
```suggestion
    del alg, x_train, x_test, y_train, y_test
    # TODO: add memory tracking before garbage collector as debug message
    gc.collect()
    mem_after, _ = tracemalloc.get_traced_memory()
```

Alexsandruss (approved):

PR #660, opened 2021-05-20
Fix problem with big estimators; refactoring tests for random forest

PetrovKP:


OnlyDeniko (approved):

daal4py/sklearn/ensemble/tests/test_decision_forest.py
```diff
@@ -31,16 +31,63 @@
     import RandomForestRegressor as DaalRandomForestRegressor
 from daal4py.sklearn._utils import daal_check_version
 
-N_TRIES = 10
-ACCURACY_RATIO = 0.85 if daal_check_version((2021, 'P', 200)) else 0.7
```
agorshk, line 35: if an user takes oneDal version lower than 2021.2, will the test failed in sklearnex?

PetrovKP: yes, but we guarantee compatibility only with the previous release version. if user want to run test with oneDal version 2021.1 for example, you need tests with daal4py 2021.1 :)

agorshk: ok, in this case no problem

agorshk (approved):
Check CI please

PR #662, opened 2021-05-21
Fix perf issues for svm's in new API

PetrovKP:


onedal/svm/svm.py
```diff
@@ -152,16 +152,20 @@ def _fit(self, X, y, sample_weight, Computer):
...
-
         X, y = _check_X_y(
             X, y, dtype=[np.float64, np.float32], force_all_finite=True)
         y = self._validate_targets(y, X.dtype)
         sample_weight = _get_sample_weight(
             X, y, sample_weight, self.class_weight_, self.classes_, self.svm_type)
 
-        self._scale_, self._sigma_ = self._compute_gamma_sigma(self.gamma, X)
+        if self.kernel == 'linear':
+            self._scale_, self._sigma_ = 1.0, 1.0
```
owerbat, line 162: Maybe in linear case we should set `self.coef0 = 0.0` too?

owerbat (approved):

PR #664, opened 2021-05-25
Export enableThreadPinning

cmsxbc:
# Description

Some application may get performance improvement by binding threads.And there is a way for `daal` to enable it which is not export to `daal4py`.

- export `daal::services::Environment::getInstance()->enableThreadPinning` to daal4py.


PetrovKP:
\intelci: run

PetrovKP (approved):
Thanks you!

PR #673, opened 2021-06-03
Fix cache size and shrinking parameters in onedal iface

PetrovKP:


michael-smirnov (commented):
@PetrovKP Don't you think that some tests on shrinking and cache_size have to be added as well?

PetrovKP:
@michael-smirnov both parameters do not functionally affect the algorithm, only the performance

michael-smirnov:
@PetrovKP performance is critical for us. I think it is worth to think how to test that all parameter values are correctly passed to oneDAL C++ API

onedal/svm/backend/svm_py.cpp
```diff
@@ -39,11 +39,11 @@ KernelDescriptor get_kernel_params(const svm_params &params) {
 }
 
 template <typename Result, typename Descriptor, typename... Args>
-Result compute_descriptor_impl(Descriptor descriptor, const svm_params &params, Args &&...args) {
+Result compute_descriptor_impl(Descriptor descriptor, const svm_params &params, Args &&... args) {
     using Task = typename Result::task_t;
     descriptor.set_accuracy_threshold(params.accuracy_threshold)
         .set_max_iteration_count(params.max_iteration_count)
-        .set_cache_size(params.shrinking)
+        .set_cache_size(params.cache_size)
```
SmirnovEgorRu, line 46: :D

michael-smirnov (approved):

PR #694, opened 2021-06-13
Replace mnist to adult dataset for SVM example

PetrovKP:


SmirnovEgorRu:
I think better to just add the additional notebook, not replace mnist.

SmirnovEgorRu (approved):

PR #709, opened 2021-06-29
Refactoring algos before scikit-learn 1.0

OnlyDeniko:


daal4py/sklearn/cluster/_dbscan.py
```diff
@@ -220,8 +224,8 @@ def fit(self, X, y=None, sample_weight=None):
 
         Returns
         -------
-        self
-
+        self : object
+            Returns a fitted instance of self.
         """
```
PetrovKP, line 229: please, define `_daal_ready` before `check_array `:

```python
        _daal_ready = self.algorithm in ['auto', 'brute'] and \
            (self.metric == 'euclidean' or (
             self.metric == 'minkowski' and self.p == 2)) and \
            not issparse(X)
       if _daal_ready :
           X = check_array(X, dtype=[np.float64, np.float32])
           ...
       else:
           super().fit(X, y, sample_weight=sample_weight)
```

PetrovKP (approved):

PR #710, opened 2021-06-29
replaced label with response

makart19:
replaced label with response for svm

### DIFF
--- a/onedal/svm/backend/svm_py.cpp
+++ b/onedal/svm/backend/svm_py.cpp
@@ -38,9 +38,7 @@ KernelDescriptor get_kernel_params(const svm_params &params) {
 
     if constexpr (std::is_same_v<typename KernelDescriptor::tag_t,
                                  sigmoid_kernel::detail::descriptor_tag>) {
-        return KernelDescriptor{}
-            .set_scale(params.scale)
-            .set_shift(params.shift);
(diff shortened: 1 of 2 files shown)


onedal/svm/backend/svm_py.cpp
```diff
@@ -320,7 +312,7 @@ void svm_infer<Task>::infer(PyObject *data,
                      .set_biases(biases_table);
     if constexpr (std::is_same_v<Task, svm::task::classification> ||
                   std::is_same_v<Task, svm::task::nu_classification>) {
-        model.set_first_class_label(0).set_second_class_label(1);
+        model.set_first_class_response(0).set_second_class_response(1);
```
michael-smirnov, line 315: But old methods with `_label` naming will work too? (until the next major release?)

michael-smirnov (commented):
If you started renaming labels to responses, please do it in pyx, pxi and py files as well

onedal/svm/backend/svm_py.cpp
makart19: yes, they were marked as deprecated (see https://github.com/oneapi-src/oneDAL/pull/1755/checks?check_run_id=2942887627)

PetrovKP (approved):

PR #730, opened 2021-07-08
Fix `get_number_of_types` failing with pandas dtypes

Yard1:
# Description
`get_number_of_types` will throw an exception if it checks a dataframe with just one dtype, and that dtype is a pandas one (eg. CategoricalDtype). As those are not numpy dtypes, it will assume it's an iterable, and raise a TypeError when it cannot be converted to set.

- Replace the check in `get_number_of_types` with a more robust try-except approach


PetrovKP (approved):
Thanks for fixing the problem!

PR #739, opened 2021-07-13
Add instruction set function

PetrovKP:


src/daal4py.cpp
```diff
@@ -829,6 +829,27 @@ extern "C"
...
+            daal::services::Environment::getInstance()->setCpuId(daal::avx512);
+        }
+        else if (instructions_set == "avx2")
+        {
+            daal::services::Environment::getInstance()->setCpuId(daal::avx2);
+        }
+        else if (instructions_set == "sse2")
+        {
+            daal::services::Environment::getInstance()->setCpuId(daal::sse2);
+        }
```
SmirnovEgorRu, line 846: Why not all possible combinations?

PR #766, opened 2021-08-09
[RF] Partial revert for fixing the test

OnlyDeniko:


daal4py/sklearn/ensemble/_forest.py
```diff
@@ -704,17 +704,28 @@ def predict_proba(self, X):
             The class probabilities of the input samples. The order of the
             classes corresponds to that in the attribute :term:`classes_`.
         """
+        X = check_array(
```
agorshk, line 707: If we fall back to stock Sklearn than there is a double check of X array

agorshk: After conversation, we decided stay it here

agorshk (approved):

Issue #767, opened 2021-08-10
Invalid results when using KNeighborsClassifier on a benchmark datasets with sklearnex 2021.3.0


ogrisel:
**Describe the bug**

Using the following benchmark script. I observe 0 classification accuracy instead of the expected ~86%.

**To Reproduce**

- `bench_knn.py`:

```python
import numpy as np
from subprocess import check_output
import threadpoolctl
from pprint import pprint
from time import perf_counter
from pathlib import Path
import sklearn
from sklearn.neighbors import KNeighborsClassifier
from sklearn.datasets import make_classification
from sklearn.model_selection import train_test_split


sklearn_path = Path(sklearn.__file__).parent
sklearn_version = sklearn.__version__
sklearn_git_branch = check_output(
    f'(cd {sklearn_path} && git symbolic-ref --short HEAD) || echo ""',
    shell=True,
).decode("utf-8").strip()
if sklearn_git_branch:
    sklearn_version = f"{sklearn_version} (git: {sklearn_git_branch})"
print(f"sklearn {sklearn_version} at {sklearn_path}")
pprint(threadpoolctl.threadpool_info())

X, y = make_classification(n_samples=int(2e6), n_features=100, random_state=0)
X_train, X_test, y_train, y_test = train_test_split(
    X,
    y,
    train_size=int(1e6),
    test_size=int(1e3),
    random_state=0,
)
clf = KNeighborsClassifier(n_neighbors=10, n_jobs=1).fit(X_train, y_train)
tic = perf_counter()
y_pred = clf.predict(X_test)
toc = perf_counter()
assert y_pred.shape == y_test.shape
accuracy = np.mean(y_pred == y_test)
print(
    f"n_samples_train: {X_train.shape[0]}, n_samples_test: {X_test.shape[0]}, acc:"
    f" {accuracy:.3f}, duration: {toc - tic:.3f} s"
)
```

**Expected behavior**

Get a high enough classification accuracy (80% or so). See below:

**Output/Screenshots**

- Output for the reference scikit-learn version

```
(sklearn-0.24.2) $ python bench_knn.py 
sklearn 0.24.2 at /scratch/ogrisel/miniforge3/envs/sklearn-0.24.2/lib/python3.9/site-packages/sklearn
[{'architecture': 'Sandybridge',
  'filepath': '/scratch/ogrisel/miniforge3/envs/sklearn-0.24.2/lib/libopenblasp-r0.3.17.so',
[...]

PR #768, opened 2021-08-11
Train_test_split fix and test for specific subset sizes

Alexsandruss:
# Description
Fix for incorrect work of train_test_split when `train_size + test_size < input_size`:
 - number of indices from RNG is corrected
 - test for case is added
 
Merged https://github.com/oneapi-src/oneDAL/pull/1817 are required for full fix.

Fixes intel/scikit-learn-intelex#767 issue.

### DIFF
--- a/daal4py/sklearn/model_selection/tests/test_split.py
+++ b/daal4py/sklearn/model_selection/tests/test_split.py
@@ -18,6 +18,7 @@ import numpy as np
 import pytest
 from sklearn.model_selection import train_test_split as skl_train_test_split
 from daal4py.sklearn.model_selection import _daal_train_test_split as d4p_train_test_split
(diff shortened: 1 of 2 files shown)


PetrovKP:
Fails in public CI, need greening

daal4py/sklearn/model_selection/tests/test_split.py
```diff
@@ -27,6 +28,8 @@
 
 @pytest.mark.parametrize('n_samples', N_SAMPLES)
 def test_results_similarity(n_samples):
+    if not daal_check_version((2021, 'P', 400)):
```
PetrovKP, line 31: use skip tests from pytest functonality

PetrovKP (approved):
after minor fix

PR #772, opened 2021-08-17
Device support in Intel(R) Extension for Scikit-learn*

michael-smirnov:
- Introduce device support in `onedal` module by passing a queue directly to the methods
- Introduce new config APIs in `sklearnex` module: `set_config`, `get_config`, `config_context`
- Patch these configs in original scikit-learn
- Implement global options for device offload: `target_offload` and `allow_fallback_to_host`, recognizable by all sklearnex algorithms
- Introduce `dpctl.tensor.usm_ndarray` support for all algorithms in `sklearnex`: now they can consume and produce data located on target device and offload computations using compute-follows-data strategy!


daal4py/sklearn/_utils.py
```diff
@@ -151,3 +152,113 @@ def get_number_of_types(dataframe):
...
+
+
+def _create_global_queue_from_sycl_context():
+    import sys
+    d4p_target = None
+    if 'daal4py.oneapi' in sys.modules:
+        from daal4py.oneapi import _get_device_name_sycl_ctxt
+        d4p_target = _get_device_name_sycl_ctxt()
+
+    if d4p_target is not None:
```
PetrovKP, line 164: Will it work correctly if there are two identical devices, for example 2 GPUs? Because you create a new queue knowing only the name of the device

daal4py/sklearn/_utils.py
```diff
@@ -151,3 +152,113 @@ def get_number_of_types(dataframe):
...
+        try:
+            from dpctl import SyclQueue
+        except ImportError:
+            raise RuntimeError("dpctl need to be installed for device offload")
+        return SyclQueue(d4p_target if d4p_target != 'host' else 'cpu')
```
PetrovKP, line 169: Why is there a limitation on the host device?

daal4py/sklearn/_utils.py
```diff
@@ -151,3 +152,113 @@ def get_number_of_types(dataframe):
...
+
+
+def _transfer_to_host(queue, *data):
+    has_usm_data = False
+
+    host_data = []
+    for item in data:
+        usm_iface = getattr(item, '__sycl_usm_array_interface__', None)
+        if usm_iface is not None:
+            import dpctl.memory as dp_mem
```
PetrovKP, line 180: Check available dpctl?

daal4py/sklearn/_utils.py
```diff
@@ -151,3 +152,113 @@ def get_number_of_types(dataframe):
...
+                                       'on single target device')
+            else:
+                queue = usm_iface['syclobj']
+
+            buffer = dp_mem.as_usm_memory(item).copy_to_host()
+            item = np.ndarray(shape=usm_iface['shape'],
+                              dtype=usm_iface['typestr'],
+                              buffer=buffer)
+            has_usm_data = True
+        elif has_usm_data and item is not None:
```
PetrovKP, line 195: In what case would it be false? Do you need this flag?

sklearnex/dispatcher.py
```diff
@@ -45,6 +51,15 @@ def get_patch_map():
         mapping['svc'] = [[(svm_module, 'SVC', SVC_sklearnex), None]]
         mapping['nusvr'] = [[(svm_module, 'NuSVR', NuSVR_sklearnex), None]]
         mapping['nusvc'] = [[(svm_module, 'NuSVC', NuSVC_sklearnex), None]]
+        mapping['set_config'] = [[(base_module,
```
PetrovKP, line 54: When one or several algorithms are patched, the config will not be patched. Configs will only be patched in two versions:
1. `patch_sklearn()`
2. `patch_sklearn(["set_config", "get_config", "config_context"])`

I think we need to patch these functions in other cases.

sklearnex/svm/svc.py
```diff
@@ -41,46 +39,58 @@ def __init__(self, *, C=1.0, kernel='rbf', degree=3, gamma='scale',
...
-                "sklearn.svm.SVC.decision_function: " + get_patch_message("sklearn"))
-            return sklearn_SVC.decision_function(self, X)
+        return dispatch(self, 'svm.SVC.decision_function', {
+            'onedal': self.__class__._onedal_decision_function,
+            'sklearn': sklearn_SVC.decision_function,
+        }, X)
+
+    def _onedal_gpu_supported(self, method_name, *data):
+        if method_name == 'svm.SVC.fit':
+            if len(data) > 1:
```
PetrovKP, line 71: else? or is this condition necessary at all?

sklearnex/svm/svc.py
```diff
@@ -41,46 +39,58 @@ def __init__(self, *, C=1.0, kernel='rbf', degree=3, gamma='scale',
...
+                import numpy as np
+                self._class_count = len(np.unique(data[1]))
+            return self.kernel in ['linear', 'rbf'] and \
+                self.class_weight is None and \
+                self._class_count == 2
```
PetrovKP, line 76: and check CSR support, for new API we don't have support CSR

sklearnex/_utils.py
```diff
@@ -32,14 +32,22 @@ def set_sklearn_ex_verbose():
...
-def get_patch_message(s):
+def get_patch_message(s, queue=None):
     import sys
     if s == "onedal":
         message = "running accelerated version on "
-        if 'daal4py.oneapi' in sys.modules:
+        if queue is not None:
+            if queue.sycl_device.is_gpu:
+                message += 'GPU'
+            elif queue.sycl_device.is_cpu:
```
PetrovKP, line 42: or host?

sklearnex/_utils.py
```diff
@@ -32,14 +32,22 @@ def set_sklearn_ex_verbose():
...
+                message += 'CPU'
+            else:
+                raise RuntimeError('Unsupported device')
```
PetrovKP, line 45: is this an unreachable branch of code?

sklearnex/_device_offload.py
```diff
@@ -0,0 +1,108 @@
+#===============================================================================
```
PetrovKP, line 1: I am a little confused that the auxiliary functionality lies at the root of the project. Can I put them in a utils folder?

sklearnex/_config.py
```diff
@@ -0,0 +1,110 @@
+#===============================================================================
```
PetrovKP, line 1: Why name file _configpy? not config.py? Seem this contain public functions?

daal4py/sklearn/_utils.py
michael-smirnov: There is no way to process this case without dpctl. So this is requirement to have it

sklearnex/_config.py
michael-smirnov: But this file is not supposed to be imported by user. This is aligned with original scikit: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/_config.py

sklearnex/_utils.py
michael-smirnov: No - for example, we do not support FPGA

sklearnex/svm/svc.py
michael-smirnov: This condition is needed to initialize `_class_count` when fit support is checked. This function is also used in predict, decision_function, etc., where `_class_count` is already initialized by prior fit call. 
I can add `else` branch with exception throwing.

daal4py/sklearn/_utils.py
PetrovKP: For  user application? Then we always carry dependency DPCPP, and again, for all applications - include only CPU applications, our library will have size more than 1GB. There are other options?

michael-smirnov:
CI: http://intel-ci.intel.com/ec0fcc92-b89d-f172-869b-8cdcd4b723a1

napetrov (approved):

michael-smirnov:

PR #789, opened 2021-09-09
Update documentation of gpu section

michael-smirnov:
6e219563-2df5-4389-b9ae-8f223a81cb76


michael-smirnov:

napetrov:
@outoftardis  - appreciate if you can take a look on this.

outoftardis:
> @outoftardis - appreciate if you can take a look on this.

@napetrov I'll review this today

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,100 @@
...
+
+   .. note::
+     All the data passed to the algorithm must reside on the same device.
+
+   .. warning::
+     Only base methods like :code:`fit`, :code:`predict`, and :code:`transform`
+     can consume :code:`usm_ndarray`. Please pay attension that only the
+     algorithms in Intel(R) Extension for Scikit-learn* provide support of
+     :code:`usm_ndarray`. Standard algorithms in stock version of Scikit-learn*
+     do not include this feature.
```
outoftardis, line 48:
```suggestion
     The :code:`usm_ndarray` can only be consumed by the base methods
     like :code:`fit`, :code:`predict`, and :code:`transform`.
     Note that only the algorithms in Intel(R) Extension for Scikit-learn* support
     :code:`usm_ndarray`. The algorithms from the stock version of scikit-learn
     do not support this feature.
```

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,100 @@
...
+2. Use global configurations of Intel(R) Extension for Scikit-learn*:
+     1. Option :code:`target_offload` can be used to set the device primarily
+        used to perform computations. Accepted values are :code:`str` and
+        :code:`dpctl.SyclQueue`. If string, expected to be "auto" (the execution
+        context is deduced from input data location), or SYCL* filter selector
+        string. Default value is "auto"
+     2. Option :code:`allow_fallback_to_host`
+        is a boolean flag, that, if set, allows to fallback computation to host
+        device in case when particular estimator does not support the selected
+        one. Default value is :code:`False`.
```
outoftardis, line 58: `Accepted values` -- you really mean `accepted data types`, no?

I also don't think my changes below are correct, because it's not clear what happens if the input is dpctl.SyclQueue.

```suggestion
     1. The :code:`target_offload` option can be used to set the device primarily
        used to perform computations. Accepted values are :code:`str` and
        :code:`dpctl.SyclQueue`. The default value is "auto", meaning that the execution
        context is deduced from the location of input data. Otherwise, a string with a
        SYCL* filter selector is expected.
     2. The :code:`allow_fallback_to_host` option
        is a Boolean flag. If set to :code:`True`, the computation is allowed to fallback to the host
        device when a particular estimator does not support the selected
        device. The default value is :code:`False`.
```

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,100 @@
...
+
+These options can be set using :code:`sklearnex.set_config()` function or
+:code:`sklearnex.config_context`. Current values of these options can be
+obtained by the call of :code:`sklearnex.get_config()`.
+
+.. note::
+     Functions :code:`set_config`, :code:`get_config` and :code:`config_context`
+     are always patched after :code:`sklearnex.patch_sklearn()` call.
```
outoftardis, line 66:
```suggestion
     are always patched after the :code:`sklearnex.patch_sklearn()` call.
```

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,100 @@
...
+
+DPC++ compiler runtime can be installed either from PyPI or Anaconda:
+
+- Install from PyPI::
+
+     pip install dpcpp-cpp-rt
+
+- Install from Anaconda::
+
+     conda install dpcpp_cpp_rt -c intel
```
outoftardis, line 100: lines 89-100 should be placed at the beginning of this topic and clearly marked as "requirements" or "prerequisites"

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,100 @@
...
+
+These options can be set using :code:`sklearnex.set_config()` function or
+:code:`sklearnex.config_context`. Current values of these options can be
+obtained by the call of :code:`sklearnex.get_config()`.
+
+.. note::
+     Functions :code:`set_config`, :code:`get_config` and :code:`config_context`
+     are always patched after :code:`sklearnex.patch_sklearn()` call.
+
+Example how to patch your code with Intel CPU/GPU optimizations:
```
outoftardis, line 68:
```suggestion
.. rubric:: Example

An example on how to patch your code with Intel CPU/GPU optimizations:
```

or

```suggestion
Example
#######

An example on how to patch your code with Intel CPU/GPU optimizations:
```

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,100 @@
...
+   X = np.array([[1., 2.], [2., 2.], [2., 3.],
+               [8., 7.], [8., 8.], [25., 80.]], dtype=np.float32)
+   with config_context(target_offload="gpu:0"):
+      clustering = DBSCAN(eps=3, min_samples=2).fit(X)
+
+.. warning::
+     For compatibility reasons, algorithms in Intel(R) Extension for
+     Scikit-learn* can be offloaded to the device with use of
+     :code:`daal4py.oneapi.sycl_context`. It is recommended to use the ways
+     described above for device offloading instead of using :code:`sycl_context`.
```
outoftardis, line 86: I think this should be placed before the example. I also don't think this is a warning, you may place this in the main text body. You may use a section header (e.g. "Compatibility considerations" or something similar) to make it stand out.

I also have questions about this paragraph.
- `compatibility reasons` -- compatibility with what? daal4py? something else?
- are you saying that the user may use sycl_context if they have to, but we do not recommend this?

outoftardis (commented):
I propose you have a clearer structure to this topic by introducing sections such as prerequisites, compatibility considerations, example. This means that you'd also need a section for the main information in this topic. What about "Offloading computations"?

doc/sources/oneapi_gpu.rst
vlad-nazarov: `I also don't think my changes below are correct, because it's not clear what happens if the input is dpctl.SyclQueue.`
It means that if the data type is a string, then it can take the values "auto" or SYCL * filter selector. For the dpctl type.SyclQueue there are no restrictions

doc/sources/oneapi_gpu.rst
vlad-nazarov: * `compatibility reasons -- compatibility with what? daal4py? something else?` 
I will clarify this question and formulate the reason more precisely
* `are you saying that the user may use sycl_context if they have to, but we do not recommend this?`
Yes, we still have sycl_context support, but it is preferable to use the methods suggested in the device offloading paragraph

vlad-nazarov: `Compatibility reasons`  - compatibility with users scripts. You can use `sycl_context`, but it's not recommended.

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,107 @@
...
+
+These options can be set using :code:`sklearnex.set_config()` function or
+:code:`sklearnex.config_context`. To obtain the current values of these options,
+call :code:`sklearnex.get_config()`.
+
+.. note::
+     Functions :code:`set_config`, :code:`get_config` and :code:`config_context`
+     are always patched after the :code:`sklearnex.patch_sklearn()` call.
+
+**Compatibility considerations**::
```
outoftardis, line 85:
```suggestion
.. rubric:: Compatibility considerations
```

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,107 @@
...
+
+For compatibility reasons, algorithms in Intel(R) Extension for
+Scikit-learn* can be offloaded to the device with use of
```
outoftardis, line 88:
```suggestion
Scikit-learn* may be offloaded to the device using
```

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,107 @@
...
+:code:`daal4py.oneapi.sycl_context`. It is recommended to use the ways
+described above for device offloading instead of using :code:`sycl_context`.
```
outoftardis, line 90:
```suggestion
:code:`daal4py.oneapi.sycl_context`. However, it is recommended to use one of the options
described above for device offloading instead of using :code:`sycl_context`.
```

doc/sources/oneapi_gpu.rst
```diff
@@ -0,0 +1,107 @@
...
+     like :code:`fit`, :code:`predict`, and :code:`transform`.
+     Note that only the algorithms in Intel(R) Extension for Scikit-learn* support
+     :code:`usm_ndarray`. The algorithms from the stock version of scikit-learn
+     do not support this feature.
+2. Use global configurations of Intel(R) Extension for Scikit-learn\*:
+     1. The :code:`target_offload` option can be used to set the device primarily
+        used to perform computations. Accepted data types are :code:`str` and
+        :code:`dpctl.SyclQueue`. If string, expected to be "auto" (the execution
+        context is deduced from input data location), or SYCL* filter selector
+        string. Default value is "auto"
```
outoftardis, line 71:
```suggestion
        :code:`dpctl.SyclQueue`. If you pass a string to :code:`target_offload`,
        it should either be ``"auto"``, which means that the execution
        context is deduced from the location of input data, or a string
        with SYCL* filter selector. The default value is ``"auto"``.
```

outoftardis: @vlad-nazarov please take a look at this wording

vlad-nazarov: @outoftardis, this wording looks good to me

vlad-nazarov (approved):

PR #803, opened 2021-10-04
Fix requirements for python 3.6

Pahandrovich:
scikit-learn 1.0 is not support python3.6

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,4 +1,5 @@
 pytest==6.2.5
-scikit-learn==1.0
+scikit-learn==1.0 ; python_version >= '3.7'
+scikit-learn==0.24 ; python_version <= '3.6'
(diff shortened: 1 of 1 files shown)


requirements-test.txt
```diff
@@ -1,4 +1,5 @@
 pytest==6.2.5
-scikit-learn==1.0
+scikit-learn==1.0 ; python_version >= '3.7'
+scikit-learn==0.24 ; python_version <= '3.6'
```
agorshk, line 3: According [Sklearn doc](https://scikit-learn.org/stable/install.html): "Scikit-learn 0.22 supported Python 3.5-3.8. Scikit-learn 0.23 - 0.24 require Python 3.6 or newer. Scikit-learn 1.0 and later requires Python 3.7 or newer."

agorshk (approved):

PR #810, opened 2021-10-09
Fix release CI [part 2]

Pahandrovich:


vmeshche:
what problem are we solving here?

PR #828, opened 2021-10-21
Update atol in sycl linreg example

vlad-nazarov:


napetrov:
are we going to merge this?

PR #841, opened 2021-10-28
XGBoost model builder fix

KalyanovD:
# Description
Fix for numpy like datasets due to xgb 1.5.0 changes:
"Python interface now supports configuring constraints using feature names instead of feature indices"


KalyanovD:

PR #854, opened 2021-11-01
Remove cast in SVM support indices result

vlad-nazarov:


vlad-nazarov:
http://intel-ci.intel.com/ec3adec8-ac04-f1ea-bd20-8cdcd4b723a1

PR #868, opened 2021-11-10
Cast to float in SVC test due to fail on GPU

vlad-nazarov:


agorshk (approved):

vlad-nazarov:
http://intel-ci.intel.com/ec427530-8916-f1bf-832f-8cdcd4b723a1

PR #882, opened 2021-11-16
Add check for double hw support

vlad-nazarov:


napetrov:
@KulikovNikita @samir-nasibli @Alexsandruss  - this looks like not harmful PR, but i'm not sure if it's sufficient for actually working on systems without FP64 support?

syakov-intel:
Fixed in #1198

PR #891, opened 2021-11-23
Update docs with distributed mode and unpatching, add context for kaggle kernels

outoftardis:
output: https://outoftardis.github.io/daal4py/

- Added a shield for supported scikit version to readme
- Added a topic about distributed mode: https://outoftardis.github.io/daal4py/distributed-mode.html
- Added a link to dbscan sample
- Added the definition for unpatching: https://outoftardis.github.io/daal4py/what-is-patching.html#term-unpatching
- Replaced the list of kaggle kernels with multiple tables that provide more context for the user (data, goal, what they can find in the kernel, which task it is, which algorithm, etc) https://outoftardis.github.io/daal4py/kaggle.html


agorshk (approved):
Good work! Especially I like structuredness and short description of Kaggle Kernels.
Also I notice similar content in sections **Intel® Extension for Scikit-learn*/Usage** and
**Patching**. Does it possible to reuse some information instead of copy?

outoftardis:
> Good work! Especially I like structuredness and short description of Kaggle Kernels. Also I notice similar content in sections **Intel® Extension for Scikit-learn/Usage** and **Patching**. Does it possible to reuse some information instead of copy?

the content in **usage** and **patching** is already reused :) if you look in the source, you'll see that it uses `include` directive to include the same pieces of info into different topics

napetrov (approved):

PR #898, opened 2021-11-30
Add memory requirements to docs

vlad-nazarov:
https://vlad-nazarov.github.io/daal4py/memory-requirements.html

### DIFF
--- a/doc/sources/memory-requirements.rst
+++ b/doc/sources/memory-requirements.rst
@@ -20,7 +20,7 @@
 Memory requirements 
 ###################
 
-By default, algorithms in |intelex| run in multi-threaded mode, which uses all available threads. 
+By default, algorithms in |intelex| run in multi-thread mode, which uses all available threads. 
 This may lead to optimized algorithms consuming more RAM than the corresponding stock scikit-learn algorithms.
 
 .. list-table::
@@ -29,14 +29,14 @@ This may lead to optimized algorithms consuming more RAM than the corresponding 
    :align: left
 
    * - Algorithm
-     - Single-threaded
-     - Multi-threaded
+     - Single-thread mode
+     - Multi-thread mode
    * - SVM
      - Both Scikit-learn and |intelex| consume approximately the same amount of RAM.
-     - In |intelex|, an algorithm with ``N`` number of threads will consume in ``N`` times more RAM.
+     - In |intelex|, an algorithm with ``N`` threads will consume ``N`` times more RAM.
 
 Memory consumption on GPU
-----------------------
+-------------------------
 In all |intelex| algorithms with GPU support, computations run on device memory. 
 The device memory must be large enough to store a copy of the entire dataset.
 Additional device memory may also be required for internal arrays used in computation.
--- a/doc/sources/memory-requirements.rst
+++ b/doc/sources/memory-requirements.rst
@@ -20,16 +20,23 @@
 Memory requirements 
 ###################
 
-By default |intelex| algorithms run in multithread mode using all available threads. 
-Depending on this the amount of RAM consumed by optimized algorithms may be more than the stock Scikit-learn.
+By default, algorithms in |intelex| run in multi-threaded mode, which uses all available threads. 
+This may lead to optimized algorithms consuming more RAM than the corresponding stock scikit-learn algorithms.
 
-SVM RAM consumption
--------------------
-In single-thread mode both Scikit-learn and |intelex| consumes approximately the same amount of RAM.
-In default |intelex| multithread mode with ``N`` number of threads algorithm will consume in ``N`` times more RAM.
+.. list-table::
+   :widths: 10 30 30
+   :header-rows: 1
+   :align: left
 
-GPU memory consumption
+   * - Algorithm
+     - Single-threaded
+     - Multi-threaded
+   * - SVM
+     - Both Scikit-learn and |intelex| consume approximately the same amount of RAM.
+     - In |intelex|, an algorithm with ``N`` number of threads will consume in ``N`` times more RAM.
+
+Memory consumption on GPU
 ----------------------
-In all |intelex| algorithms with GPU support computations run on device memory. 
-The device memory size must be large enough to copy the entire dataset.
+In all |intelex| algorithms with GPU support, computations run on device memory. 
+The device memory must be large enough to store a copy of the entire dataset.
 Additional device memory may also be required for internal arrays used in computation.
--- a/doc/sources/memory-requirements.rst
+++ b/doc/sources/memory-requirements.rst
@@ -1,0 +1,35 @@
+.. ******************************************************************************
+.. * Copyright 2021 Intel Corporation
+.. *
+.. * Licensed under the Apache License, Version 2.0 (the "License");
+.. * you may not use this file except in compliance with the License.
+.. * You may obtain a copy of the License at
+.. *
+.. *     http://www.apache.org/licenses/LICENSE-2.0
+.. *
+.. * Unless required by applicable law or agreed to in writing, software
+.. * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


doc/sources/memory-requirements.rst
```diff
@@ -0,0 +1,35 @@
...
+.. *******************************************************************************/
+
+.. _memory_requirements:
+
+###################
+Memory requirements 
+###################
+
+By default |intelex| algorithms run in multithread mode using all available threads. 
+Depending on this the amount of RAM consumed by optimized algorithms may be more than the stock Scikit-learn.
```
outoftardis, line 24:
```suggestion
By default, algorithms in |intelex| run in multithreaded mode, which uses all available threads. 
This may lead to optimized algorithms consuming more RAM than the corresponding stock scikit-learn algorithms.
```

or `multi-thread`:

```suggestion
By default, algorithms in |intelex| run in multi-thread mode, which uses all available threads. 
This may lead to optimized algorithms consuming more RAM than the corresponding stock scikit-learn algorithms.
```

doc/sources/memory-requirements.rst
```diff
@@ -0,0 +1,35 @@
...
+
+SVM RAM consumption
```
outoftardis, line 26: is it about SVM specifically? you don't mention it in the paragraph below

doc/sources/memory-requirements.rst
```diff
@@ -0,0 +1,35 @@
...
+-------------------
+In single-thread mode both Scikit-learn and |intelex| consumes approximately the same amount of RAM.
+In default |intelex| multithread mode with ``N`` number of threads algorithm will consume in ``N`` times more RAM.
+
+GPU memory consumption
```
outoftardis, line 31: maybe `Memory consumption on GPU`?

doc/sources/memory-requirements.rst
```diff
@@ -0,0 +1,35 @@
...
+Memory requirements 
+###################
+
+By default |intelex| algorithms run in multithread mode using all available threads. 
+Depending on this the amount of RAM consumed by optimized algorithms may be more than the stock Scikit-learn.
+
+SVM RAM consumption
+-------------------
+In single-thread mode both Scikit-learn and |intelex| consumes approximately the same amount of RAM.
+In default |intelex| multithread mode with ``N`` number of threads algorithm will consume in ``N`` times more RAM.
```
outoftardis, line 29:
```suggestion
In single-thread mode, both scikit-learn and |intelex| consume approximately the same amount of RAM.
In default multithread mode in |intelex|, an algorithm with ``N`` threads will consume ``N`` times more RAM.
```

is the second sentence about svm or does it apply to all algorithms? I think it should be placed elsewhere if it applies to all algorithms

also: maybe we need to highlight differences between single-thread vs. multi-thread modes somehow: for example, a table

doc/sources/memory-requirements.rst
```diff
@@ -0,0 +1,35 @@
...
+
+GPU memory consumption
+----------------------
+In all |intelex| algorithms with GPU support computations run on device memory.
```
outoftardis, line 33:
```suggestion
In all |intelex| algorithms with GPU support, computations run on device memory. 
```

doc/sources/memory-requirements.rst
```diff
@@ -0,0 +1,35 @@
...
+
+SVM RAM consumption
+-------------------
+In single-thread mode both Scikit-learn and |intelex| consumes approximately the same amount of RAM.
+In default |intelex| multithread mode with ``N`` number of threads algorithm will consume in ``N`` times more RAM.
+
+GPU memory consumption
+----------------------
+In all |intelex| algorithms with GPU support computations run on device memory. 
+The device memory size must be large enough to copy the entire dataset.
```
outoftardis, line 34:
```suggestion
The device memory must be large enough to store a copy of the entire dataset.
```

doc/sources/memory-requirements.rst
vlad-nazarov: This topic is about SVM specifically. We don't have exact information about other algorithms yet

doc/sources/memory-requirements.rst
vlad-nazarov: Maybe it's better to add table with one row and 3 columns (Algorithm, single-thread and multi-thread) and add there information about SVM? This may be more readable than a separate topic about SVM

outoftardis: yes, I agree

doc/sources/memory-requirements.rst
```diff
@@ -0,0 +1,42 @@
...
+.. list-table::
+   :widths: 10 30 30
+   :header-rows: 1
+   :align: left
+
+   * - Algorithm
+     - Single-threaded
+     - Multi-threaded
+   * - SVM
+     - Both Scikit-learn and |intelex| consume approximately the same amount of RAM.
```
agorshk, line 35: Do we have some where information in doc about how to run the algorithms in single thread-mode? If yes, it might be useful to add link here.

agorshk (approved):

doc/sources/memory-requirements.rst
```diff
@@ -20,16 +20,23 @@
...
+   :header-rows: 1
+   :align: left
 
-GPU memory consumption
+   * - Algorithm
+     - Single-threaded
+     - Multi-threaded
+   * - SVM
+     - Both Scikit-learn and |intelex| consume approximately the same amount of RAM.
+     - In |intelex|, an algorithm with ``N`` number of threads will consume in ``N`` times more RAM.
```
outoftardis, line 36:
```suggestion
     - In |intelex|, an algorithm with ``N`` threads will consume ``N`` times more RAM.
```

doc/sources/memory-requirements.rst
```diff
@@ -20,16 +20,23 @@
...
-In default |intelex| multithread mode with ``N`` number of threads algorithm will consume in ``N`` times more RAM.
+.. list-table::
+   :widths: 10 30 30
+   :header-rows: 1
+   :align: left
 
-GPU memory consumption
+   * - Algorithm
+     - Single-threaded
+     - Multi-threaded
```
outoftardis, line 33:
```suggestion
     - Single-thread mode
     - Multi-thread mode
```

doc/sources/memory-requirements.rst
```diff
@@ -20,16 +20,23 @@
 Memory requirements 
 ###################
 
-By default |intelex| algorithms run in multithread mode using all available threads. 
-Depending on this the amount of RAM consumed by optimized algorithms may be more than the stock Scikit-learn.
+By default, algorithms in |intelex| run in multi-threaded mode, which uses all available threads.
```
outoftardis, line 23:
```suggestion
By default, algorithms in |intelex| run in multi-thread mode, which uses all available threads. 
```

doc/sources/memory-requirements.rst
```diff
@@ -20,16 +20,23 @@
...
-GPU memory consumption
+   * - Algorithm
+     - Single-threaded
+     - Multi-threaded
+   * - SVM
+     - Both Scikit-learn and |intelex| consume approximately the same amount of RAM.
+     - In |intelex|, an algorithm with ``N`` number of threads will consume in ``N`` times more RAM.
+
+Memory consumption on GPU
 ----------------------
```
outoftardis, line 39: this line should be as long as the one right above it, otherwise the build will fail

doc/sources/memory-requirements.rst
vlad-nazarov: We have function to change number of threads - [daalinit](https://github.com/intel/scikit-learn-intelex/blob/master/generator/wrapper_gen.py#L165).
But no information about this in doc

outoftardis (approved):

PR #905, opened 2021-12-03
Remove patch rpath for release CI

Pahandrovich:


KalyanovD (approved):
Need to check if the problem resolved.

PR #913, opened 2021-12-10
Exclude dbscan from nightly jupyter testing

Pahandrovich:
dbscan requires more memory than is available in the Azure VM


KalyanovD:
Is there some nightly runs to check the problem was fixed?

KalyanovD (approved):

PR #941, opened 2022-01-11
Add information about daal4py in install.md file for sklearnex

Pahandrovich:


agorshk (approved):

KalyanovD (approved):

outoftardis (approved):

outoftardis (changes requested):
could you please also update the same info in docs: https://github.com/intel/scikit-learn-intelex/blob/master/doc/sources/installation.rst#build-from-sources

also note that the prerequisites there mention python version < 3.10, so I guess the same should be said here

outoftardis (approved):

Issue #963, opened 2022-02-04
An error: 'str' object has no attribute 'sycl_device' 


qiezikuangge:
Hello, when I want to use gpu accelerate my code, it raise this error. 
```
import numpy as np
import dpctl
from sklearnex import patch_sklearn, config_context
patch_sklearn()
from sklearn.cluster import KMeans
with config_context(target_offload="gpu:0"):
        kmm = KMeans(n_clusters=10,max_iter=2000,tol=1e-9)
        kmm_label = kmm.fit_predict(data)
# data is a array
```
the error message:
```
D:\Anaconda\lib\site-packages\daal4py\sklearn\_device_offload.py in wrapper_with_self(self, *args, **kwargs)
     86         @wraps(func)
     87         def wrapper_with_self(self, *args, **kwargs):
---> 88             return wrapper_impl(self, *args, **kwargs)
     89         return wrapper_with_self
     90     return decorator

D:\Anaconda\lib\site-packages\daal4py\sklearn\_device_offload.py in wrapper_impl(obj, *args, **kwargs)
     72                 usm_iface = _extract_usm_iface(*args, **kwargs)
     73                 q, hostargs, hostkwargs = _get_host_inputs(*args, **kwargs)
---> 74                 result = _run_on_device(func, q, obj, *hostargs, **hostkwargs)
     75                 if usm_iface is not None and hasattr(result, '__array_interface__'):
     76                     return _copy_to_usm(q, result)

D:\Anaconda\lib\site-packages\daal4py\sklearn\_device_offload.py in _run_on_device(func, queue, obj, *args, **kwargs)
     62             with sycl_context('gpu' if queue.sycl_device.is_gpu else 'cpu',
     63                               host_offload_on_fail=host_offload):
---> 64                 return dispatch_by_obj(obj, func, *args, **kwargs)
     65     return dispatch_by_obj(obj, func, *args, **kwargs)
     66 

D:\Anaconda\lib\site-packages\daal4py\sklearn\_device_offload.py in dispatch_by_obj(obj, func, *args, **kwargs)
     51     def dispatch_by_obj(obj, func, *args, **kwargs):
     52         if obj is not None:
---> 53             return func(obj, *args, **kwargs)
     54         return func(*args, **kwargs)
     55 

D:\Anaconda\lib\site-packages\daal4py\sklearn\cluster\_k_means_0_23.py in fit_predict(self, X, y, sample_weight)
    457     @support_usm_ndarray()
[...]

PR #966, opened 2022-02-08
fix str device

Alexsandruss:
# Description
Solves https://github.com/intel/scikit-learn-intelex/issues/963

### DIFF
--- a/sklearnex/_device_offload.py
+++ b/sklearnex/_device_offload.py
@@ -69,10 +69,14 @@ def _get_global_queue():
 
     if target != 'auto':
         if d4p_target is not None and \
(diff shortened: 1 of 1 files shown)


Alexsandruss:
@KalyanovD, looks like there is no error in CI connected to this PR.

KalyanovD:
@Alexsandruss, there are several issues on gpu related to knn which i didn't expect to see at your precommit but it can be caused by not merged latest master into your branch. Ok, i think your changes are LGTM but I will recheck if mentioned problems will disappear after merging this PR. Thanks =)

KalyanovD (approved):

PR #974, opened 2022-02-18
RF: Change test

lordoz234:
Change test in RandomForest


agorshk (approved):
The format of data a little bit different, but let's go on

PR #980, opened 2022-02-25
Fixes to support sklearn 1.1

Pahandrovich:


Alexsandruss (approved):
Minor: "complex method" warnings and notices should be disabled or fixed in future.

daal4py/sklearn/linear_model/logistic_path.py
```diff
@@ -278,18 +301,38 @@ def __logistic_regression_path(
...
+            mask_classes = np.array([-1, 1])
+            mask = (y == pos_class)
+            y_bin[~mask] = -1.
+            # for compute_class_weight
+
+            if class_weight == "balanced" and not _dal_ready:
+                class_weight_ = compute_class_weight(class_weight, classes=mask_classes,
+                                                     y=y_bin)
+                if not np.allclose(class_weight_, np.ones_like(class_weight_)):
+                    sample_weight *= class_weight_[le.fit_transform(y_bin)]
```
KalyanovD, line 335: duplicated, need to be moved below if-else.

KalyanovD (changes requested):

daal4py/sklearn/linear_model/logistic_path.py
```diff
@@ -74,6 +84,7 @@ def __logistic_regression_path(
     max_squared_sum=None,
     sample_weight=None,
     l1_ratio=None,
+    n_threads=1,
```
agorshk, line 87: Am I right that multithreads have been used? Why it was changed?

Pahandrovich: this is the setting for openmp parallelization. It was introduced in scikit-learn 1.1 and is always equal to 1 so far.

KalyanovD (approved):

daal4py/sklearn/linear_model/logistic_path.py
```diff
@@ -381,8 +445,11 @@ def __logistic_regression_path(
                     daal_extra_args_func = _daal4py_cross_entropy_loss_extra_args
                 func = _daal4py_loss_and_grad
             else:
-                def func(x, *args):
-                    return _multinomial_loss_grad(x, *args)[0:2]
+                if sklearn_check_version('1.1'):
```
KalyanovD, line 448:
```suggestion
                if sklearn_check_version('1.1') and loss is not None:
```

Pahandrovich: I think explicitly specifying `loss = None` is safer

PR #990, opened 2022-03-24
LocalOutlierFactor implementation

lordoz234:
Add LocalOutlierFactor implementation


sklearnex/neighbors/lof.py
```diff
@@ -0,0 +1,151 @@
...
+                (n_samples, n_samples) if metric='precomputed'
+            Training data.
+        y : Ignored
+            Not used, present for API consistency by convention.
+        Returns
+        -------
+        self : LocalOutlierFactor
+            The fitted local outlier factor detector.
+        """
+        logging.info("sklearn.neighbors.LocalOutlierFactor."
```
KalyanovD, line 71: same logging for score_samples() ?

sklearnex/dispatcher.py
```diff
@@ -75,6 +76,10 @@ def get_patch_map():
                                           'NearestNeighbors',
                                           NearestNeighbors_sklearnex), None]]
 
+        mapping['local_outlier'] = [[(neighbors_module,
```
KalyanovD, line 79: "lof", "local_outlier_factor" ?
@Pahandrovich, @vlad-nazarov, what would you say? Are there some rules for this mapping?

vlad-nazarov: I think we don't have any rules for mapping. I guess `lof` would be better. It is better to use abbreviations if they are well known.

vlad-nazarov:
@lordoz234, @KalyanovD, do we have any plans to add GPU support or is it a CPU-only patching?

lordoz234:
> @lordoz234, @KalyanovD, do we have any plans to add GPU support or is it a CPU-only patching?

@vlad-nazarov LOF patching include KNN patching, which support GPU

sklearnex/dispatcher.py
KalyanovD: Good suggestion! But I also want to notice that for all algorithms we have alias at this map with python class name in lower case. It was the cause of unpatch failures for new knn patching. Maybe is it better to add such an alias here for users who wants to unpatch this class for some reason?

sklearnex/neighbors/lof.py
```diff
@@ -0,0 +1,376 @@
...
+                for knn_prop_name in self._knn.__dict__.keys():
+                    if knn_prop_name not in self.__dict__.keys():
+                        setattr(self, knn_prop_name, self._knn.__dict__[knn_prop_name])
+
+                return self
+
+        def fit(self, X, y=None):
+            return dispatch(self, 'neighbors.LocalOutlierFactor.fit', {
+                'onedal': self.__class__._fit,
+                'sklearn': None,
```
samir-nasibli, line 133: Could you please explain why we do not fallback to stock scikit-learn here?

sklearnex/neighbors/__init__.py
```diff
@@ -18,5 +18,7 @@
 from .knn_classification import KNeighborsClassifier
 from .knn_unsupervised import NearestNeighbors
 from .knn_regression import KNeighborsRegressor
+from .lof import LocalOutlierFactor
 
-__all__ = ['KNeighborsClassifier', 'KNeighborsRegressor', 'NearestNeighbors']
+__all__ = ['KNeighborsClassifier', 'KNeighborsRegressor', 'NearestNeighbors',
```
samir-nasibli, line 23: Alphabetic order is prefered

PR #1006, opened 2022-05-19
Update logistic_path.py

bmg-pcl:
Multinomial multi classes were erroring with unhandled exceptions for logistic_path because n_classes hadn't been assigned. This patch fixes the issue.

# Description
This is a one-liner to set the n_classes variable. The recent changes for scikit-learn 1.1 eliminated the required assignement of class count, causing multiclass logistic regressions to fail with exception. 

- Assign n_class


daal4py/sklearn/linear_model/logistic_path.py
```diff
@@ -661,6 +661,7 @@ def _func_(x, *args):
                 n_classes = max(2, classes.size)
```
Alexsandruss, line 661: n_classes is assigned above

Issue #1013, opened 2022-06-27
KNeighborsClassifier predict_proba 'list' object has no attribute 'shape' error with newer scikit-learn-intelex (but not scikit-learn or older scikit-learn-intelex)


BethanyL:
**Describe the bug**
I fit a KNeighborsClassifier and then used the predict_proba function on a 1x1 example that has not been explicitly converted to a numpy array: neigh.predict_proba([[0.9]]). This is copied from a scikit-learn example and does not produce an error in scikit-learn. It also did not produce an error in a previous version of scikit-learn-intelex. 

**To Reproduce**
1. Install scikit-learn-intelex. I used:
```
python -m venv intelex
source intelex/bin/activate
pip install scikit-learn-intelex
pip install pytest

```
2. Run this example of k-NN classification from [scikit-learn documentation](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html) but after patching scikit-learn:

```
from sklearnex import patch_sklearn
patch_sklearn()

from sklearn.neighbors import KNeighborsClassifier
X = [[0], [1], [2], [3]]
y = [0, 0, 1, 1]
neigh = KNeighborsClassifier(n_neighbors=3)
neigh.fit(X, y)
print(neigh.predict([[1.1]]))
print(neigh.predict_proba([[0.9]]))
```

**Expected behavior**
I expect no errors, since scikit-learn does not produce an error. 

**Output/Screenshots**
Error:
```
Traceback (most recent call last):
  File "knn_standalone.py", line 12, in <module>
    print(neigh.predict_proba([[0.9]]))
  File "/Users/blusch/Documents/intelex/lib/python3.7/site-packages/sklearnex/_device_offload.py", line 182, in wrapper
    result = func(self, *args, **kwargs)
  File "/Users/blusch/Documents/intelex/lib/python3.7/site-packages/sklearnex/neighbors/knn_classification.py", line 244, in predict_proba
    }, X)
  File "/Users/blusch/Documents/intelex/lib/python3.7/site-packages/sklearnex/_device_offload.py", line 159, in dispatch
    return branches[backend](obj, *hostargs, **hostkwargs, queue=q)
  File "/Users/blusch/Documents/intelex/lib/python3.7/site-packages/sklearnex/neighbors/knn_classification.py", line 400, in _onedal_predict_proba
    return self._onedal_estimator.predict_proba(X, queue=queue)
[...]

PR #1014, opened 2022-06-28
added shape attr check in validation

PivovarA:
We ran into a problem with [predict_proba in knn](https://github.com/intel/scikit-learn-intelex/issues/1013). If the **predict_proba** method receives a **list** as input, the method will return an **error**. This is because we were trying to extract the **shape** attr from the **list**. 
I added an extra check to avoid this.


samir-nasibli (approved):

Alexsandruss:
@PivovarA @samir-nasibli are there any stoppers for merging?

Issue #1012, opened 2022-06-27
LinearSVC: INFO statements about running accelerated version or falling back to scikit-learn missing


BethanyL:
**Describe the bug**
When [verbose mode](https://intel.github.io/scikit-learn-intelex/verbose.html) is enabled (e.g. SKLEARNEX_VERBOSE=INFO), INFO statements are printed, such as "SKLEARNEX INFO: sklearn.cluster.DBSCAN.fit: running accelerated version on CPU" or "SKLEARNEX INFO: sklearn.cluster.DBSCAN.fit: fallback to original Scikit-learn." However, these info statements are missing for sklearn.svm.LinearSVC. 

**To Reproduce**
1. [Install](https://github.com/intel/scikit-learn-intelex/blob/master/INSTALL.md) Intel(R) Extension for Scikit-learn. I did:
```
python -m venv intelex
source intelex/bin/activate
pip install scikit-learn-intelex
```
on a Mac (and found the same behavior with a different installation process in Linux)

2. In the terminal:
```
source intelex/bin/activate
export SKLEARNEX_VERBOSE=INFO
python
```

3. In the Python interpreter (although I also saw this bug using a script):
```
from sklearnex import patch_sklearn
patch_sklearn()
from sklearn.svm import LinearSVC
# example adapted from https://scikit-learn.org/stable/modules/generated/sklearn.svm.LinearSVC.html#sklearn.svm.LinearSVC
from sklearn.datasets import make_classification
X, y = make_classification(n_features=4, random_state=0)
clf = LinearSVC(random_state=0, tol=1e-5)
clf.fit(X, y)
```

4. See lack of INFO message.
5. In contrast, continue by importing SVC:
```
from sklearn.svm import SVC
clf = SVC(random_state=0, tol=1e-5)
clf.fit(X, y)
```

6. See INFO message printed:
"SKLEARNEX INFO: sklearn.svm.SVC.fit: running accelerated version on CPU"

**Expected behavior**
I see that SVC support is via sklearn.svm.SVC, not sklearn.svm.LinearSVC, so I'm guessing the fallback message is the expected behavior. I expect to see printed "SKLEARNEX INFO: sklearn.svm.LinearSVC.fit: fallback to original Scikit-learn."

**Environment:**
 - MacOS Monterey 12.4
 - Python 3.7.4
 - scikit-learn 1.0.2
 - scikit-learn-intelex-2021.6.3
 - daal-2021.6.0
[...]

PR #1023, opened 2022-07-29
Extra note about verbose mode support

Alexsandruss:
Originated from #1012 to highlight that verbose mode is available only for supported set of algorithms

### DIFF
--- a/doc/sources/verbose.rst
+++ b/doc/sources/verbose.rst
@@ -23,7 +23,7 @@ Verbose Mode
 Use |intelex| in verbose mode to find out which implementation of the algorithm is currently used,
 |intelex| or original Scikit-learn.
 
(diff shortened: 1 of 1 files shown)


doc/sources/verbose.rst
```diff
@@ -23,6 +23,8 @@ Verbose Mode
 Use |intelex| in verbose mode to find out which implementation of the algorithm is currently used,
 |intelex| or original Scikit-learn.
 
+**Note:** verbose mode is only available for :ref:`supported algorithms <sklearn_algorithms>`.
```
outoftardis, line 26:
```suggestion
.. note:: Verbose mode is only available for :ref:`supported algorithms <sklearn_algorithms>`.
```

outoftardis (approved):

PR #1045, opened 2022-09-22
allow n_init param in kmeans to accept str 'auto'

md-shafiul-alam:
Update kmeans params options to match with stock scikit-learn. 
1. Allow k_means param n_init to accept string 'auto' in addition to integers, added in version 1.2
2. Change default of 'algorithm' to "lloyd" and adapt the code for the change

### DIFF
--- a/daal4py/sklearn/cluster/_k_means_0_23.py
+++ b/daal4py/sklearn/cluster/_k_means_0_23.py
@@ -174,6 +174,15 @@ def _daal4py_k_means_fit(X, nClusters, numIterations,
     if numIterations < 0:
         raise ValueError("Wrong iterations number")
 
+    def is_string(s, target_str):
+        return isinstance(s, str) and s == target_str
+
+    if is_string(n_init , 'auto'):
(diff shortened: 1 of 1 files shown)


daal4py/sklearn/cluster/_k_means_0_23.py
```diff
@@ -174,6 +174,14 @@ def _daal4py_k_means_fit(X, nClusters, numIterations,
     if numIterations < 0:
         raise ValueError("Wrong iterations number")
 
+    def is_string(s, target_str):
+        return isinstance(s, str) and s == target_str
```
Alexsandruss, line 178: Is there a reason why `isinstance` check needed?

md-shafiul-alam: "n_init" accepts int and 'auto'. "cluster_centers_0" (which is used as an alias of "init" in scikit-learn) can accept str and array-like. The check may be avoided for n_init, but I think "cluster_centers_0" needs the check.

Alexsandruss (approved):

PR #1053, opened 2022-10-20
adding checks on ridge regression arguments

md-shafiul-alam:
# Description
Adding checks on alpha, tol, max_iter 

Details of changes proposed in this pull request:
alpha: added a check to raise value error if alpha< 0
alpha: added a check to raise type error if alpha is str
max_iter: added a check to raise value error if max_iter< 1
max_iter: added a check to raise type error if max_iter is str
tol: added a check to raise value error if tol< 1
tol: added a check to raise type error if tol is str

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,31 +1,32 @@
 # Infra and tools
-.ci/                          @napetrov
-.github/                      @napetrov
-.circleci/                    @Pahandrovich
+.ci/                          @napetrov @Alexsandruss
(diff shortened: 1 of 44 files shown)


Alexsandruss:
Replaced in stock sklearn with [parameter constraints](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/linear_model/_ridge.py#LL785) and applied in https://github.com/intel/scikit-learn-intelex/pull/1145.

PR #1055, opened 2022-10-20
add checks on DBScan params

md-shafiul-alam:
# Description
Add checks on DBScan Params min_samples, leaf_size, n_jobs, p

Details of changes proposed in this pull request:
min_samples: added a check to raise value error if min_samples < 1
min_samples: added a check to raise type error if min_samples is float
leaf_size: added a check to raise value error if leaf_size < 1
leaf_size: added a check to raise type error if leaf_size is float
n_jobs: added a check to raise type error if n_jobs is float
p: added a check to raise value error if p < 0.0

### DIFF
--- a/daal4py/sklearn/cluster/_dbscan.py
+++ b/daal4py/sklearn/cluster/_dbscan.py
@@ -233,7 +233,7 @@ class DBSCAN(DBSCAN_original):
             Returns a fitted instance of self.
         """
         if self.eps <= 0.0:
-            raise ValueError("eps must be positive.")
+            raise ValueError(f"eps == {self.eps}, must be > 0.0.")
 
         if sklearn_check_version("1.0"):
             self._check_feature_names(X, reset=True)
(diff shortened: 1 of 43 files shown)


daal4py/sklearn/cluster/_dbscan.py
```diff
@@ -29,6 +29,8 @@
 from .._device_offload import support_usm_ndarray
 from .._utils import sklearn_check_version
 
+import numbers
+
```
samir-nasibli, line 33:
```suggestion
```

samir-nasibli:
please align code with pep8

daal4py/sklearn/cluster/_dbscan.py
md-shafiul-alam: There is one function "def" right after the import section (not shown in the above view). Line 33 is one of the two blank lines required before "def" by pep8.

Alexsandruss:
Replaced in stock sklearn with parameter constraints and applied in https://github.com/intel/scikit-learn-intelex/pull/1145.

PR #1071, opened 2022-12-01
Fix for n_iter_ SVM attribute and oneDAL kNN classifier result option

Alexsandruss:
# Description

- Add workaround for n_iter_ SVM attribute: oneDAL doesn't return number of iterations now, setting it to 1 or max in python interface
- Add pytest run for onedal directory
- fix result_option fail for kneighbors method in knn classifier (fix for circleci sklearn test)

### DIFF
--- a/onedal/svm/svm.py
+++ b/onedal/svm/svm.py
@@ -165,6 +165,9 @@ class BaseSVM(BaseEstimator, metaclass=ABCMeta):
 
     def _get_onedal_params(self, data):
         max_iter = 10000 if self.max_iter == -1 else self.max_iter
+        # TODO: remove this workaround
+        # when oneDAL SVM starts support of 'n_iterations' result
+        self.n_iter_ = 1 if max_iter < 1 else max_iter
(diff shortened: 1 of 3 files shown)


onedal/svm/svm.py
```diff
@@ -165,6 +165,9 @@ def _get_sample_weight(self, X, y, sample_weight):
 
     def _get_onedal_params(self, data):
         max_iter = 10000 if self.max_iter == -1 else self.max_iter
+        # TODO: remove this workaround
+        # when oneDAL SVM starts support of 'n_iterations' result
+        self.n_iter_ = 1 if max_iter < 1 else max_iter
```
samir-nasibli, line 170: If we don't have this feature lets do fallback on sklearnex side

Alexsandruss: it's not feature or parameter, but output
`n_iter_` is for number of computed svm iterations

KulikovNikita (approved):

napetrov (approved):

PR #1086, opened 2022-12-11
CI matrix and sklearn>1.1 fixes

Alexsandruss:
# Description

- Disable decision forest examples/tests for daal<2023.0.1 because of illegal instructions
- CI matrix update: add variation of python and sklearn versions
- Enable pytest for sklearnex dir
- Enable oneDAL iface building
- Add dependencies for Python 3.11
- Remove exact sklearn and pandas versions from testing requirements

Update of sklearnex with sklearn>1.1 changes:
- Deprecate `normalize` argument of linear models
- Deprecate `n_features_` property of PCA


.ci/pipeline/ci.yml
```diff
@@ -106,64 +77,96 @@ jobs:
...
-          intel-oneapi-dpcpp-cpp-compiler                       \
-          intel-oneapi-dev-utilities                            \
-          intel-oneapi-libdpstd-devel
-      sudo bash -c 'echo libintelocl.so > /etc/OpenCL/vendors/intel-cpu.icd'
-      sudo mv -f /opt/intel/oneapi/compiler/latest/linux/lib/oclfpga /opt/intel/oneapi/compiler/latest/linux/lib/oclfpga_
+  - script: sudo apt-get update && sudo apt-get install -y libtinfo5 clang-format
     displayName: 'apt-get'
   - script: |
+      # TODO: update dpcpp version on next dal release
+      wget "https://github.com/intel/llvm/releases/download/sycl-nightly%2F20220614/dpcpp-compiler.tar.gz"
```
napetrov, line 105: This is opensource dpc compiler that differs from what is released by intel. I would say it's worth to stick with Intel releases to avoid unexpected problems

.ci/pipeline/ci.yml
```diff
@@ -15,20 +15,12 @@
 #===============================================================================
 
 variables:
-  - name: WINDOWS_BASEKIT_URL
```
napetrov, line 18: Why to remove this?

requirements-dev.txt
```diff
@@ -1,7 +1,10 @@
 Cython==0.29.24 ; python_version <= '3.9'
-Cython==0.29.25 ; python_version >= '3.10'
+Cython==0.29.25 ; python_version == '3.10'
+Cython==0.29.30 ; python_version == '3.11'
 Jinja2==3.0.3
 numpy==1.19.5 ; python_version <= '3.9'
-numpy==1.21.6 ; python_version >= '3.10'
-pybind11==2.8.0
+numpy==1.21.6 ; python_version == '3.10'
+numpy==1.23.5 ; python_version == '3.11'
```
napetrov, line 7: >=  for 3.11 ?

requirements-test.txt
```diff
@@ -1,8 +1,6 @@
...
-scikit-learn==0.24 ; python_version == '3.6'
-pandas==1.1.5 ; python_version <= '3.6'
-pandas==1.2.2 ; python_version >= '3.7' and python_version < "3.10"
-pandas==1.4.0 ; python_version >= '3.10'
+pytest
 numpy>=1.19.5 ; python_version <= '3.9'
-numpy>=1.21.6 ; python_version >= '3.10'
+numpy>=1.21.6 ; python_version == '3.10'
+numpy>=1.23.5 ; python_version == '3.11'
+scikit-learn
```
napetrov, line 5: this would lead to non reproducible env. It's worth sticking to version . and just get it updated

requirements.txt
```diff
@@ -1,2 +1 @@
-daal==2021.4.0
```
napetrov, line 1: Where this does used?

setup_sklearnex.py
```diff
@@ -61,6 +61,9 @@
           "Programming Language :: Python :: 3.6",
```
napetrov, line 61: Guess we can remove 3.6 and 3.7 ?

daal4py/sklearn/decomposition/_pca.py
```diff
@@ -139,7 +139,10 @@ def _fit_full_daal4py(self, X, n_components):
         else:
             self.noise_variance_ = 0.
 
-        self.n_samples_, self.n_features_ = n_samples, n_features
+        if sklearn_check_version('1.2'):
```
napetrov, line 142: what would happen in 1.3? or we expect fixed in other places by that time?

.ci/pipeline/ci.yml
```diff
@@ -106,64 +77,96 @@ jobs:
     inputs:
       artifactName: 'documentation'
       targetPath: '$(Build.ArtifactStagingDirectory)/'
-- job: Linux_DPCPP
+- job: Linux
+  strategy:
+    matrix:
```
napetrov, line 82: suggest to have non overlaping matrix. At least for CI

.ci/pipeline/ci.yml
Alexsandruss: It was not mentioned anywhere in code.

requirements-dev.txt
Alexsandruss: I'm not sure it's meaningful because numpy-1.23.5 highly likely will not support higher versions of python

requirements.txt
Alexsandruss: ¯\_(ツ)_/¯

daal4py/sklearn/decomposition/_pca.py
Alexsandruss: `self.n_features_in_` is used everywhere in default PCA starting this (1.2) release.
`sklearn_check_version` is `>=`

requirements.txt
Alexsandruss: I mean it's used as requirements for daal4py package/installation but it's meaningless since daal4py will be dependent on daal4py==2021.4.0 which is ignored by pip

Alexsandruss: from setup.py:
```python
install_requires = []
with open('requirements.txt') as f:
    install_requires.extend(f.read().splitlines())
    if IS_MAC:
        for r in install_requires:
            if "dpcpp_cpp_rt" in r:
                install_requires.remove(r)
                break
```

.ci/pipeline/ci.yml
napetrov: hmm - interesting. ok I thought it was used - bight be in oneDAL ci

requirements-dev.txt
napetrov: yes. but at least it will give it a chance if it would.

requirements.txt
napetrov: prod package deps are handled internally. So this file might be not needed at all?

daal4py/sklearn/decomposition/_pca.py
napetrov: ok. thanks!

napetrov (approved):

PR #1092, opened 2022-12-14
Revert record filenames in conda-recipe

Alexsandruss:
# Description
Revert record filenames in conda-recipe to prevent fails of dependent building scripts


napetrov (approved):
Assuming internal CI works

PR #1145, opened 2023-01-03
Sklearn wrapper fixes

Alexsandruss:
Changes:
- Parameters constraints checks in SVM, linear model and DBSCAN (introduced in sklearn 1.2)
- Update messages in ValueErrors
- Add fallback to stock sklearn for metric='minkowski' and p<1 (not supported by oneDAL currently)
- Fix for parameters of oneDAL's kNN call


samir-nasibli (approved):
👍

PR #1152, opened 2023-01-09
sklearnex examples

napetrov:
# Description
Restructuring examples to setup new sklernex folder to have scikit examples there 

- move notebook to notebooks section
- create new sklernex folder for scikit examples
- create new unittest scripts for sklearnex and daal4py with same entry point for general examples


examples/daal4py/sycl/sklearn_sycl.py
```diff
@@ -169,7 +169,6 @@ def device_type_to_str(queue):
             devices.append(dpctl.SyclQueue('gpu'))
 
     else:
-        devices.append('host')
```
samir-nasibli, line 172: 👍

examples/sklernex/dbscan.py
```diff
@@ -0,0 +1,43 @@
...
+
+db = DBSCAN(eps=0.3, min_samples=10).fit(X)
+labels = db.labels_
+
+# Number of clusters in labels, ignoring noise if present.
+n_clusters_ = len(set(labels)) - (1 if -1 in labels else 0)
+n_noise_ = list(labels).count(-1)
+
+print("Estimated number of clusters: %d" % n_clusters_)
+print("Estimated number of noise points: %d" % n_noise_)
```
samir-nasibli, line 43: For many reasons it is better use def main(). Also it would be consistent with other examples.
 Something like this.
```python
def main():
    # my code here

if __name__ == "__main__":
    main()

```

napetrov: I think about scikit-learn examples more as clean way for showcase usage and not use them for validation, although they would be added to testing. 

Also aligning to excising daal4py examples is not necessary the goal - in case comparing to scikit-learn examples they are written in plain form

examples/daal4py/sycl/sklearn_sycl.py
napetrov: suggest to merge this separately as this PR would take time to form new examples and structure
https://github.com/intel/scikit-learn-intelex/pull/1153

napetrov:
@samir-nasibli  @Alexsandruss - please review

Alexsandruss:
CodeFactor issue should be fixed or disabled for green CI.

napetrov:
> CodeFactor issue should be fixed or disabled for green CI.

done

samir-nasibli (approved):

PR #1168, opened 2023-01-26
Setup helpers

Alexsandruss:
# Description

- Modify setup.py for linking to oneDAL libs with major binary version which are only available libs from conda/pip packages starting from 2023.0.1
- Manually fix install_name for 2023.0.1 oneDAL release on macos

### DIFF
--- a/scripts/build_backend.py
+++ b/scripts/build_backend.py
@@ -100,7 +100,7 @@ def build_cpp(cc, cxx, sources, targetprefix, targetname, targetsuffix, libs, li
     os.chdir(d4p_dir)
 
 
-def custom_build_cmake_clib(iface, cxx=None):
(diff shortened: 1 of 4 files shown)


scripts/build_backend.py
```diff
@@ -100,7 +100,7 @@ def build_cpp(cc, cxx, sources, targetprefix, targetname, targetsuffix, libs, li
     os.chdir(d4p_dir)
 
 
-def custom_build_cmake_clib(iface, cxx=None):
+def custom_build_cmake_clib(iface, cxx=None, onedal_major_binary_version=1):
```
napetrov, line 103: We might need to move this to more noticeable  place in future

napetrov (approved):

PR #1169, opened 2023-01-27
Fix install_name_tool call

Alexsandruss:


napetrov (approved):
Looks good, let's wait for build to complete

PR #1181, opened 2023-02-13
Removing Missing values check for XGBoost model builder

razdoburdin:
# Description
I have found recently that checking for missing values works incorrect for XGBoost model builder. It leads to false positive results in some cases. Both model builders for CatBoost and LightGBM don't contain such a check. Thus I suggest removing it also for XGBoost.

### DIFF
--- a/src/gbt_convertors.pyx
+++ b/src/gbt_convertors.pyx
@@ -163,7 +163,6 @@ def get_gbt_model_from_xgboost(booster: Any) -> Any:
 
     class_label = 0
     iterations_counter = 0
-    mis_eq_yes = None
     for tree in trees_arr:
         n_nodes = 1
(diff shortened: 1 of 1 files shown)


Alexsandruss (changes requested):
Add test for models with missing values

samir-nasibli:
Please rebase your branch and run tests with relevant onedal branch https://github.com/oneapi-src/oneDAL/pull/2345

razdoburdin:
> Please rebase your branch and run tests with relevant onedal branch [oneapi-src/oneDAL#2345](https://github.com/oneapi-src/oneDAL/pull/2345)

I close this PR, as far as it is not relevant yet. I will create the new one with support of missing values in all three model builders.

PR #1199, opened 2023-03-01
sklearn tests subset in CI

Alexsandruss:
**Changes:**
- Subset of sklearn tests in CI
- New deselected tests
- Filtering of deselected tests based on `sys.platform`
- Pinning of scipy version in CI to prevent sklearn test failures related to scipy API and impl. changes
- Parameters validation from sklearn 1.1 integrated to sklearnex DBSCAN, SVM and linear models.

### DIFF
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -128,7 +128,8 @@ jobs:
       conda activate CB
       sed -i -e "s/scikit-learn==1.2.0/scikit-learn==$(SKLEARN_VERSION).*/" requirements-test.txt
       sed -i -e "s/scikit-learn==1.0.2/scikit-learn==$(SKLEARN_VERSION).*/" requirements-test.txt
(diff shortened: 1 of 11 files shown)


.ci/pipeline/ci.yml
```diff
@@ -101,7 +101,7 @@ jobs:
   - script: |
       conda update -y -q conda
       if [ $(echo $(PYTHON_VERSION) | grep '3.11') ]; then export DPCTL_PACKAGE=; else export DPCTL_PACKAGE=; fi
-      conda create -q -y -n CB -c conda-forge -c intel python=$(PYTHON_VERSION) dal-devel mpich $DPCTL_PACKAGE
+      conda create -q -y -n CB -c conda-forge -c intel python=$(PYTHON_VERSION) dal-devel mpich pyyaml $DPCTL_PACKAGE
```
napetrov, line 104: Should we put conda deps in env file?

napetrov (approved):

.ci/pipeline/ci.yml
Alexsandruss: It's only for deselected tests file parsing, not build or runtime dependency

PR #1203, opened 2023-03-07
Running of sklearn tests on device=cpu

Alexsandruss:
Changes:
- Update dpcpp packages installation in CI on Linux
- Add `'Sklearn testing [device=cpu]'` step in CI on Linux
- Update sklearn tests running scripts to run with context and output correct exit code
- Add `has_aspect_fp64` attribute in `DummySyclDevice` to avoid failure in `onedal.datatypes._convert_to_supported` function

### DIFF
--- a/sklearnex/_device_offload.py
+++ b/sklearnex/_device_offload.py
@@ -43,6 +43,8 @@ class DummySyclQueue:
             self._filter_string = filter_string
             self.is_cpu = 'cpu' in filter_string
             self.is_gpu = 'gpu' in filter_string
+            # TODO: check for possibility of fp64 support
+            # on other devices in this dummy class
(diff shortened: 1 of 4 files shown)


sklearnex/_device_offload.py
```diff
@@ -43,6 +43,7 @@ def __init__(self, filter_string):
             self._filter_string = filter_string
             self.is_cpu = 'cpu' in filter_string
             self.is_gpu = 'gpu' in filter_string
+            self.has_aspect_fp64 = self.is_cpu
```
samir-nasibli, line 46: I don't like these hardcoded restriction.
At least comment is required.

Alexsandruss: Comment was added

KulikovNikita (approved):

PR #1205, opened 2023-03-08
Basic statistics algorithm interface in oneDAL

KulikovNikita:


onedal/datatypes/_data_conversion.py
```diff
@@ -76,3 +75,19 @@ def func(x):
             return x
 
         return _apply_and_pass(func, *data)
+
+
+#TODO: Update with dpctl
+#For now it will fall back to numpy
+def _convert_one_to_dataframe(policy, x):
```
samir-nasibli, line 82: I am not sure about `dataframe` in the func name.

samir-nasibli: And also do we really need this func? We already have `_convert_to_supported`... that at least could be updated.

onedal/__init__.py
```diff
@@ -46,7 +46,9 @@
     __all__.append('spmd')
 
 if daal_check_version((2023, 'P', 100)):
+    __all__.append('basic_statistics')
```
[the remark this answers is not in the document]
KulikovNikita: Yeap. I have added weights support for 2023.1. I see no need to support both branches of code here

onedal/datatypes/_data_conversion.py
KulikovNikita: `_convert_to_supported` - handles datatype, while `_convert_to_dataframe` should handle the following 3 options:
1) `list` as an input
2) `np.ndarray` as an input
3) `dpt.usm_ndarray` as an input

These things can be merged together but it will violate single responsibility principle.

Speaking about naming - I'm open to any suggestions. For me `dataframe` just sounds general enough to cover both ndarrays.

sklearnex/basic_statistics/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+from .basic_statistics import BasicStatistics
+
+__all__ = ['BasicStatistics']
```
samir-nasibli, line 20: Do we leaving this kind of functionality in `sklearnex.basic_statistics`? Are we planning to move this kind of new functionality, that is not in sklearn API to the new namespace in `sklearnex`?

examples/sklearnex/spmd_basic_statistics.py
```diff
@@ -0,0 +1,74 @@
...
+        weight_blocks.append(weights)
+        data_blocks.append(data)
+
+    data = np.concatenate(data_blocks, axis=0)
+    weights = np.concatenate(weight_blocks)
+
+    return (data, weights)
+
+
+if __name__ == "__main__":
```
samir-nasibli, line 44: As we agreed we should to create simple examples without `main` call.

onedal/__init__.py
```diff
@@ -46,7 +46,9 @@
...
     __all__.append('linear_model')
 
     if _is_dpc_backend:
+        __all__.append('spmd.basic_statistics')
         __all__.append('spmd.linear_model')
```
samir-nasibli, line 54:
```suggestion
        __all__ += ['spmd.basic_statistics', 'spmd.linear_model']
```

onedal/__init__.py
```diff
@@ -46,7 +46,9 @@
     __all__.append('spmd')
 
 if daal_check_version((2023, 'P', 100)):
+    __all__.append('basic_statistics')
     __all__.append('linear_model')
```
samir-nasibli, line 50:
```suggestion
    __all__ += ['basic_statistics', 'linear_model']
```

onedal/basic_statistics/basic_statistics.py
```diff
@@ -0,0 +1,99 @@
...
+            'method': self.algorithm, 'result_option': options,
+        }
+
+    def _compute(self, data, weights, module, queue):
+        policy = self._get_policy(queue, data, weights)
+
+        data_loc, weights_loc = _convert_to_dataframe(policy, data, weights)
+
+        data_loc, weights_loc = _convert_to_supported(
+            policy, data_loc, weights_loc)
```
samir-nasibli, line 70: With this flow we probably firstly converting it to `numpy.float64` and then converting it to `numpy.float32` type ndarray. It is better do conversions at ones.

samir-nasibli: I am still voting for reusing `_convert_to_supported`.

onedal/__init__.py
KulikovNikita: Fixed

onedal/__init__.py
KulikovNikita: Done as well

examples/sklearnex/spmd_basic_statistics.py
KulikovNikita: Fixed and simplified

sklearnex/basic_statistics/__init__.py
KulikovNikita: I believe we have had an agreement on this topic. The thing is that we do not have any obligations on such functionality since it does not have any counterparts in the vanilla scikit-learn. Thus we can introduce pretty much anything in this namespace as long as it only extends existing API.

onedal/basic_statistics/basic_statistics.py
KulikovNikita: I insist. It is part of functionality that will allow us to support different types of input. This part of library is crucial for now. I believe that even if we will find workaround and decide to delete this function - it still will be much faster & easier to identify such places by its name. Another thing is that it is clearly miscellaneous part of the library because it's name starts with an underscore.

samir-nasibli (approved):

PR #1208, opened 2023-03-10
KNN SPMD python interfaces

ethanglaser:
# Description

-
-
-


.ci/pipeline/ci.yml
```diff
@@ -100,7 +100,7 @@ jobs:
     displayName: 'System info'
   - script: |
       conda update -y -q conda
-      if [ $(echo $(PYTHON_VERSION) | grep '3.7') ] || [ $(echo $(PYTHON_VERSION) | grep '3.11') ]; then export DPCPP_PACKAGE="dpcpp-cpp-rt>=2023.0.0"; else export DPCPP_PACKAGE="dpctl>=0.14"; fi
+      if [ $(echo $(PYTHON_VERSION) | grep '3.7') ] || [ $(echo $(PYTHON_VERSION) | grep '3.11') ]; then export DPCPP_PACKAGE="dpcpp-cpp-rt>=2023.1.0"; else export DPCPP_PACKAGE="dpctl>=0.14"; fi
```
samir-nasibli, line 103: Why we doing it in `KNN SPMD python interfaces` PR?

examples/sklearnex/spmd_knn_bf_classification.py
```diff
@@ -0,0 +1,73 @@
...
+    size = comm.Get_size()
+
+    model_spmd, y_predict, y_test = run_example(rank, 100000, 100, 8, n_neighbors=20)
+
+    print("Brute Force Distributed kNN classification results:")
+    print("Ground truth (first 5 observations on rank {}):\n{}".format(rank, y_test[:5]))
+    print("Classification results (first 5 observations on rank {}):\n{}"
+          .format(rank, y_predict[:5]))
+    print("Accuracy for entire rank {} (256 classes): {}"
+          .format(rank, np.mean(np.equal(y_test, y_predict))))
```
samir-nasibli, line 73: Please avoid calling main. We agreed to simplify our sklearnex examples. See: https://github.com/intel/scikit-learn-intelex/blob/789d5478971b5bece05188eb35bae2b924b1dafa/examples/sklearnex/spmd_basic_statistics.py in #1205

.ci/pipeline/ci.yml
ethanglaser: I was just trying something, reverted now

examples/sklearnex/spmd_knn_bf_classification.py
ethanglaser: Following Nikita's example for consistency (which was merged): https://github.com/intel/scikit-learn-intelex/blob/master/examples/sklearnex/spmd_linear_regression.py. Should I remove main?

samir-nasibli: Yeah. It is already updated on #1205. Try to follow. https://github.com/intel/scikit-learn-intelex/blob/master/examples/sklearnex/spmd_linear_regression.py also will be refactored and simplified. Thanks!

onedal/__init__.py
```diff
@@ -50,3 +50,4 @@
 
     if _is_dpc_backend:
         __all__.append('spmd.linear_model')
+        __all__.append('spmd.neighbors')
```
samir-nasibli, line 53:
```suggestion
        __all__ += ['spmd.linear_model', 'spmd.neighbors']
```

onedal/neighbors/neighbors.py
```diff
@@ -44,6 +44,12 @@
 
 
 class NeighborsCommonBase(metaclass=ABCMeta):
+    def _get_train_policy(self, queue, *data):
+        return _get_policy(queue, *data)
+
+    def _get_infer_policy(self, queue, *data):
+        return _get_policy(queue, *data)
```
samir-nasibli, line 51: Could you please comment why you are duplicating this? The same method with different names...

onedal/neighbors/neighbors.py
```diff
@@ -568,14 +575,17 @@ def _onedal_predict(self, model, X, params, queue):
...
+        X = _convert_to_supported(policy, X)
         backend = _backend.neighbors.regression if gpu_device \
             else _backend.neighbors.search
 
         if hasattr(self, '_onedal_model'):
             model = self._onedal_model
         else:
             model = self._create_model(backend)
+        if gpu_device and "responses" not in params["result_option"]:
+            params["result_option"] += "|responses"
```
samir-nasibli, line 588: Move it to `_get_onedal_params` please.

onedal/spmd/neighbors/neighbors.py
```diff
@@ -0,0 +1,43 @@
...
+        raise NotImplementedError("predict_proba not supported in distributed mode.")
+
+
+class KNeighborsRegressor(NeighborsBaseSPMD, KNeighborsRegressor_Batch):
+    def predict(self, X, queue=None):
+        return self._predict_gpu(X, queue=queue)
+
+
+class NearestNeighbors(NeighborsBaseSPMD):
+    pass
```
samir-nasibli, line 43: If we don't support SPMD, there is no need adding it here

KulikovNikita (approved):
Everything looks fine for me. But I also propose to simplify examples a little bit

samir-nasibli:
Please rebase your branch and address public CI test fails.

onedal/spmd/neighbors/neighbors.py
ethanglaser: I think it should be left since it uses NeighborsBaseSPMD/NeighborsCommonBaseSPMD

onedal/neighbors/neighbors.py
ethanglaser: It is dependent on gpu_device which depends on queue

samir-nasibli: > It is dependent on gpu_device which depends on queue

Function `_get_onedal_params` also have `queue` as an arg.

onedal/spmd/neighbors/neighbors.py
samir-nasibli: We don't need `NeighborsBaseSPMD` as well here.

samir-nasibli: Remove it, please. Leave only NeighborsCommonBaseSPMD, the code where you really changing behavior. Remove `NearestNeighbors` from init file as well. If we don't support SPMD interface, no need to put this class here.

sklearnex/spmd/__init__.py
```diff
@@ -14,4 +14,4 @@
 # limitations under the License.
 #===============================================================================
 
-__all__ = ['linear_model', 'basic_statistics']
+__all__ = ['linear_model', 'basic_statistics', 'neighbors']
```
samir-nasibli, line 17: alphabetic order.

sklearnex/spmd/neighbors/neighbors.py
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+from onedal.spmd.neighbors import (
+    KNeighborsClassifier,
+    KNeighborsRegressor,
+    NearestNeighbors
```
samir-nasibli, line 20: Remove it from here.

sklearnex/spmd/neighbors/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+from .neighbors import KNeighborsClassifier, KNeighborsRegressor, NearestNeighbors
+
+__all__ = ['KNeighborsClassifier', 'KNeighborsRegressor', 'NearestNeighbors']
```
samir-nasibli, line 19: Remove please algos from the list if we don't support SPMD interface.

onedal/spmd/neighbors/neighbors.py
ethanglaser: We do support it. It's the same circumstance as in [linear regression](https://github.com/intel/scikit-learn-intelex/blob/master/onedal/spmd/linear_model/linear_model.py#L27). No extensions of the class itself are introduced but there are extensions to classes it inherits from, so I believe it is necessary here.

samir-nasibli: There is no need for inheriting `NeighborsBaseSPMD`. This are not public. there is no need for such `NeighborsBaseSPMD`

onedal/spmd/neighbors/neighbors.py
```diff
@@ -0,0 +1,43 @@
...
```
samir-nasibli, line 43:
```suggestion
class NeighborsCommonBaseSPMD(ABC):
    def _get_policy(self, queue, *data):
        return _get_spmd_policy(queue)


class KNeighborsClassifier(NeighborsCommonBaseSPMD, KNeighborsClassifier_Batch):
    def predict_proba(self, X, queue=None):
        raise NotImplementedError("predict_proba not supported in distributed mode.")


class KNeighborsRegressor(NeighborsCommonBaseSPMD, KNeighborsRegressor_Batch):
    def predict(self, X, queue=None):
        return self._predict_gpu(X, queue=queue)


class NearestNeighbors(NeighborsCommonBaseSPMD):
    pass
```

examples/sklearnex/spmd_knn_bf_classification.py
```diff
@@ -0,0 +1,64 @@
...
+model_spmd.fit(X_train, y_train, queue=q)
+
+y_predict = model_spmd.predict(X_test, queue=q)
+
+print("Brute Force Distributed kNN classification results:")
+print("Ground truth (first 5 observations on rank {}):\n{}".format(rank, y_test[:5]))
+print("Classification results (first 5 observations on rank {}):\n{}"
+      .format(rank, y_predict[:5]))
+print("Accuracy for entire rank {} (256 classes): {}"
+      .format(rank, np.mean(np.equal(y_test, y_predict))))
```
Alexsandruss, line 64: Use sklearn's accuracy score instead or even better classification score from sklearn.metrics

```suggestion
      .format(rank, np.mean(np.equal(y_test, y_predict))))
```

examples/sklearnex/spmd_knn_bf_regression.py
```diff
@@ -0,0 +1,69 @@
...
+
+comm = MPI.COMM_WORLD
+rank = comm.Get_rank()
+size = comm.Get_size()
+
+if size < 2:
+    warn("This example was intentionally "
+         "designed to run in distributed mode only", RuntimeWarning)
+
+q = dpctl.SyclQueue("gpu")
```
Alexsandruss, line 46: GPU unavailability should be handled. For example, use exit with warning

onedal/neighbors/neighbors.py
```diff
@@ -549,7 +552,8 @@ def _onedal_fit(self, X, y, queue):
 
             return train_alg(**params).compute(X, y).model
 
-        policy = _get_policy(queue, X, y)
+        policy = self._get_policy(queue, X, y)
+        X, y = _convert_to_supported(policy, X, y)
```
Alexsandruss, line 556: Is there any performance measurement to prove that this code didn't affect performance especially for small data shapes cases?

Alexsandruss (changes requested):

onedal/neighbors/neighbors.py
ethanglaser: It does not currently have queue as an arg. I can add it as an optional arg and move lines into function or I can leave it as is to keep consistency with _get_onedal_params across all algos.

ethanglaser: But in current master there is a similar param conditional outside of _get_onedal_params which is part of the reason I configure it the way it is: https://github.com/intel/scikit-learn-intelex/blob/master/onedal/neighbors/neighbors.py#L430-L431

samir-nasibli: This is internal use func, there is no need saving consistency for these methods. It is not a public. So it could be and should extended in this case. Since it is on master you can keep it as is.
BTW kNN master is not good enough for the good code reference.

samir-nasibli: And also I didn't get why we adding condition for GPU device?

examples/sklearnex/spmd_knn_bf_regression.py
ethanglaser: How does this seem - https://github.com/ethanglaser/scikit-learn-intelex/blob/dev/eglaser-spmd-knn/examples/sklearnex/spmd_knn_bf_regression.py#L46-L49

onedal/neighbors/neighbors.py
ethanglaser: The addition was only necessary for distributed version so I think the latest commit most cleanly implements the logic and addresses your comments - https://github.com/ethanglaser/scikit-learn-intelex/blob/dev/eglaser-spmd-knn/onedal/spmd/neighbors/neighbors.py#L37-L41

ethanglaser:
Thank you for your thorough reviews. Addressed most comments and finally external CI is green (was related to onedal version >= 20230100).

onedal/neighbors/neighbors.py
ethanglaser: The inclusion or exclusion of this line does not impact performance (runtime, results) for examples of various sizes from a few basic experiments I have tried. @KulikovNikita have you observed improvements?

examples/sklearnex/spmd_knn_bf_regression.py
Alexsandruss: I meant it should have better explanation why example can't be run. dpctl itself produces error saying 'device is unavailable'. @KulikovNikita, what do think about adding warning `GPU device is unavailable. Currently, SPMD execution mode is implemented only for this device type.`

setup.py
```diff
@@ -475,9 +475,11 @@ def run(self):
           ] if ONEDAL_VERSION >= 20230100 else []
          ) + (
         ['onedal.spmd',
-         'onedal.spmd.linear_model',
-         'onedal.spmd.basic_statistics'
-         ] if build_distribute else [])),
+         'onedal.spmd.basic_statistics',
+         'onedal.spmd.linear_model'
+         ] + (['onedal.spmd.neighbors']
```
jmhatre-work, line 480: ['onedal.spmd',
         'onedal.spmd.linear_model',
         'onedal.spmd.basic_statistics',
        'onedal.spmd.neighbors'
         ] instead of  +(['onedal.spmd.neighbors'])

ethanglaser: I think the formatting makes the conditionals confusing, what the existing logic amounts to is:
if build_distributed:
  append(onedal.spmd, onedal.spmd.linear_model, onedal.spmd.basic_statistics]
  if onedal_version >= 20230100:
    append(onedal.spmd.neighbors)

examples/sklearnex/spmd_knn_bf_classification.py
```diff
@@ -0,0 +1,64 @@
+#===============================================================================
```
samir-nasibli, line 1: Please rename your examples like `examples/skleaernex/<algo_name>_spmd.py`.

samir-nasibli:
Hi just add GPU device and librariry requirements for your examples here https://github.com/intel/scikit-learn-intelex/blob/master/tests/run_examples.py#L127:L156:
```
...
req_device['knn_bf_classification_spmd.py'] = ["gpu"]
req_device['knn_bf_regression_spmd.py'] = ["gpu"]

...

req_library['knn_bf_classification_spmd.py'] = ['catboost']
req_library['knn_bf_regression_spmd.py''] = ['dpctl', 'mpi4py']
```
And also rename your examples and make `spmd` as suffix in the file names.

onedal/spmd/neighbors/neighbors.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class KNeighborsClassifier(NeighborsCommonBaseSPMD, KNeighborsClassifier_Batch):
+    def predict_proba(self, X, queue=None):
+        raise NotImplementedError("predict_proba not supported in distributed mode.")
+
+
+class KNeighborsRegressor(NeighborsCommonBaseSPMD, KNeighborsRegressor_Batch):
+    def predict(self, X, queue=None):
+        return self._predict_gpu(X, queue=queue)
```
samir-nasibli, line 35: What if it is not GPU queue?

Alexsandruss (approved):

ethanglaser:
Should these interfaces be compatible with using dpctl tensors for X and y without queue argument? Because currently it is not. And it seems like other algos may not be also, not sure if this is needed at the moment.

samir-nasibli:
> Should these interfaces be compatible with using dpctl tensors for X and y without queue argument? Because currently it is not. And it seems like other algos may not be also, not sure if this is needed at the moment.

This is good question and it is common for all SPMD iface algos. We are working on DCPTL adoption, so it will be aligned on  https://github.com/intel/scikit-learn-intelex/pull/1216.

This PR is done :)

samir-nasibli (approved):

PR #1210, opened 2023-03-13
This PR replicates the changes made in PR #1187

Vika-F:
Add versioning for csr_table class.
Use dal::detail::csr_table for oneDAL 2023.1 and older; use dal::csr_table starting from oneDAL 2023.1.1.


onedal/datatypes/data_conversion.cpp
```diff
@@ -20,14 +20,26 @@
...
+
+#if ONEDAL_VERSION <= 20230100
+    #include "oneapi/dal/table/detail/csr.hpp"
+#else
+    #include "oneapi/dal/table/csr.hpp"
+#endif
 
 namespace oneapi::dal::python {
 
+#if ONEDAL_VERSION <= 20230100
```
KulikovNikita, line 37:
```suggestion
#if ONEDAL_VERSION <= 20230101
```

It should depend on 2023.01.01 build since you have pushed updates recently and they'll be only in this version of OneDAL

samir-nasibli: I think this is for the old interface, and should be ok.

KulikovNikita:
@samir-nasibli - I believe duplication was done on purpose. Speaking specifically - previous PR have some issues on the GitHub side

samir-nasibli:
> @samir-nasibli - I believe duplication was done on purpose. Speaking specifically - previous PR have some issues on the GitHub side

Yes, exactly. We early discussed about with author.

KulikovNikita (approved):

PR #1213, opened 2023-03-14
Update tests after feature selection change

ahuber21:
With https://github.com/oneapi-src/oneDAL/pull/2292 some changes are introduced that need to be reflected on the scikit-learn-intelex side

* Add `useConstFeatures` algorithm option
  * The previously hardcoded value can now be changed as a parameter. The option defaults to the previously hardcoded value `false`
* Feature sampling was changed: Sampling features for calculating the best node split was optimized, which results in different  numerical values in the predictions. The overall result of the ensemble is still the same.

```
// from the commit message
    chore: update unit test reference values
    
    details: feature selection for node splitting was changed which results in
    different numerics for the prediction. mean and variance are still in good
    agreement
    
    mean:
      old -> 22.088
      new -> 22.104
    
    variance:
      old -> 49.4695
      new -> 49.4311
```

Edit:
During development I ran a `RandomForestClassifier` and I'm still producing similar results

```
Accuracy
  old -> 0.9313
  new -> 0.9316

Confusion matrix
  old
    [[6922    4]
     [ 511   63]]

  new
    [[6921    5]
     [ 508   66]]
```

However, the training time is greatly improved

```
old: 433.87 seconds
new: 17.25 seconds
```


tests/unittest_data/decision_forest_regression_batch.csv
```diff
@@ -1,127 +1,127 @@
-36.70242652
```
Alexsandruss, line 1: What is the need for changing data file?

Alexsandruss (changes requested):
`sklearnex/preview/ensemble/forest` wrappers should be updated too

tests/unittest_data/decision_forest_regression_batch.csv
ahuber21: See the PR description. Node splitting will be changed with https://github.com/oneapi-src/oneDAL/pull/2292, resulting in different values. As far as I can tell, the algorithm is still producing correct results.

ahuber21:
@Alexsandruss Concerning the `sklearnex/preview/ensemble/forest` wrappers: I'm not sure about the naming. Variable names in `daal4py` are (mostly) camelCase (`maxBins`, minBinSize`, ...) while the onedal params are snake_case (`max_bins`, `min_bin_size`, ...). I only modified [cpp/daal/include/algorithms/decision_forest/decision_forest_training_parameter.h](https://github.com/oneapi-src/oneDAL/pull/2292/files#diff-8f6391e5dda2288905529e3197ed9c6ff15a614e2f68b18507d1201f5c97dfe4), not sure if adding the option to onedal would even work

daal4py/sklearn/ensemble/tests/test_decision_forest.py
```diff
@@ -33,7 +33,12 @@
 
 ACCURACY_RATIO = 0.95 if daal_check_version((2021, 'P', 400)) else 0.85
 MSE_RATIO = 1.07
-LOG_LOSS_RATIO = 1.4 if daal_check_version((2021, 'P', 400)) else 1.55
+if daal_check_version((2023, 'P', 101)):
+    LOG_LOSS_RATIO = 1.5
```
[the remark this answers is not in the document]
ahuber21: It is discussed in https://github.com/oneapi-src/oneDAL/pull/2292 and I am investigating the issue. I think my changes pronounce an existing bug for one particular test case even more. But I really would like to tackle speedup and accuracy in separate PRs

tests/test_examples.py
```diff
@@ -185,6 +185,12 @@ def test_svm_batch(self):
      'decision_forest_regression_batch.csv', lambda r: r[1].prediction, (2023, 'P', 1)),
     ('decision_forest_regression_hist_batch',
      'decision_forest_regression_batch.csv', lambda r: r[1].prediction, (2023, 'P', 1)),
+    ('decision_forest_regression_default_dense_batch',
+     'decision_forest_regression_batch_20230101.csv',
+     lambda r: r[1].prediction, (2023, 'P', 101)),
+    ('decision_forest_regression_hist_batch',
+     'decision_forest_regression_batch_20230101.csv',
+     lambda r: r[1].prediction, (2023, 'P', 101)),
```
Alexsandruss, line 193: This set of examples will fail when version updates from 2023.1.0 to 2023.1.1 because two sets of regression examples will be run.

Alexsandruss (approved):

PR #1217, opened 2023-03-15
TEST: SPMD sklearnex examples run

samir-nasibli:
# Description
Added running SPMD sklearnex examples


examples/daal4py/run_examples.py
```diff
@@ -52,9 +54,9 @@
 assert 8 * struct.calcsize('P') in [32, 64]
 
 if 8 * struct.calcsize('P') == 32:
-    logdir = jp(exdir, '_results', 'ia32')
+    logdir = jp(exdirs[0], '_results', 'ia32')
```
samir-nasibli, line 57: I am leaving logs in the `examples/daal4py` dir.
Do we need to move it into `examples`?

KulikovNikita:
```suggestion
    logdir = jp(jp(rootdir, 'daal4py'), '_results', 'ia32')
```

examples/daal4py/run_examples.py
```diff
@@ -52,9 +54,9 @@
...
 else:
-    logdir = jp(exdir, '_results', 'intel64')
+    logdir = jp(exdirs[0], '_results', 'intel64')
```
KulikovNikita, line 59:
```suggestion
    logdir = jp(jp(rootdir, 'daal4py'), '_results', 'intel64')
```

examples/daal4py/run_examples.py
napetrov: Might be it worth to completely move run_examples script to tests directory? It's a bit strange for running sklernex examples from daal4py directory.

examples/daal4py/run_examples.py
```diff
@@ -158,6 +166,13 @@ def get_exe_cmd(ex, nodist, nostream):
             return None
         if not check_library(req_library[os.path.basename(ex)]):
             return None
+    if os.path.dirname(ex).endswith("sklearnex"):
```
napetrov, line 169: i guess there is a way to rewrite this in a bit cleaner logic - for scikit you need only check for device additionally, but version check is universal for example and can be run elsewhere.

examples/daal4py/run_examples.py
samir-nasibli: I was thinking about it, so that is why I ask. Moving it may require some additional changes. But it is really better, let me try.

examples/daal4py/run_examples.py
samir-nasibli: Generally, for `sklearnex` examples I think we don't need additional restrictions.
I believe we will have some other examples, not only SPMD, and some of them on CPU as well. Here I am just checking env for exact spmd examples in `examples.sklearnex` dir.
I think here it's just required to specify condition like this:
```suggestion
    if os.path.dirname(ex).endswith("sklearnex") and not nodist and ex.endswith('spmd.py'):
```

examples/daal4py/run_examples.py
```diff
@@ -129,14 +131,20 @@ def check_library(rule):
 req_version['decision_forest_classification_default_dense_batch.py'] = (2023, 'P', 1)
 req_version['decision_forest_classification_traverse_batch.py'] = (2023, 'P', 1)
 req_version['decision_forest_regression_hist_batch.py'] = (2021, 'P', 200)
+req_version['basic_statistics_spmd.py'] = (2023, 'P', 1)
+req_version['linear_regression_spmd.py'] = (2023, 'P', 1)
 
 req_device = defaultdict(lambda: [])
+req_device['basic_statistics_spmd.py'] = ["gpu"]
```
napetrov, line 138: Are we targeting examples for GPU only? in this case we could run them only on GPU enabled systems, but i would assume that technically we can run this on cpus as well?

samir-nasibli: We are targeting only these exact examples on GPU. https://github.com/intel/scikit-learn-intelex/blob/master/examples/sklearnex/spmd_basic_statistics.py#L43

napetrov: yes for now, but more generic question is that we can run them on CPU device as well for purpose of validation. This might require change in example

samir-nasibli: Yeah, current examples requires GPU devices, so that is why we have required device check here. It is possible run them on CPU as well, if algo supported on distr CPU. But it requires changes in examples and updates for devices checks in this runner.

napetrov (changes requested):
You shouldn't threat examples as scripts. 
Please keep them as they were at top level. Also there is no expectation that users would be using run_examples script - this is test automation starter for us, not interface for users 

I suggest to move to to tests and keep examples structure as is

napetrov (approved):

PR #1220, opened 2023-03-17
Fix build error introduced after adding versioning for csr_table class (PR #1210)

Vika-F:


Vika-F:
http://intel-ci.intel.com/edc4acea-369c-f1a4-97be-a4bf010d0e2e

samir-nasibli (approved):

PR #1226, opened 2023-03-24
Update Azure pipelines scope

Alexsandruss:
Changes:
- Move CI steps to os-specific templates
- Add nightly jobs with run of all tests on sklearn `main` branch
- Update Ubuntu and Python versions in nightly
- Use run_sklearn_tests.sh script for testing in release jobs
- Pin pandas version to 1.3 for sklearn 0.24 and scipy 1.6 to fix compatibility of sparse formats
- Add option to run sklearn tests in run_sklearn_tests.sh script
- Move dpcpp installation to separate script
- Fix typo in test_global_patch.py
- Fix KDTree valid metrics getting for sklearn>=1.3
- Deselect LinReg tests failure on MacOS because of impossibility to converge

### DIFF
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -127,12 +127,14 @@ jobs:
   - script: |
       . /usr/share/miniconda/etc/profile.d/conda.sh
       conda activate CB
+      export SELECTED_TESTS=all
       bash .ci/scripts/run_sklearn_tests.sh
     displayName: 'Sklearn testing'
     continueOnError: true
   - script: |
       . /usr/share/miniconda/etc/profile.d/conda.sh
       conda activate CB
+      export SELECTED_TESTS=all
       bash .ci/scripts/run_sklearn_tests.sh cpu
     displayName: 'Sklearn testing [device=cpu]'
     continueOnError: true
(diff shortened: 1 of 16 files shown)


.ci/pipeline/ci.yml
```diff
@@ -79,21 +77,16 @@ jobs:
       Python3.11_Sklearn1.2:
         PYTHON_VERSION: '3.11'
         SKLEARN_VERSION: '1.2'
+      Python3.11_SklearnMain:
```
napetrov, line 80: Isn't this overkill for CI? might be worth to keep it just in Nightly?

napetrov:
also think that we can update https://github.com/intel/scikit-learn-intelex/blob/master/.ci/pipeline/release.yml ?  Or would wait for 2023.1.1 release?

.ci/pipeline/ci.yml
Alexsandruss: This step is in PR scope now to check if it works. But I think it is not real overkill because this step takes ~10 minutes and ~25% CI time resources additionally.

napetrov: Yes. but there might be problem with scikti that would be blocking PRs. 
so in case of nightly runs we would have daily validation but failures against scikit master wouldn't be blocking PRs

Alexsandruss:
> also think that we can update https://github.com/intel/scikit-learn-intelex/blob/master/.ci/pipeline/release.yml ? Or would wait for 2023.1.1 release?

I updated set of sklearn tests used in release jobs. 3.10 python version might be added to test matrix too. Did you mean any other update?

napetrov (approved):

PR #1231, opened 2023-03-31
feat: add function to check patching status of an instance

ahuber21:
Had to run auto-formatting on the files, diff is best viewed with w=1 https://github.com/intel/scikit-learn-intelex/pull/1231/files?w=1

* Provides a new function `is_patched_instance(instance)`, which returns `True` if `instance` is loaded from `sklearnex`, and `False` otherwise
* Tests for all currently patched classes added in `sklearnex/tests/test_patching.py`
* Removed implicit imports from `sklearnex/__init__.py` - currently discussing with @samir-nasibli 
  * Update: Added them back. We follow the sklearn's design choices here

### DIFF
--- a/sklearnex/__init__.py
+++ b/sklearnex/__init__.py
@@ -27,17 +27,34 @@ from .dispatcher import (
 )
 
 __all__ = [
-    "utils",
+    "basic_statistics",
+    "cluster",
     "config_context",
+    "decomposition",
+    "ensemble",
     "get_config",
-    "set_config",
     "get_patch_map",
     "get_patch_names",
     "is_patched_instance",
+    "linear_model",
+    "manifold",
+    "metrics",
+    "neighbors",
     "patch_sklearn",
+    "set_config",
     "sklearn_is_patched",
+    "sklearn_is_patchedget_patch_map",
+    "svm",
     "unpatch_sklearn",
+    "utils",
 ]
+
+
+from onedal import _is_dpc_backend
+
+if _is_dpc_backend:
+    __all__.append("spmd")
+
 
 from ._utils import set_sklearn_ex_verbose
 
--- a/sklearnex/__init__.py
+++ b/sklearnex/__init__.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-#===============================================================================
+# ===============================================================================
 # Copyright 2021 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


sklearnex/__init__.py
```diff
@@ -13,27 +13,32 @@
...
+    "utils",
+    "config_context",
+    "get_config",
+    "set_config",
+    "get_patch_map",
+    "get_patch_names",
+    "is_patched_instance",
+    "patch_sklearn",
+    "sklearn_is_patched",
+    "unpatch_sklearn",
```
Alexsandruss, line 39: Submodules (cluster, model_selection, etc) should not be removed from `__all__` so module behavior is aligned with [sklearn](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/__init__.py#L85) and submodules can be imported through `*` symbol (`from sklearn[ex] import *`)

Alexsandruss (changes requested):

sklearnex/__init__.py
```diff
@@ -13,27 +13,32 @@
...
 ]
 
-from onedal import _is_dpc_backend
-
-if _is_dpc_backend:
-    __all__.append('spmd')
-
```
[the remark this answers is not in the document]
ahuber21: Thanks for flagging it. It's probably not valid, and I wanted to chat with @samir-nasibli about it. We're implicitly importing submodules right now, which confused me. It looks like sklearn is doing the same thing. I'll understand how and why this is okay and add everything back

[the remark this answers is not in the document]
ahuber21: Added it back. Sorry for the big diff, the old version erroneously used CLRF line termination.

sklearnex/tests/test_patching.py
```diff
@@ -41,50 +44,110 @@ def run_parse(mas, result):
...
+        NuSVC(),
+        NuSVR(),
+        PCA(),
+        RandomForestClassifier(),
+        RandomForestRegressor(),
+        Ridge(),
+        SVC(),
+        SVR(),
+        TSNE(),
+    ]
```
Alexsandruss, line 136: It is better to get estimators from sklearnex.dispatcher::get_patch_map without need to update list of estimators manually. Entry of mapping contains sklearn submodule and estimator name which you can get:
```
>>> mapping['pca']
[[(<module 'sklearn.decomposition' from 'miniconda3/envs/build_py39/lib/python3.9/site-packages/sklearn/decomposition/__init__.py'>, 'PCA', <class 'daal4py.sklearn.decomposition._pca.PCA'>), None]]
```

sklearnex/dispatcher.py
```diff
@@ -1,225 +1,277 @@
...
+            mapping["random_forest_classifier"] = [
+                [
+                    (
+                        ensemble_module,
+                        "RandomForestClassifier",
+                        RandomForestClassifier_sklearnex,
+                    ),
+                    None,
+                ]
+            ]
```
Alexsandruss, line 94: Is there any better non-sparse formatting for this mapping entry like this:
```suggestion
            mapping["random_forest_classifier"] = [[(
                ensemble_module,
                "RandomForestClassifier",
                RandomForestClassifier_sklearnex,
            ), None]]
```

ahuber21: I think stuff like this should be handled in a `pyproject.toml` which we don't yet have
See for example https://github.com/psf/black/blob/main/pyproject.toml#L8

sklearnex/tests/test_patching.py
ahuber21: Trying to add it, but I really don't get the structure of this thing. Why the heck is it a list of lists of tuples. There is only a single tuple. Why not flatten this?

ahuber21: It does not produce the correct instances

```python
    models = [
        patch_infos[0][0][2]()
        for key, patch_infos in get_patch_map().items()
        if key
        not in (
            "distances",
            "logistic",
            "train_test_split",
            "fin_check",
            "roc_auc_score",
            "set_config",
            "get_config",
            "config_context",
        )
    ]
```

image

ahuber21: Okay, I think I see now what you mean. Will try importlib

sklearnex/tests/test_patching.py
```diff
@@ -41,50 +44,85 @@ def run_parse(mas, result):
...
+        not in (
+            "distances",
+            "logistic",
+            "train_test_split",
+            "fin_check",
+            "roc_auc_score",
+            "set_config",
+            "get_config",
+            "config_context",
+        )
```
Alexsandruss, line 110: Use `inspect.isclass` or `issubclass(estimator_class, sklearn.BaseEstimator)` function instead

Alexsandruss (approved):

PR #1234, opened 2023-04-05
English corrections to deselected_tests.yaml

icfaust:
# Description
English is corrected in the same file for presentability.

- English fixes

### DIFF
--- a/deselected_tests.yaml
+++ b/deselected_tests.yaml
@@ -186,7 +186,7 @@ deselected_tests:
   # https://github.com/oneapi-src/oneDAL/issues/495
   - linear_model/tests/test_coordinate_descent.py::test_warm_start_convergence_with_regularizer_decrement <0.24
 
-  # oneAPI Data Analytics Library (oneDAL) doesn't support sample_weight (back to sklearn),
+  # oneAPI Data Analytics Library (oneDAL) doesn't support sample_weight (back to scikit-learn),
   # sufficient accuracy (similar to previous cases)
   - linear_model/tests/test_coordinate_descent.py::test_enet_sample_weight_consistency >=0.23
 
--- a/deselected_tests.yaml
+++ b/deselected_tests.yaml
@@ -213,7 +213,6 @@ deselected_tests:
   # sklearnex implementation builds different trees compared to sklearn
   # Comparison of tree forest will fail
   - ensemble/tests/test_forest.py::test_class_weights[RandomForestClassifier]
-  - ensemble/tests/test_forest.py::test_poisson_vs_mse
   - inspection/tests/test_permutation_importance.py::test_robustness_to_high_cardinality_noisy_feature >=0.23
   - tests/test_common.py::test_estimators[RandomForestClassifier()-check_classifiers_train]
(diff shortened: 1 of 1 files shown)


deselected_tests.yaml
```diff
@@ -296,17 +300,17 @@ deselected_tests:
...
   # This test fails because with patch config_context with new options, but the
-  # test checks that only exact number of options is used
+  # test checks only the exact number of options that are used
   - tests/test_config.py::test_config_context
 
   # HalvingGridSearchCV with Ridge and PCA didn't have feature_name_in. Need to fix.
   - tests/test_common.py::test_pandas_column_name_consistency >=1.0.1
 
-  # Some sklearnex docstrings differ from scikit-learn.
+  # Some sklearnex docstrings differ from sklearn.
```
samir-nasibli, line 313: I am not sure about this changes. It is better leave it as is.

samir-nasibli:
1) Generally, logically it is better separate into different PRs refactoring work and enhancement/bug fixing even in the same file.
2) If we tests is non-relevant now, we should also cover functionality with the new one as well.

deselected_tests.yaml
```diff
@@ -71,21 +71,21 @@ deselected_tests:
...
-  # kdtree kNN rarely misses 0-distance points when kneighbors is used on same fitting data
+  # KDtree kNN rarely misses 0-distance points when kneighbors is used on same-fitting data
   - manifold/tests/test_spectral_embedding.py::test_precomputed_nearest_neighbors_filtering
 
-  # cache directory is not accessible on some systems
+  # Cache directory is not accessible on some systems
   - utils/tests/test_validation.py::test_check_memory
 
-  # oneDAL doesn't throw error if resulting coeffs are not finite
+  # OneDAL doesn't throw error if resulting coeffs are not finite
```
Alexsandruss, line 80: `oneDAL/sklearn/sklearnex` are short names of libraries and should start from lower case even in the beginning of sentence.

deselected_tests.yaml
icfaust: I believe it would be good to go with one or the other here. Let me know which you prefer.

deselected_tests.yaml
icfaust: > 1. Generally, logically it is better separate into different PRs refactoring work and enhancement/bug fixing even in the same file.
> 2. If we tests is non-relevant now, we should also cover functionality with the new one as well.

I've created an Epic just for this purpose. A new ticket has been included to address this issue in detail, among 11 others within the file.

icfaust: Changed.

icfaust:
I have split off the change for test_poisson_vs_mse to another PR based on @samir-nasibli 's suggestion

icfaust:
Should I change the copyright to 2023?

deselected_tests.yaml
samir-nasibli: From my point of view it is better use `scikit-learn`, `scikit-learn-intelex`, as these are package names

KulikovNikita (approved):

PR #1245, opened 2023-04-24
MacOS CI fix

Alexsandruss:


napetrov (approved):
overall ok for mac part, but we need to address linux failures

PR #1249, opened 2023-04-26
Update requirements

Alexsandruss:
# Description
- Update development and testing requirements.
- Increase relative error in oneDAL BasicStatistics and LinearRegression tests for float32
- Fix max_features default value for oneDAL RandomForest


napetrov:
there are failures in windows in CI. 

Also it would be good to fix linux public failures to ensure we are doing good here

.ci/pipeline/ci.yml
```diff
@@ -121,6 +127,9 @@ jobs:
       Python3.11_Sklearn1.2:
         PYTHON_VERSION: '3.11'
         SKLEARN_VERSION: '1.2'
+      Python3.11_SklearnMain:
+        PYTHON_VERSION: '3.11'
+        SKLEARN_VERSION: 'main'
```
KulikovNikita, line 132: what if it will be updated with API break\?

KulikovNikita (approved):

.ci/pipeline/ci.yml
Alexsandruss: It's temporary scope for this PR (https://github.com/intel/scikit-learn-intelex/pull/1249/commits/0ae69474b8d92e3fe34333ca2c85f12b8f81ae44) and will be removed before merge

Issue #960, opened 2022-02-02
HELP: Missing values are not supported in daal4py Gradient Boosting Trees


dragarok:
Can somebody tell me how can I resolve this issue? I have tried looking at my data and find nulls. I can't resolve this issue. This is just a help post. Any help to guide me to the correct approach is appreciated.

dragarok:
I found an issue here in github regarding this. I checked my whole dataframe for null values using `df.isnull().values.any()` but it all results to false.

Alexsandruss:
Can you provide XGBoost training parameters or reproducer code?
Problems with missing values are known, one possible reason is 'gpu_hist' training method which output XGBoost booster with active missing values indicators even if training data hasn't them.

dragarok:
> Can you provide XGBoost training parameters or reproducer code? Problems with missing values are known, one possible reason is 'gpu_hist' training method which output XGBoost booster with active missing values indicators even if training data hasn't them.

Oh yes. I am using 'gpu_hist' as training tree method. So, is there no way to train on gpu and get model for sure to work with daal?

Alexsandruss:
For now, xgboost models from `gpu_hist` method are not working with daal4py. This behavior is not expected and probably bug, but its source (daal4py or xgboost) is non-obvious since `hist` is working. I will investigate it.

dklein0:
@Alexsandruss - Is there any workaround?

Alexsandruss:
It looks like problem was solved from XGBoost side: boosters created by `gpu_hist` tree method are easily translated to DAAL models now if no missing values are presented (DAAL inference is not supporting them).


SW used to check this case: Python 3.9.13, XGBoost 1.6.1 pip package, daal4py/oneDAL 2021.5.0 conda-forge packages.


Testing script:
```python
import xgboost as xgb
import daal4py as d4p
import numpy as np
from sklearn.datasets import make_classification


x, y = make_classification(n_samples=10000, n_features=16, n_classes=2)
xgb_clsf = xgb.XGBClassifier(tree_method='gpu_hist')
xgb_clsf.fit(x, y)

booster = xgb_clsf.get_booster()

xgb_prediction = xgb_clsf.predict(x)
xgb_errors_count = np.count_nonzero(xgb_prediction - y)

daal_model = d4p.get_gbt_model_from_xgboost(booster)

daal_predict_algo = d4p.gbt_classification_prediction(
    nClasses=2,
    resultsToEvaluate="computeClassLabels",
[...]

PR #1276, opened 2023-05-05
Missing values support for inference

razdoburdin:
# Description
This PR is a continuation of [oneDAL missing values support](https://github.com/oneapi-src/oneDAL/pull/2345) from scikit-learn-intelex side.

- Changing the logic of all three model builders (XGBoost/LightGBM/Catboost) to extract information about desired behavior in case of missing values and pass it to oneDAL.
- Removing checking for missing values, as far as it produced [false positive results](https://github.com/intel/scikit-learn-intelex/issues/960) and is not required now.


samir-nasibli:
I think these changes depend on onedal 2023.2 version, so please add versioning here as well

daal4py/sklearn/_utils.py
```diff
@@ -19,7 +19,7 @@
 import os
 import warnings
 
-from daal4py import _get__daal_link_version__ as dv
+from daal4py._daal4py import _get__daal_link_version__ as dv
```
samir-nasibli, line 22: test

src/gbt_model_builder.pyx
```diff
@@ -18,6 +18,9 @@
 # The model builder object is retrieved through calling model_builder.
 # We will extend this once we know how other model builders will work in DAAL
 
+from daal4py.sklearn._utils import daal_check_version
```
samir-nasibli, line 21: I am not sure if this will work in cython src files for daal4py.

It would be better predefine `missing_values_are_supported` in `gbt_model_builder.h`:
```c++
#if __INTEL_DAAL__ >= 2023 && INTEL_DAAL_VERSION >= 20230200
#define missing_values_are_supported 1
#else
#define missing_values_are_supported 0
#endif
```
And import it from the header in this `.pyx` file.

daal4py/sklearn/_utils.py
samir-nasibli, line 22: I think this changes are not relevant. `_get__daal_link_version__` attribute is in `daal4py`. Could you please share issue details?

src/gbt_model_builder.pyx
razdoburdin: I have tried this solution on local machine. It works well. Need I to made more tests?

daal4py/sklearn/_utils.py
razdoburdin: When I tried to import it from `src/gbt_model_builder.pyx`, I got a cyclic import error. This fix solved the problem.

src/gbt_model_builder.pyx
samir-nasibli: outdated

razdoburdin:
Restarted CI, after update of oneDAL master:
https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/edf64f3e-f3af-f15d-aeed-a4bf010d0e2e

samir-nasibli (approved):

samir-nasibli (approved):

PR #1299, opened 2023-05-10
Fix for scan warning in kNN dispatching

Alexsandruss:


sklearnex/neighbors/common.py
```diff
@@ -68,14 +68,15 @@ def _onedal_supported(self, device, method_name, *data):
...
+            y = None
+            # To check multioutput, might be overhead
+            if len(data) > 1:
+                y = np.asarray(data[1])
+                if is_classifier:
+                    class_count = len(np.unique(y))
+            if hasattr(self, '_onedal_estimator'):
+                y = self._onedal_estimator._y
+            if y is not None and hasattr(y, 'ndim') and hasattr(y, 'shape'):
                 is_single_output = y.ndim == 1 or y.ndim == 2 and y.shape[1] == 1
```
KulikovNikita, line 80: What about the native python lists as an input?

Alexsandruss: y is ndarray in all cases. It's converted to ndarray on training

ethanglaser (approved):

KulikovNikita (approved):

PR #1307, opened 2023-05-25
sklearnex GPU CI updates

ethanglaser:
# Description
Fixing dpctl has_gpu_devices to be function call (it is not an attribute so currently always returns true)

### DIFF
--- a/examples/daal4py/sycl/sklearn_sycl.py
+++ b/examples/daal4py/sycl/sklearn_sycl.py
@@ -165,7 +165,7 @@ if __name__ == "__main__":
     if dpctl_available:
         devices.append(None)
         devices.append(dpctl.SyclQueue('cpu'))
-        if dpctl.has_gpu_devices:
+        if dpctl.has_gpu_devices():
             devices.append(dpctl.SyclQueue('gpu'))
 
     else:


napetrov:
Here is examples code - the problem is that there is no dpctx package - it should be dpctl, so examples would run through sycl_context in daal4py instead of dpctl. 

Also in device_context suposed to be deprecated in dpctl so we can't relay on it. 

So we do have dpctl compute follows data aproach then we are getting dpctl tensor as input and getting queue from it. 
Second interface is config_context that should be tested.   

Also i think that testing should relay also on unitestests in pytest run on GPU and not examples. 

try:
    from dpctx import device_context, device_type
    with device_context(device_type.gpu, 0):
        gpu_available = True
except:
    try:
        from daal4py.oneapi import sycl_context
        with sycl_context('gpu'):
            gpu_available = True
    except:
        gpu_available = False

samir-nasibli:
@ethanglaser @napetrov #1315

napetrov (approved):

examples/daal4py/sycl/sklearn_sycl.py
```diff
@@ -165,7 +165,7 @@ def device_type_to_str(queue):
     if dpctl_available:
         devices.append(None)
         devices.append(dpctl.SyclQueue('cpu'))
-        if dpctl.has_gpu_devices:
+        if dpctl.has_gpu_devices():
```
ahuber21, line 168:
```suggestion
        if dpctl.has_gpu_devices() is True:
```

ahuber21 (approved):
Comment is just a suggestion to protect against such errors

PR #1313, opened 2023-05-29
Fix sklearn version in CI and wrappers

Alexsandruss:
Changes:
- Fix for sklearn version used in CI/nightly jobs
- Apply new tree nodes format from sklearn 1.3: add `missing_go_to_left` field
- Fix logloss ratio in iris test for decision forest
- Add new deselected tests

### DIFF
--- a/daal4py/sklearn/_utils.py
+++ b/daal4py/sklearn/_utils.py
@@ -19,6 +19,8 @@ import sys
 import os
 import warnings
 
+from numpy.lib.recfunctions import require_fields
+
 from daal4py import _get__daal_link_version__ as dv
 from sklearn import __version__ as sklearn_version
 try:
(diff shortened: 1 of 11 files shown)


daal4py/sklearn/_utils.py
```diff
@@ -176,6 +178,23 @@ def get_number_of_types(dataframe):
         return 1
 
 
+def check_tree_nodes(tree_nodes):
```
samir-nasibli, line 181: Did you checked perf with this new checker?

ahuber21 (approved):

daal4py/sklearn/_utils.py
Alexsandruss: Perf ratio for `sklearnex.ensemble` and `sklearnex.preview.ensemble` is nearly 1, so perf check is passed

PR #1315, opened 2023-05-31
DEP: deprecate `dpctx` using in examples

samir-nasibli:
# Description
Deprecated using `dpctx` `device_context` in daal4py sycl examples.

Updates on `sycl_contex` and `config_context` will be considered on other PRs.

### DIFF
--- a/examples/daal4py/sycl/decision_forest_regression_batch.py
+++ b/examples/daal4py/sycl/decision_forest_regression_batch.py
@@ -33,16 +33,11 @@ except Exception:
         return np.loadtxt(f, usecols=c, delimiter=',', ndmin=2, dtype=t)
 
 try:
-    from dpctx import device_context, device_type
-    with device_context(device_type.gpu, 0):
+    from daal4py.oneapi import sycl_context
+    with sycl_context('gpu'):
         gpu_available = True
 except Exception:
-    try:
-        from daal4py.oneapi import sycl_context
-        with sycl_context('gpu'):
-            gpu_available = True
-    except Exception:
-        gpu_available = False
+    gpu_available = False
 
 
 # Commone code for both CPU and GPU computations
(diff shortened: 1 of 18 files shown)


samir-nasibli:
http://intel-ci.intel.com/edff5b1b-a024-f100-beac-a4bf010d0e2e

examples/daal4py/sycl/decision_forest_regression_batch.py
```diff
@@ -33,16 +33,11 @@ def read_csv(f, c, t=np.float64):
...
-    with device_context(device_type.gpu, 0):
+    from daal4py.oneapi import sycl_context
+    with sycl_context('gpu'):
         gpu_available = True
 except Exception:
-    try:
-        from daal4py.oneapi import sycl_context
-        with sycl_context('gpu'):
-            gpu_available = True
-    except Exception:
```
ahuber21, line 44: Also `Exception` is only slightly better. We should catch what is raised and that's apparently a `RuntimeError`

ahuber21 (dismissed):
Please change the `except` statements as requested.
Generally, I don't like the code duplication / spaghetti code. We have a lot of duplication.

We should determine the devices and then run the test for each one instead. To me it's okay if we do this separately, since you didn't touch this code.

For example:
```python
devices = ("cpu", "gpu") if gpu_available else ("cpu",)

for device in devices:
        with sycl_context(device):
            sycl_train_data = sycl_buffer(train_data)
            sycl_train_labels = sycl_buffer(train_labels)
            sycl_predict_data = sycl_buffer(predict_data)

            compute_result= compute(sycl_train_data, sycl_train_labels,
                                         sycl_predict_data, nClasses)
            assert np.allclose(compute_result.prediction,
                               predict_result_classic.prediction)
```

samir-nasibli:
> Please change the `except` statements as requested. Generally, I don't like the code duplication / spaghetti code. We have a lot of duplication.
> 
> We should determine the devices and then run the test for each one instead. To me it's okay if we do this separately, since you didn't touch this code.
> 
> For example:
> 
> ```python
> devices = ("cpu", "gpu") if gpu_available else ("cpu",)
> 
> for device in devices:
>         with sycl_context(device):
>             sycl_train_data = sycl_buffer(train_data)
>             sycl_train_labels = sycl_buffer(train_labels)
>             sycl_predict_data = sycl_buffer(predict_data)
> 
>             compute_result= compute(sycl_train_data, sycl_train_labels,
>                                          sycl_predict_data, nClasses)
>             assert np.allclose(compute_result.prediction,
>                                predict_result_classic.prediction)
> ```

I am totally agree with code duplication point. But this is an example, for users, and here code should be much simpler and explicit, so that is why I am leaving this duplication as is.
I was thinking about duplication here, just need other feedbacks for being sure. So I think we can avoid it here at least.

napetrov (approved):

napetrov:
I'd like Andreas suggestion but think that we can merge this now as cleanup of obvious problems. And probably have to have discussion on examples and how we would like to handle them overall

PR #1320, opened 2023-06-06
Model builders API update

razdoburdin:
I want to start discussion about changing the Model builders API. 
The main challenge is support of both scikit-learn style estimators and non-scikit style booster objects.
In the presented PR I added a new class `GBTDAALModel` for python interface in non-scikit style. It allow to hide all worldly syntax from the old API.

In case of using `XGBoost`, for regression tasks user-side code changes from:
`d4p_model = daal4py.get_gbt_model_from_xgboost(booster)`
 `d4p_prediction = daal4py.gbt_regression_prediction().compute(X_test, d4p_model).prediction`
to
`d4p_model = daal4py.mb.convert_model(booster)`
`d4p_prediction =d4p_model.predict()`

In case of classification problem user-side code changes from:
`d4p_model = daal4py.get_gbt_model_from_xgboost(booster)`
`d4p_prediction=daal4py.gbt_classification_prediction(nClasses=n_classes).compute(X_test, d4p_model).prediction`
to
`d4p_model = daal4py.mb.convert_model(booster)`
`d4p_prediction = d4p_model.predict()`

For support of scikit-style estimators, I updated `GBTDAALClassifier `and `GBTDAALRegressor `classes.
One can use them like this (example for  `XGBoost`, regression task):
`from daal4py.sklearn.ensemble import GBTDAALRegressor`
`reg = xgb.XGBRegressor()`
`reg.fit(X, y)`
`d4p_predt = GBTDAALRegressor.convert_model(reg).predict(X)`


napetrov:
the only thing that i don't fully like - CamelCase on .GbtModel.  Might be GBTModel would be better?

src/gbt_model_builder.pyx
```diff
@@ -179,3 +179,33 @@ def gbt_reg_model_builder(n_features, n_iterations):
...
     '''
     return gbt_regression_model_builder(n_features, n_iterations)
+
+class GBTModel():
+    '''
+    Daal4py GBT Model
+    '''
+
+    def __init__(self, model):
+        framework_name = (model.__class__.__module__, model.__class__.__name__)
```
Alexsandruss, line 189: It doesn't look like framework name, better rename to `submodule_name` and `class_name`

src/gbt_model_builder.pyx
```diff
@@ -179,3 +179,33 @@ def gbt_reg_model_builder(n_features, n_iterations):
...
+        if framework_name == ("lightgbm.basic", "Booster"):
+            self._model, self._n_classes = get_gbt_model_from_lightgbm(model)
+        elif framework_name == ("xgboost.core", "Booster"):
```
Alexsandruss, line 192: Support of xgb.XGBClassifier/Regressor is required. Same for other frameworks if they contain sklearn-like estimators

src/gbt_model_builder.pyx
```diff
@@ -179,3 +179,33 @@ def gbt_reg_model_builder(n_features, n_iterations):
...
+            self._model, self._n_classes = get_gbt_model_from_xgboost(model)
+        elif framework_name == ("catboost.core", "CatBoost"):
+            self._model, self._n_classes = get_gbt_model_from_catboost(model)
+        else:
+            raise TypeError(f"Unknown model format {framework_name}")
+        
+        self._is_regression = isinstance(self._model, gbt_regression_model)
+
+    def predict(self, x, resultsToEvaluate = 'computeClassLabels'):
```
Alexsandruss, line 201: In sklearn terms, `predict` outputs class or regression value while `predict_proba` outputs class probabilities. Thus, resultsToEvaluate should be internal for these methods instead of passed argument

Alexsandruss, line 201: Input validation is required

src/gbt_model_builder.pyx
```diff
@@ -179,3 +179,33 @@ def gbt_reg_model_builder(n_features, n_iterations):
     :param size_t n_iterations: Number of trees in the model
     '''
     return gbt_regression_model_builder(n_features, n_iterations)
+
+class GBTModel():
```
Alexsandruss, line 183: Should be derived from sklearn BaseEstimator. Since sklearn is runtime dependency, this class should not be compiled during build

```suggestion
class GBTModel(BaseEstimator):
```

Alexsandruss (changes requested):

inteldimitrius:
The rest looks good to me!

src/gbt_model_builder.pyx
```diff
@@ -178,34 +178,4 @@ def gbt_reg_model_builder(n_features, n_iterations):
...
-            return gbt_regression_prediction().compute(x, self._model).prediction
-        else:
-            output = gbt_classification_prediction(nClasses=self._n_classes, resultsToEvaluate=resultsToEvaluate).compute(x, self._model)
-            if resultsToEvaluate == 'computeClassLabels':
-                return output.prediction
-            elif resultsToEvaluate == 'computeClassProbabilities':
-                return output.probabilities
-            else:
-                raise TypeError(f"Value of resultsToEvaluate = {resultsToEvaluate} is unknown")
+    return gbt_regression_model_builder(n_features, n_iterations)
```
napetrov, line 181:
```suggestion
    return gbt_regression_model_builder(n_features, n_iterations)

```

src/gbt_model_builder.pyx
razdoburdin: done

razdoburdin: done

src/gbt_model_builder.pyx
razdoburdin: done

razdoburdin:
Dear all,
I have updated the PR. Currently, the scikit-style estimators are supported. Please see the updated description for details.

inteldimitrius:
MacOS and Linux CI checks are failing. Is it okay?

Alexsandruss:
NaN errors are not the only ones:
```
======================================================================
ERROR: test_gbt_cls_model_create_from_catboost_batch (test_examples.TestExNpyArray)
----------------------------------------------------------------------
  File "/home/vsts/work/1/s/tests/test_examples.py", line 96, in testit
    result = self.call(ex)
  File "/home/vsts/work/1/s/tests/test_examples.py", line 269, in call
    return ex.main(readcsv=np_read_csv)
  File "/home/vsts/work/1/s/examples/daal4py/gbt_cls_model_create_from_catboost_batch.py", line 73, in main
    daal_prediction = daal_predict_algo.compute(X_test, daal_model)
TypeError: Argument 'model' has incorrect type (expected daal4py._daal4py.gbt_classification_model, got tuple)
```

src/gbt_model_builder.pyx
```diff
@@ -179,3 +181,110 @@ def gbt_reg_model_builder(n_features, n_iterations):
     :param size_t n_iterations: Number of trees in the model
     '''
     return gbt_regression_model_builder(n_features, n_iterations)
+
+
+class GBTDAALModel:
```
Alexsandruss, line 186: It's not obvious why this class should be in compiled file, not usual python script.

src/gbt_model_builder.pyx
```diff
@@ -179,3 +181,110 @@ def gbt_reg_model_builder(n_features, n_iterations):
...
+        return predict_result.prediction.ravel()
+
+    def predict(self, X, fptype="float", resultsToEvaluate="computeClassLabels"):
+        if self.is_regression_:
+            return self._predict_regression(X, fptype)
+        else:
+            return self._predict_classification(X, fptype, resultsToEvaluate)
+
+def gbt_build_model(model):
+    return GBTDAALModel().build_model(model)
```
Alexsandruss, line 290: Why do not use static method in GBTDAALModel instead of separate function?

src/gbt_model_builder.pyx
```diff
@@ -179,3 +181,110 @@ def gbt_reg_model_builder(n_features, n_iterations):
...
+                "Call 'fit' with appropriate arguments before using this method.").format(
+                    type(self).__name__))
+
+        # Prediction
+        predict_algo = gbt_regression_prediction(fptype=fptype)
+        predict_result = predict_algo.compute(X, self.daal_model_)
+
+        return predict_result.prediction.ravel()
+
+    def predict(self, X, fptype="float", resultsToEvaluate="computeClassLabels"):
```
Alexsandruss, line 283: Where is `predict_proba` and `predict_log_proba` methods?

razdoburdin: > Where is `predict_proba` and `predict_log_proba` methods?

This class is an analog of booster object from xgboost/lightgbm/catboost. It is not scikit-style estimator. Look [here](https://github.com/intel/scikit-learn-intelex/blob/3f34dcb5b2f97182f279d8bbbb5cb8eb36f2d298/daal4py/sklearn/ensemble/GBTDAAL.py#L191) for scikit-style classifier and regressor.

src/gbt_model_builder.pyx
razdoburdin: > Why do not use static method in GBTDAALModel instead of separate function?

Separate function is less wordily and looks more clear for the end user.

src/gbt_model_builder.pyx
razdoburdin: > It's not obvious why this class should be in compiled file, not usual python script.

What benefits do you expect from moving this class from a compile file?

src/gbt_model_builder.pyx
Alexsandruss: From example:
```python
    # Conversion to daal4py
    daal_model = d4p.gbt_build_model(lgb_model)

    # daal4py prediction
    daal_prediction = daal_model.predict(X_test)
```
Is there a reason why `daal_model` should not be sklearn-like estimator with full set of prediction functions?

src/gbt_model_builder.pyx
Alexsandruss: It's not obvious what is returned from this function while static method will make association with estimator class
Also, `gbt_build_model` and `GBTDAALModel` use abbreviature while sklearn API/docs prefer [full name](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.GradientBoostingClassifier.html#sklearn-ensemble-gradientboostingclassifier).

src/gbt_model_builder.pyx
razdoburdin: The reason is that the frameworks (xgboost/lightgbm/catboost) have two parallel versions of API: functional and scikit-style. The problem is that one can't made a scikit-estimator from the Booster object from functional API for xgboost and lightgbm. We need to support model builders not only for scikit-style API users, but also for functional API users of the GBT Frameworks. GBTDAALModel is analog of the Booster object from functional API.

Alexsandruss: > GBTDAALModel is analog of the Booster object from functional API

What is the need to mimic Booster object if most of it's methods are not implemented?

razdoburdin: > 

There is no any standard methods sets for Booster. Each framework has it's own set of methods. We need to support Booster-based API, as far as frameworks support it, and some users use it.

razdoburdin:
> @razdoburdin - we probably also need examples to be added with new API

I updated the existed examples and added the new ones for scikit-style API.

Alexsandruss:
@mergifyio rebase

examples/sklearnex/gbt_cls_model_create_from_catboost_batch.py
```diff
@@ -0,0 +1,89 @@
+#===============================================================================
+# Copyright 2021 Intel Corporation
```
napetrov, line 2: Looks that those older names were not removed

examples/sklearnex/model_builders_catboost.py
```diff
@@ -0,0 +1,89 @@
+#===============================================================================
+# Copyright 2021 Intel Corporation
```
napetrov, line 2:
```suggestion
# Copyright 2023 Intel Corporation
```

examples/sklearnex/model_builders_lightgbm.py
```diff
@@ -0,0 +1,84 @@
+#===============================================================================
+# Copyright 2020 Intel Corporation
```
napetrov, line 2:
```suggestion
# Copyright 2023 Intel Corporation
```

examples/sklearnex/model_builders_xgboost.py
```diff
@@ -0,0 +1,85 @@
+#===============================================================================
+# Copyright 2014 Intel Corporation
```
napetrov, line 2:
```suggestion
# Copyright 2023 Intel Corporation
```

examples/sklearnex/gbt_cls_model_create_from_catboost_batch.py
razdoburdin: fixed

examples/sklearnex/model_builders_xgboost.py
```diff
@@ -0,0 +1,85 @@
...
+import xgboost as xgb
+import numpy as np
+import pandas as pd
+
+
+def pd_read_csv(f, c=None, t=np.float64):
+    return pd.read_csv(f, usecols=c, delimiter=',', header=None, dtype=t)
+
+
+def main(readcsv=pd_read_csv, method='defaultDense'):
```
Alexsandruss, line 29: Method is not used. Same for second example.
```suggestion
def main(readcsv=pd_read_csv):
```

Alexsandruss (changes requested):

daal4py/sklearn/ensemble/GBTDAAL.py
```diff
@@ -27,7 +27,7 @@
 from .._utils import getFPType
 
 
-class GBTDAALBase(BaseEstimator):
```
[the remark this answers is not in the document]
razdoburdin: > Implement nan tag dispatching for base class:
> 
> ```python
> def _more_tags(self):
>     return {"allow_nan": self.allow_nan_}
> ```

done

daal4py/mb/ModelBuilders.py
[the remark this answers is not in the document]
razdoburdin: done

examples/sklearnex/model_builders_xgboost.py
razdoburdin: > Method is not used. Same for second example.

done

napetrov:
@razdoburdin  test_model_builders_xgboost/ test_model_builders_lightgbm are failing internally

src/gbt_convertors.pyx
```diff
@@ -131,7 +152,8 @@ def get_gbt_model_from_xgboost(booster: Any) -> Any:
     booster.feature_names = [str(i) for i in lst]
 
     trees_arr = booster.get_dump(dump_format="json")
-    xgb_config = json.loads(booster.save_config())
+    if xgb_config is None:
+        xgb_config =get_xgboost_params(booster)
```
Alexsandruss, line 156:
```suggestion
        xgb_config = get_xgboost_params(booster)

```

daal4py/mb/model_builders.py
```diff
@@ -0,0 +1,195 @@
...
+        predict_result = predict_algo.compute(X, self.daal_model_)
+
+        return predict_result.prediction.ravel()
+
+
+class GBTDAALModel(GBTDAALBaseModel):
+    def __init__(self):
+        pass
+
+    def predict(self, X, fptype="float"):
```
Alexsandruss, line 176: `fptype` should be deduced from input data.

Alexsandruss (approved):

samir-nasibli:
@razdoburdin you have changed examples names, removed suffix `_batch`. Examples runner will ignore this such file names https://github.com/intel/scikit-learn-intelex/blob/master/tests/run_examples.py#L219:L222 .
Please fix it.

tests/run_examples.py
```diff
@@ -157,9 +157,9 @@ def check_library(rule):
 
 req_library = defaultdict(lambda: [])
 req_library['basic_statistics_spmd.py'] = ['dpctl', 'mpi4py']
-req_library['gbt_cls_model_create_from_lightgbm_batch.py'] = ['lightgbm']
-req_library['gbt_cls_model_create_from_xgboost_batch.py'] = ['xgboost']
-req_library['gbt_cls_model_create_from_catboost_batch.py'] = ['catboost']
+req_library['model_builders_lightgbm.py'] = ['lightgbm']
+req_library['model_builders_xgboost.py'] = ['xgboost']
+req_library['model_builders_catboost.py'] = ['catboost']
```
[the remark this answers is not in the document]
napetrov: No. Batch is strange and miningless suffix. If runner will not pick up them, then launcher should be fixed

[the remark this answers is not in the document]
samir-nasibli: Let's decide how we can group-by this kind of examples then.
All examples [scikit-learn-intelex/examples/](https://github.com/intel/scikit-learn-intelex/tree/master/examples/) are all grouped logically into 4 groups: `spmd.py`, `streaming.py`, `stream.py`, `batch.py`.
If none of them is suitable, then:
* we will come up with a new suffix for such examples.
* update whole examples runner logic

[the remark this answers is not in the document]
razdoburdin: I have just made a brief look in the run_examples.py. May we just add some sort of default behavior in case the filename doesn't end on `*spmd.py`, `*streaming.py`, `*stream.py` or `*batch.py`?
As far as I see we could switch to this default behavior for all the `*_batch.py` and `*_stream.py` files. In my mind, it will make naming of examples more clear.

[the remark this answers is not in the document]
napetrov: @razdoburdin i like that idea more, than using suffixes.  

@samir-nasibli - those are logical only if you look on this from some testing perspective. From end user perspective this looks strange.

I would go with base examples without batch because it make no sense to specify this. On stream/streaming - looks like this should have be single group. And for spmd - let's leave it as is

@Alexsandruss, @KulikovNikita   - your thoughts?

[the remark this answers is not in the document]
samir-nasibli: > I have just made a brief look in the run_examples.py. May we just add some sort of default behavior in case the filename doesn't end on `*spmd.py`, `*streaming.py`, `*stream.py` or `*batch.py`?

Sure we can. That means that we need update examples runner logic little bit (second option).

PR #1330, opened 2023-06-13
Remove check for NAN for GBT Estimators

razdoburdin:
# Description
Faced with this problem, while worked on https://github.com/intel/scikit-learn-intelex/pull/1320
Checking for NAN is required for sklearn-style estimators. However GBT DAAL inference [support missing values](https://github.com/intel/scikit-learn-intelex/pull/1276) in form of NAN. Thus, I suggest removing the corresponding check from `test_estimator.py`.

See also the related issues for XGBoost estimators: [[1](https://github.com/dmlc/xgboost/issues/3580), [2](https://github.com/dmlc/xgboost/issues/5641)].

### DIFF
--- a/tests/test_estimators.py
+++ b/tests/test_estimators.py
@@ -69,9 +69,15 @@ def _restore_from_saved(md, saved_dict):
         setattr(md, check_f, saved_dict[check_f])
 
 
+def _check_estimator(estimator, skip_list):
+    checks = check_estimator(estimator, generate_only=True)
+    for (estimator, check) in checks:
+        if check.func.__name__ not in skip_list:
+            check(estimator)
+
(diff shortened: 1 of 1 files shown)


inteldimitrius (approved):
Looks good to me.

razdoburdin:
> This is not compliant with sklearn's way to define NaN support. Check [hist_gradient_boosting](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/ensemble/_hist_gradient_boosting/gradient_boosting.py#L1168) for example.

The approach you have suggested, affect the `fit()` method as well. But daal4py GBT doesn't support NAN in the fitting stage yet.

PR #1333, opened 2023-06-15
Performance improvement for preview ensemble algorithms (RF and ET)

icfaust:
# Description
This corrects performance problems associated with estimators_ construction in the random forest and extra trees fit methods for regression and classification, and corrects a performance problem associated with a default construction of a sample_weights vector (which is unnecessary). These yield significant performance improvements.

- Add the ability to the 'fit' method to operate without a sample_weights vector, which has impacts on branch of code operation on CPU (15-20% performance impact) and in part prevented GPU operation (sample_weights not supported in oneDAL GPU)
- Set preview Random Forest Regressor back to using the 'hist' method rather than 'dense' (*already done in other PR)
- Fix pybind11/python accesses in DFS, streamline slowdowns from abstraction in call() functions
- Set estimators_ attribute to being lazy evaluated
- Fixes memory issue when initializing skl_tree_node arrays causing issues in leaf nodes, and creates speedup through C++ initialization rather than copy


onedal/primitives/tree_visitor.cpp
```diff
@@ -153,52 +184,31 @@ to_sklearn_tree_object_visitor<Task>::to_sklearn_tree_object_visitor(std::size_t
...
-
     OVERFLOW_CHECK_BY_MULTIPLICATION(std::size_t, this->node_count, this->class_count);
-    double* value_ar_ptr =
-        new double[this->node_count * 1 *
-                   this->class_count](); // oneDAL only supports scalar responses for now
-
-    this->node_ar = py::array_t<skl_tree_node>(node_ar_shape, node_ar_strides, node_ar_ptr);
-    this->value_ar = py::array_t<double>(value_ar_shape, value_ar_strides, value_ar_ptr);
+    this->node_ar = new skl_tree_node[this->node_count];
+    this->value_ar = new double[this->node_count*1*this->class_count](); // oneDAL only supports scalar responses for now
```
KulikovNikita, line 189:
```suggestion
    this->value_ar = new double[this->node_count * this->class_count](); // oneDAL only supports scalar responses for now
```

onedal/primitives/tree_visitor.cpp
```diff
@@ -267,7 +268,7 @@ bool to_sklearn_tree_object_visitor<df::task::classification>::call(
     }
     _onLeafNode(info);
     OVERFLOW_CHECK_BY_ADDING(std::size_t, node_id * 1 * this->class_count, info.get_response());
-    value_ar_ptr[node_id * 1 * this->class_count + info.get_response()] += info.get_sample_count();
+    this->value_ar[node_id * 1 * this->class_count + info.get_response()] += info.get_sample_count();
```
KulikovNikita, line 271:
```suggestion
    this->value_ar[node_id * this->class_count + info.get_response()] += info.get_sample_count();
```

onedal/ensemble/forest.py
ahuber21, line 1: In a perfect world, we'd have some unit tests for `forest.py` in which you could check your changes to `_validate_targets()`. Does such a thing exist / would it make sense to add it?

onedal/primitives/tree_visitor.cpp
```diff
@@ -75,14 +75,45 @@ class skl_tree_node {
 // We only expose the minimum information to python
 template <typename T>
```
ahuber21, line 76: What's the difference between `tree_state` and `tree_state_py` ? Is the former still required after your changes?

onedal/primitives/tree_visitor.cpp
```diff
@@ -153,52 +184,31 @@ to_sklearn_tree_object_visitor<Task>::to_sklearn_tree_object_visitor(std::size_t
...
-
-
     OVERFLOW_CHECK_BY_MULTIPLICATION(std::size_t, this->node_count, this->class_count);
-    double* value_ar_ptr =
-        new double[this->node_count * 1 *
-                   this->class_count](); // oneDAL only supports scalar responses for now
-
-    this->node_ar = py::array_t<skl_tree_node>(node_ar_shape, node_ar_strides, node_ar_ptr);
-    this->value_ar = py::array_t<double>(value_ar_shape, value_ar_strides, value_ar_ptr);
+    this->node_ar = new skl_tree_node[this->node_count];
```
ahuber21, line 188: Who owns this memory / how is it cleaned? I don't see any corresponding frees.

ahuber21 (changes requested):

onedal/ensemble/forest.py
```diff
@@ -533,7 +521,7 @@ def __init__(self,
                  voting_mode='weighted',
                  error_metric_mode='none',
                  variable_importance_mode='none',
-                 algorithm='dense',
+                 algorithm='hist',
```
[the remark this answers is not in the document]
icfaust: So far its only hardcoded to match daal4py (which has been always hardcoded to 'hist'). As to why daal4py has it hardcoded, that's very unclear to me. Maybe to guarantee performance gains? I expect its not so straightforward

onedal/ensemble/forest.cpp
```diff
@@ -216,6 +216,17 @@ void init_train_ops(py::module_& m) {
               train_ops ops(policy, input_t{ data, responses, weights }, params2desc{});
               return fptype2t{ method2t{ Task{}, ops } }(params);
           });
+    m.def("train",
+          [](const Policy& policy,
+             const py::dict& params,
+             const table& data,
+             const table& responses) {
+              using namespace decision_forest;
```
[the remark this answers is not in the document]
icfaust: I'll toy around with it, full disclosure I just copied the "train" def from above and removed weights, and I don't know pybind11 very well.

onedal/primitives/tree_visitor.cpp
icfaust: Ahh, so I added tree_state_py which matches behavior in daal4py's gettree: https://github.com/intel/scikit-learn-intelex/blob/master/src/gettree.pyx#L58 . The pyTreeState is used in a weird location: https://github.com/intel/scikit-learn-intelex/blob/master/generator/wrapper_gen.py#L1259 but does it at the very end.  The changes are related to the py::array_t only be used at the very end (thus internally only c++ data structures are used).

onedal/primitives/tree_visitor.cpp
icfaust: Maybe also a bug in the daal4py side? I did it to match this: https://github.com/intel/scikit-learn-intelex/blob/master/src/tree_visitor.h#L211

onedal/ensemble/forest.py
```diff
@@ -305,40 +305,21 @@ def _validate_targets(self, y, dtype):
...
-        elif isinstance(sample_weight, Number):
-            sample_weight = np.full(n_samples, sample_weight, dtype=dtype)
-        else:
-            sample_weight = _check_array(
-                sample_weight, accept_sparse=False, ensure_2d=False,
-                dtype=dtype, order="C"
-            )
-            if sample_weight.ndim != 1:
-                raise ValueError("Sample weights must be 1D array or scalar")
+        sample_weight = _check_array(
```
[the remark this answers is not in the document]
samir-nasibli: this is good question . And the quick answer is yes.
https://github.com/intel/scikit-learn-intelex/blob/9eb6afe124bc0c5daa7a4597391b68dc98c55f3b/onedal/datatypes/validation.py#L113

Actually there is a argument `copy` that is `False` by default. But if you will go through the implementation this arg used only once. This `copy` arg also should be checked here:
* https://github.com/intel/scikit-learn-intelex/blob/9eb6afe124bc0c5daa7a4597391b68dc98c55f3b/onedal/datatypes/validation.py#L138
* https://github.com/intel/scikit-learn-intelex/blob/9eb6afe124bc0c5daa7a4597391b68dc98c55f3b/onedal/datatypes/validation.py#L143


I think this primitive could be fixed in separate PR.

onedal/primitives/tree_visitor.cpp
samir-nasibli: In both it is owned on python side.

ahuber21: And how/where does the memory get freed and cleaned up?

onedal/primitives/tree_visitor.cpp
ahuber21: I didn't notice the subtle differences between in the data members `tree_state` and `tree_state_py`. Looks are both a justified.

onedal/primitives/tree_visitor.cpp
samir-nasibli: * For `onedal` implementation:
https://github.com/intel/scikit-learn-intelex/blob/master/onedal/primitives/tree_visitor.cpp#L301:L302
See `py::return_value_policy::take_ownership`. Please read about it https://pybind11-jagerman.readthedocs.io/en/stable/advanced.html#return-value-policies

* For `daal4py` implementation:
During daal4py building stage `getree.pyx` copies into `daal4py_cy.pyx` and then cythonizing gives `daal4py_cy.cpp`, find it in build dir.
It is really hard debug cython auto generated c code, I find this code from there.
```c
static void __pyx_tp_dealloc_7daal4py_8_daal4py_pyTreeState(PyObject *o) {
  struct __pyx_obj_7daal4py_8_daal4py_pyTreeState *p = (struct __pyx_obj_7daal4py_8_daal4py_pyTreeState *)o;
  #if CYTHON_USE_TP_FINALIZE
  if (unlikely(PyType_HasFeature(Py_TYPE(o), Py_TPFLAGS_HAVE_FINALIZE) && Py_TYPE(o)->tp_finalize) && !_PyGC_FINALIZED(o)) {
    if (PyObject_CallFinalizerFromDealloc(o)) return;
  }
  #endif
  PyObject_GC_UnTrack(o);
  Py_CLEAR(p->node_ar);
  Py_CLEAR(p->value_ar);
  (*Py_TYPE(o)->tp_free)(o);
}
```
So it seems that Cython also allows python ownership this.
 
**In any way I am agree that we should check mem leak.** (With valgrind or some other dynamic analysis tools)

ahuber21: Okay, especially the `daal4py` approach seems a bit magical. I agree that we should check for mem leaks. @icfaust is this something you could do or do you need help?

Alexsandruss:
@mergifyio rebase

onedal/primitives/tree_visitor.cpp
icfaust: Just as info @samir-nasibli @ahuber21 , @ahuber21 had it correct in terms of memory leaking.  Unfortunately the pybind11 documentation is extremely limited on the subject. If we look at the pybind11/numpy.h , the way that was implemented using new calls would create a memory leak due to the inherent copy of the array class in pybind11.  https://github.com/pybind/pybind11/blob/master/include/pybind11/numpy.h#L747  The way to do it is to pass a dummy base python object, of which py::none() is sufficient. This is definitely a work around that others have faced https://github.com/pybind/pybind11/issues/323

icfaust: Also PyArray_NewFromDescr will not initialize the memory to the default values of the class, so there was a bug in skl_tree_nodes for split nodes where certain values (like threshold) were not set to their defaults.

onedal/ensemble/forest.cpp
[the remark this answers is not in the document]
icfaust: I think I may write this down and have a discussion with @samir-nasibli @Alexsandruss about the design guidelines for the onedal module. This is done in SVM, KNN, and it every object in this file. I agree with your position, but I think this should be made into a rule for the codebase. Until then I want to stay consistent with the other files in the onedal module.

onedal/ensemble/forest.py
```diff
@@ -383,16 +357,24 @@ def _fit(self, X, y, sample_weight, module, queue):
...
+            policy = self._get_policy(queue, X, y, sample_weight)
+            X, y, sample_weight = _convert_to_supported(policy, X, y, sample_weight)
+            params = self._get_onedal_params(X)
+            train_result = module.train(policy, params, *to_table(X, y, sample_weight))
+
+        else:
+            policy = self._get_policy(queue, X, y)
+            X, y = _convert_to_supported(policy, X, y)
+            params = self._get_onedal_params(X)
+            train_result = module.train(policy, params, *to_table(X, y))
```
Alexsandruss, line 376:
```suggestion
        if sample_weight is not None and len(sample_weight) > 0:
            sample_weight = self._get_sample_weight(sample_weight, X)
            data = (X, y, sample_weight)
        else:
            data = (X, y)
        policy = self._get_policy(queue, *data)
        data = _convert_to_supported(policy, *data)
        params = self._get_onedal_params(data[0])
        train_result = module.train(policy, params, *to_table(*data))
```

onedal/primitives/tree_visitor.cpp
```diff
@@ -82,7 +82,7 @@ struct tree_state {
     std::size_t leaf_count;
     std::size_t class_count;
 };
-
+
```
Alexsandruss, line 85: Delete extra spaces
```suggestion

```

onedal/primitives/tree_visitor.cpp
```diff
@@ -278,8 +274,8 @@ template <typename Task>
 void init_get_tree_state(py::module_& m) {
     using namespace decision_forest;
     using model_t = model<Task>;
-    using tree_state_t = tree_state<Task>;
-
+    using tree_state_t = tree_state<Task>;    
+
```
Alexsandruss, line 278:
```suggestion
    using tree_state_t = tree_state<Task>;

```

Alexsandruss (changes requested):

samir-nasibli:
@icfaust please rebase and enable this tests #1393 , if it's required.

Alexsandruss:
Common PreCommit job with https://github.com/oneapi-src/oneDAL/pull/2441: http://intel-ci.intel.com/ee459771-cca8-f11a-93c5-a4bf010d0e2e

Alexsandruss:
http://intel-ci.intel.com/ee45d6a3-be9d-f123-adec-a4bf010d0e2e

Alexsandruss (approved):

PR #1334, opened 2023-06-15
Align `p` kNN parameter behavior with sklearn

Alexsandruss:
`p` kNN parameter should be assigned to default value of 2 even if `metric` value is set to one which doesn't mean `p=2` (for example, `l1` or `manhattan`). Additionally, fitting input and parameters validation was moved to common function for kNN estimators.
Required to produce same output from 'get_params' method as stock sklearn.

sklearn kNN API: https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html#sklearn.neighbors.KNeighborsClassifier


sklearnex/neighbors/common.py
```diff
@@ -15,15 +15,118 @@
...
+                is_n_neighbors_valid_for_brute = self.n_neighbors is not None and \
+                    self.n_neighbors >= self._fit_X.shape[0] // 2
+                if self._fit_X.shape[1] > 15 or is_n_neighbors_valid_for_brute:
+                    self._fit_method = "brute"
+                else:
+                    if self.effective_metric_ in VALID_METRICS["kd_tree"]:
+                        self._fit_method = "kd_tree"
+                    elif callable(self.effective_metric_) or \
+                        self.effective_metric_ in \
+                            VALID_METRICS["ball_tree"]:
```
KulikovNikita, line 78: What about completely removing `ball_tree` from the default branch? It is not implemented on the oneDAL side.

Another question - can we disable `kd_tree` for GPUs?

Alexsandruss: `self._fit_method="ball_tree"` is required later to decide if oneDAL patching is applicable. `kd_tree` on GPU is disabled - see `_onedal_supported` function

inteldimitrius (approved):

PR #1350, opened 2023-07-05
Fix inference for xgboost fitted with early-stopping

avolkov-intel:
# Description
Earlier there was a difference in results between xgboost predictions and predictions of our modelbuilder if xgboost was fitted using early-stopping parameter. This pull request is to fix this issue

### DIFF
--- a/src/gbt_convertors.pyx
+++ b/src/gbt_convertors.pyx
@@ -152,7 +152,8 @@ def get_gbt_model_from_xgboost(booster: Any) -> Any:
     else:
         is_regression = True
 
-    n_iterations = int(len(trees_arr) / (n_classes if n_classes > 2 else 1))
+    n_iterations = booster.best_ntree_limit
+    trees_arr = trees_arr[: n_iterations * (n_classes if n_classes > 2 else 1)]
 
     # Create + base iteration
     if is_regression:
--- a/src/gbt_convertors.pyx
+++ b/src/gbt_convertors.pyx
@@ -152,7 +152,7 @@ def get_gbt_model_from_xgboost(booster: Any) -> Any:
     else:
         is_regression = True
 
-    n_iterations = booster.best_ntree_limit
+    n_iterations = booster.best_iteration + 1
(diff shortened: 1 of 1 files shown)


avolkov-intel:
@Alexsandruss can we merge it?

src/gbt_convertors.pyx
```diff
@@ -152,7 +152,8 @@ def get_gbt_model_from_xgboost(booster: Any) -> Any:
     else:
         is_regression = True
 
-    n_iterations = int(len(trees_arr) / (n_classes if n_classes > 2 else 1))
+    n_iterations = booster.best_ntree_limit
```
razdoburdin, line 155: I think we should use [booster.best_iteration](https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBRegressor.best_iteration), not 
booster.best_ntree_limit.
From [this part of the code](https://github.com/dmlc/xgboost/blob/663136aa08c00598d8b49adf5901e4cb2ce187da/python-package/xgboost/core.py#L1143) I concluded, that best_ntree_limit will be removed soon.

Please take into account that best_iteration is starts from 0, while best_ntree_limit starts from 1. Thus we will have a bias equals to 1.

napetrov:
Also a broader question - can we add tests for this case?  we might do this as a second iteration after enabling fix due to release timelines

avolkov-intel:
> Also a broader question - can we add tests for this case? we might do this as a second iteration after enabling fix due to release timelines

Yes, I think we should do this in different pull request to merge that one before release

samir-nasibli:
> > Also a broader question - can we add tests for this case? we might do this as a second iteration after enabling fix due to release timelines
> 
> Yes, I think we should do this in different pull request to merge that one before release

I am expecting that we are testing the fix in the PR, so it is better to add simple test case in this PR as well.

napetrov:
Manual testing is fine for this PR integration, but should be added anyway

razdoburdin (approved):
Looks good

PR #1353, opened 2023-07-07
MAINT: refactor device offloading

samir-nasibli:
# Description
refactored `_get_backend` function:
removed unused variable and return value.

### DIFF
--- a/sklearnex/_device_offload.py
+++ b/sklearnex/_device_offload.py
@@ -128,23 +128,23 @@ def _transfer_to_host(queue, *data):
 def _get_backend(obj, queue, method_name, *data):
     cpu_device = queue is None or queue.sycl_device.is_cpu
     gpu_device = queue is not None and queue.sycl_device.is_gpu
-    cpu_fallback = False
+    onedal_cpu_supported = obj._onedal_cpu_supported(method_name, *data)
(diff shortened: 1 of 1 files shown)


KulikovNikita (approved):
Looks fine to me

sklearnex/_device_offload.py
```diff
@@ -128,23 +128,23 @@ def _transfer_to_host(queue, *data):
...
 
-    if (cpu_device and obj._onedal_cpu_supported(method_name, *data)) or \
-       (gpu_device and obj._onedal_gpu_supported(method_name, *data)):
-        return 'onedal', queue, cpu_fallback
+    if (cpu_device and onedal_cpu_supported) or \
+       (gpu_device and onedal_gpu_supported):
+        return 'onedal', queue
     if cpu_device:
-        return 'sklearn', None, cpu_fallback
+        return 'sklearn', None
```
Alexsandruss, line 138: This change introduce overhead for gpu support check: `_onedal_gpu_supported` was not called before if first condition for cpu is True.

samir-nasibli: Actually, it calls all checks here, but in any way overhead point is correct. Thanks!

Alexsandruss (approved):

PR #1356, opened 2023-07-11
feat: add black GHA for changed files

ahuber21:
Provide a github action that runs black code formatting in PRs.
I'm trying to configure it such that only changed files are checked.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 23.7.0
+    hooks:
+      - id: black
+        language_version: python3.10
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,3 +4,8 @@ repos:
     hooks:
       - id: black
         language_version: python3.10
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.12.0
+    hooks:
+      - id: isort
+        language_version: python3.10
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,19 @@
+#===============================================================================
+# Copyright 2023 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
 repos:
   - repo: https://github.com/psf/black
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,11 @@
+repos:
```
napetrov, line 1: let's move this file from root to .github

ahuber21: Is this a strong preference? Having it in the root directory is the convention and any contributor thinking about pre-commit hooks would expect it there. I think moving it to .github would defeat the point as no one would adapt using it.

napetrov: sorry - miss that this is pre-commit config. 

But in general i would move towards clean and tidy first level directory layout that would feet on one screen vs inflating it with more files

ahuber21:
I agree. Refactoring the requirements management will help a lot here!

.pre-commit-config.yaml
ahuber21: I agree. Refactoring the requirements management will help a lot here!

napetrov: yes. and in case on this repo - probably at some point doing a better split/separation of daal4py/scikit-learn-intelex

setup.py
```diff
@@ -15,24 +15,27 @@
...
-from scripts.version import get_onedal_version
+import setuptools.command.develop as orig_develop
+from Cython.Build import cythonize
+from setuptools import Extension, setup
+from setuptools.command.build_ext import build_ext as _build_ext
+
 import scripts.build_backend as build_backend
 from scripts.package_helpers import get_packages_with_tests
-from concurrent.futures import ThreadPoolExecutor
+from scripts.version import get_onedal_version
```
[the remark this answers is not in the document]
samir-nasibli: Sorry I have reviewed this part before, just forget submit review commit. Now it looks outdated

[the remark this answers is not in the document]
ahuber21: This is exactly what `isort` is doing and where the changes came from. Sometimes (when paths are not configured properly) it mixes third-party and local imports. Did you see that happening here? In this case I can update the config. But not in this PR, since I removed the changes in `setup.py` anyway.

samir-nasibli:
Thanks @ahuber21! Looks good to me. @napetrov @Alexsandruss could you please review?

Alexsandruss (changes requested):
It's not clear how `black` will affect code: for example, it changed license header when I manually applied it.
Using of it for only changed files will cause formatting inconsistency between changed part and old part of repo.

ahuber21:
Well it reformats the license because is not pep8 conform

> Block comments generally apply to some (or all) code that follows them, and are indented to the same level as that code. Each line of a block comment starts with a # and a single space (unless it is indented text inside the comment).

Therefore `black` adds the missing whitespace.

The effect of `black` is well defined and according to the docs, the coding style "can be viewed as a strict subset of PEP 8."
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html

We already have many style inconsistencies. I would prefer conserving a reasonable commit history for files that do not change often which is why I'm adding `black` only for changed files. I think the benefits outweigh the downsides (like more inconsistencies). Frequently used/changed files will quickly be aligned to the consistent coding style.

PR #1359, opened 2023-07-12
Add test for xgboost modelbuiler

avolkov-intel:
# Description
In this PR https://github.com/intel/scikit-learn-intelex/pull/1350 fix for xgboost modelbuilder was presented. The issue was about inference of model fitted with early_stopping_rounds parameter. In this PR I present the test that shows that there is no disrepancy between xgboost and xgboost modelbuilders.


avolkov-intel:
@razdoburdin, please, review this PR

examples/daal4py/model_builders_xgboost_early_stop_batch.py
```diff
@@ -0,0 +1,119 @@
+#===============================================================================
+# Copyright 2014 Intel Corporation
```
napetrov, line 2:
```suggestion
# Copyright 2023 Intel Corporation
```

napetrov:
@avolkov-intel  - let's not use examples as mean for testing - this is foe sure doesn't worth separate example to be created

samir-nasibli:
@avolkov-intel please also update [doc/daal4py/model-builders.rst](https://github.com/intel/scikit-learn-intelex/blob/master/doc/daal4py/model-builders.rst) for this example.

samir-nasibli (changes requested):
Please address CI fails

examples/daal4py/model_builders_xgboost_early_stop_batch.py
```diff
@@ -0,0 +1,119 @@
...
+        'verbosity': 0,
+        'objective': 'multi:softmax',
+        'learning_rate': 0.3,
+        'num_class': 5,
+        'early_stopping_rounds': 5
+    }
+
+    # Training
+    xgb_clf = xgb.XGBClassifier(**params)
+    xgb_clf.fit(X_train, y_train, eval_set=[(X_test, y_test)])
```
samir-nasibli, line 64: Do we really want the validation set is printed to stdout at each boosting stage?

avolkov-intel: You should provide eval_set to fit xgboost with early_stopping_rounds (this feature is used to avoid overfitting so dataset different from train should be provided)

avolkov-intel:
> @avolkov-intel please add unit test for that feature. Let's do it on separate PR

I suppose I will just remove example and add test in this PR since we do not need to add new example as Nikolay said

avolkov-intel:
@razdoburdin, please review the PR

razdoburdin (changes requested):
The test itself is fine. The problem is the failure of the private-ci. XGBoost is not installed in private-ci environment, so we need to skip the test in this case somehow.

napetrov:
@razdoburdin  @homksei  - might be it worth to just add xgboost package internally?

razdoburdin:
> @razdoburdin @homksei - might be it worth to just add xgboost package internally?

it will be the best option for us

tests/test_xgboost_mb.py
```diff
@@ -0,0 +1,87 @@
...
+reason += str(daal_version)
+
+
+class XgboostModelBuilder(unittest.TestCase):
+    @unittest.skipUnless(all([
+        hasattr(d4p, 'get_gbt_model_from_xgboost'),
+        hasattr(d4p, 'gbt_classification_prediction'),
+        daal_check_version(((2021, 'P', 1)))]), reason)
+    @unittest.skipUnless(importlib.util.find_spec('xgboost')
+                         is not None, 'xgoost library is not installed')
```
razdoburdin, line 40:
```suggestion
                         is not None, 'xgboost library is not installed')
```

samir-nasibli (changes requested):
I don't see that this tests are passed:
```
test_earlystop (test_xgboost_mb.XgboostModelBuilder) ... skipped 'xgboost library is not installed'
```
Please address this issue first

samir-nasibli:
Sorry, I see that this test passed on **public** CI:
```
test_earlystop (test_xgboost_mb.XgboostModelBuilder) ... DAAL version: (2023, 'P', 101)
```

samir-nasibli (approved):
looks good to me.

razdoburdin (approved):

PR #1360, opened 2023-07-12
MAINT: fixing mb examples names

samir-nasibli:
# Description
Address https://github.com/intel/scikit-learn-intelex/pull/1320#issuecomment-1633109340

PR #1364, opened 2023-07-14
Remove the batch suffix from examples

razdoburdin:
# Description
In continue of discussion started by @samir-nasibli in https://github.com/intel/scikit-learn-intelex/pull/1320#issuecomment-1633109340 and https://github.com/intel/scikit-learn-intelex/pull/1360.

I prepared an update for tests/run_examples.py:
- Now example names without suffix are supported.
- The processing of examples without suffix equivalent to the examples with `_batch` suffix in the priv version.
- List of skipping files is introduced.

**Issue:**
I found, that `log_reg_model_builder.py` return error. This file wasn't launching before. I have temporarily added it to the skipping list. We should fix this example or remove it.


tests/run_examples.py
```diff
@@ -171,6 +171,13 @@ def check_library(rule):
 
 req_os = defaultdict(lambda: [])
 
+skiped_files = ['__init__.py',
+                'spmd_utils.py',
+                'log_reg_model_builder.py',
+                'n_jobs.py',
```
napetrov, line 177: but we would like to run those examples to validate that they are work

razdoburdin: > but we would like to run those examples to validate that they are work

Most of these files are not the examples, but some helper functions. For this reason we shouldn't run it as examples. The only exception is `log_reg_model_builder.py`. For some unknown reason this file didn't have any suffix, for this reason, it wasn't launching before at all. So we have two independent tasks here:
1. Remove `_batch` suffix for more clear naming (this PR).
2. Investigate the problem with `log_reg_model_builder.py` and prepare a fix for this problem (some other PR). After the problem will be fixed, we can remove this file from the list of skipping examples.

This plan looks reasonable from my point of view, what is your opinion?

samir-nasibli: @razdoburdin I think as an option you can move non-example, helpers scripts into utils module `examples/utils`.

samir-nasibli:
@razdoburdin let's split the changes from this PR.
Please create separate PR with refactoring `examples` directory (isort and black fixes).
We will merge that PR first.
After rebasing in this PR, your changes will be clear for the review :)

samir-nasibli:
> @razdoburdin let's split the changes from this PR. Please create separate PR with refactoring `examples` directory (isort and black fixes). We will merge that PR first. After rebasing in this PR, your changes will be clear for the review :)

Please review https://github.com/intel/scikit-learn-intelex/pull/1386

examples/daal4py/log_reg_model_builder.py
```diff
@@ -13,6 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===============================================================================
+import sys
+
+print("KNOWN BUG IN EXAMPLES. TODO: fixme")
+sys.exit()
```
samir-nasibli, line 19: This will return `PASS` for the current examples.
It is better to add it to the skip list.

tests/run_examples.py
```diff
@@ -233,42 +232,51 @@ def run(exdir, logdir, nodist=False, nostream=False):
...
-            if any(
-                script.endswith(x)
-                for x in ["spmd.py", "streaming.py", "stream.py", "batch.py"]
-            ):
+            if script.endswith(".py") and script not in ["__init__.py"]:
                 n += 1
-                logfn = jp(logdir, script.replace(".py", ".res"))
-                with open(logfn, "w") as logfile:
+                if script in skiped_files:
+                    success += 1
```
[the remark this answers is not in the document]
razdoburdin: Agree. But CI compare `n` with the `success `and produce error in case of difference. For improving this logic we need to modify CI....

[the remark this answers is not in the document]
samir-nasibli: In this case it seems, we also have to update CI scripts for that

razdoburdin:
GPU CI job run is [here](http://intel-ci.intel.com/ee322b54-ad25-f18f-80d6-a4bf010d0e2e).

ethanglaser:
Should examples in daal4py/sycl also have the batch suffix removed?

razdoburdin:
> Should examples in daal4py/sycl also have the batch suffix removed?

I think yes, I suggest removing it in a separate PR.

ethanglaser:
> > Should examples in daal4py/sycl also have the batch suffix removed?
> 
> I think yes, I suggest removing it in a separate PR.

I think it could be added to this one if we are already doing both daal4py and sklearnex examples in this one

samir-nasibli:
GPU CI job run is [here](http://intel-ci.intel.com/ee435cf6-6ec2-f140-919d-a4bf010d0e2e).

samir-nasibli:
> > GPU CI job run is [here](http://intel-ci.intel.com/ee35ed43-6a11-f1fb-93ef-a4bf010d0e2e).
> 
> For reference, these are the examples that run in this job that did not run previously:
> 
> daal4py/log_reg_model_builder.py daal4py/model_builders_lightgbm.py daal4py/model_builders_catboost.py daal4py/model_builders_xgboost.py daal4py/spmd_utils.py daal4py/sycl/sklearn_sycl.py sklearnex/patch_sklearn.py sklearnex/verbose_mode.py sklearnex/n_jobs.py
> 
> Of these, it seems like the model builders should be ran. We have a separate step in recipe testing for daal4py/sycl/sklearn_sycl.py so this should be excluded from run_examples file, as well as spmd_utils since its just helper functions. Are sklearnex/ njobs, patch_sklearn, and verbose_mode intended to be ran with run_examples.py (@samir-nasibli)?

`spmd_utils.py` should be renamed or restructured. since this is not helper module.

napetrov:
daal4py/log_reg_model_builder.py - we can technically drop this if there are problems there. 

meanwhile lot of failures with  - TypeError: create_pytest_switches() missing 1 required positional argument: 'preview'

napetrov:
@Mergify rebase

samir-nasibli:
GPU CI job run is [here](http://intel-ci.intel.com/ee46767f-5695-f14e-82b6-a4bf010d0e2e).

ethanglaser:
> > GPU CI job run is [here](http://intel-ci.intel.com/ee35ed43-6a11-f1fb-93ef-a4bf010d0e2e).
> 
> For reference, these are the examples that run in this job that did not run previously:
> 
> daal4py/log_reg_model_builder.py daal4py/model_builders_lightgbm.py daal4py/model_builders_catboost.py daal4py/model_builders_xgboost.py daal4py/spmd_utils.py daal4py/sycl/sklearn_sycl.py sklearnex/patch_sklearn.py sklearnex/verbose_mode.py sklearnex/n_jobs.py
> 
> Of these, it seems like the model builders should be ran. We have a separate step in recipe testing for daal4py/sycl/sklearn_sycl.py so this should be excluded from run_examples file, as well as spmd_utils since its just helper functions. Are sklearnex/ njobs, patch_sklearn, and verbose_mode intended to be ran with run_examples.py (@samir-nasibli)?

daal4py/sycl/sklearn_sycl.py
sklearnex/patch_sklearn.py
sklearnex/verbose_mode.py
sklearnex/n_jobs.py
are all still running during run_examples step - I don't think this is desired behavior

razdoburdin:
> 

done

napetrov:
/intelci: run
set infra_branch=razdoburdin-patch-1

napetrov (approved):
ok with changes - assuming CI would be green

PR #1369, opened 2023-07-19
MAINT: restructure `onedal` utils

samir-nasibli:
# Description
* moved validation utils functions from `onedal.datatype` into separate `onedal.utils` module


ahuber21 (approved):
lgtm

PR #1382, opened 2023-07-31
Update dpcpp dependencies to 2023.2.0

Alexsandruss:
Changes:
- Update dpcpp build and runtime deps in CI to 2023.2.0

### DIFF
--- a/.ci/pipeline/build-and-test-lnx.yml
+++ b/.ci/pipeline/build-and-test-lnx.yml
@@ -24,7 +24,7 @@ steps:
     displayName: 'System info'
   - script: |
       conda update -y -q conda
-      if [ $(echo $(PYTHON_VERSION) | grep '3.7') ] || [ $(echo $(PYTHON_VERSION) | grep '3.11') ]; then export DPCPP_PACKAGE="dpcpp-cpp-rt>=2023.2.0"; else export DPCPP_PACKAGE="dpctl>=0.14 dpcpp-cpp-rt>=2023.2.0"; fi
+      if [ $(echo $(PYTHON_VERSION) | grep '3.7') ] || [ $(echo $(PYTHON_VERSION) | grep '3.11') ]; then export DPCPP_PACKAGE="dpcpp-cpp-rt>=2023.2.0"; else export DPCPP_PACKAGE="dpctl>=0.14 dpnp>=0.12.1 dpcpp-cpp-rt>=2023.2.0"; fi
       conda create -q -y -n CB -c conda-forge -c intel python=$(PYTHON_VERSION) dal-devel mpich pyyaml $DPCPP_PACKAGE
     displayName: 'Conda create'
   - script: |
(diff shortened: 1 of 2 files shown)


.ci/pipeline/build-and-test-lnx.yml
```diff
@@ -47,7 +47,6 @@ steps:
       bash .ci/scripts/setup_sklearn.sh $(SKLEARN_VERSION)
       pip install --upgrade -r requirements-test.txt -r requirements-test-optional.txt
       pip install $(python .ci/scripts/get_compatible_scipy_version.py)
-      if [ $(echo $(PYTHON_VERSION) | grep '3.8\|3.9\|3.10') ]; then conda install -q -y -c intel dpnp; fi
```
samir-nasibli, line 50: let's leave dpnp installation for testing. just specify version dpnp>=0.12*

.ci/pipeline/build-and-test-lnx.yml
```diff
@@ -24,7 +24,7 @@ steps:
     displayName: 'System info'
   - script: |
       conda update -y -q conda
-      if [ $(echo $(PYTHON_VERSION) | grep '3.7') ] || [ $(echo $(PYTHON_VERSION) | grep '3.11') ]; then export DPCPP_PACKAGE="dpcpp-cpp-rt>=2023.1.0"; else export DPCPP_PACKAGE="dpctl>=0.14 dpcpp-cpp-rt>=2023.1.0"; fi
+      if [ $(echo $(PYTHON_VERSION) | grep '3.7') ] || [ $(echo $(PYTHON_VERSION) | grep '3.11') ]; then export DPCPP_PACKAGE="dpcpp-cpp-rt>=2023.2.0"; else export DPCPP_PACKAGE="dpctl>=0.14.5 dpnp>=0.12.1 dpcpp-cpp-rt>=2023.2.0"; fi
```
samir-nasibli, line 27: We don't need dpnp on building, so there is no reason install it on conda creation step, I think

.ci/pipeline/build-and-test-lnx.yml
Alexsandruss: dpnp installation in testing stage will mix conda and pip commands which can result in env consistency problems in future. I prefer to preserve next order: conda env creation -> conda packages installation -> pypi packages installation

.ci/pipeline/build-and-test-lnx.yml
Alexsandruss: dpnp installation in testing stage will mix conda and pip commands which can result in env consistency problems in future. I prefer to preserve next order: conda env creation -> conda packages installation -> pypi packages installation

.ci/pipeline/build-and-test-lnx.yml
samir-nasibli: On main branch we already installing `dpnp` on testing step, and everything works. Here we can just specify dpnp version: `dpnp>=0.12*`

I propose to solve future problems as they come.

In addition, I vote for the purity of the flow. We are building sklearnex without dpnp and therefore I see no reason to add it to the build env.

Alexsandruss: dpnp is not essential scope of this PR so I removed it and we will discuss it later

PR #1391, opened 2023-08-07
chore: migrate linting from GHA to Azure

ahuber21:
We're experiencing long queue times for the GH linting action. Move to Azure for mitigation


ahuber21:
The CodeFactor failure is a `#TODO` comment in a formatted file which must be addressed separately.

samir-nasibli (changes requested):
Sorry, let's separate linting migration and code refactoring/style alignment.

ahuber21:
No problem @samir-nasibli, I've added a filter so that the CI targets only the files that we have already reformatted. If we decide to format the remaining files, the filter in `pyproject.toml` needs to be updated.

Alexsandruss (approved):

pyproject.toml
```diff
@@ -19,7 +19,9 @@
 line-length = 90
 target-version = ['py37', 'py38', 'py39', 'py310', 'py311']
 extend-ignore = 'E203'
+include = "(examples|onedal|sklearnex)/.*py$"
```
samir-nasibli, line 22: We have py sources not only in these mentioned directories.

samir-nasibli:
I am ok with the rest of the changes, but I think enabling format check only for specific directories is not right. Let's avoid.

ahuber21:
> I am ok with the rest of the changes, but I think enabling format check only for specific directories is not right. Let's avoid.

We cannot do both i) separate this migration and linting of files into two PRs (as you requested in your first comment), and ii) filter out files that weren't linted before.
Please make a suggestion of your preferred order of operations. I will provide the necessary changes.

samir-nasibli (approved):
We already discussed it. Let's do remaining formatting in separate branches and remove the filter then.

PR #1396, opened 2023-08-10
infra: adding assert_gpu flag to run_examples

ethanglaser:
# Description
run_examples script will parse for assert_gpu flag and raise error if GPU expected but not detected

also minor spelling availabe -> available


tests/run_examples.py
ahuber21, line 1: Can you please move all code that is executed into a 

```python
if __name__ == "__main__":
   ...
```

code block? What I typically do is create a `main(args)` that I call like

```python
if __name__ == "__main__":
  parser = argparse.ArgumentParser()
  # add arguments
  main(parser.parse_args()
```

tests/run_examples.py
```diff
@@ -25,6 +26,10 @@
 from daal4py import __has_dist__
 from daal4py.sklearn._utils import get_daal_version
 
+parser = argparse.ArgumentParser()
+parser.add_argument("--assert_gpu", action="store_true")
```
ahuber21, line 30: I think it's more typical to use only hyphens in CLI arguments, i.e. `--assert-gpu`. The variable is still available as `args.assert_gpu`, so you can rename it here without further changes.

```suggestion
parser.add_argument("--assert-gpu", action="store_true")
```

tests/run_examples.py
```diff
@@ -65,7 +70,7 @@
     (jp(examples_rootdir, "sklearnex"), jp(logdir, "sklearnex")),
 ]
 
-availabe_devices = []
+available_devices = []
```
ahuber21, line 73: When we have an explicit list of available devices, do we even still need the flag `gpu_available`? Wouldn't it be equivalent to saying `"gpu" in available_devices` ?

tests/run_examples.py
```diff
@@ -65,7 +70,7 @@
...
 
 try:
```
ahuber21, line 75: General question: Why is there no utility function like `environment.has_sycl_gpu_device()`? Maybe we could even put it into `sycl_context`, i.e. `sycl_context.has_gpu_available()` or whatever.

tests/run_examples.py
```diff
@@ -197,7 +205,9 @@ def get_exe_cmd(ex, nodist, nostream):
             req_version["sycl/" + os.path.basename(ex)], get_daal_version()
         ):
             return None
```
ahuber21, line 207: Can we just remove all the `None`s from the return? They're redundant.

ahuber21 (commented):
Looks good, I just have a few "Pythonic" comments.

tests/run_examples.py
```diff
@@ -197,7 +205,9 @@ def get_exe_cmd(ex, nodist, nostream):
...
-        if not check_device(req_device["sycl/" + os.path.basename(ex)], availabe_devices):
+        if not check_device(
```
ahuber21, line 208: Maybe one final question. What does `check_device` do? It's such a bad name, I don't understand what it's supposed to happen. Can we rename it or at least add a docstring to the function that helps me understand what's going on?

tests/run_examples.py
ethanglaser: You bring up several valid points, however I think most are related to the general structure and logic of the file instead of the changes being made, so they would probably be best addressed in a separate PR to keep this one concise to the changes being made. And perhaps best answered by @samir-nasibli as I believe he is the original author of this file.

ahuber21 (approved):

PR #1399, opened 2023-08-11
Add SHAP calculation to GBT regression

ahuber21:
Pendant to https://github.com/oneapi-src/oneDAL/pull/2460. This PR enables to new SHAP value functionality in scikit-learn-intelex.

The API is going to be similar to XGBoosts predict() method

```python
xgb_model = xgb.XGBRegressor(max_depth=6, n_estimators=100, random_state=3)
xgb_model.fit(X_train, y_train)
booster = xgb_model.get_booster()

d4p_model = daal4py.mb.convert_model(booster)
d4p_model.predict(X_test, pred_contributions=True)  # new: pred_contributions
d4p_model.predict(X_test, pred_interactions=True)  # new: pred_interactions
```


generator/gen_daal4py.py
```diff
@@ -272,19 +321,22 @@ class 'cls' and all its parents.
...
-                        cls
-                    ].iface != self.namespace_dict[pns].classes[parentclass].iface:
-                sanep = '{}::{}'.format(self.namespace_dict[ns].classes[cls].iface, sanep)
+            if (
+                pns is not None
+                and "interface" not in parent
+                and ns == pns
+                and self.namespace_dict[ns].classes[cls].iface
+                != self.namespace_dict[pns].classes[parentclass].iface
+            ):
```
KulikovNikita, line 338: Can you please use separate variables for all this conditions?

src/gbt_convertors-recursive-append.pyx
```diff
@@ -0,0 +1,755 @@
+# ===============================================================================
+# Copyright 2020 Intel Corporation
```
KulikovNikita, line 2: Year

src/gbt_model_builder.h
```diff
@@ -22,10 +22,12 @@
...
-#if (((MAJOR_VERSION == 2023) && (MINOR_VERSION >= 2)) || (MAJOR_VERSION > 2023))
-#define _gbt_inference_has_missing_values_support 1
+#if (((MAJOR_VERSION == 2024) && (MINOR_VERSION >= 1)) || (MAJOR_VERSION > 2024))
+    #define _gbt_inference_api_version 2
+#elif (((MAJOR_VERSION == 2023) && (MINOR_VERSION >= 2)) || (MAJOR_VERSION > 2023))
+    #define _gbt_inference_api_version 1
 #else
-#define _gbt_inference_has_missing_values_support 0
+    #define _gbt_inference_api_version 0
 #endif
```
razdoburdin, line 31: I suggest adding some comments about the meaning of this numbers. Like "version 1 added support of missing values, version 2 added support of shap-values".

src/gbt_convertors-recursive-append.pyx
```diff
@@ -0,0 +1,755 @@
...
+        split: Optional[float] = None,
+        value: Optional[List[float]] = None,
+        right: Optional[int] = None,
+        left: Optional[float] = None,
+    ) -> None:
+        self.split = split
+        self.value = value
+        self.right = right
+        self.left = left
+
```
razdoburdin, line 39: It looks like some logic in gbt_convertors-recursive-append duplicates the logic in gbt_convertors. Is it possible to avoid duplication?

ahuber21:
http://intel-ci.intel.com/ee674475-2b8f-f13d-9e94-a4bf010d0e2e

ahuber21:
Updated CI 
http://intel-ci.intel.com/ee691f19-524a-f1d7-8419-a4bf010d0e2e

napetrov:
i would also add example and documentation explaining new API

requirements-test.txt
```diff
@@ -6,3 +6,8 @@ scikit-learn==1.2.2 ; python_version == '3.8'
 scikit-learn==1.3.1 ; python_version >= '3.9'
 pandas==2.0.1 ; python_version == '3.8'
 pandas==2.1.1 ; python_version >= '3.9'
+xgboost==1.7.6; python_version <= '3.9'
```
napetrov, line 9: should it be test vs requirements-test-optional.txt?

ahuber21: Needed here, in `test_model_builders.py` (part of Intel CI) we are now comparing against XGBoost, LightGBM, CatBoost reference predictions.

ahuber21:
> i would also add example and documentation explaining new API

Added `examples/daal4py/model_builders_xgboost_shap.py` - where should I add the documentation?

requirements-test.txt
napetrov: well, same as it was for test_xgboost_mb

The problem is that  test requirements imply that this is test requirements for sklearnex while they are not.  

tests-optional is not great name and might be worth renaming to target daal4py only.  This might be also good for testing steps separation so daal4py and sklearnex would be tested in separate steps.  
This is also true for tests placement - now current design is that we have all test files in same place and xgboost_mb check if xgboost package is available and skip test if it's not. 

The better approach would be in separation of test files between scikit-learn and daal4py testing and making them non optional

napetrov:
> > i would also add example and documentation explaining new API
> 
> Added `examples/daal4py/model_builders_xgboost_shap.py` - where should I add the documentation?

https://github.com/intel/scikit-learn-intelex/blob/2c4dbb0593cbf3e5eff55a10cbc5fed1685b1505/doc/daal4py/model-builders.rst

requirements-test.txt
napetrov: More on that
Here are daal4py test deps  and test commands
https://github.com/conda-forge/daal4py-feedstock/blob/main/recipe/meta.yaml#L42

and this is scikit
https://github.com/conda-forge/scikit-learn-intelex-feedstock/blob/main/recipe/meta.yaml#L31

those have to be cleaned for 2024.0 in terms of commands but from tests perspective we don't need to bring GBT dependencies to scikit

ahuber21:
@napetrov done - updated the RST

napetrov:
> > i would also add example and documentation explaining new API
> 
> Added `examples/daal4py/model_builders_xgboost_shap.py` - where should I add the documentation?

Should we consider adding MB examples to separate directory? at same level as scikit/daal4py?  @razdoburdin  - might be you can share thoughts as well

The problem with keeping in daal4py examples is that it's hard to find, while in case of modelbuilders separate folder we can put all MB related examples in same place.  

I think that majority of users are not overlapping between MB use and daal4py use

razdoburdin:
> Should we consider adding MB examples to separate directory? at same level as scikit/daal4py? @razdoburdin - might be you can share thoughts as well

That is a good idea. In my mind the structure of examples should reproduce the structure of the [python imports](https://github.com/intel/scikit-learn-intelex/tree/master/daal4py). Currently we are not close to this structure. But if we want to move in this direction, we should place all the model builders related examples in examples/daal4py/mb folder.

ahuber21:
@napetrov @razdoburdin I have created the `model_builders` subdirectory. Did I catch all the examples?

razdoburdin:
> @napetrov @razdoburdin I have created the `model_builders` subdirectory. Did I catch all the examples?

Yes, they are all in place.
But I think we should move the subdirectory inside daal4py. @ahuber21 @napetrov, how do you think?

napetrov:
> `model_builders`

i would keep at same level as daal4py/scikit . Although technically this is part of daal4py but from usage scenario it ether of those.

The only question is this should be model_builders but ether "Model builders"  or MB?   The point - there is no places there we have model_builders format used.

ahuber21:
Don't have a strong opinion here @napetrov. Do you want me to rename the directory to `mb`?

napetrov:
> Don't have a strong opinion here @napetrov. Do you want me to rename the directory to `mb`?

I'm not sure on corrects one - but version with underline doesn't match other usages.  mb would be consistent with module name

napetrov:
Also i would assume you would be fixing ci to pass?

ahuber21:
@napetrov public CI won't build properly because my changes are targeting 2024.0.1 which is already set in the master branch, but so far the onDAL side is not merged. The latest Intel CI run is this one
https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/ee71a0d5-b79b-f185-804b-a4bf010d0e2e
The recent changes seem to have triggered a new failure - working on it.

doc/daal4py/model-builders.rst
```diff
@@ -61,22 +61,47 @@ CatBoost::
...
+
+The returned prediction will have shape ``(n_rows, n_features + 1)`` and ``(n_rows, n_features + 1, n_features + 1)``
+for SHAP contributions and interactions, respectively. Here, ``n_rows`` is the number of rows (i.e., observations) in
+``test_data``, and ``n_features`` is the number of features in the dataset.
+
+The prediction result for SHAP contributions comprises one feature attribution value per feature and a bias term for
+each observation.
+
+The prediction result for SHAP interactions comprises ``(n_features + 1) x (n_features + 1)`` values for all possible
+feature combinations, as well as the accompanying bias terms.
```
aepanchi, line 93:
```suggestion
feature combinations, along with their corresponding bias terms.
```

doc/daal4py/model-builders.rst
```diff
@@ -61,22 +61,47 @@ CatBoost::
...
+
+.. note:: The shapes of SHAP contributions and interactions are consistent with XGBoost's* results.
```
aepanchi, line 95:
```suggestion
.. note:: The shapes of SHAP contributions and interactions are consistent with the XGBoost results.
```

doc/daal4py/model-builders.rst
```diff
@@ -61,22 +61,47 @@ CatBoost::
...
+  In contrast, the popular `SHAP Python package* <https://shap.readthedocs.io/en/latest/>`_ drops bias terms, resulting
```
aepanchi, line 96:
```suggestion
  In contrast, the `SHAP Python package <https://shap.readthedocs.io/en/latest/>`_ drops bias terms, resulting
```

doc/daal4py/model-builders.rst
```diff
@@ -88,16 +113,17 @@ Limitations
 ------------------
 Model Builders support only base inference with prediction and probabilities prediction. The functionality is to be extended.
 Therefore, there are the following limitations:
-- The categorical features are not supported for conversion and prediction. 
+- The categorical features are not supported for conversion and prediction.
 - The multioutput models are not supported for conversion and prediction.
-- The tree SHAP calculations are not supported.
+- SHAP values can only be calculated for regression models.
```
aepanchi, line 118:
```suggestion
- SHAP values can be calculated for regression models only.
```

.ci/pipeline/build-and-test-lnx.yml
```diff
@@ -15,58 +15,58 @@
...
-      bash .ci/scripts/setup_sklearn.sh $(SKLEARN_VERSION)
-      pip install --upgrade -r requirements-test.txt -r requirements-test-optional.txt
-      pip install $(python .ci/scripts/get_compatible_scipy_version.py)
-      if [ $(echo $(PYTHON_VERSION) | grep '3.8\|3.9\|3.10') ]; then conda install -q -y -c intel dpnp; fi
+      . /usr/share/miniconda/etc/profile.d/conda.sh && \
+      conda activate CB && \
+      bash .ci/scripts/setup_sklearn.sh $(SKLEARN_VERSION) && \
+      pip install --upgrade -r requirements-test.txt -r requirements-test-optional.txt && \
+      pip install $(python .ci/scripts/get_compatible_scipy_version.py) && \
+      if [ $(echo $(PYTHON_VERSION) | grep '3.8\|3.9\|3.10') ]; then conda install -q -y -c intel dpnp; fi && \
```
Alexsandruss, line 50: Is there a smarter way to immediately exit script on fail like using of shell init script to set it?

requirements-test.txt
napetrov: let's keep single requirements-test file and drop one specific for dppy and test-optional?

setup.py
```diff
@@ -47,6 +48,7 @@
 IS_LIN = False
 
 dal_root = os.environ.get("DALROOT")
+n_threads = int(os.environ.get("NTHREADS", os.cpu_count() or 1))
```
napetrov, line 51: Do we need this for SHAP?

examples/mb/model_builders_xgboost_shap.py
```diff
@@ -0,0 +1,80 @@
+# ==============================================================================
```
napetrov, line 1: Will example from mb directory actually run?  I think new directory should be added to examples launcher script

requirements-test.txt
ahuber21: I actually can't find a place where `requirements-dppy.txt` is used. I'm fine with removing it.

.ci/pipeline/build-and-test-lnx.yml
ahuber21: Didn't find anything in the azure docs about it :( In any case, I'm moving this into a separate PR
https://github.com/intel/scikit-learn-intelex/pull/1552

setup.py
ahuber21: No, it is going into a separate PR https://github.com/intel/scikit-learn-intelex/pull/1553/files

.ci/pipeline/build-and-test-lnx.yml
ahuber21: More precisely. I think `set -e` could work (didn't test it). But using `&& \` is pretty common in Dockerfiles or similar script files. That's why I chose this syntax. It works and it is explicit. I like it.

requirements-test.txt
```diff
@@ -6,3 +6,9 @@ scikit-learn==1.2.2 ; python_version == '3.8'
...
+xgboost==2.0.0; python_version >= '3.10'
+lightgbm==4.1.0
+catboost==1.2.2; python_version <= '3.11'
+catboost>=1.2.2; python_version >= '3.12'
```
napetrov, line 13: i would drop this - there is not 3.12 python supported catboost and >= statement would also be resulting in not pinned version selection 
```suggestion
```

tests/test_model_builders.py
```diff
@@ -0,0 +1,780 @@
...
+from sklearn.linear_model import LogisticRegression
+from sklearn.model_selection import train_test_split
+
+import daal4py as d4p
+from daal4py.sklearn._utils import daal_check_version
+
+try:
+    import catboost as cb
+
+    cb_available = True
```
napetrov, line 37: i think we might end up in cases then lightgbm and xgboost will not be avaialable so would have same  skipps. 

The question, how we will be forcing actual tests execution, so we will not endup in state there steps will be green, but nothing will be tested. 

We can merge this one and discuss how to handle this

requirements-test.txt
ahuber21: Checked in with @samir-nasibli he wants to keep it for now. We can always clean up separately. Does not have to happen in this PR.

tests/test_model_builders.py
ahuber21: Yes, I was thinking about it. I first had a comment in `requirements-test.txt`. Looked like this

```text
catboost==1.2.2; python_version <= '3.11' ; FIXME: Add to 3.12 when supported
shap==0.42.1; python_version <= '3.11'     ; FIXME: Add to 3.12 when supported
```

But I did not like the idea of `FIXME` comments in requirements files. For now, I will just add a reminder for myself.

napetrov: This is not a problem- fix me is fine and I would expect that with dependencies updates those restrictions would be validated by person integrating.

But the question on validation - package install might fail or done incorrectly across all platforms and we will not catch this because ci would be green.

ahuber21: First, in this case I'm adding the `FIXME` comments back.

Second, I see the point. We might have to include such a testing gap in the release notes.
Something that says "Functionality dependent on SHAP and CatBoost packages not fully tested"

Alternatively, we could issue warnings on 3.12 if CatBoost functionality is used (like model building from CB models).
(Same for SHAP, but that is really only used in testing, so the user would never get any warnings.)

napetrov (approved):

PR #1414, opened 2023-08-23
Fix cmake wrapper call in setup.py

Alexsandruss:
Fix for dpc iface cmake wrapper in setup script: previously, `ONEDAL_MAJOR_BINARY_VERSION` was passed as `cxx` argument instead of `onedal_major_binary_version`.


ethanglaser (approved):

ethanglaser:
Looks ready to be merged along with the other two, assuming they look good as well.

PR #1423, opened 2023-08-30
sycl_context deprecation workaround

ethanglaser:
# Description
Workaround to avoid failures in sklearnex functionality due to oneDAL cpp sycl deprecation
 
- modifications to src to create default arg `from_python` that defaults to true when passed to oneDAL SyclExecutionContext for backwards compatibility
- moving examples/daal4py/sycl to test and updating accordingly

Note: public CI will be red until https://github.com/oneapi-src/oneDAL/pull/2480 is merged


src/oneapi/oneapi.pyx
```diff
@@ -24,7 +24,7 @@ from libcpp.string cimport string as std_string
 
 cdef extern from "oneapi/oneapi.h":
     cdef cppclass PySyclExecutionContext:
-        PySyclExecutionContext(const std_string & dev) except +
+        PySyclExecutionContext(const std_string & dev, const bool from_python) except +
```
samir-nasibli, line 27: I think, there is no need for such updates. Adding warnings here https://github.com/intel/scikit-learn-intelex/blob/164708ac2160c722a4eb11f233f7a44a84163c37/src/oneapi/oneapi.pyx#L114:L119 would be enough in my opinion.

src/oneapi/oneapi.pyx
```diff
@@ -111,11 +111,11 @@ from contextlib import contextmanager
 
 
 @contextmanager
-def sycl_context(dev='host', host_offload_on_fail=False):
+def sycl_context(dev='host', host_offload_on_fail=False, from_python=True):
     # Code to acquire resource
```
samir-nasibli, line 115: Just add warning here about the deprecation please.

tests/daal4py/sycl/covariance_streaming.py
```diff
@@ -37,6 +36,42 @@
 except:
```
napetrov, line 36:
```suggestion
except Exception:
```

tests/daal4py/sycl/covariance_streaming.py
```diff
@@ -37,6 +36,42 @@
...
 
+try:
+    import pandas
+
+    def read_csv(f, c=None, s=0, n=None, t=np.float64):
+        return pandas.read_csv(
+            f, usecols=c, delimiter=",", header=None, skiprows=s, nrows=n, dtype=t
+        )
+
+except:
```
napetrov, line 47:
```suggestion
except Exception:
```

tests/daal4py/sycl/low_order_moms_streaming.py
```diff
@@ -37,6 +36,42 @@
 except:
```
napetrov, line 36:
```suggestion
except Exception:
```

napetrov:
there are 2 more except hits  + there are 2 todo hits for one of sycl examples. I would say it safe to remove them as those will never be fixed and even test would be removed later on. make sense?

src/oneapi/oneapi.pyx
```diff
@@ -111,11 +111,11 @@ from contextlib import contextmanager
 
 
 @contextmanager
-def sycl_context(dev='host', host_offload_on_fail=False):
```
Alexsandruss, line 114: Add deprecation notice.

src/oneapi/oneapi.pyx
samir-nasibli: I vote not to change this function signature. Its already deprecated, we can use it as is.

samir-nasibli (commented):
Generally fine, added comment below.

src/oneapi/oneapi.pyx
samir-nasibli: last comment can be ignored. since this pr follow changes from https://github.com/oneapi-src/oneDAL/pull/2480

src/oneapi/oneapi.pyx
```diff
@@ -111,11 +111,11 @@ from contextlib import contextmanager
...
+def sycl_context(dev='host', host_offload_on_fail=False, from_python=True):
```
samir-nasibli, line 114: Please add warning here about the deprecation.

samir-nasibli (approved):
lgtm

ethanglaser:
Thanks for reviews. Will add deprecation notice to sycl_context - but one thing to note is that it will also raise the warning when users use GPU algos as intended (i.e. with config_context("gpu"): random_forest.fit() -> "warning, sycl_context deprecated") because some of our GPU interfaces use this under the hood.

napetrov:
> Thanks for reviews. Will add deprecation notice to sycl_context - but one thing to note is that it will also raise the warning when users use GPU algos as intended (i.e. with config_context("gpu"): random_forest.fit() -> "warning, sycl_context deprecated") because some of our GPU interfaces use this under the hood.

i would not add this for now because of those extra warnings

PR #1462, opened 2023-09-08
[bug] correct decision tree threshold default values for leaf nodes in daal4py and onedal

icfaust:
# Description
Decision tree thresholds do not match sklearn's convention in daal4py or onedal. This characteristic of setting leaf node thresholds to -2 exists since at least scikit-learn 0.22, and should be changed for all circumstances. See (sklearn 0.22) https://github.com/scikit-learn/scikit-learn/blob/0.22.X/sklearn/tree/_tree.pyx#L757 and (sklearn 1.4) https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/tree/_tree.pyx#L493

- change threshold default value to -2


KulikovNikita:
Looks correct to me.

KulikovNikita (approved):

PR #1480, opened 2023-09-18
Initial support of n_jobs parameter

Alexsandruss:
- Add `n_jobs` parameter to all sklearnex estimators using decorators


icfaust:
Could you send me a copy of the arch meeting presentation for this over email/chat? This looks like a reasonably large API enablement.

Alexsandruss:
> Could you send me a copy of the arch meeting presentation for this over email/chat? This looks like a reasonably large API enablement.

It's not an API change in terms of sklearn. `n_jobs` parameter was listed in some of estimators while sklearnex doesn't take it in consideration currently. The only difference is sklearn supporting this parameter for subset of estimators while sklearnex will support it for all estimators.

icfaust:
This is a substantive change in the sklearnex codebase, and even though it may exist in sklearn API it should have been a patching condition check in sklearnex.  From my perspective, I will have to now understand for the preview implementation how this will interact with the oneDAL backend, which isn't trivial for making sure that it matches this implementation. This also impact most (if not all) sklearnex algorithms, and I believe the structure of this warrants a discussion in the Architecture meeting.  At a minimum there should be some sort of understanding why/how this was implemented this way, such that others can work well with it in the future. @samir-nasibli @KulikovNikita, what do you think?

napetrov:
let's also update example - https://github.com/intel/scikit-learn-intelex/blob/master/examples/sklearnex/n_jobs.py

examples/sklearnex/n_jobs.py
```diff
@@ -14,13 +14,8 @@
...
 
-# sklearnex doesn't have interface for threading configuration and not following
-# scikit-learn n_jobs yet. Thus it's requered to use daal4py package to set this.
-# nthreads parameter define number of threads used by sklearnex.
-# Without this code sklearnex would be using all system cores
-import daal4py
-
-daal4py.daalinit(nthreads=2)
+# sklearnex support `n_jobs` parameter for all patched estimators
+# even if original sklearn estimator doesn't
```
Vika-F, line 18: I'd leave a bit modified sentence from the previous version of the code:

"n_jobs parameter defines the number of threads used by sklearnex.
Without this code sklearnex would be using all system cores."

Vika-F: I mean in addition to the comment you've added.

Vika-F (commented):
The code looks good. I have only minor comments regarding the descriptions.

daal4py/sklearn/_utils.py
```diff
@@ -257,3 +266,119 @@ def get_status(self, logs=False):
         if logs:
             self.write_log()
         return self.patching_is_enabled
+
+
+# decorator for addition of "n_jobs" parameter to estimator's init
+def support_init_with_n_jobs(init_function):
```
ahuber21, line 272: I generally prefer function descriptions to be provided like this. I would argue it's more Pythonic and more widely used. Do we have a guideline on this? We should have one ...

```suggestion
def support_init_with_n_jobs(init_function):
    """decorator for addition of "n_jobs" parameter to estimator's init"""
```

daal4py/sklearn/_utils.py
```diff
@@ -257,3 +266,119 @@ def get_status(self, logs=False):
...
+                "'Threading' parallel backend is not supported by "
+                "Intel(R) Extension for Scikit-learn*. "
+                "Falling back to usage of all available threads."
+            )
+            result = method(self, *args, **kwargs)
+            return result
+        # multiprocess parallel backends branch
+        method_name = (
+            f"{self.__class__.__module__}.{self.__class__.__name__}.{method.__name__}"
+        )
```
ahuber21, line 351: This will result in code that's more readable, even after auto-formatting
```suggestion
        cls = self.__class__
        method_name = ".".join([cls.__module__, cls.___name__, method.__name__])
```

ahuber21 (commented):
Great work! I just have a few stylistic comments.

daal4py/sklearn/_utils.py
```diff
@@ -257,3 +266,119 @@ def get_status(self, logs=False):
...
+        if (
+            hasattr(self, "__doc__")
+            and isinstance(self.__doc__, str)
+            and "n_jobs : int" not in self.__doc__
+        ):
+            parameters_doc_tail = "\n    Attributes"
+            n_jobs_doc = """
+    n_jobs : int, default=None
+        The number of jobs to use in parallel for the computation.
+        ``None`` means 1 unless in a :obj:`joblib.parallel_backend` context.
```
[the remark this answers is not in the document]
Alexsandruss: Done.

daal4py/sklearn/_utils.py
```diff
@@ -257,3 +266,119 @@ def get_status(self, logs=False):
...
+
+    return init_with_n_jobs
+
+
+# Note: getting controller in global scope of this module is required
+# to avoid overheads by its initialization per each function call
+threadpool_controller = threadpoolctl.ThreadpoolController()
+
+
+def get_suggested_n_threads(n_cpus):
```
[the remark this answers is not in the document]
Alexsandruss: Done.

daal4py/sklearn/_utils.py
Alexsandruss: Done. Btw, VSCode thinks `cls` is reserved word, so changed it to `cl`

daal4py/sklearn/_utils.py
Alexsandruss: Done.
No guidelines exist for this currently.

daal4py/sklearn/_utils.py
```diff
@@ -257,3 +266,119 @@ def get_status(self, logs=False):
...
+def support_init_with_n_jobs(init_function):
+    @wraps(init_function)
+    def init_with_n_jobs(self, *args, n_jobs=None, **kwargs):
+        if sklearn_check_version("1.2") and hasattr(self, "_parameter_constraints"):
+            parameter_constraints = self._parameter_constraints
+            if "n_jobs" not in parameter_constraints:
+                parameter_constraints["n_jobs"] = [Integral, None]
+        init_function(self, *args, **kwargs)
+        # add n_jobs to __doc__ string if needed
+        if (
```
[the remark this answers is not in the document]
Alexsandruss: Thanks for inspiration, I changed init method decorator to class decorator. Now, `n_jobs` is partially suggested or highlighted in IDEs like VSCode, but static type checking tools can't fully deduct it.

Vika-F (approved):
From my point of view this PR is ready to be merged.

sklearnex/ensemble/_forest.py
```diff
@@ -1122,6 +1128,7 @@ def predict(self, X):
     predict.__doc__ = sklearn_ForestRegressor.predict.__doc__
 
 
+@support_n_jobs
```
icfaust, line 1131: Please remove @support_n_jobs for all of the classes in _forest.py, n_jobs has been supported by an inherited sklearn class since at least 0.24 https://github.com/scikit-learn/scikit-learn/blob/0.24.X/sklearn/ensemble/_forest.py#L204. All the actions in that function are unnecessary. If it still must be included, use ForestClassifier and ForestRegressor classes instead.

daal4py/sklearn/_utils.py
```diff
@@ -257,3 +266,131 @@ def get_status(self, logs=False):
...
+        return None
+
+
+def run_with_n_jobs(method):
+    """Decorator for running of methods containing oneDAL kernels with 'n_jobs'"""
+
+    @wraps(method)
+    def method_wrapper(self, *args, **kwargs):
+        # threading parallel backend branch
+        if not isinstance(threading.current_thread(), threading._MainThread):
```
icfaust, line 353: Possibly a dumb question: In the case that someone spawns a sklearnex routine in additional thread, it this will just run as normal?

daal4py/sklearn/_utils.py
```diff
@@ -257,3 +266,131 @@ def get_status(self, logs=False):
...
+            if n_jobs != old_n_threads:
+                logger = logging.getLogger("sklearnex")
+                logger.debug(
+                    f"{method_name}: setting {n_jobs} threads (previous - {old_n_threads})"
+                )
+                set_n_threads(n_jobs)
+        # run method
+        result = method(self, *args, **kwargs)
+        # reset number of threads to old one
+        if n_jobs is not None:
```
icfaust, line 391: similar for this if statement.

sklearnex/ensemble/_forest.py
```diff
@@ -46,7 +46,9 @@
 from daal4py.sklearn._utils import (
     check_tree_nodes,
     daal_check_version,
+    run_with_n_jobs,
     sklearn_check_version,
+    support_n_jobs,
```
icfaust, line 51: will probably need to remove this from the import tuple

icfaust (changes requested):
Please message me when changes are made and I will re-review

daal4py/sklearn/_utils.py
Alexsandruss: It will just run as normal as it was before this change. Another issue is an oversubscription which might occur by calling oneDAL in different threads.

daal4py/sklearn/_utils.py
```diff
@@ -257,3 +266,131 @@ def get_status(self, logs=False):
...
+        # using sklearn rules and `n_threads` from upper parallelism context
+        if n_jobs is None or n_jobs == 0:
+            n_jobs = n_threads
+        elif n_jobs < 0:
+            if n_threads is None:
+                n_jobs = max(1, n_cpus + n_jobs + 1)
+            else:
+                n_jobs = max(1, n_threads + n_jobs + 1)
+        # set number of threads
+        if n_jobs is not None:
```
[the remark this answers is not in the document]
Alexsandruss: n_jobs might be assigned earlier to n_threads which is Union[None, int], so these lines make sense.

sklearnex/ensemble/_forest.py
Alexsandruss: It wasn't clear this decorator is targeted to control n_jobs behaviour in all estimators patched by oneDAL so I renamed it to `control_n_jobs` and extended comment.

daal4py/sklearn/_utils.py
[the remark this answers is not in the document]
icfaust:
```python
if n_threads is None:
   if n_jobs is None or n_jobs == 0:
     return method(self, *args, **kwargs)
   else:
     n_threads = cpu_count()

if n_jobs is None or n_jobs == 0:
   n_jobs = n_threads
elif n_jobs < 0:
   n_jobs = max(1, n_threads + n_jobs + 1)
```
Reduces if statement checks and only finds cpu_count if necessary

daal4py/sklearn/_utils.py
```diff
@@ -257,3 +266,131 @@ def get_status(self, logs=False):
...
+        original_params.append(Parameter("n_jobs", Parameter.KEYWORD_ONLY, default=None))
+        init_with_n_jobs.__signature__ = sig.replace(parameters=original_params)
+
+    original_class.__init__ = init_with_n_jobs
+
+    # add n_jobs to __doc__ string if needed
+    if (
+        hasattr(original_class, "__doc__")
+        and isinstance(original_class.__doc__, str)
+        and "n_jobs : int" not in original_class.__doc__
```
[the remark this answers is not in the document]
Alexsandruss: - This decorator works during imports
- It takes less than 1 ms

icfaust (changes requested):

sklearnex/ensemble/_forest.py
icfaust: I now understand the necessity of doing so. For the patched estimators in _forest, they are thin APIs over ForestClassifier and ForestRegressor. I believe for maintainability purposes, it is best to switch these decorators to those classes. I will agree to the changes to the direct patched algos if that was the agreed upon rule in the arch meeting. In general, this will then need to be written somewhere for future introduced algos.

daal4py/sklearn/_utils.py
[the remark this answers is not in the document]
icfaust: Maybe:

```python
        if n_jobs is None or n_jobs == 0:
            if n_threads is None:
                return method(self, *args, **kwargs)
            else:
                n_jobs = n_threads
        elif n_jobs < 0:
            if n_threads is None:
                n_jobs = max(1, n_cpus + n_jobs + 1)
            else:
                n_jobs = max(1, n_threads + n_jobs + 1)

```

icfaust (approved):

deselected_tests.yaml
```diff
@@ -32,6 +32,20 @@ deselected_tests:
...
+  - tests/test_common.py::test_estimators[PCA()-check_no_attributes_set_in_init]
+  - tests/test_common.py::test_estimators[KMeans()-check_no_attributes_set_in_init]
+  - tests/test_common.py::test_estimators[Lasso()-check_no_attributes_set_in_init]
+  - tests/test_common.py::test_estimators[Ridge()-check_no_attributes_set_in_init]
+  - tests/test_common.py::test_estimators[ElasticNet()-check_no_attributes_set_in_init]
+  - tests/test_common.py::test_estimators[NuSVC()-check_no_attributes_set_in_init]
+  - tests/test_common.py::test_estimators[NuSVR()-check_no_attributes_set_in_init]
+  - tests/test_common.py::test_estimators[SVC()-check_no_attributes_set_in_init]
+  - tests/test_common.py::test_estimators[SVR()-check_no_attributes_set_in_init]
+
```
samir-nasibli, line 48: I still don't understand the logic of this test deselection.
Are we going to fix it?

PR #1491, opened 2023-09-22
[workaround] fix cpu offloading

samir-nasibli:
# Description
Addressed to fix examples/sklearnex: `n_jobs`, `patch_sklearn` and `verbose_mode` run with the new compiler env.

# TODO
will be cleaned and updated if checks success


samir-nasibli:
@napetrov It seems to me that we can merge this fix.
In follow up I am going to check if `_Daal4PyContextReset` is still required. If this is not used, I will remove it at all.

napetrov (approved):

PR #1440, opened 2023-09-06
[Enhancement] Unify Random Forest and Extra Trees preview algorithms

icfaust:
# Description
This unifies Random Forest and Extra Trees to a single file from the same inheritance classes to match stock scikit-learn. This reduces overall code size, and prepares for movement out of preview. This improves preview RF to meet some of the standards of sklearnex in terms of testing and verbosity. It is still an open question whether to rename extra_trees.py to forest.py, though this would be temporary before the move out of preview.

Changes proposed in this pull request:
- Fix TODO associated with _estimator base_estimator_ and estimator_
- add sklearn v1.4 monotonic_cst keyword and API input
- move RF algos to extra_trees.py, rename to _forest.py (matches sklearn/daal4py naming convention)
- remove forest.py
- subtle reorder of patching_status to exactly match daal4py RF order
- correct issue associated with 'balanced_subsample' in class_weight
- remove _onedal_model attribute and focus on _onedal_estimator instead
- fix testing issue with test_min_samples_leaf for daal4py
- spoof inheritance of RF/ET algos to allow for proper isinstance checking
- fix fallback issue for warm_start caused by estimators_ lazy evaluation
- Re-add CI tests from deselected_tests for RF and ET
- splitter_mode kwarg in preview RF is removed, splitter mode is checked by the passed tree object to match sklearn
- Change import in RF spmd interface (generalization)
- NOTE: binning strategy (see daal4py RF) not implemented

PR #1509, opened 2023-09-28
[refactor] Import corrections using isort listed in CI

icfaust:
# Description
These changes are raised as errors in the Lint CI step, but doesn't cause a CI failure (aka still green). This is a cosmetic change to follow/using isort.

It looks like half of the errors are my fault... Sorry


ethanglaser:
Is it a follow up to https://github.com/intel/scikit-learn-intelex/pull/1440? Looks like lint step passed there in public CI

PR #1515, opened 2023-09-28
chore: apply isort after fixing CI step

ahuber21:


napetrov (approved):

napetrov:
> Before we merge this, it looks like #1510 didn't fix it, as more recent PRs still pass lint even with these issues. Maybe a follow up PR linked to this would be sufficient.

i see for last PR step is failing 
image

PR #1507, opened 2023-09-28
[testing] Fixing Extra Trees and Random Forest preview test_common.py issues

icfaust:
# Description
Unify and modernize aspects of preview ensemble for test coverage and maintainability

Changes proposed in this pull request:
- _estimators_ to BaseForest
- onedal_fit to BaseForest (matches sklearn)
- reduce code by using matching sklearn native functions (for max_samples test)
- use class attributes rather than instance attributes for clarity and testing for sklearn <1.1
- fixed base_estimator setting for sklearn versions <1.2
- cleanup in cpu_supported to better match data prep in sklearn
- fix issues with fit() and y shape switching back and forth from 2 dimensions to 1 causing an errant warning
- fixed test_common.py test failure caused by an issue in predict(), check_array was casting data to float32 instead of float64, causing an estimator test to fail (only occurred when not taking pandas or numpy arrays)
- fixed test_common.py test failure caused when y=None, would trigger a different warning than desired
- Changed sklearn_check_versions to better match the sklearn versions (i.e. validate_data from 0.24)

This PR solves the following unit test failures in public Nightly CI:
inspection/tests/test_permutation_importance.py::test_permutation_importance_equivalence_array_dataframe[0.5-None]
inspection/tests/test_permutation_importance.py::test_permutation_importance_equivalence_array_dataframe[0.5-1]
inspection/tests/test_permutation_importance.py::test_permutation_importance_equivalence_array_dataframe[0.5-2]
inspection/tests/test_permutation_importance.py::test_permutation_importance_equivalence_array_dataframe[1.0-None]
inspection/tests/test_permutation_importance.py::test_permutation_importance_equivalence_array_dataframe[1.0-1]
inspection/tests/test_permutation_importance.py::test_permutation_importance_equivalence_array_dataframe[1.0-2]
tests/test_common.py::test_estimators[ExtraTreesClassifier()-check_requires_y_none]
tests/test_common.py::test_estimators[ExtraTreesRegressor()-check_regressor_data_not_an_array]
tests/test_common.py::test_estimators[ExtraTreesRegressor()-check_requires_y_none]
tests/test_common.py::test_estimators[RandomForestRegressor()-check_regressor_data_not_an_array]

PR #1528, opened 2023-10-06
[testing, bug] include sklearn fallback for normal non-invertible matrices in preview linearRegression

icfaust:
# Description
Daal contains several [methods for linear regression](https://github.com/oneapi-src/oneDAL/blob/master/cpp/daal/include/algorithms/linear_regression/linear_regression_training_types.h#L58). oneDAL uses daal for CPU, but only has the [norm_eq method available](https://github.com/oneapi-src/oneDAL/blob/master/cpp/oneapi/dal/algo/linear_regression/parameters/cpu/train_parameters.cpp#L42). Daal4py linear regression [includes a fallback](https://github.com/intel/scikit-learn-intelex/blob/master/daal4py/sklearn/linear_model/_linear.py#L61) for non-invertible normal matrices:  using the other method before falling back to sklearn. sklearnex.preview.linearRegression does not have either of these fallbacks, meaning it cannot handle norm_eq non-invertible matrices. Ideally additional work should be done on the oneDAL side to include the QRdense method, and then expose it to sklearnex. This PR can be considered a stopgap bugfix as it falls directly back to sklearn.

This causes unit test failures for sklearn preview in public Nightly:
inspection/tests/test_partial_dependence.py::test_partial_dependence_easy_target[1-est0]
inspection/tests/test_partial_dependence.py::test_partial_dependence_easy_target[1-est1]
inspection/tests/test_partial_dependence.py::test_partial_dependence_easy_target[1-est2]
inspection/tests/test_partial_dependence.py::test_partial_dependence_easy_target[1-est3]
inspection/tests/test_partial_dependence.py::test_partial_dependence_easy_target[2-est0]
inspection/tests/test_partial_dependence.py::test_partial_dependence_easy_target[2-est1]
inspection/tests/test_partial_dependence.py::test_partial_dependence_easy_target[2-est2]
inspection/tests/test_partial_dependence.py::test_partial_dependence_easy_target[2-est3]
- Include a try catch for a Runtime error, fallback to sklearn due to non-invertibility.
- Fixed issue with sklearnex get_patch_message which wasn't implemented but necessary for bugfix

### DIFF
--- a/sklearnex/preview/linear_model/linear.py
+++ b/sklearnex/preview/linear_model/linear.py
@@ -305,7 +305,7 @@ if daal_check_version((2023, "P", 100)):
                 self._save_attributes()
 
             except RuntimeError:
-                logging.warning(get_patch_message('sklearn_after_daal'))
+                logging.warning(get_patch_message("sklearn_after_daal"))
 
                 del self._onedal_estimator
                 super().fit(X, y)
--- a/sklearnex/preview/linear_model/linear.py
+++ b/sklearnex/preview/linear_model/linear.py
@@ -305,7 +305,7 @@ if daal_check_version((2023, "P", 100)):
                 self._save_attributes()
 
             except RuntimeError:
-                logging.getlogger("sklearnex").info(
+                logging.getLogger("sklearnex").info(
                     f"{self.__class__.__name__}.fit "
                     + get_patch_message("sklearn_after_onedal")
                 )
--- a/sklearnex/preview/linear_model/linear.py
+++ b/sklearnex/preview/linear_model/linear.py
@@ -305,7 +305,10 @@ if daal_check_version((2023, "P", 100)):
                 self._save_attributes()
 
             except RuntimeError:
-                logging.warning(get_patch_message("sklearn_after_daal"))
+                logging.getlogger("sklearnex").info(
+                    f"{self.__class__.__name__}.fit "
+                    + get_patch_message("sklearn_after_onedal")
+                )
 
                 del self._onedal_estimator
                 super().fit(X, y)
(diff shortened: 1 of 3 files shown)


sklearnex/preview/linear_model/linear.py
```diff
@@ -300,9 +300,13 @@ def _onedal_fit(self, X, y, sample_weight, queue=None):
...
             self._initialize_onedal_estimator()
-            self._onedal_estimator.fit(X, y, queue=queue)
+            try:
+                self._onedal_estimator.fit(X, y, queue=queue)
 
-            self._save_attributes()
+                self._save_attributes()
+            except RuntimeError:
+                del self._onedal_estimator
+                super().fit(X, y)
```
KulikovNikita, line 309: Looks weird. Yes, it will fix the problem for now. Unfortunately we don't quite understand what is causing the problem on onedal side

Alexsandruss, line 309: No warning about fallback to stock implementation is provided.

Alexsandruss (changes requested):

sklearnex/preview/linear_model/linear.py
```diff
@@ -300,9 +300,18 @@ def _onedal_fit(self, X, y, sample_weight, queue=None):
                 )
 
             self._initialize_onedal_estimator()
-            self._onedal_estimator.fit(X, y, queue=queue)
+            try:
+                self._onedal_estimator.fit(X, y, queue=queue)
+                self._save_attributes()
+
+            except RuntimeError:
+                logging.getlogger("sklearnex").info(
```
Alexsandruss, line 308:
```suggestion
                logging.getLogger("sklearnex").info(
```

icfaust:
manual rerun of private CI http://intel-ci.intel.com/ee644fea-91ee-f107-96d4-a4bf010d0e2e

icfaust:
Pre-commit failure is intermittent and unrelated to this PR, see manual rereun

icfaust:
image

icfaust:
image
The failed tests shown here are solved in #1507

Alexsandruss (approved):

PR #1543, opened 2023-10-20
[performance] remove _assert_all_finite() call from onedal/utils/_check_array

icfaust:
# Description
sklearn's check_array() contains force_all_finite=True as default. It also [handles pandas dataframes](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/validation.py#L835) before doing finite checks which could be the cause of a performance degradation on random forest preview predict.

- Remove if statement entirely

### DIFF
--- a/onedal/utils/validation.py
+++ b/onedal/utils/validation.py
@@ -125,11 +125,6 @@ def _check_array(
     ensure_2d=True,
     accept_large_sparse=True,
 ):
-    if force_all_finite:
-        if sp.issparse(array):
-            if hasattr(array, "data"):
-                _assert_all_finite(array.data)
-                force_all_finite = False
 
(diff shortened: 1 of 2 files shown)


icfaust:
@Alexsandruss Should we also remove the [if statement entirely](https://github.com/intel/scikit-learn-intelex/blob/master/onedal/utils/validation.py#L129)? Given that check_array has [_assert_all_finite](https://github.com/scikit-learn/scikit-learn/blob/d99b728b3/sklearn/utils/validation.py#L957), and that in [assert_all_finite it has a a sparsity check similar to the if statement](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/validation.py#L204)?

PR #1546, opened 2023-10-23
Remove 'filter_selector' deprecated in SYCL 2020 specification

Vika-F:
'filter_selector' usage was removed.


icfaust:
As far as I found, there isn't a [plug-in-replacement](https://intel.github.io/llvm-docs/doxygen/filter__selector_8hpp_source.html) for filter_selector in dpcpp for SYCL2020...

Vika-F:
@icfaust yes, but this is not a reason to have a deprecated functionality in the codebase. We will need to implement a replacement anyway (as simple removal does not work).

PR #1552, opened 2023-10-25
fix: update CI scripts to fail on intermediate failures

ahuber21:
The CI scripts would only trigger a failed step if the last command in each `script` returned a non-zero exit code. By logically `and`ing all commands, intermediate commands can fail the entire step.

### DIFF
--- a/.ci/pipeline/build-and-test-lnx.yml
+++ b/.ci/pipeline/build-and-test-lnx.yml
@@ -15,58 +15,58 @@
 #===============================================================================
 steps:
   - script: sudo apt-get update && sudo apt-get install -y clang-format
-    displayName: 'apt-get'
+    displayName: "apt-get"
   - script: |
       bash .ci/scripts/install_dpcpp.sh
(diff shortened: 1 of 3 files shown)


napetrov:
Should "failOnStderr: string # Fail the task if output is sent to Stderr?  " be concidered instead?

https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-bash?view=azure-pipelines

ahuber21:
> Should "failOnStderr: string # Fail the task if output is sent to Stderr? " be concidered instead?
> 
> https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-bash?view=azure-pipelines

Python warnings are printed to `stderr`, so we can't use this option consistently. I researched a bit and couldn't find a better option than the one proposed in this PR.

napetrov (approved):
Ok, as intermediate improvement. Need broader fix and cleanup though

ethanglaser (approved):

PR #1554, opened 2023-10-25
CI: Add SHELLOPTS for immediate exit from pipeline script

Alexsandruss:
Add `SHELLOPTS='errexit:pipefail'` on Unix systems for immediate exit from pipeline step script if last command has non-zero exit code.
For example of work, see jobs in https://github.com/intel/scikit-learn-intelex/pull/1554/commits/08210a559dd99b6fe4f89fb54e8f916bd14dba32 commit.


ahuber21 (approved):
I'm fine if it works the same way

PR #1561, opened 2023-10-31
Update `sed` command for sklearn version

Alexsandruss:
# Description
`sed` doesn't replace sklearn version correctly, this PR aims to fix it.

PR #1563, opened 2023-11-01
[deps] Restore sklearn version condition

ethanglaser:
# Description

-
-
-

### DIFF
--- a/.ci/scripts/setup_sklearn.sh
+++ b/.ci/scripts/setup_sklearn.sh
@@ -26,10 +26,12 @@ sklearn_version=${1:-main}
 if [ "$sklearn_version" == "main" ]; then
     # remove sklearn version from test requirements file
     sed -i.bak -e "s/scikit-learn==1.3.2/scikit-learn/" requirements-test.txt
+    sed -i.bak -e "s/scikit-learn==1.2.2/scikit-learn/" requirements-test.txt
     # install sklearn build dependencies
     pip install threadpoolctl joblib scipy
     # install sklearn from main branch of git repo
     pip install git+https://github.com/scikit-learn/scikit-learn.git@main
 else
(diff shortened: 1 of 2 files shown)


Alexsandruss (changes requested):
Reapplying of https://github.com/intel/scikit-learn-intelex/pull/1561 is required

.ci/scripts/setup_sklearn.sh
```diff
@@ -26,10 +26,12 @@ sklearn_version=${1:-main}
...
     # remove sklearn version from test requirements file
     sed -i.bak -e "s/scikit-learn==1.3.2/scikit-learn/" requirements-test.txt
+    sed -i.bak -e "s/scikit-learn==1.2.2/scikit-learn/" requirements-test.txt
     # install sklearn build dependencies
     pip install threadpoolctl joblib scipy
     # install sklearn from main branch of git repo
     pip install git+https://github.com/scikit-learn/scikit-learn.git@main
 else
     sed -i.bak -e "s/scikit-learn==1.3.2/scikit-learn==${sklearn_version}.*/" requirements-test.txt
+    sed -i.bak -e "s/scikit-learn==1.2.2/scikit-learn==${sklearn_version}.*/" requirements-test.txt
```
napetrov, line 36: should we just replace this for any version? not some particular? 
I mean we will not be updating this in case of next version update in requirements 

We ether in case of main would like to go with just scikit-learn without version specification.

and in case of second we will be just forcing version defined 

i.e. with 1.3.3 version we will get this broken

napetrov (approved):
Looks good to me 

there is strange timeout for daal4py testing, but i don't think they are related to this PR

ethanglaser:
> Looks good to me
> 
> there is strange timeout for daal4py testing, but i don't think they are related to this PR

Thanks, yes timeout is related to model builders - I think resolution is in progress

ahuber21 (approved):

Alexsandruss (approved):

PR #1566, opened 2023-11-04
Fix LightGBM test_examples runtime & cleanup resulting errors

ahuber21:
Cleanup of run_examples.py and test_examples.py


ahuber21:
Updated CI: http://intel-ci.intel.com/ee7f073e-81da-f106-b4a0-a4bf010d0e2e

ahuber21:
http://intel-ci.intel.com/ee7f1dde-e480-f15a-9024-a4bf010d0e2e

tests/test_examples.py
```diff
@@ -63,7 +61,9 @@ def check_libraries(rule):
 
 
 # function reading file and returning numpy array
-def np_read_csv(f, c=None, s=0, n=np.iinfo(np.int64).max, t=np.float64):
+def np_read_csv(f, c=None, s=0, n=np.iinfo(np.int64).max, t=np.float64, **kwargs):
+    n = kwargs.get("nrows") or n
+    print(f"np_read_csv: {n=}")
```
ethanglaser, line 66:
```suggestion
```
if its for debugging

ethanglaser (approved):
LGTM and removes problematic CI examples...any other opinions on examples refactoring?

ahuber21:
There is no loss in removing the examples for now, as they never ran previously due to missing dependencies. So the coverage is unaffected. Of course, the original intent was always to run examples as part of regular tests and we need to restore that. Separate issue though.

PR #1571, opened 2023-11-10
Onedal-wrapper for LogisticRegression

avolkov-intel:
# Description
Implement onedal.linear_model.LogisticRegression class

- Use oneDAL LogisticRegression interfaces for GPU computatiuon
- Implement tests for algorithm


onedal/__init__.py
```diff
@@ -64,3 +64,6 @@
 
     if _is_dpc_backend:
         __all__ += ["spmd.cluster"]
+
+    
+__all__ += ["logistic_regression"]
```
samir-nasibli, line 69: please add extra line. applicable for other files

onedal/datatypes/numpy_helpers.hpp
```diff
@@ -81,18 +81,24 @@
...
             break;                              \
         }                                       \
+        case NPY_LONGLTR:                       \
+        case NPY_LONGLONGLTR:                   \
         case NPY_INT64: {                       \
             _FUNCT(std::int64_t);               \
             break;                              \
         }                                       \
+        case NPY_ULONGLTR:                      \
+        case NPY_ULONGLONGLTR:                  \
```
samir-nasibli, line 101: Why are you updating this helper in current PR? this should be separate branch work

onedal/linear_model/logistic_regression.py
```diff
@@ -0,0 +1,215 @@
...
+            "max_iter": self.max_iter,
+            "C" : self.C,
+            "optimizer": self.solver,
+            "result_option": (intercept + "coefficients|iterations_count"),
+        }
+
+    def _fit(self, X, y, module, queue):
+        policy = self._get_policy(queue, X, y)
+
+        X_loc, y_loc = X, y
```
samir-nasibli, line 63: no need for having copies or synonyms at all.

onedal/linear_model/logistic_regression.py
```diff
@@ -0,0 +1,215 @@
...
+        if not isinstance(X, np.ndarray):
```
samir-nasibli, line 64: numpy.asarray already checks for numpy ndarray instance, no need for that. Please remove

onedal/linear_model/logistic_regression.py
```diff
@@ -0,0 +1,215 @@
...
+            X_loc = np.asarray(X)
+
+        dtype = get_dtype(X_loc)
+        if dtype not in [np.float32, np.float64]:
+            dtype = np.float64
+            X_loc = X_loc.astype(dtype, copy=self.copy_X)
+        y_loc = np.asarray(y_loc)
+
+        # Finiteness is checked in the sklearnex wrapper
```
samir-nasibli, line 73: why you have this comment here?

onedal/linear_model/logistic_regression.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def _predict(self, X, module, queue):
+        _check_is_fitted(self)
+
+        policy = self._get_policy(queue, X)
+
+        if isinstance(X, np.ndarray):
+            X_loc = np.asarray(X)
+        else:
+            X_loc = X
```
samir-nasibli, line 164: see related comment above.

onedal/linear_model/logistic_regression.py
```diff
@@ -0,0 +1,215 @@
...
+        params = self._get_onedal_params(get_dtype(X_loc))
+
+        X_table = to_table(X_loc)
+        result = module.infer(policy, params, model, X_table)
+        y = from_table(result.responses)
+
+        if not isinstance(self.coef_, np.ndarray):
+            coefficients = np.asarray(self.coef_)
+        else:
+            coefficients = self.coef_
```
samir-nasibli, line 188: From: https://numpy.org/doc/stable/reference/generated/numpy.asarray.html
> Array interpretation of a. No copy is performed if the input is already an ndarray with matching dtype and order. If a is a subclass of ndarray, a base class ndarray is returned.

Please remove this check. They just mess up the code.

onedal/linear_model/tmp.py
```diff
@@ -0,0 +1,89 @@
...
+import pandas as pd
+from sklearn.datasets import load_breast_cancer
+from sklearn.model_selection import train_test_split
+
+from sklearn.linear_model import LogisticRegression as sklearn_logreg
+from onedal.linear_model import LogisticRegression as onedal_logreg
+
+from sklearnex import patch_sklearn
+
+#from sklearnex import config_context
```
samir-nasibli, line 29: please move this example into proper directory

onedal/datatypes/numpy_helpers.hpp
avolkov-intel: We need this fix to convert numpy array of np.int32 to table in this algorithm

onedal/linear_model/logistic_regression.py
avolkov-intel: Because we have parameter false for force_all_finite

ethanglaser:
Please address codefactor hit

ethanglaser (changes requested):
Please address codefactor hit

onedal/__init__.py
```diff
@@ -64,3 +64,6 @@
 
     if _is_dpc_backend:
         __all__ += ["spmd.cluster"]
+
+if daal_check_version((2024, "P", 1)):
+    __all__ += ["logistic_regression"]
```
ethanglaser, line 69: logistic_regression is within the linear_model module - not sure if this is necessary

onedal/datatypes/numpy_helpers.hpp
icfaust: Could you link the lines of code which utilizes the int table conversion?

onedal/linear_model/logistic_regression.py
```diff
@@ -0,0 +1,239 @@
...
+        X = _convert_to_supported(policy, X)
+        params = self._get_onedal_params(get_dtype(X))
+
+        X_table = to_table(X)
+        result = module.infer(policy, params, model, X_table)
+        return result
+
+    def _predict(self, X, module, queue):
+        result = self._infer(X, module, queue)
+        y = from_table(result.responses)
```
icfaust, line 183: It would seem to me that two functions on the cpp side would be better served than to have two attributes of a result object which are never used together (_predict doesn't use result.probabilities and _predict_proba doesnt use result.responses)

icfaust (changes requested):

onedal/linear_model/logistic_regression.py
```diff
@@ -0,0 +1,239 @@
...
+    def _infer(self, X, module, queue):
+        _check_is_fitted(self)
+
+        policy = self._get_policy(queue, X)
+
+        X = np.asarray(X)
+
+        # Finiteness is checked in the sklearnex wrapper
+        X = _check_array(
+            X, dtype=[np.float64, np.float32], force_all_finite=False, ensure_2d=False
```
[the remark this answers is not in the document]
avolkov-intel: No, we will perform the finiteness check but on a higher level in sklearnex wrapper but we want to avoid doing this check twice.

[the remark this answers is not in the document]
icfaust: However, that would imply that no one would use these classes directly.  I'm not sure what our design guideline is for this. I'll leave it up to @samir-nasibli @Alexsandruss

ethanglaser:
> Please address codefactor hit

We can ignore this

ethanglaser:
internal CI will be more green if you rebase and rerun

onedal/linear_model/logistic_regression.py
avolkov-intel: I noticed that for now we cannot specify whether we can

icfaust (commented):
Deferring to others on this now

icfaust (approved):

onedal/linear_model/logistic_regression.cpp
```diff
@@ -0,0 +1,256 @@
...
+                onedal_options = onedal_options | result_options::coefficients;
+            }
+            else if (match.str() == "iterations_count") {
+                onedal_options = onedal_options | result_options::iterations_count;
+            }
+            else
+                ONEDAL_PARAM_DISPATCH_THROW_INVALID_VALUE(result_option);
+        }
+    }
+    catch (std::regex_error& e) {
```
ethanglaser, line 102: unreferenced local variable warning 'e' in daal4py build log

ethanglaser (approved):
Looks good, no objections from me but please fix windows build warning mentioned in previous comment

onedal/datatypes/numpy_helpers.hpp
samir-nasibli: > this fix to convert numpy array of np.int32 to table in this algorithm

the is no problem for the fix. Why it is in the current PR? This kind of changes should be separated

avolkov-intel: There was a problem during the build if these lines were not added

avolkov-intel: You can see that all double and float types already had case NPY_DOUBLELTR:

PR #1582, opened 2023-11-21
chore: Refactor test_examples

ahuber21:
This cleans up the test cases that are generated from `daal4py/examples`. The main testing script was renamed `tests/test_examples.py → tests/test_daal4py_examples.py`, as it only targets `daal4py` examples.

In a separate effort, `test/run_examples.py` (mainly there for `sklearnex` validation) should be updated to a similar implementation.

This PR comes with 2 new features

#### Test auto-discovery

All `.py` files in the `daal4py` examples dir are added. Most have existing `Config` objects. But some don't. A minimalistic default test will be run for those cases. The overall test count increased from 183 to 243

#### Test suspension

Previously, one test was commented out with a TODO-style comment to bring it back once something is fixed---which never happened. Now, the `Config` object support test suspension. It accepts two keyword arguments

* `suspended_on`: This is a date in the format `(YEAR, MONTH, DAY)`, for example `(2023, 11, 1)`
* `suspended_for_n_days`: The number of days to suspend the test, default `30`

This prevents us from forgetting re-enabling tests :)


samir-nasibli (commented):
It is better to move out `daal4py/tests/readcsv.py` into `tests/utils`.
Generally it could be usefull for all tests, examples of both sklearnex and daal4py.

ahuber21:
@samir-nasibli  `tests/utils` is not a package, so we can't import it from everywhere. I've moved `readcsv` to `daal4py.sklearn.utils`

examples/tests/test_daal4py_examples.py
```diff
@@ -0,0 +1,400 @@
+# ==============================================================================
```
napetrov, line 1: I would suggest not to use examples/tests at all, to keep examples folder clean and tidy for users. 

If we consider this as tests, then lets keep in tests folder

daal4py/sklearn/_utils.py
```diff
@@ -17,13 +17,16 @@
...
 import warnings
+from typing import Any, Tuple
 
 import numpy as np
 from numpy.lib.recfunctions import require_fields
 from sklearn import __version__ as sklearn_version
 
 from daal4py import _get__daal_link_version__ as dv
 
+DaalVersionTuple = Tuple[int, str, int]
```
napetrov, line 28: i guess this have to be 4 elements 
Major - 2024
Minor - 0 
Patch - 1
Status - P 

Might be play with outputs/prints in case of this.

ahuber21: For some reason the convention in `daal4py` is `(Major, Status, Minor+Patch)` - so only three elements

examples/tests/test_daal4py_examples.py
ahuber21: I was faster than your comment :) It's back in the "old" `tests` folder.

samir-nasibli:
> @samir-nasibli `tests/utils` is not a package, so we can't import it from everywhere. I've moved `readcsv` to `daal4py.sklearn.utils`

It wasn't the part of the daal4py package functionality (it was used only in our examples), so exactly that is why I was asking moving it out.


And in my opinion, any way utils functions, not of the part of testing, is better to move utils named folder/module.

daal4py/sklearn/_utils.py
napetrov: > For some reason the convention in `daal4py` is `(Major, Status, Minor+Patch)` - so only three elements

ok, just to ensure that <= comparison would actually work. Like how Minor+Patch is done. I guess it should be ok, but a point to doublecheck

daal4py/sklearn/utils/readcsv.py
```diff
@@ -0,0 +1,121 @@
+from pathlib import Path
```
napetrov, line 1: i guess this would be considered publicAPI for daal4py. 
I don't think that this is functionality that we would like  to expose?

ahuber21: I don't have a good solution for not putting it into a package _and_ making it reusable. This is a limitation in Python's import system for which I don't have a good workaround. Maybe @samir-nasibli has an idea.

Initially that was why I created the `tests` namespace inside `daal4py` - I didn't think it would hurt there.

daal4py/sklearn/_utils.py
ahuber21: I did some tests locally and they all worked as expected. And I didn't change any logic. I just updated the types and the comments to align with the code (which looks like it was changed some time without updating the comments).

daal4py/sklearn/utils/readcsv.py
napetrov: but you can put it just in to tests folder in the root of repo? 
or we have https://github.com/intel/scikit-learn-intelex/tree/master/scripts already 

@Alexsandruss - might be any suggestions?

ahuber21:
Interesting side effect from the more explicit structure of the updated code.
There are duplicates in the current master.
Because of how the code in master is written, the version **without** `20230101` are simply overloaded and never run.
My code avoids such shadowing scenarios which is why the old check ran (and failed) again. This was addressed by removing the legacy checks in [433debb](https://github.com/intel/scikit-learn-intelex/pull/1582/commits/433debbfceaf6b966d1a9ae65fccc7c6b8f31ee2).

```python
gen_examples = [
# ...
    (
        "decision_forest_regression_default_dense",
        "decision_forest_regression.csv",
        lambda r: r[1].prediction,
        (2023, "P", 1),
    ),
    (
        "decision_forest_regression_hist",
        "decision_forest_regression.csv",
        lambda r: r[1].prediction,
        (2023, "P", 1),
    ),
    (
        "decision_forest_regression_default_dense",
        "decision_forest_regression_20230101.csv",
        lambda r: r[1].prediction,
        (2023, "P", 101),
    ),
    (
        "decision_forest_regression_hist",
        "decision_forest_regression_20230101.csv",
        lambda r: r[1].prediction,
        (2023, "P", 101),
    ),
# ...
]
```

daal4py/sklearn/utils/readcsv.py
napetrov: Looks good overall

napetrov:
> Interesting side effect from the more explicit structure of the updated code. There are duplicated tests in the current master. Because of how the code in master is written, the versions **without** `20230101` are simply overloaded and never run. My code avoids such shadowing scenarios which is why the old check ran (and failed) again. This was addressed by removing the legacy checks in [433debb](https://github.com/intel/scikit-learn-intelex/pull/1582/commits/433debbfceaf6b966d1a9ae65fccc7c6b8f31ee2).
> 
> ```python
> gen_examples = [
> # ...
>     (
>         "decision_forest_regression_default_dense",
>         "decision_forest_regression.csv",
>         lambda r: r[1].prediction,
>         (2023, "P", 1),
>     ),
>     (
>         "decision_forest_regression_hist",
>         "decision_forest_regression.csv",
>         lambda r: r[1].prediction,
>         (2023, "P", 1),
>     ),
>     (
>         "decision_forest_regression_default_dense",
>         "decision_forest_regression_20230101.csv",
>         lambda r: r[1].prediction,
>         (2023, "P", 101),
>     ),
>     (
>         "decision_forest_regression_hist",
>         "decision_forest_regression_20230101.csv",
>         lambda r: r[1].prediction,
>         (2023, "P", 101),
>     ),
> # ...
> ]
> ```

should we just keep 1 version of csv files?

examples/daal4py/pca.py
```diff
@@ -19,23 +19,12 @@
...
-        return pandas.read_csv(f, usecols=c, delimiter=",", header=None, dtype=t)
-
-except ImportError:
-    # fall back to numpy loadtxt
-    def read_csv(f, c=None, t=np.float64):
-        return np.loadtxt(f, usecols=c, delimiter=",", ndmin=2)
-
-
-def main(readcsv=read_csv, method="svdDense"):
+def main(readcsv=pd_read_csv, method="svdDense"):
```
ethanglaser, line 27: remove method here? assuming the removals are when its unused in the example

examples/daal4py/pca_transform.py
```diff
@@ -16,29 +16,20 @@
...
-except ImportError:
-    # fall back to numpy loadtxt
-    def read_csv(f, c, t=np.float64):
-        return np.loadtxt(f, usecols=c, delimiter=",", ndmin=2)
+import daal4py as d4p
 
 
-def main(readcsv=read_csv, method="svdDense"):
-    dataFileName = "data/batch/pca_transform.csv"
+def main(readcsv=pd_read_csv, method="svdDense"):
```
ethanglaser, line 26: unused method here as well

examples/daal4py/pivoted_qr.py
```diff
@@ -19,23 +19,12 @@
...
-        return pandas.read_csv(f, usecols=c, delimiter=",", header=None, dtype=t)
-
-except ImportError:
-    # fall back to numpy loadtxt
-    def read_csv(f, c=None, t=np.float64):
-        return np.loadtxt(f, usecols=c, delimiter=",", ndmin=2)
-
-
-def main(readcsv=read_csv, method="svdDense"):
+def main(readcsv=pd_read_csv, method="svdDense"):
```
ethanglaser, line 27: same

examples/utils/readcsv.py
```diff
@@ -0,0 +1,121 @@
...
+        file_path: Path,
+        usecols: Optional[Sequence[int]] = None,
+        skip_header: int = 0,
+        dtype: DType = np.float32,
+        max_rows: Optional[int] = None,
+    ) -> Any:
+        ...
+
+
+class NpReadCsv:
```
ethanglaser, line 27: what is advantage of making these classes and not functions?

tests/test_daal4py_spmd_examples.py
```diff
@@ -0,0 +1,196 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import os
+
+# if d4p.__has_dist__:
```
ethanglaser, line 19: can be removed

daal4py/sklearn/_utils.py
```diff
@@ -68,21 +71,18 @@ def set_idp_sklearn_verbose():
...
-        for rule_item in rule:
-            if rule_item > target:
-                return False
-            if rule_item[0] == target[0]:
-                break
-    return True
+    # Second item is status - B
+    # Third is minor+patch - 0110,
+    if isinstance(required_version[0], (list, tuple)):
+        return any(map(daal_check_version, required_version))
```
ethanglaser, line 83: its more concise but I have no idea what's going on here ;)

tests/test_daal4py_examples.py
```diff
@@ -0,0 +1,423 @@
...
+            result.mean,
+            result.secondOrderRawMoment,
+            result.variance,
+            result.standardDeviation,
+            result.variation,
+        )
+    )
+
+
+examples = [
```
ethanglaser, line 247: I guess there isn't really a way to unify these

examples/daal4py/pca.py
ahuber21: No I am removing duplicated code. The CSV read helpers are now defined in a single place `examples/utils/readcsv.py`

examples/utils/readcsv.py
ahuber21: See the inline comment where I defined the functions. I'm defining a protocol class. It makes type checking easier.

daal4py/sklearn/_utils.py
ahuber21: I have provided a better description and a `test_utils.py` that shows how it's supposed to work. Hope that helps. Just need to verify the `test_utils.py` actually runs in CI.

daal4py/sklearn/utils/test/test_utils.py
```diff
@@ -0,0 +1,30 @@
...
+        ((2020, "C", 99), True),
+        ((2021, "C", 100), False),
+        ((2020, "D", 100), False),
+        ((2020, "C", 101), False),
+        ((2019, "B", 99), True),
+        ((2021, "D", 101), False),
+        (((2020, "C", 100), (2021, "D", 101)), True),
+        (((2019, "C", 100), (2018, "C", 100)), True),
+        (((2019, "C", 100), (2018, "C", 100)), True),
+        (((2021, "C", 100), (2022, "C", 100)), False),
```
ethanglaser, line 25: This is helpful and nice to have. In terms of scope of tuples I think realistic values that could be worth having checks for with the 3rd number are 0, 1, 2, 100, 101, 102, 200, 201, 202 (where minor and patch are the hundreds and ones digits). Not that all of these have to be tested but maybe worth having a 1 instead of a 99, and swapping one of these hundreds for something with minor version 0 (single digit). Also I am not sure what the letter is but I've only ever seen "P"

examples/daal4py/pca.py
ethanglaser: I mean removing the method="svdDense" part. Since it seems like many other files removed method when it is unused. Just a minor thing.

examples/mb/readcsv.py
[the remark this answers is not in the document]
ahuber21: That's a compromise to hide away ugly importing practices in one place. We cannot use relative imports and we cannot import from daal4py as we're not adding the read csv utils to the public API.

examples/daal4py/pca.py
ahuber21: Good call, looks like I didn't remove unused `svdDense` methods. Fixed.

ethanglaser:
Will need to update these: https://github.com/search?q=repo%3Aintel%2Fscikit-learn-intelex%20spmd*.py&type=code

ethanglaser:
With infra update: http://intel-ci.intel.com/ee93bc22-fdb1-f1c1-93c6-a4bf010d0e2e

napetrov:
> @napetrov reworking `run_examples.py` is out of scope for this PR. When I get to `run_examples.py` I'll consider this scenario.

o

ahuber21:
One example still timed out - I've reduced its complexity. Should be fine now.

ahuber21:
We can review based on this CI (the latest commit only contains infra changes) http://intel-ci.intel.com/ee983735-5f11-f192-9e8c-a4bf010d0e2e

ethanglaser (approved):
Let's merge it with infra update. If spmd example main function skips can be addressed then can add those first.

PR #1593, opened 2023-11-28
Create a more informative error in assert in _get_policy.py for the dpc backend

icfaust:
# Description
Users can run into this assert statement when the dpc backend isn't created. There is no information on what is wrong, which is that they need to have compiled scikit-learn-intelex with the proper compiler for the dpc backend.  I have run into this unfortunately and the bare assert was uninformative.  This also come across when dpctl isn't installed or the improper version is used (i.e. any case without the dpc backend).

- Add an informative RuntimeError with a string

### DIFF
--- a/onedal/common/_policy.py
+++ b/onedal/common/_policy.py
@@ -32,7 +32,9 @@ def _get_policy(queue, *data):
             return _DataParallelInteropPolicy(data_queue)
         return _DataParallelInteropPolicy(queue)
     if not (data_queue is None and queue is None):
-        raise RuntimeError("Operation using the requested SYCL queue requires the DPC backend")
+        raise RuntimeError(
+            "Operation using the requested SYCL queue requires the DPC backend"
+        )
     return _HostInteropPolicy()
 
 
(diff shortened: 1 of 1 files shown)


onedal/common/_policy.py
```diff
@@ -31,7 +31,10 @@ def _get_policy(queue, *data):
                 return _HostInteropPolicy()
             return _DataParallelInteropPolicy(data_queue)
         return _DataParallelInteropPolicy(queue)
-    assert data_queue is None and queue is None
+    if not (data_queue is None and queue is None):
+        raise RuntimeError(
+            "Operation using the requested SYCL queue requires the DPC backend"
+        )
```
samir-nasibli, line 37: I think it worth to add else here. This not abvous that this is related with dpc_backend
```suggestion
    else:
        if not (data_queue is None and queue is None):
            raise RuntimeError(
                "Operation using the requested SYCL queue requires the DPC backend"
            )
```

ethanglaser (approved):

PR #1596, opened 2023-11-29
[bug] fix scikit-learn upper bound dependency for conda public Releases CI

icfaust:
# Description
This will check the installed scikit-learn version, and will reinstall scikit-learn as necessary to the last-supported scikit-learn version for the currently available scikit-learn-intelex on the conda channel. Currently, it will test scikit-learn-intelex against a scikit-learn version that was not supported at the time of the scikit-learn-intelex release.

This will solve the Releases CI failures, which are errant faults (and may be covering true errors in recent merges).

I recommend we add an upper bound to the conda channels to avoid this problem in a more standardized way.

- add another scikit-learn install to the releases.ci
- include a python program "get_compatible_sklearn_version.py" to check conda for the proper supported sklearn (since an upper bound is not included in most scikit-learn-intelex conda releases).

### DIFF
--- a/.ci/scripts/get_compatible_sklearn_version.py
+++ b/.ci/scripts/get_compatible_sklearn_version.py
@@ -1,8 +1,8 @@
 import json
 import os
+import re
 import subprocess
 import sys
-import re
 
 # Find current Scikit-learn major versions in release order
 conda_call = subprocess.run(
--- a/.ci/scripts/get_compatible_sklearn_version.py
+++ b/.ci/scripts/get_compatible_sklearn_version.py
@@ -1,0 +1,55 @@
+import json
+import os
+import subprocess
+import sys
+import re
+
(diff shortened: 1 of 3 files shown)


.ci/scripts/get_compatible_sklearn_version.py
```diff
@@ -0,0 +1,55 @@
+import json
+import os
+import re
+import subprocess
+import sys
+
+# Find current Scikit-learn major versions in release order
+conda_call = subprocess.run(
+    ["conda", "search", "scikit-learn", "-c", sys.argv[1]], capture_output=True
+)
```
samir-nasibli, line 10: In this case, it’s worth answering Codefactor hits.

napetrov:
HAve a discussion offline on this. This is a bit more complex, so not sure if this PR should be considered in this form

PR #1607, opened 2023-12-08
[testing] Check for docstrings for patched vs unpatched classes

icfaust:
# Description
We are inconsistent on enforcing docstrings for our patched classes in sklearnex. This is a test which will verify that the class and its user-facing class attributes (those without '_' at the beginning or the end) have docstrings which match sklearn.

This is likely to fail pretty spectacularly on the first CI run.

- add test_docstring_patching_match
- Change list to dict so that pytest failure titles are more informative

Tasks
- [x] implement test
- [x] correct failures


icfaust:
Something is weird with public CI windows sklearnex checks. It has fails but is showing as green.

sklearnex/tests/test_patching.py
```diff
@@ -116,7 +116,32 @@ def _load_all_models(patched):
...
+        if not i.startswith("_") and not i.endswith("_")
+    }
+    unpatched_docstrings = {
+        i: getattr(unpatched, i).__doc__
+        for i in dir(unpatched)
+        if not i.startswith("_") and not i.endswith("_")
+    }
+
+    # check class docstring match
+    assert patched.__doc__ == unpatched.__doc__
```
KulikovNikita, line 143: The question is - should they match 1-to-1 or should our documentation be an **extension** over what is in the vanilla sklearn?

KulikovNikita: Because in my mind we can have additional parameters as well as some caveats that should be exposed to the user like the list of conditions for the actual patching

icfaust: I totally agree. I think I may change it to making sure its not 'None', which means that to pass CI it must be set to something, and hopefully then it will spur a conversation.

Alexsandruss (commented):
There are at least 15 warning for docstring mismatches. What is the stopper from fixing them?

sklearnex/neighbors/_lof.py
```diff
@@ -162,6 +201,4 @@ def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
         )
 
     fit.__doc__ = sklearn_LocalOutlierFactor.fit.__doc__
-    fit_predict.__doc__ = sklearn_LocalOutlierFactor.fit_predict.__doc__
-    predict.__doc__ = sklearn_LocalOutlierFactor.predict.__doc__
```
samir-nasibli, line 166: Why you are not using it from sklearn?

samir-nasibli (commented):
Looks good to me

sklearnex/tests/test_patching.py
```diff
@@ -20,6 +20,7 @@
 import re
 import subprocess
 import sys
+import warnings
```
ethanglaser, line 23: is it used?

sklearnex/neighbors/_lof.py
icfaust: Available_if blocks it unfortunately. It's really a bad construction on sklearn's side

sklearnex/tests/test_patching.py
icfaust: good eye, there was something before that used it, but available_if ruined it.

PR #1609, opened 2023-12-09
[ci,bug] add errorcode persistence to run_test.bat

icfaust:
# Description
See PR #1607 for windows azure pipelines CI failed pytests without CI step failure

- make an errorcode check to run_test.bat


icfaust:
This should pass while #1607 with the same changes but with CI errors at that step will properly fail now.

napetrov (approved):

PR #1611, opened 2023-12-11
feat: add nosklearnex option to run_examples

ahuber21:
Fixes the problem described in https://github.com/conda-forge/daal4py-feedstock/pull/57

### DIFF
--- a/tests/run_examples.py
+++ b/tests/run_examples.py
@@ -182,6 +182,10 @@ req_library["random_forest_classifier_spmd.py"] = ["dpctl", "mpi4py"]
 req_library["random_forest_regressor_dpnp.py"] = ["dpnp"]
 req_library["random_forest_regressor_spmd.py"] = ["dpctl", "dpnp", "mpi4py"]
 
+# add sklearnex as required library for all sklearnex examples
+for fname in os.listdir(jp(examples_rootdir, "sklearnex")):
+    req_library[fname].append("sklearnex")
+
 req_os = defaultdict(lambda: [])
 
 skiped_files = []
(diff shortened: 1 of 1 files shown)


tests/run_examples.py
```diff
@@ -302,4 +298,29 @@ def run_all(nodist=False, nostream=False):
...
+        action="store_true",
+        default=False,
+        help="Skip examples with data streaming",
+    )
+    parser.add_argument(
+        "--nosklearnex",
+        action="store_true",
+        default=False,
+        help="Skip sklearnex examples",
+    )
```
samir-nasibli, line 319: When we are using this new argunment?

samir-nasibli (commented):
Now it is better. Thank you!

tests/run_examples.py
ahuber21: We should use it in conda recipe building to fix the problem discussed in the link from the description.

napetrov:
Looks great - the even better one would be to add switch to run just skleranex components in case of scikit-learn-intelex recipe. 
In this case we don't need or even can running MB/daal4py examples and need just scikit-leanex. 

Might be this should be --sklernex  and --daal4py switches .

ahuber21:
Adding `--sklearnex` and `--daal4py` switches to actively enable tests (with the default being disabled) would be incompatible with the current logic that has all tests enabled by default. But I added a `--nodaal4py` flag for symmetry.

tests/run_examples.py
```diff
@@ -203,27 +196,31 @@ def get_exe_cmd(ex, nodist, nostream):
             return None
 
     if os.path.dirname(ex).endswith("daal4py"):
```
napetrov, line 198: i think we also have to have if for "mb" directory 
```suggestion
    if os.path.dirname(ex).endswith("daal4py") or os.path.dirname(ex).endswith("mb"):
```

Or might be add as separate if line

ahuber21: Good call!

napetrov (approved):

PR #1616, opened 2023-12-19
IncrementalCovariance interface

olegkkruglov:
- Incremental covariance algorithm is added to onedal package (API is the same as in oneDAL)
- Incremental covariance algorithm is added to sklearnex package (API is like in IncrementalPCA algorithm in sklearn)

### DIFF
--- a/examples/sklearnex/incremental_covariance.py
+++ b/examples/sklearnex/incremental_covariance.py
@@ -1,5 +1,5 @@
 # ==============================================================================
-# Copyright 2023 Intel Corporation
+# Copyright 2024 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/examples/sklearnex/incremental_covariance.py
+++ b/examples/sklearnex/incremental_covariance.py
@@ -18,12 +18,17 @@ import numpy as np
 
 from sklearnex.covariance import IncrementalEmpiricalCovariance
 
-X = np.array([[0, 1], [0, 1]])
-X_split = np.array_split(X, 2)
-
 inccov = IncrementalEmpiricalCovariance()
-for i in range(2):
-    result = inccov.partial_fit(X_split[i])
+
+# We do partial_fit for each batch and then print final result
+X_1 = np.array([[0, 1], [0, 1]])
+result = inccov.partial_fit(X_1)
+
+X_2 = np.array([[1, 2]])
+result = inccov.partial_fit(X_2)
+
+X_3 = np.array([[1, 1], [1, 2], [2, 3]])
+result = inccov.partial_fit(X_3)
 
 print(f"Covariance matrix:\n{result.covariance_}")
 print(f"Means:\n{result.location_}")
--- a/examples/sklearnex/incremental_covariance.py
+++ b/examples/sklearnex/incremental_covariance.py
@@ -18,7 +18,7 @@ import numpy as np
 
 from sklearnex.covariance import IncrementalEmpiricalCovariance
 
-inccov = IncrementalEmpiricalCovariance()
+inccov = IncrementalEmpiricalCovariance(batch_size=3)
 
 # We do partial_fit for each batch and then print final result
 X_1 = np.array([[0, 1], [0, 1]])
@@ -29,6 +29,14 @@ result = inccov.partial_fit(X_2)
 
 X_3 = np.array([[1, 1], [1, 2], [2, 3]])
 result = inccov.partial_fit(X_3)
+
+print(f"Covariance matrix:\n{result.covariance_}")
+print(f"Means:\n{result.location_}")
+
+# We put the whole data to fit method, it is splitted automatically and then
+# partial_fit is called for each batch
+X = np.array([[0, 1], [0, 1], [1, 2], [1, 1], [1, 2], [2, 3]])
+result = inccov.fit(X)
 
 print(f"Covariance matrix:\n{result.covariance_}")
 print(f"Means:\n{result.location_}")
--- a/examples/sklearnex/incremental_covariance.py
+++ b/examples/sklearnex/incremental_covariance.py
@@ -1,0 +1,29 @@
+# ==============================================================================
+# Copyright 2023 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import numpy as np
+
+from sklearnex.covariance import IncrementalEmpiricalCovariance
+
+X = np.array([[0, 1], [0, 1]])
+X_split = np.array_split(X, 2)
+
+inccov = IncrementalEmpiricalCovariance()
+for i in range(2):
+    result = inccov.partial_fit(X_split[i])
+
+print(f"Covariance matrix:\n{result.covariance_}")
+print(f"Means:\n{result.location_}")
--- a/onedal/covariance/covariance.cpp
+++ b/onedal/covariance/covariance.cpp
@@ -57,6 +57,33 @@ struct params2desc {
 };
 
 template <typename Policy, typename Task>
+void init_finalize_compute_ops(pybind11::module_& m) {
+    using namespace dal::covariance;
+    using input_t = partial_compute_result<Task>;
+    m.def("finalize_compute", [](const Policy& policy, const pybind11::dict& params, const input_t& data) {
+        finalize_compute_ops ops(policy, data, params2desc{});
+        return fptype2t{ method2t{ Task{}, ops } }(params);
+    });
+}
+
+template <typename Policy, typename Task>
+void init_partial_compute_ops(pybind11::module_& m) {
+    using prev_result_t = dal::covariance::partial_compute_result<Task>;
+    m.def("partial_compute", [](
+        const Policy& policy,
+        const pybind11::dict& params,
+        const prev_result_t& prev,
+        const table& data) {
+            using namespace dal::covariance;
+            using input_t = partial_compute_input<Task>;
+            partial_compute_ops ops(policy, input_t{prev, data}, params2desc{});
+            return fptype2t{ method2t{ Task{}, ops } }(params);
+        }
+    );
+}
+
+
+template <typename Policy, typename Task>
 void init_compute_ops(py::module_& m) {
 #if defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240000
     using compute_hyperparams_t = dal::covariance::detail::compute_parameters<Task>;
@@ -96,6 +123,17 @@ inline void init_compute_result(py::module_& m) {
         .def_property("means", &result_t::get_means, &result_t::set_means);
 }
 
+template <typename Task>
+inline void init_partial_compute_result(pybind11::module_& m) {
+    using namespace dal::covariance;
+    using result_t = partial_compute_result<Task>;
+    pybind11::class_<result_t>(m, "partial_compute_result")
+        .def(pybind11::init())
+        .def_property("partial_n_rows", &result_t::get_partial_n_rows, &result_t::set_partial_n_rows)
+        .def_property("partial_crossproduct", &result_t::get_partial_crossproduct, &result_t::set_partial_crossproduct)
+        .def_property("partial_sums", &result_t::get_partial_sum, &result_t::set_partial_sum);
+}
+
 #if defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240000
 
 template <typename Task>
@@ -116,7 +154,10 @@ void init_compute_hyperparameters(py::module_& m) {
 #endif // defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240000
 
 ONEDAL_PY_DECLARE_INSTANTIATOR(init_compute_result);
+ONEDAL_PY_DECLARE_INSTANTIATOR(init_partial_compute_result);
 ONEDAL_PY_DECLARE_INSTANTIATOR(init_compute_ops);
+ONEDAL_PY_DECLARE_INSTANTIATOR(init_finalize_compute_ops);
+ONEDAL_PY_DECLARE_INSTANTIATOR(init_partial_compute_ops);
 #if defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240000
     ONEDAL_PY_DECLARE_INSTANTIATOR(init_compute_hyperparameters);
 #endif // defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240000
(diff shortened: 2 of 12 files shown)


sklearnex/covariance/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===============================================================================
+
+from .incremental_covariance import IncrementalEmpiricalCovariance
+
+__all__ = ["IncrementalEmpiricalCovariance"]
```
ethanglaser, line 19: So we will have online covariance out of preview but standard covariance in preview?

ethanglaser:
I recommend adding an example (examples/sklearnex/incremental_covariance.py) with a simple demonstration of usage

examples/sklearnex/incremental_covariance.py
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+# ==============================================================================
+
+import numpy as np
+
+from sklearnex.covariance import IncrementalEmpiricalCovariance
+
+X = np.array([[0, 1], [0, 1]])
+X_split = np.array_split(X, 2)
+
```
ethanglaser, line 23: Since this will be user facing, I recommend setting up a more realistic use case. For example, you could create 2 variables for the data (perhaps "original_chunk" and "new_chunk") with maybe 3 rows in each (can still keep small 2 columns for simplicity). Run partial fit not in a loop, but on the first chunk of data, maybe a comment describing, and then run partial fit on the second chunk after. Would it make sense to add finalize as well?

olegkkruglov: regarding finalize - it is sklearnex interface which is designed to be like in sklearn where is no call finalize or something similar. finalize is calling when user tries to call any attribute with result data in case if it is necessary to compute these data. also it is calling directly in onedal-like interfaces. (onedal.covariance.IncrementalEmpiricalCovariance)

ethanglaser: Thanks for clarification. So we can leave any finalize call out of the example, but I would still suggest removing loop and creating 2 separate 3x2 chunks (an original and new) to better illustrate usage.

sklearnex/covariance/incremental_covariance.py
```diff
@@ -0,0 +1,128 @@
...
+    location_ : ndarray of shape (n_features,)
+        Estimated location, i.e. the estimated mean.
+
+    covariance_ : ndarray of shape (n_features, n_features)
+        Estimated covariance matrix
+
+    _need_to_finalize : bool
+        If True then finalize compute should be called to
+        obtain covariance_ or location_ from partial
+        compute data
```
samir-nasibli, line 49: Do we need this for api?

sklearnex/covariance/__init__.py
samir-nasibli: @olegkkruglov Please leave here some description why you are making this as resolved

samir-nasibli (commented):
Partially reviewed

onedal/covariance/tests/test_incremental_covariance.py
```diff
@@ -0,0 +1,39 @@
...
+    inccov = IncrementalEmpiricalCovariance()
+    for i in range(2):
+        inccov.partial_compute(X_split[i], queue=queue)
+    
+    result = inccov.finalize_compute()
+    expected_covariance = np.array([[0, 0], [0, 0]])
+    expected_means = np.array([0, 1])
+
+    assert_allclose(expected_covariance, result.covariance_)
+    assert_allclose(expected_means, result.location_)
```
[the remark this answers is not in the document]
olegkkruglov: there are empty lines at EOF for all files, you can see it by green CI Lint step. i don't know why they are not showed in the github interface.

sklearnex/covariance/incremental_covariance.py
olegkkruglov: i don't understand the question, clarify please. the usage of this attribute is described in the doc string

sklearnex/preview/covariance/covariance.py
```diff
@@ -28,7 +29,7 @@
 
 @register_hyperparameters({"fit": get_hyperparameters("covariance", "compute")})
 @control_n_jobs
-class EmpiricalCovariance(sklearn_EmpiricalCovariance):
+class EmpiricalCovariance(sklearn_EmpiricalCovariance, BaseEstimator):
```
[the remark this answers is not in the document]
olegkkruglov: this inheritance was here before, it is necessary to patch EmpiricalCovariance. it was wrongly put to onedal4py part. i rewrote inheritance structure while adding BaseEmpiricalCovariance class and decided that sklearn classes should be in sklearnex part.

sklearnex/covariance/__init__.py
olegkkruglov: i have answered in private channel.

sklearnex/covariance/incremental_covariance.py
samir-nasibli: This attribute is private, non-public.

Due to : [PEP8 documenting](https://peps.python.org/pep-0008/#documentation-strings)

> Write docstrings for all public modules, functions, classes, and methods. Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does. This comment should appear after the "def" line.

Also applicable for attributes.
So it is better to leave just comment, not a docstring description for a such non-public attribute.

sklearnex/covariance/__init__.py
samir-nasibli: It's always better to make it as clear as possible why the discussion is over.
It could be very useful for other reviewers

sklearnex/preview/covariance/covariance.py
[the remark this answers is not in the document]
samir-nasibli: scikit-learn's `EmpiricalCovariance` already has this know BaseEstimator ancestor. https://github.com/scikit-learn/scikit-learn/blob/3f89022fa04d293152f1d32fbc2a5bdaaf2df364/sklearn/covariance/_empirical_covariance.py#L109

So why are you doing it again here?

onedal/covariance/tests/test_incremental_covariance.py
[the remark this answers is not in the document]
samir-nasibli: outdated.

sklearnex/preview/covariance/covariance.py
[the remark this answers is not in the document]
olegkkruglov: you are right, it is really redundant, i'll remove it. thank you for pointing out

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -0,0 +1,60 @@
...
+    get_dataframes_and_queues,
+)
+
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+def test_sklearnex_import_incremental_covariance(dataframe, queue):
+    from sklearnex.covariance import IncrementalEmpiricalCovariance
+
+    X = np.array([[0, 1], [0, 1]])
+    X = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
```
samir-nasibli, line 32:
```suggestion
    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
```

sklearnex/covariance/incremental_covariance.py
```diff
@@ -0,0 +1,131 @@
...
+        Returns
+        -------
+        self : object
+            Returns the instance itself.
+        """
+        X = check_array(X, dtype=[np.float64, np.float32])
+        self._onedal_partial_compute(X)
+        return self
+
+    @support_usm_ndarray()
```
ethanglaser, line 105: what does support_usm_ndarray do to these functions?

ethanglaser (commented):
CI looks ready, just have a few questions to be answered

sklearnex/covariance/incremental_covariance.py
```diff
@@ -0,0 +1,131 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===============================================================================
+
+import numpy as np
+from sklearn.utils import check_array, gen_batches
+
+from daal4py.sklearn._device_offload import support_usm_ndarray
```
samir-nasibli, line 20: Use onedal4py impl.

onedal/covariance/incremental_covariance.py
```diff
@@ -0,0 +1,57 @@
...
+from ..datatypes import _convert_to_supported, from_table, to_table
+from .covariance import BaseEmpiricalCovariance
+
+
+class IncrementalEmpiricalCovariance(BaseEmpiricalCovariance):
+    def __init__(self, method="dense", bias=False):
+        super().__init__(method, bias)
+        self._partial_result = _backend.covariance.partial_compute_result()
+
+    def partial_compute(self, X, queue=None):
```
samir-nasibli, line 30: please, use sklearn like naming. `partial_fit`

onedal/covariance/covariance.py
```diff
@@ -26,7 +25,27 @@
 from ..datatypes import _convert_to_supported, from_table, to_table
 
 
-class EmpiricalCovariance(BaseEstimator):
+class BaseEmpiricalCovariance(metaclass=ABCMeta):
```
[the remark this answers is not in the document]
olegkkruglov: BaseEmpiricalCovariance is abstract class designed to decrease code duplication between EmpiricalCovariance and IncrementalEmpiricalCovariance classes

sklearnex/covariance/incremental_covariance.py
olegkkruglov: it takes queue from data dpctl/dpnp container and put it as queue to the method args

examples/sklearnex/incremental_covariance.py
```diff
@@ -0,0 +1,34 @@
+# ==============================================================================
+# Copyright 2023 Intel Corporation
```
Alexandr-Solovev, line 2: please update year to 2024 in all new files

onedal/covariance/covariance.cpp
```diff
@@ -56,6 +56,33 @@ struct params2desc {
...
+    using namespace dal::covariance;
+    using input_t = partial_compute_result<Task>;
+    m.def("finalize_compute", [](const Policy& policy, const pybind11::dict& params, const input_t& data) {
+        finalize_compute_ops ops(policy, data, params2desc{});
+        return fptype2t{ method2t{ Task{}, ops } }(params);
+    });
+}
+
+template <typename Policy, typename Task>
+void init_partial_compute_ops(pybind11::module_& m) {
```
Alexandr-Solovev, line 70: it could be an option to support other constructors from oneDAL, if its necessary

onedal/covariance/tests/test_incremental_covariance.py
```diff
@@ -0,0 +1,65 @@
...
+
+from onedal.tests.utils._device_selection import get_queues
+
+
+@pytest.mark.parametrize("queue", get_queues())
+def test_onedal_import_covariance(queue):
+    from onedal.covariance import IncrementalEmpiricalCovariance
+
+    X = np.array([[0, 1], [0, 1]])
+    X_split = np.array_split(X, 2)
```
Alexandr-Solovev, line 29: is it possible to add tests with random split count? like 1, 2, 8, 10, and extend input data/generate it with bigger dimensions

sklearnex/covariance/incremental_covariance.py
```diff
@@ -0,0 +1,132 @@
...
+    Incremental estimator for covariance.
+    Allows to compute empirical covariance estimated by maximum
+    likelihood method if data are splitted into batches.
+
+    Parameters
+    ----------
+    batch_size : int, default=None
+        The number of samples to use for each batch. Only used when calling
+        ``fit``. If ``batch_size`` is ``None``, then ``batch_size``
+        is inferred from the data and set to ``5 * n_features``, to provide a
```
Alexandr-Solovev, line 39: if I correctly understood, we split input data into 5 batches by default, but whats the reason of it? May be its better to avoid splitting, due to have a clear understanding that we can provide blocks with significantly different sizes

Alexandr-Solovev (commented):
LGTM! Great work, just one suggestion related to this pr:
1)If its possible, please extend tests with synthetic data, you can check oneDAL tests for covariance as example

sklearnex/covariance/incremental_covariance.py
```diff
@@ -0,0 +1,132 @@
...
+        return self._onedal_estimator.covariance_
+
+    @property
+    def location_(self):
+        if self._need_to_finalize:
+            self._onedal_finalize_fit()
+        return self._onedal_estimator.location_
+
+    @support_usm_ndarray()
+    def partial_fit(self, X, queue=None):
```
samir-nasibli, line 88: This is not a scope of the current PR, it is required some follow up work here.
queue param here is only used for `support_usm_ndarray`.
It is better update to `support_usm_ndarray` and update passing queue into onedal4py backend.

sklearnex/covariance/incremental_covariance.py
Alexandr-Solovev: I got your point, so, looks like we have to add it, but my suggestion is to set it equals to 1 by default in intelex. 
Correct me, if I am wrong, but it seems like:
1)we get batch from non-onedal array/table
2)convert it to onedal type
3)call partial compute
4)convert computed batch in non-onedal array/table
5)take the second batch
and in my understanding it can significantly influence on the performance, and if you can its better to measure both options and chose the best one

onedal/covariance/covariance.cpp
```diff
@@ -56,6 +56,33 @@ struct params2desc {
     }
 };
 
+template <typename Policy, typename Task>
+void init_finalize_compute_ops(pybind11::module_& m) {
```
Alexandr-Solovev, line 60: its better to order firstly partial_compute after finalize_compute

onedal/covariance/covariance.cpp
olegkkruglov: Other constructors are not necessary to support current API. It can be added later if it will become important.

sklearnex/covariance/incremental_covariance.py
olegkkruglov: The usage is the following: if the user has already split data then partial_fit method should be used. if user has the whole data then it is possible to use fit method which automatically splits data and then calls partial_fit for each batch. I updated the example to highlight it.

Alexandr-Solovev (approved):
LGTM! Good work with oneDAL online covariance implementation

onedal/covariance/covariance.cpp
```diff
@@ -56,6 +56,7 @@ struct params2desc {
     }
 };
 
+
```
samir-nasibli, line 59:
```suggestion
```

onedal/covariance/incremental_covariance.py
```diff
@@ -0,0 +1,61 @@
...
+class IncrementalEmpiricalCovariance(BaseEmpiricalCovariance):
+    """
+    Incremental Covariance oneDAL implementation
+    """
+
+    def __init__(self, method="dense", bias=False):
+        super().__init__(method, bias)
+        self._partial_result = _backend.covariance.partial_compute_result()
+
+    def partial_fit(self, X, queue=None):
```
samir-nasibli, line 34: not so critical but make sense adding docstrings for the public methods.

onedal/covariance/incremental_covariance.py
```diff
@@ -0,0 +1,61 @@
...
+        if not hasattr(self, "_policy"):
+            self._policy = self._get_policy(queue, X)
```
samir-nasibli, line 36: Make sense on refactoring update _get_policy method.
Could be done as a follow up work

onedal/covariance/tests/test_incremental_covariance.py
```diff
@@ -0,0 +1,122 @@
...
+        inccov.partial_fit(X_split[i], queue=queue)
+    result = inccov.finalize_fit()
+
+    expected_covariance = np.array([[0, 0], [0, 0]])
+    expected_means = np.array([0, 1])
+
+    assert_allclose(expected_covariance, result.covariance_)
+    assert_allclose(expected_means, result.location_)
+
+    X = np.array([[1, 2], [3, 6]])
```
samir-nasibli, line 44: Maybe it make sense parameterization for the different input data

samir-nasibli (commented):
Thank you @olegkkruglov !
Looks very good to me. Just minor comments that could be addressed here or in follow-up PR

samir-nasibli (approved):

ethanglaser:
CI with oneDAL fix branch: http://intel-ci.intel.com/eeba1089-5c89-f171-a0af-a4bf010d0e2e

onedal/covariance/covariance.cpp
```diff
@@ -86,6 +87,33 @@ void init_compute_ops(py::module_& m) {
...
+void init_finalize_compute_ops(pybind11::module_& m) {
+    using namespace dal::covariance;
+    using input_t = partial_compute_result<Task>;
+    m.def("finalize_compute", [](const Policy& policy, const pybind11::dict& params, const input_t& data) {
+        finalize_compute_ops ops(policy, data, params2desc{});
+        return fptype2t{ method2t{ Task{}, ops } }(params);
+    });
+}
+
+
```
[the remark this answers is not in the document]
olegkkruglov: it is added to more clearly divide groups of functions with the similar meaning

onedal/covariance/tests/test_incremental_covariance.py
olegkkruglov: it would not significantly decrease code size

PR #1619, opened 2023-12-21
feat: SHAP support for symmetric CatBoost models

ahuber21:
This PR provides support for SHAP value calculation from CatBoost models with symmetric trees.

The values **are different** from the CatBoost internal calculation due to inconsistencies in CatBoost's SHAP algorithm. This is under investigation: https://github.com/catboost/catboost/issues/2556

Therefore, I disabled the corresponding check in `test_model_builders.py`. The test isn't suspended forever, it will need to be snoozed once per month.

### DIFF
--- a/daal4py/mb/model_builders.py
+++ b/daal4py/mb/model_builders.py
@@ -283,8 +283,13 @@ class GBTDAALModel(GBTDAALBaseModel):
 
 
 def convert_model(model):
-    gbm = GBTDAALModel()
-    gbm._convert_model(model)
+    try:
+        gbm = GBTDAALModel()
+        gbm._convert_model(model)
+    except TypeError as err:
+        if "Only GBTDAALRegressor can be created" not in str(err):
+            raise
+        gbm = d4p.sklearn.ensemble.GBTDAALRegressor.convert_model(model)
 
     gbm._is_regression = isinstance(gbm.daal_model_, d4p.gbt_regression_model)
 
--- a/daal4py/mb/model_builders.py
+++ b/daal4py/mb/model_builders.py
@@ -287,9 +287,12 @@ def convert_model(model):
         gbm = GBTDAALModel()
         gbm._convert_model(model)
     except TypeError as err:
-        if "Only GBTDAALRegressor can be created" not in str(err):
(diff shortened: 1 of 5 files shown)


daal4py/mb/model_builders.py
```diff
@@ -283,8 +283,13 @@ def predict_proba(self, X):
 
 
 def convert_model(model):
-    gbm = GBTDAALModel()
-    gbm._convert_model(model)
+    try:
```
razdoburdin, line 286: Could you please describe, why do you need this try-except block?

ahuber21: Our current version of dispatching doesn't account for all scenarios. We trying to support scikit-style model objects, but also native types from XGBoost, LightGBM, CatBoost. In some cases we are creating a `GBTDAALBase` object when we actually expect a `GBTDAALRegressor` or `GBTDAALClassifier`.
The solution in this PR is to "ask for forgiveness, not permission", but I have created a follow-up task to consolidate the dispatching.
For now, this is an improvement, as it increases the compatibility of the `convert_model()` function to more types of input model types.

razdoburdin (approved):
Overall looks good, but we should refactor the `convert_model` in future.

daal4py/mb/model_builders.py
```diff
@@ -283,8 +283,16 @@ def predict_proba(self, X):
...
+        gbm = GBTDAALModel()
+        gbm._convert_model(model)
+    except TypeError as err:
+        if "Only GBTDAALRegressor can be created" in str(err):
+            gbm = d4p.sklearn.ensemble.GBTDAALRegressor.convert_model(model)
+        elif "Only GBTDAALClassifier can be created" in str(err):
+            gbm = d4p.sklearn.ensemble.GBTDAALClassifier.convert_model(model)
```
samir-nasibli, line 293: Is it possible to check conditions for this kind of conversion avoiding exception?

ahuber21: As discussed with @razdoburdin, this will be done in a rework of the dispatching logic. It's not the only thing we're doing suboptimally.

PR #1622, opened 2023-12-21
Fix validation of n_jobs parameter

Alexsandruss:


icfaust:
The daal4py kmeans code is ugly. @Alexsandruss  Is the adding 'self' for making it n_jobs compliant? Its ugly enough that I wonder when a small refactor would be warranted when we modify it.

icfaust:
Looks like LocalOutlierFactor is missing n_jobs support (shown in #1632 )

icfaust:
There is an issue within the NuSVC and SVC algorithms, there is a hardcoded n_jobs value (which is probably later overwritten) here https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/svm/_common.py#L110 . This came up in #1632 , there is a roundabout way in which it is calling onedal code in nusvc and svc which should be covered by other checks of @run_with_n_jobs . I have removed a "@run_with_n_jobs" from _onedal_predict_proba in that PR

icfaust:
daal4py TSNE has similar issues as daal4py means. That should probably be fixed in this PR as well.

icfaust:
LocalOutlierFactor n_jobs support has been fixed in #1640

Alexsandruss (commented):
@mergifyio rebase

Alexsandruss:
@icfaust

> The daal4py kmeans code is ugly. @Alexsandruss Is the adding 'self' for making it n_jobs compliant? Its ugly enough that I wonder when a small refactor would be warranted when we modify it.

Yes, `self` is needed for decorator which uses this argument.
daal4py.kmeans might be replaced with onedal+sklearnex.kmeans sooner or later, so refactoring is unnecessary

> daal4py TSNE has similar issues as daal4py kmeans. That should probably be fixed in this PR as well.

I don't see any issues with TSNE: tsne-specific oneDAL functions, NearestNeighbors and PCA run with correct number of threads.

icfaust:
@Alexsandruss 

https://github.com/intel/scikit-learn-intelex/blob/main/daal4py/sklearn/manifold/_t_sne.py#L129

Shown in new tests in  #1640

Alexsandruss:
@icfaust

> https://github.com/intel/scikit-learn-intelex/blob/main/daal4py/sklearn/manifold/_t_sne.py#L129
> 
> Shown in new tests in #1632. This is covered up because TSNE sklearn natively supports n_jobs.

Runtime code analysis of Python function is not suitable for this test: https://github.com/intel/scikit-learn-intelex/pull/1632#pullrequestreview-1813057433

icfaust:
@Alexsandruss 

The 'runtime code analysis' will only be unsuitable depending on the rules associated with control_n_jobs and run_with_n_jobs and therefore the statement is debatable. If we place run_with_n_jobs beside any call to the cython generated daal4py code in daal4py/sklearn (i.e. at the lowest level possible), it should work as any indirection or nesting is removed. It is the high levels (at 'fit' or 'predict') that we should worry.

As per the documentation of run_with_n_jobs (https://github.com/intel/scikit-learn-intelex/blob/main/daal4py/sklearn/_utils.py#L365), for every part of the code which calls the daal4py or onedal modules, it should be decorated. This can be done in direct way associated with the code which calls out to the daal4py module (low level like in #1632 ), or placed at high levels (like 'fit'). However, there can be circumstances where the daal4py code can be circuitously accessed when done at high levels.

I think the comment I made in the initial n_jobs PR still is necessary, namely that a fixed design rule should be made and documented so that other developers can evaluate the suitability and capability of n_jobs support in scikit-learn-intelex pull requests (new estimators, features, etc.).

I think the daal4py test in #1632 can be deleted as long as there is a documented design rule or test to replace it. However, I do think low level use of run_with_n_jobs is safer. I am open to discussing it further.

Addenum:

Where #1632 fails is the inconsistent implementation of methods or functions which call daal4py, and the code analysis test assumes that daal4py is called from an instance method. 

> @icfaust
> 
> > https://github.com/intel/scikit-learn-intelex/blob/main/daal4py/sklearn/manifold/_t_sne.py#L129
> > Shown in new tests in #1632. This is covered up because TSNE sklearn natively supports n_jobs.
> 
> Runtime code analysis of Python function is not suitable for this test: [#1632 (review)](https://github.com/intel/scikit-learn-intelex/pull/1632#pullrequestreview-1813057433)

Alexsandruss:
@icfaust, I moved decoration by `run_with_n_jobs` to class decorator implementation using list of methods as argument to simplify control and minimize code. I guess it will be enough to test for class decorator with these changes.

daal4py/sklearn/_utils.py
```diff
@@ -411,3 +378,66 @@ def method_wrapper(self, *args, **kwargs):
...
     return method_wrapper
+
+
+def control_n_jobs(decorated_methods: list):
+    """Decorator for the control of 'n_jobs' parameter in estimator class. It applied
+    for all estimators with and without support of parameter in original sklearn.
+    In case of estimator without 'n_jobs' support, this decorator adds it.
+    Methods listed in 'decorated_methods' argument are only methods
+    to be run with 'n_jobs'.
+    """
```
ahuber21, line 389: I think we can improve this description

```suggestion
"""Decorator for controlling the 'n_jobs' parameter in an estimator class.

This decorator is designed to be applied to both estimators with and without native support for the 'n_jobs' parameter in the original scikit-learn implementation. When applied to an estimator without 'n_jobs' support in its original '__init__' method, this decorator adds the 'n_jobs' parameter.

Additionally, this decorator allows for fine-grained control over which methods should be executed with the 'n_jobs' parameter. The methods specified in the 'decorated_methods' argument will run with 'n_jobs', while all other methods remain unaffected.

Parameters:
    decorated_methods (list): A list of method names to be executed with 'n_jobs'.

Example:
    @control_n_jobs(decorated_methods=['fit', 'predict'])
    class MyEstimator:
        def __init__(self, *args, **kwargs):
            # Your original __init__ implementation here

        def fit(self, *args, **kwargs):
            # Your original fit implementation here

        def predict(self, *args, **kwargs):
            # Your original predict implementation here

        def other_method(self, *args, **kwargs):
            # Methods not listed in decorated_methods will not be affected by 'n_jobs'
            pass
"""
```

And why not call the decorator simply `n_jobs_decorator`?

daal4py/sklearn/_utils.py
```diff
@@ -411,3 +378,66 @@ def method_wrapper(self, *args, **kwargs):
         return result
 
     return method_wrapper
+
+
+def control_n_jobs(decorated_methods: list):
```
ahuber21, line 383: I think we should add a few tests.

* Does a class with `n_jobs` in `__init__()` behave as expected
* Does a class without `n_jobs` in `__init__()` behave as expected
* Are the functions in `decorated_methods` modified as expected
* Are functions *not* in `decorated_methods` untouched

daal4py/sklearn/_utils.py
```diff
@@ -362,7 +318,9 @@ def get_suggested_n_threads(n_cpus):
 
 
 def run_with_n_jobs(method):
```
ahuber21, line 320: And what about tests for this method? We should add least add the most basic closures, like checking if attributes that should be added are there, functions are decorated, etc.

ahuber21 (changes requested):
I feel like we should add basic closure tests before merging

icfaust (changes requested):
I'd like to see a test for this for coverage purposes (or let me know if I should adapt the n_jobs test PR for this one).

daal4py/sklearn/_utils.py
```diff
@@ -362,7 +318,9 @@ def get_suggested_n_threads(n_cpus):
...
-    """Decorator for running of methods containing oneDAL kernels with 'n_jobs'"""
+    """Decorator for running of methods containing oneDAL kernels with 'n_jobs'.
+    This decorator is not expected to be used manually.
```
olegkkruglov, line 322: It might be better to write few words here about how it should be used then. If it shouldn't be used by user then it's probably worth to name it with leading underscore to highlight this fact.

daal4py/sklearn/_utils.py
```diff
@@ -129,6 +129,10 @@ def sklearn_check_version(ver):
     return res
 
 
+if sklearn_check_version("1.2"):
+    from sklearn.utils._param_validation import validate_parameter_constraints
```
samir-nasibli, line 133: Better to move it on top of the file.

samir-nasibli (commented):
Thank you everyone for the review!
New primitives should be well documented, even for private utils funcst good docstrings is required.
Please add more and more tests. It seems that sklearn test validation is not enough for that/

daal4py/sklearn/_utils.py
Alexsandruss: `sklearn_check_version` is defined in same file after all other imports.

samir-nasibli (commented):
It make sense to move out from utils module all `n_jobs` related primitives into separate file/module.
Like we did it for device_offloading

daal4py/sklearn/_utils.py
Alexsandruss: Description was extended.

Alexsandruss: It's not common to name decorators as `*_decorator` in sklearn at least.

daal4py/sklearn/_utils.py
Alexsandruss: Basic test to check sklearnex logs was added.

daal4py/sklearn/_utils.py
```diff
@@ -411,3 +378,66 @@ def method_wrapper(self, *args, **kwargs):
...
+        @wraps(original_init)
+        def init_with_n_jobs(self, *args, n_jobs=None, **kwargs):
+            original_init(self, *args, **kwargs)
+            self.n_jobs = n_jobs
+
+        # add "n_jobs" parameter to signature of wrapped init
+        # if estimator doesn't originally support it
+        sig = signature(original_init)
+        original_params = list(sig.parameters.values())
+        if "n_jobs" not in list(map(lambda param: param.name, original_params)):
```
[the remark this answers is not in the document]
Alexsandruss: `sig.replace` doesn't accept OrderedDict, but I simplified other mentioned code.

daal4py/sklearn/_utils.py
```diff
@@ -411,3 +378,66 @@ def method_wrapper(self, *args, **kwargs):
...
+            original_params.append(
+                Parameter("n_jobs", Parameter.KEYWORD_ONLY, default=None)
+            )
+            init_with_n_jobs.__signature__ = sig.replace(parameters=original_params)
+            original_class.__init__ = init_with_n_jobs
+
+        # add n_jobs to __doc__ string if needed
+        if (
+            hasattr(original_class, "__doc__")
```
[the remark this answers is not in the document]
Alexsandruss: Depending on this PR and #1607 readiness, it should be changed in last PR to merge.

daal4py/sklearn/_utils.py
Alexsandruss: Description was extended

daal4py/sklearn/_utils.py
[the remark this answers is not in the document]
Alexsandruss: Underscore prefix was added, function is now hidden.

sklearnex/neighbors/lof.py
```diff
@@ -38,6 +39,7 @@
 
 if sklearn_check_version("1.0"):
 
+    @control_n_jobs(decorated_methods=[])
```
samir-nasibli, line 42: It will be prettier if we have either `None` and `list` for a `control_n_jobs` input.
`None` would be better instead of having empty list:
```suggestion
    @control_n_jobs()
```

icfaust (commented):
Thanks for adding the test and catching lof. Depending on if this or the lof refactor comes in,  either of us will need to merge conflicts. If you can, please dismiss my changes requested, that way I can re-review when necessary or approve later on

sklearnex/neighbors/lof.py
Alexsandruss: Done.

icfaust:
One thing that I've been thinking about is how to insure that the right methods (ideally those that are patched) have been properly decorated. Is there any way to adapt/ add a test which verifies that the right methods have been addressed by n_jobs?

daal4py/sklearn/_utils.py
Alexsandruss: Decoration check was added in https://github.com/intel/scikit-learn-intelex/pull/1622/commits/25725d8901abdbf3b472cb80846e4471e1f52095

Alexsandruss:
@icfaust @ahuber21 @samir-nasibli, all comments applied and CI is green. Any other thoughts?

icfaust (approved):

sklearnex/tests/test_n_jobs_support.py
```diff
@@ -14,18 +14,76 @@
...
+
+
+@pytest.mark.parametrize("estimator_class", ESTIMATORS)
+@pytest.mark.parametrize("n_jobs", [None, -1, 1, 2])
+def test_n_jobs_support(estimator_class, n_jobs):
+    def check_estimator_doc(estimator):
+        if estimator.__doc__ is not None:
+            assert "n_jobs" in estimator.__doc__
+
+    def get_sklearnex_logger_stream(level="DEBUG"):
```
ahuber21, line 46: Did you look into https://docs.pytest.org/en/latest/how-to/logging.html#caplog-fixture

Is there anything special about our logs that we have to reinvent the wheel?

ahuber21: By the way, why are we testing logs in the first place? Is this an important message that must be shown to the user. Or is it rather a proxy to check if some statements were executed? If it is the latter, we should maybe do this test differently.

Alexsandruss: Updated to caplog usage (https://github.com/intel/scikit-learn-intelex/pull/1622/commits/4b53223206733813e1924e1a5e0e3bc6b2ad5a14).

Alexsandruss: These messages are important for debugging.

ahuber21:
Approving assuming CI still passes :)

samir-nasibli (approved):
Looks good. Thank you!
Let's wait green CI and merge

PR #1625, opened 2023-12-29
infra: disable MacOS public CI tests

ethanglaser:
# Description

-
-
-

### DIFF
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -37,7 +37,6 @@ pr:
     - .ci/pipeline/docs.yml
 
 variables:
-  MACOSX_DEPLOYMENT_TARGET: '10.15'
   PYTHON: 'python'
   ARGS: '1'
   SHELLOPTS: 'errexit:pipefail'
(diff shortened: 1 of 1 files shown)


md-shafiul-alam (approved):
looks good to me

ethanglaser:
I suppose this could be removed as its not currently being used: https://github.com/intel/scikit-learn-intelex/blob/main/.ci/pipeline/build-and-test-mac.yml. But if we are viewing this as a temporary disable then may not be necessary.

napetrov:
> I suppose this could be removed as its not currently being used: https://github.com/intel/scikit-learn-intelex/blob/main/.ci/pipeline/build-and-test-mac.yml. But if we are viewing this as a temporary disable then may not be necessary. Any thoughts?

As i see this - if this is temporal - then will be recovered with revert, if it's permanent, then it's better to do complete removal

ethanglaser:
> > I suppose this could be removed as its not currently being used: https://github.com/intel/scikit-learn-intelex/blob/main/.ci/pipeline/build-and-test-mac.yml. But if we are viewing this as a temporary disable then may not be necessary. Any thoughts?
> 
> As i see this - if this is temporal - then will be recovered with revert, if it's permanent, then it's better to do complete removal

Will leave it then, since goal would be to re-enable

PR #1640, opened 2024-01-08
[bug] Fix LocalOutlierFactor kneighbors method via refactor

icfaust:
# Description
Initial design of LocalOutlierFactor in sklearnex does not patch the kneighbors method which will cause failures. It also improperly patches sklearn for other methods, and does not have n_jobs support. In general it also does not follow sklearnex coding conventions (_onedal_estimator, etc.). This PR fixes the kneighbors method, properly checks for patching, and meets design rules while reducing the code length.

- Inherit KNeighborsDispatchingBase
- Only fully implement fit_predict, predict, kneighbors and fit, overwrite _predict to allow for use of score_subsample.
- Fix sklearn_check_version use from code duplication
- Add n_jobs support
- add docstrings
- enable GPU LocalOutlierFactor sklearn tests
- solves dpctl dpnp support for kneighbors method (also with re-enabled tests)


sklearnex/neighbors/__init__.py
KulikovNikita, line 1: What was the motivation to rename the '_lof' module?

sklearnex/neighbors/_lof.py
KulikovNikita, line 1: Please change year in the header

sklearnex/neighbors/_lof.py
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env python
+# ===============================================================================
+# Copyright 2023 Intel Corporation
```
KulikovNikita, line 3:
```suggestion
# Copyright 2024 Intel Corporation
```

sklearnex/neighbors/_lof.py
```diff
@@ -0,0 +1,180 @@
...
+        try:
+            requires_y = self._get_tags()["requires_y"]
+        except KeyError:
+            requires_y = False
+
+        self._onedal_estimator = onedal_NearestNeighbors(**onedal_params)
+        self._onedal_estimator.requires_y = requires_y
+        self._onedal_estimator.effective_metric_ = self.effective_metric_
+        self._onedal_estimator.effective_metric_params_ = self.effective_metric_params_
+        self._onedal_estimator.fit(X, y, queue=queue)
```
KulikovNikita, line 85: Should we make it a separate function? It looks like a separate sequence of actions

sklearnex/neighbors/_lof.py
```diff
@@ -0,0 +1,180 @@
...
+        if self.contamination != "auto":
+            if not (0.0 < self.contamination <= 0.5):
+                raise ValueError(
+                    "contamination must be in (0, 0.5], " "got: %f" % self.contamination
+                )
+
+        n_samples = self.n_samples_fit_
+
+        if self.n_neighbors > n_samples:
+            warnings.warn(
```
KulikovNikita, line 98: Is there a way to steal this error message from vanilla scikit?

sklearnex/neighbors/_lof.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+
+        # Compute lof score over training samples to define offset_:
+        lrd_ratios_array = self._lrd[_neighbors_indices_fit_X_] / self._lrd[:, np.newaxis]
+
+        self.negative_outlier_factor_ = -np.mean(lrd_ratios_array, axis=1)
+
+        if self.contamination == "auto":
+            # inliers score around -1 (the higher, the less abnormal).
+            self.offset_ = -1.5
```
KulikovNikita, line 124: Looks very unintuitive.can it be let say -2.0?

KulikovNikita (approved):
Looks good to me except minor issues

sklearnex/neighbors/__init__.py
icfaust: To better match sklearn convention. I think we should make a rule about this (personal opinion).

sklearnex/neighbors/_lof.py
icfaust: You caught me, its a copy-paste of the NearestNeighbors._onedal_fit() function into this

sklearnex/neighbors/_lof.py
icfaust: Direct from sklearn unfortunately

sklearnex/neighbors/_lof.py
```diff
@@ -121,5 +147,33 @@ def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
...
+        # In that case mask the first duplicate.
+        dup_gr_nbrs = np.all(sample_mask, axis=1)
+        sample_mask[:, 0][dup_gr_nbrs] = False
+
+        neigh_ind = np.reshape(neigh_ind[sample_mask], (n_queries, n_neighbors - 1))
+
+        if return_distance:
+            neigh_dist = np.reshape(neigh_dist[sample_mask], (n_queries, n_neighbors - 1))
+            return neigh_dist, neigh_ind
+        return neigh_ind
```
samir-nasibli, line 176: Please move all this into NearestNeighbors._onedal_kneighbors.
If dispatch will go with stock sklearn branch this will be duplication.

onedal/neighbors/neighbors.py
```diff
@@ -131,20 +131,20 @@ def _get_onedal_params(self, X, y=None):
...
-            "neighbor_count": self.n_neighbors,
+            "neighbor_count": self.n_neighbors if n_neighbors is None else n_neighbors,
             "metric": self.effective_metric_,
             "p": self.p,
             "metric_params": self.effective_metric_params_,
             "result_option": "indices|distances" if y is None else "responses",
         }
 
-    def _get_daal_params(self, data):
+    def _get_daal_params(self, data, n_neighbors=None):
```
samir-nasibli, line 141: use private attribute self._n_neighbors instead of extending this funcs signature

samir-nasibli (changes requested):
Partially reviewed, please address comments :)
I really don't like to have all this changes in one PR.
Even if feature is one line code, I would like to have it in separate PR with separate validation.

onedal/neighbors/neighbors.py
icfaust: This is necessary to prevent wrong values being stored in self.n_neighbors when using the fitted X values.

sklearnex/neighbors/_lof.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
icfaust: This exists in most sklearnex headers

icfaust:
@samir-nasibli I agree. In fixing the LocalOutlierFactor bugs, I have uncovered/fixed a general KNN bug with dpnp dpctl arrays here, I will split them into another PR which will need to be reviewed first. This is currently just for testing at the moment.

sklearnex/neighbors/common.py
```diff
@@ -71,9 +66,12 @@ def _fit_validation(self, X, y=None):
...
         if not isinstance(X, (KDTree, BallTree, sklearn_NeighborsBase)):
-            self._fit_X = self._check_array(
-                X, dtype=[np.float64, np.float32], accept_sparse=True
-            )
+            if not hasattr(X, "__sycl_usm_array_interface__"):
+                self._fit_X = _check_array(
+                    X, dtype=[np.float64, np.float32], accept_sparse=True
+                )
+            else:
+                self._fit_X = X
```
samir-nasibli, line 74: This is really looks like not a good workaround.
Try avoid checking this attribute for input data.
We have to use only numpy arrays internally now.

sklearnex/neighbors/common.py
```diff
@@ -262,64 +272,12 @@ def _onedal_supported(self, device, method_name, *data):
...
-            neigh_dist = np.reshape(neigh_dist[sample_mask], (n_queries, n_neighbors - 1))
-            return neigh_dist, neigh_ind
-        return neigh_ind
-
-    def _onedal_gpu_supported(self, method_name, *data):
-        return self._onedal_supported("gpu", method_name, *data)
-
-    def _onedal_cpu_supported(self, method_name, *data):
-        return self._onedal_supported("cpu", method_name, *data)
+        return queue
```
samir-nasibli: It is weird and unexpected that we have this checker here?
What is KNN case? Why we have this work around and can not to use common approach?

sklearnex/neighbors/common.py
```diff
@@ -262,64 +272,12 @@ def _onedal_supported(self, device, method_name, *data):
...
-                raise TypeError(
-                    "n_neighbors does not take %s value, "
-                    "enter integer value" % type(n_neighbors)
+    def _fit_queue_check(self, X, queue=None):
+        if X is None and hasattr(self, "_fit_queue"):
+            if queue is None:
+                queue = self._fit_queue
+            elif queue != self._fit_queue:
+                raise RuntimeError(
+                    "Input data shall be located " "on single target device"
```
samir-nasibli, line 285: It is weird and unexpected that we have this checker here?
What is KNN case? Why we have this work around and can not to use common approach?

icfaust: This is necessary because I need to preserve the queue used in making any NN fit, the kneighbors method is special as it will reuse the X data if called like 'kneighbors()', so then it is import to know the queue if it is a dpnp or dpctl array/tensor. I was going to preserve the dpnp or dpctl array, but you had said we should use numpy internally, so storing the queue for later use is the only way.

If the following situation were to occur:
```python
q = dpctl.SyclQueue("gpu:0")

X = dpnp.asarray(np.array([[1., 2.], [2., 2.], [2., 3.],
                            [8., 7.], [8., 8.], [25., 80.]], dtype=np.float32),
                            usm_type="device", sycl_queue=q)

est = NearestNeighbors().fit(X)

with config_context(target_offload="gpu:1"):
    est.kneighbors()
```
This should fail. I will need to recheck the queue since I am avoiding checks in _device_offload.dispatch.

I think because of the depth of issues in this I will make a meeting to clarify everything. This is still for testing at this point and will be split into two PRs.

sklearnex/_device_offload.py
```diff
@@ -67,7 +69,9 @@ def get_filter_string(self):
             return self._filter_string
 
     def __init__(self, filter_string):
-        self.sycl_device = self.DummySyclDevice(filter_string)
+        if not filter_string in self.__class__._cache:
+            self.__class__._cache[filter_string] = self.DummySyclDevice(filter_string)
+        self.sycl_device = self.__class__._cache[filter_string]
```
samir-nasibli, line 74: What is the purpose of such changes? And why it is in this PR?

onedal/neighbors/neighbors.py
icfaust: @samir-nasibli I guess I don't understand. It would just match the value in self.n_neighbors in most cases, and then we would still have to if statement between self.n_neighbors and self._n_neighbors.  Could you explain a little more?

icfaust:
While I could get it to pass all tests (including gpu support) at half the code size and functions, dpnp and dpctl support requires adding lots of duplicated code in order to wrap or subtly change the sklearn code.  We need to probably add a test which goes through every unpatched public method and test for dpnp and dpctl conformance. As far as I can see, we only do it in special cases in the custom sklearnex tests. Sort of sad really.

sklearnex/neighbors/_lof.py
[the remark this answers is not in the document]
ethanglaser: I think its not resolved yet

onedal/neighbors/neighbors.py
```diff
@@ -349,9 +348,6 @@ def _kneighbors(self, X=None, n_neighbors=None, return_distance=True, queue=None
         # If the query data is the same as the indexed data, we would like
```
ethanglaser, line 348: Can remove the comment as well if the code is being removed as well

ethanglaser:
Main question is do we have any test coverage of this functionality? I don't see anything in internal CI (no lof/outlier in tests, examples, and deselected in conformance)

onedal/neighbors/neighbors.py
icfaust: I added some separation from the call before for clarity, but comes from here: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/neighbors/_base.py#L910

icfaust:
I am going to activate all the possible lof tests and run CIs. The sklearn tests are activated [here](https://github.com/intel/scikit-learn-intelex/blob/main/.ci/scripts/select_sklearn_tests.py#L46), but in [sklearn](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/neighbors/tests/test_lof.py) they aren't extensive.

icfaust:
re-run due to windows timeouts http://intel-ci.intel.com/eeb5f368-df77-f1b2-b1e2-a4bf010d0e2e

icfaust:
This PR currently includes #1662 out of necessity

sklearnex/neighbors/_lof.py
```diff
@@ -0,0 +1,167 @@
...
+            self.offset_ = -1.5
+        else:
+            self.offset_ = np.percentile(
+                self.negative_outlier_factor_, 100.0 * self.contamination
+            )
+
+        return self
+
+    def fit(self, X, y=None):
+        self._fit_validation(X, y)
```
icfaust, line 103: Move onedal_fit

icfaust: https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/neighbors/knn_unsupervised.py#L128, also exists in knn, and parameters are used in _onedal_cpu_supported and _onedal_gpu_supported :(

icfaust:
Fails are unrelated omp issues.

deselected_tests.yaml
```diff
@@ -761,17 +760,10 @@ gpu:
   - manifold/tests/test_t_sne.py::test_binary_perplexity_stability
   - manifold/tests/test_t_sne.py::test_gradient_bh_multithread_match_sequential
   - neighbors/tests/test_kde.py::test_kernel_density_sampling
-  - neighbors/tests/test_lof.py
```
ethanglaser, line 764: Niiice

deselected_tests.yaml
```diff
@@ -1172,22 +1162,6 @@ gpu:
...
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_estimators_nan_inf]
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_estimators_overwrite_params]
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_estimators_pickle]
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_methods_sample_order_invariance]
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_methods_subset_invariance]
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_fit2d_1sample]
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_fit2d_1feature]
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_dict_unchanged]
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_dont_overwrite_parameters]
-  - tests/test_common.py::test_estimators[LocalOutlierFactor()-check_fit2d_predict1d]
```
ethanglaser, line 1190: Also very nice

sklearnex/neighbors/tests/test_neighbors.py
ethanglaser, line 1: Also great work

ethanglaser (approved):
LGTM

samir-nasibli (approved):
@icfaust Thank you very much for a such great work done!
:rocket:

PR #1644, opened 2024-01-10
ENH: `BasicStatistics` API change

olegkkruglov:
- Refactored basic_statistics.cpp file
- Changed API for BasicStatistics class to save results to class attribute instead of returning dict with result to be consistent with IncrementalBasicStatistics and other estimators. Saved attributes are arrays if input data is 2d and numbers if input data is 1d (like it is done in numpy)
- Updated tests
- Added support of n_jobs and usm_ndarray to BasicStatistics
- Rename compute_raw method to _compute_raw since it is private and designed for internal use

### DIFF
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -99,7 +99,7 @@ class BasicStatistics(BaseBasicStatistics):
     def compute(self, data, weights=None, queue=None):
         warnings.warn(
             "Method `compute` was deprecated in version 2024.6 and will be "
-            "removed in 2026.0. Use `fit` instead."
+            "removed in 2025.0. Use `fit` instead."
         )
 
         is_csr = _is_csr(data)
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -18,8 +18,6 @@ from abc import ABCMeta, abstractmethod
 
 import numpy as np
 
-from onedal import _backend
-
 from ..common._base import BaseEstimator
 from ..datatypes import _convert_to_supported, from_table, to_table
 from ..utils import _is_csr
@@ -30,7 +28,6 @@ class BaseBasicStatistics(BaseEstimator, metaclass=ABCMeta):
     def __init__(self, result_options, algorithm):
         self.options = result_options
         self.algorithm = algorithm
-        self._module = _backend.basic_statistics
 
     @staticmethod
     def get_all_result_options():
@@ -94,8 +91,9 @@ class BasicStatistics(BaseBasicStatistics):
         return self
 
     def _compute_raw(self, data_table, weights_table, policy, dtype=np.float32, is_csr=False):
+        module = self._get_backend("basic_statistics")
         params = self._get_onedal_params(is_csr, dtype)
-        result = self._module.compute(policy, params, data_table, weights_table)
+        result = module.compute(policy, params, data_table, weights_table)
         options = self._get_result_options(self.options).split("|")
 
         return {opt: getattr(result, opt) for opt in options}
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -98,7 +98,7 @@ class BasicStatistics(BaseBasicStatistics):
 
     def compute(self, data, weights=None, queue=None):
         warnings.warn(
-            "Method `compute` was deprecated in version 2024.6 and will be "
+            "Method `compute` was deprecated in version 2024.7 and will be "
             "removed in 2025.0. Use `fit` instead."
         )
 
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
+import warnings
 from abc import ABCMeta, abstractmethod
 
 import numpy as np
@@ -21,6 +22,7 @@ import numpy as np
 from ..common._base import BaseEstimator
 from ..datatypes import _convert_to_supported, from_table, to_table
 from ..utils import _is_csr
+from ..utils.validation import _check_array
 
 
 class BaseBasicStatistics(BaseEstimator, metaclass=ABCMeta):
@@ -73,21 +75,47 @@ class BasicStatistics(BaseBasicStatistics):
         policy = self._get_policy(queue, data, sample_weight)
 
         is_csr = _is_csr(data)
+
         if not (data is None) and not is_csr:
-            data = np.asarray(data)
+            data = _check_array(data, ensure_2d=False)
         if not (sample_weight is None):
-            sample_weight = np.asarray(sample_weight)
+            sample_weight = _check_array(sample_weight, ensure_2d=False)
 
         data, sample_weight = _convert_to_supported(policy, data, sample_weight)
+        is_single_dim = data.ndim == 1
         data_table, weights_table = to_table(data, sample_weight)
 
         dtype = data.dtype
         raw_result = self._compute_raw(data_table, weights_table, policy, dtype, is_csr)
         for opt, raw_value in raw_result.items():
             value = from_table(raw_value).ravel()
-            setattr(self, opt, value)
+            if is_single_dim:
+                setattr(self, opt, value[0])
+            else:
+                setattr(self, opt, value)
 
         return self
+
+    def compute(self, data, weights=None, queue=None):
+        warnings.warn(
+            "Method `compute` was deprecated in version 2024.3 and will be "
+            "removed in 2024.5. Use `fit` instead."
+        )
+
+        is_csr = _is_csr(data)
+
+        if not (data is None):
+            data = _check_array(data, ensure_2d=False)
+        if not (weights is None):
+            weights = _check_array(weights, ensure_2d=False)
+
+        policy = self._get_policy(queue, data, weights)
+        data, weights = _convert_to_supported(policy, data, weights)
+        data_table, weights_table = to_table(data, weights)
+        dtype = data.dtype
+        res = self._compute_raw(data_table, weights_table, policy, dtype, is_csr)
+
+        return {k: from_table(v).ravel() for k, v in res.items()}
 
     def _compute_raw(self, data_table, weights_table, policy, dtype=np.float32, is_csr=False):
         module = self._get_backend("basic_statistics")
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -15,7 +15,6 @@
 # ==============================================================================
 
 from abc import ABCMeta, abstractmethod
-from numbers import Number
 
 import numpy as np
 
@@ -31,6 +30,7 @@ class BaseBasicStatistics(metaclass=ABCMeta):
     def __init__(self, result_options, algorithm):
         self.options = result_options
         self.algorithm = algorithm
+        self._module = _backend.basic_statistics
 
     @staticmethod
     def get_all_result_options():
@@ -63,19 +63,16 @@ class BaseBasicStatistics(metaclass=ABCMeta):
             "result_option": options,
         }
 
-    def _compute_raw(
-        self, data_table, weights_table, module, policy, dtype=np.float32, is_csr=False
-    ):
-        params = self._get_onedal_params(is_csr, dtype)
 
-        result = module.train(policy, params, data_table, weights_table)
+class BasicStatistics(BaseBasicStatistics):
+    """
+    Basic Statistics oneDAL implementation.
+    """
 
-        options = self._get_result_options(self.options)
-        options = options.split("|")
+    def __init__(self, result_options="all", algorithm="by_default"):
+        super().__init__(result_options, algorithm)
 
-        return {opt: getattr(result, opt) for opt in options}
-
-    def _compute(self, data, weights, module, queue):
+    def fit(self, data, weights=None, queue=None):
         policy = self._get_policy(queue, data, weights)
 
         is_csr = _is_csr(data)
@@ -86,36 +83,19 @@ class BaseBasicStatistics(metaclass=ABCMeta):
             weights = np.asarray(weights)
 
         data, weights = _convert_to_supported(policy, data, weights)
-
         data_table, weights_table = to_table(data, weights)
 
         dtype = data.dtype
-        res = self._compute_raw(data_table, weights_table, module, policy, dtype, is_csr)
+        raw_result = self._compute_raw(data_table, weights_table, policy, dtype, is_csr)
+        for opt, raw_value in raw_result.items():
+            value = from_table(raw_value).ravel()
+            setattr(self, opt, value)
 
-        return {k: from_table(v).ravel() for k, v in res.items()}
+        return self
 
+    def _compute_raw(self, data_table, weights_table, policy, dtype=np.float32, is_csr=False):
+        params = self._get_onedal_params(is_csr, dtype)
+        result = self._module.compute(policy, params, data_table, weights_table)
+        options = self._get_result_options(self.options).split("|")
 
-class BasicStatistics(BaseEstimator, BaseBasicStatistics):
-    """
-    Basic Statistics oneDAL implementation.
-    """
-
-    def __init__(self, result_options="all", *, algorithm="by_default", **kwargs):
-        super().__init__(result_options, algorithm)
-
-    def compute(self, data, weights=None, queue=None):
-        return super()._compute(
-            data, weights, self._get_backend("basic_statistics", "compute", None), queue
-        )
-
-    def compute_raw(
-        self, data_table, weights_table, policy, dtype=np.float32, is_csr=False
-    ):
-        return super()._compute_raw(
-            data_table,
-            weights_table,
-            self._get_backend("basic_statistics", "compute", None),
-            policy,
-            dtype,
-            is_csr,
-        )
+        return {opt: getattr(result, opt) for opt in options}
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -76,9 +76,9 @@ class BasicStatistics(BaseBasicStatistics):
 
         is_csr = _is_csr(data)
 
-        if not (data is None) and not is_csr:
+        if data is not None and not is_csr:
             data = _check_array(data, ensure_2d=False)
-        if not (sample_weight is None):
+        if sample_weight is not None:
             sample_weight = _check_array(sample_weight, ensure_2d=False)
 
         data, sample_weight = _convert_to_supported(policy, data, sample_weight)
@@ -104,9 +104,9 @@ class BasicStatistics(BaseBasicStatistics):
 
         is_csr = _is_csr(data)
 
-        if not (data is None):
+        if data is not None:
             data = _check_array(data, ensure_2d=False)
-        if not (weights is None):
+        if weights is not None:
             weights = _check_array(weights, ensure_2d=False)
 
         policy = self._get_policy(queue, data, weights)
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -25,7 +25,7 @@ from ..datatypes import _convert_to_supported, from_table, to_table
 from ..utils import _is_csr
 
 
-class BaseBasicStatistics(metaclass=ABCMeta):
+class BaseBasicStatistics(BaseEstimator, metaclass=ABCMeta):
     @abstractmethod
     def __init__(self, result_options, algorithm):
         self.options = result_options
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -117,7 +117,9 @@ class BasicStatistics(BaseBasicStatistics):
 
         return {k: from_table(v).ravel() for k, v in res.items()}
 
-    def _compute_raw(self, data_table, weights_table, policy, dtype=np.float32, is_csr=False):
+    def _compute_raw(
+        self, data_table, weights_table, policy, dtype=np.float32, is_csr=False
+    ):
         module = self._get_backend("basic_statistics")
         params = self._get_onedal_params(is_csr, dtype)
         result = module.compute(policy, params, data_table, weights_table)
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -98,8 +98,8 @@ class BasicStatistics(BaseBasicStatistics):
 
     def compute(self, data, weights=None, queue=None):
         warnings.warn(
-            "Method `compute` was deprecated in version 2024.3 and will be "
-            "removed in 2024.5. Use `fit` instead."
+            "Method `compute` was deprecated in version 2024.6 and will be "
+            "removed in 2026.0. Use `fit` instead."
         )
 
         is_csr = _is_csr(data)
--- a/onedal/basic_statistics/basic_statistics.py
+++ b/onedal/basic_statistics/basic_statistics.py
@@ -69,18 +69,17 @@ class BasicStatistics(BaseBasicStatistics):
     def __init__(self, result_options="all", algorithm="by_default"):
         super().__init__(result_options, algorithm)
 
-    def fit(self, data, weights=None, queue=None):
-        policy = self._get_policy(queue, data, weights)
+    def fit(self, data, sample_weight=None, queue=None):
+        policy = self._get_policy(queue, data, sample_weight)
 
         is_csr = _is_csr(data)
         if not (data is None) and not is_csr:
             data = np.asarray(data)
+        if not (sample_weight is None):
+            sample_weight = np.asarray(sample_weight)
 
-        if not (weights is None):
-            weights = np.asarray(weights)
-
-        data, weights = _convert_to_supported(policy, data, weights)
-        data_table, weights_table = to_table(data, weights)
+        data, sample_weight = _convert_to_supported(policy, data, sample_weight)
+        data_table, weights_table = to_table(data, sample_weight)
 
         dtype = data.dtype
         raw_result = self._compute_raw(data_table, weights_table, policy, dtype, is_csr)
--- a/onedal/basic_statistics/incremental_basic_statistics.py
+++ b/onedal/basic_statistics/incremental_basic_statistics.py
@@ -1,5 +1,5 @@
 # ==============================================================================
-# Copyright 2024 Intel Corporation
+# Copyright 2023 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


onedal/basic_statistics/basic_statistics.py
```diff
@@ -66,17 +66,16 @@ def _get_onedal_params(self, dtype=np.float32):
...
+    """
+    Basic Statistics oneDAL implementation.
+    """
 
-        return {opt: getattr(result, opt) for opt in options}
+    def __init__(self, result_options="all", algorithm="by_default"):
+        super().__init__(result_options, algorithm)
 
-    def _compute(self, data, weights, module, queue):
+    def compute(self, data, weights=None, queue=None):
```
samir-nasibli, line 78:
```suggestion
    def fit(self, data, weights=None, queue=None):
```
Let's make it more scikit-learn-like

onedal/basic_statistics/basic_statistics.py
```diff
@@ -85,27 +84,18 @@ def _compute(self, data, weights, module, queue):
...
 
     def compute_raw(self, data_table, weights_table, policy, dtype=np.float32):
-        return super()._compute_raw(
-            data_table, weights_table, _backend.basic_statistics.compute, policy, dtype
-        )
+        params = self._get_onedal_params(dtype)
+        result = self._module.compute(policy, params, data_table, weights_table)
+        options = self._get_result_options(self.options).split("|")
+
+        return {opt: getattr(result, opt) for opt in options}
```
samir-nasibli, line 101: Generally, for public methods it is better to have defaults values for all params, except input data.

But for this function it will be hard to provide public iface, because for user it would very complicated to define policy value and tables.

It is better to leave it as is private, since it is only internally used.

onedal/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -67,11 +49,36 @@ def test_option_uniform(queue, option, dtype):
...
+        res, gtr = getattr(res, result_option), function(data, axis=0)
 
         tol = fp32tol if res.dtype == np.float32 else fp64tol
         assert_allclose(gtr, res, rtol=tol)
 
+    @pytest.mark.parametrize("queue", get_queues())
+    @pytest.mark.parametrize("dtype", [np.float32, np.float64])
+    def test_multiple_options_uniform(queue, dtype):
+        seed = 42
+        s_count, f_count = 70000, 29
```
samir-nasibli, line 61: Do we really need such big data for the tests?

setup_sklearnex.py
```diff
@@ -84,6 +84,7 @@
 
 packages_with_tests = [
     "sklearnex",
+    # "sklearnex.basic_statistics",
```
samir-nasibli, line 87: ?

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -14,4 +14,86 @@
...
+        if self.options == "all":
+            result_options = onedal_BasicStatistics.get_all_result_options()
+        else:
+            result_options = self.options
+
+        for option in result_options:
+            setattr(self, option, getattr(self._onedal_estimator, option))
+
+    @run_with_n_jobs
+    def _onedal_compute(self, X, weights=None, queue=None):
```
samir-nasibli, line 73:
```suggestion
    def _onedal_fit(self, X, weights=None, queue=None):
```

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -14,4 +14,86 @@
...
+        sum_squares_centered : ndarray of shape (n_features,)
+            Centered sum of squares for each feature over all samples.
+        second_order_raw_moment : ndarray of shape (n_features,)
+            Second order moment of each feature over all samples.
+    """
+
+    def __init__(self, result_options="all"):
+        self.options = result_options
+
+    _onedal_incremental_covariance = staticmethod(onedal_BasicStatistics)
```
KulikovNikita, line 59: Is it really relevant? Foe me personally it looks more like an artifact produced by a copy/paste😅

KulikovNikita: Another question - are we okay with using attributes starting with one underscore in this scope? Should it be accessible from outside?

onedal/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -32,30 +32,12 @@
...
-        tol = 2e-5 if res_mean.dtype == np.float32 else 1e-7
-        assert_allclose(gtr_mean, res_mean, rtol=tol)
-
     @pytest.mark.parametrize("queue", get_queues())
     @pytest.mark.parametrize("option", options_and_tests)
     @pytest.mark.parametrize("dtype", [np.float32, np.float64])
     def test_option_uniform(queue, option, dtype):
         seed = 77
-        s_count, f_count = 19999, 31
+        s_count, f_count = 1000, 31
```
KulikovNikita, line 40: I believe we can afford testing such a simple functionality on a larger datasets

onedal/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -65,13 +47,38 @@ def test_option_uniform(queue, option, dtype):
...
+        res = alg.fit(data, queue=queue)
 
-        res, gtr = res[result_option], function(data, axis=0)
+        res, gtr = getattr(res, result_option), function(data, axis=0)
 
         tol = fp32tol if res.dtype == np.float32 else fp64tol
         assert_allclose(gtr, res, rtol=tol)
 
+    @pytest.mark.parametrize("queue", get_queues())
+    @pytest.mark.parametrize("dtype", [np.float32, np.float64])
```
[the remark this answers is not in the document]
samir-nasibli: good catch

sklearnex/basic_statistics/basic_statistics.py
olegkkruglov: > Is it really relevant? Foe me personally it looks more like an artifact produced by a copy/paste😅

yes, it is. thanks for pointing out.

> Another question - are we okay with using attributes starting with one underscore in this scope? Should it be accessible from outside?

do you mean it should start with two underscores? it is exactly what we discussed with @samir-nasibli and it looks like we have an agreement about naming "private" methods with one underscore to simplify development

onedal/basic_statistics/tests/test_basic_statistics.py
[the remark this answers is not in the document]
olegkkruglov: what do you mean by reordering? i have just added a new test test_multiple_options_uniform to make sure it works well if result_options is a list type

onedal/spmd/basic_statistics/basic_statistics.py
```diff
@@ -29,5 +29,5 @@ def _get_policy(self, queue, *data):
 
 class BasicStatistics(BaseBasicStatisticsSPMD, BasicStatistics_Batch):
     @support_usm_ndarray()
-    def compute(self, data, weights=None, queue=None):
-        return super().compute(data, weights, queue)
+    def fit(self, data, weights=None, queue=None):
```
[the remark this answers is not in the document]
olegkkruglov: due to the previous comments by @samir-nasibli the naming should be the same as in scikit-learn, where fit is used everywhere instead of compute

onedal/basic_statistics/tests/test_basic_statistics.py
olegkkruglov: your comment is again in contradiction with @samir-nasibli 's one. as for me it does not really matter to be honest

olegkkruglov: but have to say that the test starts failing after the change.

samir-nasibli: > your comment is again in contradiction with @samir-nasibli 's one. as for me it does not really matter to be honest

Sorry, where is the contradiction? this number is really big for a test suit.

olegkkruglov: please read carefully

samir-nasibli: @olegkkruglov if your are refering this comment: https://github.com/intel/scikit-learn-intelex/pull/1644#discussion_r1448139817
It is a question, not a proposal change :)

icfaust:
I'm wondering if BasicStatistics and IncrementalBasicStatistics shouldn't be functions, possibly located in 'metrics'. @KulikovNikita @olegkkruglov @samir-nasibli

onedal/basic_statistics/basic_statistics.py
```diff
@@ -62,52 +59,38 @@ def _get_onedal_params(self, dtype=np.float32):
...
+    def __init__(self, result_options="all", algorithm="by_default"):
+        super().__init__(result_options, algorithm)
 
-    def _compute(self, data, weights, module, queue):
-        policy = self._get_policy(queue, data, weights)
+    def fit(self, data, sample_weight=None, queue=None):
+        policy = self._get_policy(queue, data, sample_weight)
 
         if not (data is None):
             data = np.asarray(data)
```
md-shafiul-alam, line 75: asarray() is not compatible with csr data. Even though this PR does not deal with sparse data, it will be better to use check_array function.

onedal/basic_statistics/basic_statistics.py
```diff
@@ -62,52 +59,38 @@ def _get_onedal_params(self, dtype=np.float32):
...
-        if not (weights is None):
-            weights = np.asarray(weights)
+        if not (sample_weight is None):
+            sample_weight = np.asarray(sample_weight)
```
md-shafiul-alam, line 77: scikit-learn has a check_sample_weight, might be a good option here

ethanglaser:
This would drop basic statistics compute completely and break API - I guess we would need to deprecate it first?

ethanglaser:
> > This would drop basic statistics compute completely and break API - I guess we would need to deprecate it first?
> 
> I would say that it is not breaking but more like fixing and making more consistent. As far as I understand, we have to use fit instead of compute in all estimators and compute was used here accidentally. But if we still need to use deprecation in such case then do we have any guidelines about how should it be done?

Agreed that it is fixing, but that doesn't mean its not breaking - anyone that is using our BasicStatistics API will be impacted by this change. @Alexsandruss how do we generally handle this? (ie should we add deprecation warning message but leave compute() and just call fit() from it initially or can we remove compute() immediately)

I would also expect that this may require an accompanying docs update, spmd interfaces, etc.

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -14,4 +14,130 @@
...
+        -------
+        self : object
+            Returns the instance itself.
+        """
+        dispatch(
+            self,
+            "fit",
+            {
+                "onedal": self.__class__._onedal_fit,
+                "sklearn": None,
```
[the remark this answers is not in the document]
olegkkruglov: it will fail. there are few such algos in our repo. probably it is better to resolve (add exception raising, I guess) in a separate pr. we have jira ticket about that.

onedal/basic_statistics/incremental_basic_statistics.py
```diff
@@ -110,11 +66,11 @@ class IncrementalBasicStatistics(BaseBasicStatistics):
 
     def __init__(self, result_options="all"):
         super().__init__(result_options, algorithm="by_default")
-        module = _backend.basic_statistics.compute
+        module = self._get_backend("basic_statistics")
```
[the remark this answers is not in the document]
olegkkruglov: it is used in the next row. or do you mean that it is better to clue these two rows in one? this style (with separate variable for module) had been chosen before I did my changes. but i can change here if you insist (probably will be better to change in all other occurrencies ).

samir-nasibli:
> I'm wondering if BasicStatistics and IncrementalBasicStatistics shouldn't be two functions, possibly located in 'metrics' which returns a dict. @KulikovNikita @olegkkruglov @samir-nasibli

I think this is good idea to discuss. In case if we will accept this, deprecation warnings also needed here.
As an option we can also add function for BS in metrics, just reusing original BS.
Adding just function for IncrementalBasicStatistics idea will be harder.

Personally I am good with current solution as well.

olegkkruglov:
> I would also expect that this may require an accompanying docs update, spmd interfaces, etc.

@ethanglaser deprecation is added, spmd part is updated, but I can't see any docs containing information about old interfaces which should be updated according to these changes. Am I missing something?

samir-nasibli:
@olegkkruglov please rebase you branch

onedal/basic_statistics/basic_statistics.py
```diff
@@ -63,59 +62,67 @@ def _get_onedal_params(self, is_csr, dtype=np.float32):
...
+            else:
+                setattr(self, opt, value)
 
-        return {k: from_table(v).ravel() for k, v in res.items()}
+        return self
+
+    def compute(self, data, weights=None, queue=None):
+        warnings.warn(
+            "Method `compute` was deprecated in version 2024.3 and will be "
+            "removed in 2024.5. Use `fit` instead."
```
samir-nasibli, line 102: Please update in which exact version it will be removed.

onedal/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -14,155 +14,222 @@
...
+    ("standard_deviation", expected_standard_deviation, (2e-3, 2e-3)),
+    ("second_order_raw_moment", expected_second_order_raw_moment, (1e-6, 1e-7)),
+]
 
-        data = data.astype(dtype=dtype)
+options_and_tests_csr = [
+    ("sum", "sum", (5e-6, 1e-9)),
+    ("min", "min", (0, 0)),
+    # There is a bug in oneDAL's max computations on GPU
+    #         ("max", "max", (0, 0)),
```
samir-nasibli, line 87: please better use pytest.mark.skip with the reason defined.

onedal/spmd/basic_statistics/basic_statistics.py
```diff
@@ -23,4 +25,13 @@
 class BasicStatistics(BaseEstimatorSPMD, BasicStatistics_Batch):
     @support_usm_ndarray()
     def compute(self, data, weights=None, queue=None):
+        warnings.warn(
+            "Method `compute` was deprecated in version 2024.3 and will be "
+            "removed in 2024.5. Use `fit` instead."
```
samir-nasibli, line 30: please update in which exact version it will be removed

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -14,4 +14,130 @@
 # limitations under the License.
 # ==============================================================================
 
-from onedal.basic_statistics import BasicStatistics
+import numpy as np
+from sklearn.base import BaseEstimator
+from sklearn.utils import check_array, deprecated
```
samir-nasibli, line 19: please use onedal.utils._validation _check_array instead
https://github.com/intel/scikit-learn-intelex/blob/01df2bb15360481441736882d9847d9adf24c811/onedal/utils/validation.py#L125

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -14,4 +14,130 @@
...
+        onedal_params = {
+            "result_options": self.options,
+        }
+
+        if not hasattr(self, "_onedal_estimator"):
+            self._onedal_estimator = self._onedal_basic_statistics(**onedal_params)
+        self._onedal_estimator.fit(X, sample_weight, queue)
+        self._save_attributes()
+
+    def compute(self, data, weights=None, queue=None):
```
samir-nasibli, line 110: Probably we need deprecation warning here?

sklearnex/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -0,0 +1,251 @@
...
+
+    assert_allclose(expected_mean, result.mean)
+    assert_allclose(expected_max, result.max)
+    assert_allclose(expected_min, result.min)
+
+    result = BasicStatistics().fit(X_df, sample_weight=weights_df)
+
+    expected_weighted_mean = np.array([0.25, 0.25])
+    expected_weighted_min = np.array([0, 0])
+    expected_weighted_max = np.array([0.5, 0.5])
```
samir-nasibli, line 56: Am I right, even dpctl/dpnp given, results are numpy arrays?

samir-nasibli: I expect that we have full interobility.
If dpnp/dpctl given, the same dataframe type with the same sycl context are expected as results

samir-nasibli (commented):
Please also rebase your branch.

onedal/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -14,155 +14,222 @@
 # limitations under the License.
 # ==============================================================================
 
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
```
Vika-F, line 19: Maybe I am missing something but it looks like CSR data testing was deleted from this file in this PR.
Need to test scipy.csr_array inputs as well, not only Numpy.

Vika-F (changes requested):
Please add scipy.csr_array inputs testing on CPU and GPU.

onedal/basic_statistics/tests/test_basic_statistics.py
olegkkruglov: looks like rebase issue, must be fixed now. thanks for pointing out.

sklearnex/basic_statistics/basic_statistics.py
olegkkruglov: currently check_array is used much more often than _check_array in sklearnex folder. why should it be changed here? probably the best way is to do follow-up PR with standardization of the usage across the whole repo.

Vika-F (approved):
The changes look good to me now. Approving in case the internal CI is also good.

samir-nasibli (dismissed):
Expecting other reviewers approvals. Currently blocking merge.

onedal/basic_statistics/basic_statistics.py
```diff
@@ -63,59 +62,67 @@ def _get_onedal_params(self, is_csr, dtype=np.float32):
...
 
         is_csr = _is_csr(data)
-        if not (data is None) and not is_csr:
-            data = np.asarray(data)
-
-        if not (weights is None):
-            weights = np.asarray(weights)
 
-        data, weights = _convert_to_supported(policy, data, weights)
+        if not (data is None) and not is_csr:
```
ethanglaser, line 79:
```suggestion
        if data is not None and not is_csr:
```
Generally this would be preferrable for python, there are a few places where it can be changed

sklearnex/basic_statistics/tests/test_basic_statistics.py
samir-nasibli: I see that this common question for some other estimators as well. Make sense to work on this separately. Covered by ticket.

onedal/basic_statistics/basic_statistics.py
samir-nasibli, line 1: Comment for the `_get_result_options`:
Inputs should be validated. For example invalid option name or empty string input.
This cases should be covered by tests.

I think make sense updated in this PR or in other smaller one either.

onedal/basic_statistics/basic_statistics.py
samir-nasibli: Also please cover this with test for checking this expected warning for the certain daal4py version. 
This will help us remove the warning from latest versions.

onedal/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -14,87 +14,181 @@
...
+    ("sum_squares_centered", expected_sum_squares_centered, (2e-4, 1e-7)),
+    ("standard_deviation", expected_standard_deviation, (2e-3, 2e-3)),
+    ("second_order_raw_moment", expected_second_order_raw_moment, (1e-6, 1e-7)),
+]
+
+
+@pytest.mark.parametrize("queue", get_queues())
+@pytest.mark.parametrize("option", options_and_tests)
+@pytest.mark.parametrize("row_count", [100, 1000])
+@pytest.mark.parametrize("column_count", [10, 100])
```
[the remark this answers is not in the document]
samir-nasibli: Is there any updates for this question?
If this not supported please make sure that this kind of inputs are validated

onedal/spmd/basic_statistics/basic_statistics.py
samir-nasibli: Based on the answer https://github.com/intel/scikit-learn-intelex/pull/1644#discussion_r1652781106 make sense to remove warning from here.

onedal/basic_statistics/basic_statistics.py
olegkkruglov: I've never seen it was done anywhere in our repo

onedal/basic_statistics/tests/test_basic_statistics.py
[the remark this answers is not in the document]
olegkkruglov: you can see it in the tests

onedal/cluster/kmeans.py
```diff
@@ -90,8 +90,7 @@ def _tolerance(self, rtol, X_table, policy, dtype=np.float32):
         is_sparse = False
         dummy = to_table(None)
         bs = self._get_basic_statistics_backend("variance")
-
-        res = bs.compute_raw(X_table, dummy, policy, dtype, is_sparse)
+        res = bs._compute_raw(X_table, dummy, policy, dtype, is_sparse)
```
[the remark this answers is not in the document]
olegkkruglov: I would say that it is more important to highlight that the method is not for user than strictly follow the rule that it should be called only inside the class. btw the rule is not stated in PEP8. it says that leading underscore is weak “internal use” indicator. the method is definitely for internal use

onedal/basic_statistics/tests/test_basic_statistics.py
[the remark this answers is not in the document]
olegkkruglov: tests are extended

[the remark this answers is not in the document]
olegkkruglov: if you mean the reason of removing if statement then it is removed because the check is duplicated in setup.py

onedal/cluster/kmeans.py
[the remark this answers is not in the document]
ethanglaser: Fair enough

onedal/basic_statistics/tests/test_basic_statistics.py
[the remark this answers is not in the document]
ethanglaser: its just that the file has a very large diff - was wondering if it was mostly just removal of indents from removing if statement (in which case could review less thoroughly) or if there were many changes to the actual tests

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -14,4 +14,130 @@
...
+                setattr(self, option, getattr(self._onedal_estimator, option))
+
+    def _onedal_supported(self, method_name, *data):
+        patching_status = PatchingConditionsChain(
+            f"sklearnex.basic_statistics.{self.__class__.__name__}.{method_name}"
+        )
+        return patching_status
+
+    _onedal_cpu_supported = _onedal_supported
+    _onedal_gpu_supported = _onedal_supported
```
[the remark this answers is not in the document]
olegkkruglov: yes. it was added for adding this class to patch map

onedal/basic_statistics/tests/test_basic_statistics.py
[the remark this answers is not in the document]
samir-nasibli: > if you mean the reason of removing the "if" statement then it is removed because the check is duplicated in setup.py

This check is duplicated for the basic_statistic.py file as well please return the check for consistency.

onedal/basic_statistics/tests/test_basic_statistics.py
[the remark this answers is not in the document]
samir-nasibli: Could you please bring more details what exactly do you mean?

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -14,4 +14,130 @@
...
+
+    def _onedal_fit(self, X, sample_weight=None, queue=None):
+        if sklearn_check_version("1.0"):
+            X = self._validate_data(X, dtype=[np.float64, np.float32], ensure_2d=False)
+        else:
+            X = check_array(X, dtype=[np.float64, np.float32], ensure_2d=False)
```
samir-nasibli, line 96:
```suggestion
        X = self._validate_data(X, dtype=[np.float64, np.float32], ensure_2d=False)
```
We are not targeted to support sklearn < 1.0

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -14,4 +14,130 @@
...
+
+        if sample_weight is not None:
+            sample_weight = _check_sample_weight(sample_weight, X)
```
samir-nasibli, line 99: Good point to highlight interesting aspect of scikit-learn-intelex.
sklearnex module is mostly for sklearn alignment and dispatching, the rest I would like to move on onedal4py side, since onedal4py interface is used for spmd ifaces, and I would like this iface to be being good validated for the inputs.

Since no basic statistics on sklearn side, it make sense to move all validation from line 93-99 to onedal4py side. Just make sure to remove `sklearn_version_check`.

samir-nasibli (dismissed):
@olegkkruglov Nice work done! Thank you!
Please address some minor comments. We are good to integrate it soon.

onedal/basic_statistics/tests/test_basic_statistics.py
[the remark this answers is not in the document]
olegkkruglov: i didn't get what do you mean. could you clarify please?

sklearnex/basic_statistics/basic_statistics.py
olegkkruglov: then I guess we need to remove it across the whole repo. I'll create the ticket.

onedal/basic_statistics/basic_statistics.py
```diff
@@ -63,59 +62,67 @@ def _get_onedal_params(self, is_csr, dtype=np.float32):
...
+            else:
+                setattr(self, opt, value)
 
-        return {k: from_table(v).ravel() for k, v in res.items()}
+        return self
+
+    def compute(self, data, weights=None, queue=None):
+        warnings.warn(
+            "Method `compute` was deprecated in version 2024.6 and will be "
+            "removed in 2026.0. Use `fit` instead."
```
samir-nasibli, line 102:
```suggestion
            "Method `compute` was deprecated in version 2024.7 and will be "
            "removed in 2026.0. Use `fit` instead."
```

sklearnex/basic_statistics/basic_statistics.py
samir-nasibli: Thank you! Addressed https://github.com/intel/scikit-learn-intelex/pull/1906

onedal/basic_statistics/tests/test_basic_statistics.py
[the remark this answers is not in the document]
samir-nasibli: Sorry, I misunderstood the changes to the file. This comment of mine can be ignored. I mark as resolved

onedal/basic_statistics/basic_statistics.py
samir-nasibli: Please don't take this comment as a nitpick. Let's try to follow secure code development.

In scikit-learn-intelex mostly parameters are validated by sklearn primitives. BS has specific things to check. Worth to add some validation this string input.

onedal/basic_statistics/basic_statistics.py
```diff
@@ -63,59 +62,67 @@ def _get_onedal_params(self, is_csr, dtype=np.float32):
...
+                setattr(self, opt, value[0])
+            else:
+                setattr(self, opt, value)
 
-        return {k: from_table(v).ravel() for k, v in res.items()}
+        return self
+
+    def compute(self, data, weights=None, queue=None):
+        warnings.warn(
+            "Method `compute` was deprecated in version 2024.6 and will be "
```
samir-nasibli, line 101:
```suggestion
            "Method `compute` was deprecated in version 2024.7 and will be "
```

samir-nasibli:
@olegkkruglov please rebase your branch and run internal CI as well.
Dear reviewers, expecting your approvals. If the comments are addressed please mark them as resolved.

onedal/basic_statistics/basic_statistics.py
samir-nasibli: By the way it make sense mention this deprecation here https://github.com/intel/scikit-learn-intelex/blob/main/doc/sources/deprecation.rst
Could be done in the separate PR.

samir-nasibli:
@olegkkruglov by the way I see that there is not information about online algos in our docs, that should be addressed.

This is not the scope of the current PR, just wanted to highlight.

onedal/basic_statistics/basic_statistics.py
olegkkruglov: the problem still exists not only here but for all algos without sklearn analogues and then goes beyond this PR scope I guess. I'll create ticket for that

samir-nasibli: Even if there are problems, I would like us to start this. This PR is a good starting point for project changes. Please add a minimum check for results_option, since API of BS is updated here.
Use with pytest.raise(ValueError) to check error raise on empty string input for results_option

Minimal validation for the input should be covered.

PR #1645, opened 2024-01-10
MAINT: leaving hashbang line where it is necessary

samir-nasibli:
# Description
leaving hashbang line where it is necessary


ethanglaser:
Where is it necessary? CI and build scripts only?

samir-nasibli:
> Where is it necessary? CI and build scripts only?

Probably, yes. But for other scripts definitely not required.

ethanglaser (approved):

PR #1650, opened 2024-01-16
feat: add model_type property to GBTDAALBaseModel

ahuber21:
# Description

* Puts the `_is_regression` check into the base class and makes it a property.
* Adds a new attribute `model_type`, similar to the [SHAP python package](https://github.com/shap/shap/blob/7bd33ff5dd5b7571cdec2dc097c7dc9facbf113b/shap/explainers/pytree.py). This attribute can be used if a model builder feature supports only certain model types

### DIFF
--- a/daal4py/mb/model_builders.py
+++ b/daal4py/mb/model_builders.py
@@ -16,6 +16,8 @@
 
 # daal4py Model builders API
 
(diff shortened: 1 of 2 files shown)


razdoburdin (approved):
nice :)

PR #1652, opened 2024-01-17
[bug,enh] add exception handling to daal4py LogisticRegression.fit monkeypatch

icfaust:
Initial fix for daal4py's LogisticRegression fit, which uses a monkeypatch design pattern to calculate the fit. This currently doesn't have a reversion capability in the case of an error. This means any exception will not properly return sklearn to the original state. Secondly, this method will be used by the new LogisticRegression (#1649) but without inheritance, requiring moving from a super statement to something more explicit such that it can be used without inheritance.

- Add a try finally statement to guarantee that monkeypatch is reverted before returning
- switch from super().fit() to LogisticRegression_original.fit(self...) for use by non-inherited classes

Tasks
- [x] implement
- [x] test
- [x] benchmark

### DIFF
--- a/daal4py/sklearn/linear_model/logistic_path.py
+++ b/daal4py/sklearn/linear_model/logistic_path.py
@@ -1034,9 +1034,10 @@ if sklearn_check_version("0.24"):
             setattr(which, what, replacer)
             try:
                 clf = LogisticRegression_original.fit(self, X, y, sample_weight)
-            except Exception as e:
+            except Exception as error:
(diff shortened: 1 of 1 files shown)


avolkov-intel (approved):

icfaust:
private CI run http://intel-ci.intel.com/eeb5c99c-4119-f1d6-8ddd-a4bf010d0e2e

samir-nasibli (approved):
Looks good to me.
internal ci fails not related with this changes.
Many thanks!

PR #1656, opened 2024-01-18
IncrementalLinearRegression interface

olegkkruglov:
- IncrementalLinearRegression added to onedal/sklearnex
- Contains changes from https://github.com/intel/scikit-learn-intelex/pull/1713


onedal/linear_model/linear_model.py
```diff
@@ -198,8 +161,43 @@ def __init__(
...
+        self.n_features_in_ = _num_features(X_loc, fallback_1d=True)
+
+        X_loc, y_loc = _convert_to_supported(policy, X_loc, y_loc)
+        params = self._get_onedal_params(get_dtype(X_loc))
+        X_table, y_table = to_table(X_loc, y_loc)
+
+        hparams = get_hyperparameters("linear_regression", "train")
+        if hparams is not None and not hparams.is_default:
+            result = module.train(policy, params, hparams.backend, X_table, y_table)
+        else:
```
samir-nasibli, line 191: How this linear regression refactoring/updating related with incremental impl?
Each changes on previous estimator shoudl be done on separate branch with separate validation , perf validation

samir-nasibli (changes requested):
refactoring/update of LinearRegression should be patched in separate branch/PR.

onedal/linear_model/linear_model.py
olegkkruglov: The methods in BaseLinearRegression class should be intersection of methods used in LinearRegression and IncrementalLinearRegression. Otherwise, this class is redundant. API for the estimator remains the same, the code also remains the same, methods just moved from one class to another. I don't think it is worth to open new PR here

sklearnex/linear_model/tests/test_incremental_linear.py
```diff
@@ -0,0 +1,61 @@
...
+from onedal.tests.utils._dataframes_support import (
+    _as_numpy,
+    _convert_to_dataframe,
+    get_dataframes_and_queues,
+)
+
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+def test_sklearnex_import_linear(dataframe, queue):
+    from sklearnex.linear_model import IncrementalLinearRegression
```
Alexandr-Solovev, line 32: please add usage example of fit

Alexandr-Solovev (commented):
in general looks good to me, but please add hyperparameters, its already exists in oneDAL

onedal/linear_model/linear_model.py
samir-nasibli: Thank you for clarification!
In any way this will affect on other LinearRegression estimator that is already on main branch.
I worth to make this common changes first in seperate branch, validate it via tests, perf and then merge.
After this increment better to add IncrementalLR

olegkkruglov: My point is that there is no affect on executed code, so, perf is not expected to be changed, tests for the estimator are still green, and considering our capacity for review PRs it might be worth to leave it as is. Moreover, the changes are necessary to introduce IncrementalLinearRegression estimator, that's why it makes sense to keep it in this PR as well.

onedal/linear_model/linear_model.py
```diff
@@ -198,8 +161,43 @@ def __init__(
         super().__init__(fit_intercept=fit_intercept, copy_X=copy_X, algorithm=algorithm)
 
     def fit(self, X, y, queue=None):
-        return super()._fit(X, y, _backend.linear_model.regression, queue)
```
samir-nasibli, line 201: please leave it as is here, do not move _fit impl from base class

onedal/linear_model/linear_model.py
samir-nasibli: I don't understand why are you that sure. The only thing that we have to base is numbers and tests validations.
Doing this changes incrementally is better for review and integration and safety

onedal/linear_model/linear_model.py
olegkkruglov: base class should contain only methods which are used in both LinearRegression and IncrementalLinearRegression. in my understanding it is designed exactly for this purpose. _fit method is not of this type. so, _fit method should be moved to LinearRegression class. LinearRegression class already contains method fit which only calls _fit method and nothing else and _fit method is not called anywhere else. in this situation I am completely sure that _fit method is redundant and its content can be safely moved to fit method to increase code readability.

onedal/linear_model/linear_model.py
olegkkruglov: Test validations are here. About perf validations, after the changes the interpreter goes through exactly the same rows and instructions, so, using common sense one can conclude that perf must remain exactly the same.

samir-nasibli: I see a lot of common code for fit and partial_fit, so worth to have common fit for this.
If you think it will more time to impl, then you can consider to leave.

onedal/linear_model/linear_model.py
```diff
@@ -198,8 +161,43 @@ def __init__(
...
+        module = _backend.linear_model.regression
 
-    def predict(self, X, queue=None):
-        y = super()._predict(X, _backend.linear_model.regression, queue)
-        return y
+        policy = self._get_policy(queue, X, y)
+
+        X_loc, y_loc = X, y
+        if not isinstance(X, np.ndarray):
+            X_loc = np.asarray(X)
```
icfaust, line 170: X_loc looks to be unnecessary.
```suggestion
        X = np.asanarray(X)
```
This will guarantee that it should have a dtype attribute.

onedal/linear_model/linear_model.py
```diff
@@ -198,8 +161,43 @@ def __init__(
...
+
+        dtype = get_dtype(X_loc)
+        if dtype not in [np.float32, np.float64]:
+            dtype = np.float64
+            X_loc = X_loc.astype(dtype, copy=self.copy_X)
```
icfaust, line 175:
```suggestion
        if X.dtype not in [np.float64, np.float32]:
            X = X.astype(np.float64, copy=self.copy_X)
```

onedal/linear_model/linear_model.py
```diff
@@ -198,8 +161,43 @@ def __init__(
...
+            result = module.train(policy, params, hparams.backend, X_table, y_table)
+        else:
+            result = module.train(policy, params, X_table, y_table)
+
+        self._onedal_model = result.model
+
+        coeff = from_table(result.model.packed_coefficients)
+        self.coef_, self.intercept_ = coeff[:, 1:], coeff[:, 0]
+
+        if self.coef_.shape[0] == 1 and y_loc.ndim == 1:
```
icfaust, line 199: I feel like there is a more elegant solution for this

icfaust (commented):
Please give a more in-depth description. This looks like something that doesn't exist in sklearn (i.e. no patch).  Could you give some more info on this?  I see some issues with the coding on sklearnex, the comments also can be duplicated for the onedal code (since they are closely related.

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,177 @@
...
+        intercept_ : float or array of shape (n_targets,)
+            Independent term in the linear model. Set to 0.0 if
+            `fit_intercept = False`.
+
+        n_features_in_ : int
+            Number of features seen during :term:`fit`.
+
+        """
+
+        def __init__(self, fit_intercept=True, copy_X=True, n_jobs=None, batch_size=None):
```
Alexsandruss, line 69: What is the purpose of `batch_size`? It can be deducted from input data.

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,177 @@
...
+            self.fit_intercept = fit_intercept
+            self.copy_X = copy_X
+            self.n_jobs = n_jobs
+            self._need_to_finalize = False
+            self.batch_size = batch_size
+
+        _onedal_incremental_linear = staticmethod(onedal_IncrementalLinearRegression)
+
+        @run_with_n_jobs
```
Alexsandruss, line 78: `run_with_n_jobs` is not expected to be used manually.

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,177 @@
...
+    from sklearn.utils import check_array, gen_batches
+
+    from daal4py.sklearn._utils import control_n_jobs, run_with_n_jobs
+    from onedal._device_offload import support_usm_ndarray
+    from onedal.linear_model import (
+        IncrementalLinearRegression as onedal_IncrementalLinearRegression,
+    )
+
+    @control_n_jobs
+    class IncrementalLinearRegression:
```
Alexsandruss, line 29: Where is predict for this class?

Alexsandruss, line 29: What is the need for separate (from usual LinearRegression) class?

Alexsandruss (changes requested):

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,177 @@
...
+    import numpy as np
+    from sklearn.utils import check_array, gen_batches
+
+    from daal4py.sklearn._utils import control_n_jobs, run_with_n_jobs
+    from onedal._device_offload import support_usm_ndarray
+    from onedal.linear_model import (
+        IncrementalLinearRegression as onedal_IncrementalLinearRegression,
+    )
+
+    @control_n_jobs
```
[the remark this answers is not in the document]
Alexsandruss: @control_n_jobs(decorated_methods=["fit", "partial_fit", "predict"])

onedal/linear_model/tests/test_incremental_linear_regression.py
```diff
@@ -0,0 +1,147 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from daal4py.sklearn._utils import daal_check_version, sklearn_check_version
+
+if daal_check_version((2023, "P", 100)):
```
samir-nasibli, line 19: No need for adding this version checker here.
Having this test files in a module already managed on setup file.

sklearnex/linear_model/incremental_linear.py
olegkkruglov: it is described in docstring few lines above. this is copy from sklearn.IncrementalPCA interface

onedal/linear_model/linear_model.py
olegkkruglov: it is not my code. I just rearranged lines Nikita used. I agree that it is unnecessary but since this PR is not related to LinearRegression class I tried to do minimal changes to this class. Probably it makes sense to refactor LinearRegression later

onedal/linear_model/linear_model.py
olegkkruglov: https://github.com/intel/scikit-learn-intelex/pull/1656#discussion_r1467600543

onedal/linear_model/linear_model.py
olegkkruglov: https://github.com/intel/scikit-learn-intelex/pull/1656#discussion_r1467600543

sklearnex/linear_model/incremental_linear.py
olegkkruglov: to align with sklearn where PCA and IncrementalPCA are different classes

onedal/linear_model/linear_model.py
```diff
@@ -198,8 +161,43 @@ def __init__(
...
+        X_loc, y_loc = X, y
+        if not isinstance(X, np.ndarray):
+            X_loc = np.asarray(X)
+
+        dtype = get_dtype(X_loc)
+        if dtype not in [np.float32, np.float64]:
+            dtype = np.float64
+            X_loc = X_loc.astype(dtype, copy=self.copy_X)
+
+        y_loc = np.asarray(y_loc).astype(dtype=dtype)
```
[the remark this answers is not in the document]
olegkkruglov: https://github.com/intel/scikit-learn-intelex/pull/1656#discussion_r1467600543

sklearnex/linear_model/incremental_linear.py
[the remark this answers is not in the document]
olegkkruglov: i would say that it should be @control_n_jobs(decorated_methods=["fit", "partial_fit", "predict", "_onedal_finalize_fit"])

onedal/linear_model/linear_model.py
icfaust: I think @samir-nasibli is correct then in this case in https://github.com/intel/scikit-learn-intelex/pull/1656#pullrequestreview-1837303634 . By moving things around you are refactoring the code, you can do it separately in a different PR, or you take the full responsibility in this one for both tasks.

olegkkruglov: i don't really think it is the same. in my current changes i kept all variables and external methods calls and their call order the same. i just removed from BaseLinearRegressionClass methods which are not necessary in IncrementalLinearRegression and put it to LinearRegression class. more serious code changes probably really have to be perf-measured.

samir-nasibli:
@olegkkruglov please rebase your branch.

It make sense think first about general online algos test coverage.
Especially that estimators that we are not patching. I expect, that we will probably add some sklearn like tests.
Some basic pikling, n_jobs, testing via different dataframes and device contexts.

Generally it is required first update primitives in case of new online algos. For example, this IncrementalLinReg will not work with `config_context` for device offloading. etc.

sklearnex/dispatcher.py
```diff
@@ -255,6 +258,18 @@ def get_patch_map():
...
+            [
+                (
+                    linear_model_module,
+                    "IncrementalLinearRegression",
+                    IncrementalLinearRegression_sklearnex,
+                ),
+                None,
+            ]
+        ]
+
```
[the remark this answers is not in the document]
icfaust: I found the mistake as forgetting a None in a getattr call. I have pushed to this repo @olegkkruglov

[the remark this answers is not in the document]
icfaust: The patch map addition will provide useful testing like pickle-ability (via sklearn's test_common.py) and memory leakage testing via sklearnex's test_memory_usage.py. I will make a quick PR to solve the issue once the latest public CI run goes through.

icfaust:
So far I am pretty impressed by this PR.

icfaust:
Ideally this should inherit sklearns' RegressorMixin too, just like BaseEstimator.

icfaust:
New PR introduced to manage the issues with test_patching and _utils.py.  We need to talk to @Alexsandruss about n_jobs testing and how that interacts with partial_fit

onedal/linear_model/linear_model.py
```diff
@@ -14,28 +14,31 @@
...
-from ..common._base import BaseEstimator as onedal_BaseEstimator
 from ..common._estimator_checks import _check_is_fitted
-from ..common._mixin import RegressorMixin
+from ..common._policy import _get_policy
 from ..common.hyperparameters import get_hyperparameters
 from ..datatypes import _convert_to_supported, from_table, to_table
 from ..utils import _check_array, _check_n_features, _check_X_y, _num_features
 
 
-class BaseLinearRegression(onedal_BaseEstimator, metaclass=ABCMeta):
```
samir-nasibli, line 32: Please use `onedal_BaseEstimator`

samir-nasibli, line 32: Please use `onedal_BaseEstimator`

samir-nasibli (dismissed):

ethanglaser:
conformance internal CI fails:
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_estimators_pickle]
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_fit2d_1sample]
FAILED tests/test_common.py::test_estimators[LinearRegression()-check_estimators_unfitted]

onedal/linear_model/linear_model.py
icfaust: Ideally we should add a test which verifies via issubclass() for all newly added classes in the onedal directory.

icfaust:
Sorry, accidental button mash

sklearnex/linear_model/tests/test_incremental_linear.py
```diff
@@ -0,0 +1,202 @@
...
+        inclin.partial_fit(X_split_df, y_split_df)
+
+    assert inclin.n_features_in_ == 2
+    assert_allclose(inclin.coef_, [1.0, 2.0], rtol=1e-5, atol=1e-5)
+    if fit_intercept:
+        assert_allclose(inclin.intercept_, 3.0, rtol=1e-5, atol=1e-5)
+
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+    y_pred = inclin.predict(X_df)
+    # TODO Implement more clear way to transfer dpctl/dpnp arrays to host
```
icfaust, line 136: I am puzzled why this isn't working out of the box... Message me about this

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,369 @@
...
+@register_hyperparameters(
+    {
+        "fit": get_hyperparameters("linear_regression", "train"),
+        "partial_fit": get_hyperparameters("linear_regression", "train"),
+    }
+)
+@control_n_jobs(
+    decorated_methods=["fit", "partial_fit", "predict", "_onedal_finalize_fit"]
+)
+class IncrementalLinearRegression(BaseEstimator):
```
icfaust, line 48:
```suggestion
class IncrementalLinearRegression(RegressorMixin, BaseEstimator):
```

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,369 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===============================================================================
+
+import numbers
+import warnings
+
+import numpy as np
+from sklearn.base import BaseEstimator
```
icfaust, line 21:
```suggestion
from sklearn.base import BaseEstimator, RegressorMixin
```

onedal/linear_model/tests/test_incremental_linear_regression.py
```diff
@@ -0,0 +1,151 @@
...
+    X_split = np.array_split(X, num_blocks)
+    y_split = np.array_split(y, num_blocks)
+
+    model = IncrementalLinearRegression(fit_intercept=False)
+    for i in range(num_blocks):
+        model.partial_fit(X_split[i], y_split[i], queue=queue)
+    model.finalize_fit()
+
+    if queue.sycl_device.is_gpu:
+        tol = 3e-3 if model.coef_.dtype == np.float32 else 1e-7
```
icfaust, line 116: why is the error here so high for fp32 vs fp64? Thats 3.5 orders of magnitude...

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,369 @@
...
+
+        n_samples, n_features = X.shape
+        if self.batch_size is None:
+            self.batch_size_ = 5 * n_features
+        else:
+            self.batch_size_ = self.batch_size
+
+        self.n_samples_seen_ = 0
+        if hasattr(self, "_onedal_estimator"):
+            self._onedal_estimator._reset()
```
icfaust, line 224: how well does this wipe the memory on oneDAL side? does it start from scratch?

sklearnex/tests/test_memory_usage.py
```diff
@@ -94,7 +94,10 @@ def remove_duplicated_estimators(estimators_list):
     return estimators_map.values()
 
 
-BANNED_ESTIMATORS = ("TSNE",)  # too slow for using in testing on common data size
+BANNED_ESTIMATORS = (
+    "IncrementalLinearRegression",  # TODO fix memory leak issue
```
icfaust, line 98: I am very happy to see the comment here. As we discussed, make sure to send me the ticket.

sklearnex/linear_model/tests/test_incremental_linear.py
```diff
@@ -0,0 +1,202 @@
...
+
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
+from onedal.tests.utils._dataframes_support import (
+    _convert_to_dataframe,
+    get_dataframes_and_queues,
+)
+from sklearnex._device_offload import _transfer_to_host
```
icfaust, line 25:
```suggestion
import sklearn.linear_model.tests.test_base
from sklearnex._device_offload import _transfer_to_host
```
I want to see some of sklearn's linear regression tests be applied to IncrementalLinearRegression via pytest monkeypatch. There are are lot of tests here, but I want to look specifically at mathematical verification.  Tests like test_linear_regression, test_fit_intercept, test_linear_regression_multiple_outcome

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,369 @@
...
+    Incremental estimator for linear regression.
+    Allows to train linear regression if data are splitted into batches.
+
+    Parameters
+    ----------
+    fit_intercept : bool, default=True
+    Whether to calculate the intercept for this model. If set
+    to False, no intercept will be used in calculations
+    (i.e. data is expected to be centered).
+
```
icfaust, line 59: Any idea if sample_weights, positive keyword will be supported?

icfaust (commented):
Just make sure as well to address the test_common.py failures in private CI as well. This is a large and impressive PR. Please message me with the tickets about pickling and memory leaks. @samir-nasibli @Alexsandruss I'd like these changes addressed before approving.

sklearnex/linear_model/tests/test_incremental_linear.py
olegkkruglov: all of them are not supported by c++ onedal side since there are not enough samples in their datasets

onedal/linear_model/tests/test_incremental_linear_regression.py
olegkkruglov: i just copied that from linear_regression_test in the main and i guess here the tolerance should be the same. i don't think resolving such issues is in the scope of this PR. i can add TODO comment if you want

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,369 @@
...
+            Samples.
+        Returns
+        -------
+        C : array, shape (n_samples, n_targets)
+            Returns predicted values.
+        """
+
+        return dispatch(
+            self,
+            "fit",
```
icfaust, line 363: For clearer debugging, change "fit" to "predict"

sklearnex/linear_model/incremental_linear.py
olegkkruglov: I didn't think about it. I guess python garbage collector should resolve it somehow. this method I just creates new empty partial_result and put it instead the existing to class attribute

sklearnex/linear_model/tests/test_incremental_linear.py
icfaust: Good to know, is this stated or checked anywhere on the python-side? I can't seem to find it. What is the value of the sample limit?

olegkkruglov: https://github.com/olegkkruglov/scikit-learn-intelex/blob/d5233d8f71b6e827079f4b5e7fec994c6ffe2b74/sklearnex/linear_model/incremental_linear.py#L194

olegkkruglov: and what do you mean by mathematical verification btw? i hope here
https://github.com/olegkkruglov/scikit-learn-intelex/blob/inc-lin/sklearnex/linear_model/tests/test_incremental_linear.py
and here
https://github.com/olegkkruglov/scikit-learn-intelex/blob/inc-lin/onedal/linear_model/tests/test_incremental_linear_regression.py
must be enough tests to check correctness of the algorithm

sklearnex/linear_model/tests/test_incremental_linear.py
olegkkruglov: in this comment I mean that it would be good to have transfer_to_host method which doesn't return any redundant data and just converts dpctl to numpy array

sklearnex/linear_model/incremental_linear.py
olegkkruglov: they are not supported on onedal side, so I don't think they will be in this pr

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,369 @@
...
+                X,
+                dtype=[np.float64, np.float32],
+                copy=self.copy_X,
+                force_all_finite=False,
+            )
+
+        assert hasattr(self, "_onedal_estimator")
+        if self._need_to_finalize:
+            self._onedal_finalize_fit()
+        X = check_array(X, dtype=[np.float64, np.float32])
```
icfaust, line 146: Why is this here again?

sklearnex/linear_model/tests/test_incremental_linear.py
icfaust: @olegkkruglov is the "is_good_for_onedal" a carryover from the onedal LinearRegression implementation, or could you explain to me the limitations on onedal side for the algorithm to necessitate it? From what I understand, it was necessary in LinearRegression due to the use of QR decomposition , but what about in IncrementalLinearRegression?

olegkkruglov: it is exactly the same algorithm with the same limitations

onedal/linear_model/tests/test_incremental_linear_regression.py
icfaust: I agree. I would say a ticket in the backlog is better than a TODO in this circumstance. Just shoot me the ticket and I would call this good.

ethanglaser:
Internal CI conformance fails including but not limited to:
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_estimators_nan_inf]
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_regressors_train]
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_regressors_train(readonly_memmap=True)]
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_regressors_train(readonly_memmap=True,X_dtype=float32)]
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_regressor_data_not_an_array]
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_supervised_y_2d]
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_estimators_unfitted]
FAILED tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_fit1d]
FAILED tests/test_common.py::test_check_param_validation[IncrementalLinearRegression()]

deselected_tests.yaml
```diff
@@ -1182,6 +1190,14 @@ gpu:
...
   - decomposition/tests/test_pca.py::test_pca_n_components_mostly_explained_variance_ratio
 
+  # Deselected tests for incremental algorithms
+  # Need to rework getting policy to correctly obtain it for method without data (finalize_fit)
+  # and avoid keeping it in class attribute, also need to investigate how to implement
+  # partial result serialization
+  - tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_estimators_pickle]
+  # There are not enough data to run onedal backend
+  - tests/test_common.py::test_estimators[IncrementalLinearRegression()-check_fit2d_1sample]
+
```
ethanglaser, line 1200: can remove these. if they are deselected in main section they will automatically be deselected for gpu

onedal/linear_model/incremental_linear_model.py
ethanglaser, line 1: incremental_linear_regression.py would probably make more sense

olegkkruglov: naming was chosen to be consistent with onedal/linear_model/linear_model.py

deselected_tests.yaml
```diff
@@ -1182,6 +1191,7 @@ gpu:
   # originated with pca dpctl/dpnp fit, to be re-assesed with pca out-of-preview
   - decomposition/tests/test_pca.py::test_pca_n_components_mostly_explained_variance_ratio
 
+
```
samir-nasibli, line 1194:
```suggestion
```

onedal/linear_model/incremental_linear_model.py
ethanglaser: yeah, good for now

onedal/linear_model/linear_model.py
```diff
@@ -193,12 +166,66 @@ def __init__(
...
+
+        Returns
+        -------
+        self : object
+            Fitted Estimator.
+        """
+        module = self._get_backend("linear_model", "regression")
+
+        if X.shape[1] + int(self.fit_intercept) > X.shape[0]:
+            raise TypeError("Not enough samples to fit")
```
ethanglaser, line 190: Leads to fail of previously successful linear regression spmd example. This condition should be only raised in cases where it will certainly fail (ie only incremental?)

olegkkruglov: Our C++ implementation is unstable if number of samples does not satisfy this condition and it was even stated before in the code, but only in sklearnex part
https://github.com/intel/scikit-learn-intelex/blob/263b9c17475314689498539b45b3d63c16454e09/sklearnex/linear_model/linear.py#L244
so I decided that it is better to add this check here as well to make this limitation more clear until it will be well investigated. JIRA task about investigation has been created

olegkkruglov:
https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/eedd31b2-8f0f-f1ce-bd49-a4bf010d0e2e

examples/sklearnex/incremental_linear_regression.py
```diff
@@ -0,0 +1,45 @@
...
+# We put the whole data to fit method, it is split automatically and then
+# partial_fit is called for each batch.
+inclin = IncrementalLinearRegression(batch_size=3)
+X, y = np.array([[0, 1], [1, 2], [2, 3], [0, 2], [1, 3], [2, 4]]), np.array(
+    [2, 4, 6, 3, 5, 7]
+)
+result = inclin.fit(X, y)
+
+print(f"Coefs:\n{result.coef_}")
+print(f"Intercept:\n{result.intercept_}")
```
Alexsandruss, line 45: Is there a reason to not compare coeffs between two ways of inc_lin_reg calling in code?

sklearnex/tests/test_memory_usage.py
```diff
@@ -97,7 +97,10 @@ def remove_duplicated_estimators(estimators_list):
...
```
Alexsandruss, line 101: Mention of failed memory leak tests range is required here.

examples/sklearnex/incremental_linear_regression.py
olegkkruglov: is there a reason to compare? we have tests for this purpose. there are a lot of tests for both onedal4py and sklearnex layers for this algorithms which allow us to make sure that the result is correct for both ways.

Alexsandruss: To demonstrate in example they are equal bit-to-bit or not.

sklearnex/tests/test_memory_usage.py
olegkkruglov: what do you mean by range?

examples/sklearnex/incremental_linear_regression.py
olegkkruglov: they are equal and it is well enough illustrated by example's output. we don't have any kind of such comparison in other examples

sklearnex/linear_model/tests/test_incremental_linear.py
```diff
@@ -0,0 +1,202 @@
...
+    assert_allclose(y_pred, y, rtol=1e-6, atol=1e-6)
+
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+@pytest.mark.parametrize("num_samples", [100, 1000])
+@pytest.mark.parametrize("num_features", [5, 10])
+@pytest.mark.parametrize("num_targets", [1, 2])
+@pytest.mark.parametrize("num_blocks", [1, 10])
+@pytest.mark.parametrize("macro_block", [None, 1024])
+@pytest.mark.parametrize("dtype", [np.float32])
```
icfaust, line 148:
```suggestion
@pytest.mark.parametrize("dtype", [np.float32, np.float64])
```

sklearnex/tests/test_memory_usage.py
icfaust: Also, please send me the ticket for the follow-on for the memory leak investigation/fix

olegkkruglov: @icfaust the ticket should have been created and assigned to you, please check your JIRA

icfaust:
please merge master and rerun private CI

onedal/linear_model/incremental_linear_model.py
```diff
@@ -0,0 +1,150 @@
...
+
+from daal4py.sklearn._utils import get_dtype
+
+from ..common.hyperparameters import get_hyperparameters
+from ..datatypes import _convert_to_supported, from_table, to_table
+from ..utils import _check_X_y, _num_features
+from .linear_model import BaseLinearRegression
+
+
+class IncrementalLinearRegression(BaseLinearRegression):
```
Alexsandruss, line 27: TODO comment to not forget this topic. 

```suggestion

# TODO: make arch. decision about Incremental APIs
class IncrementalLinearRegression(BaseLinearRegression):
```

onedal/linear_model/incremental_linear_model.py
```diff
@@ -0,0 +1,150 @@
...
+
+    copy_X : bool, default=True
+        If True, X will be copied; else, it may be overwritten.
+
+    algorithm : string, default="norm_eq"
+        Algorithm used for computation on oneDAL side
+    """
+
+    def __init__(self, fit_intercept=True, copy_X=False, algorithm="norm_eq"):
+        module = self._get_backend("linear_model", "regression")
```
Alexsandruss, line 46: Warning is required since this class is experimental and subject to change.

```suggestion
    def __init__(self, fit_intercept=True, copy_X=False, algorithm="norm_eq"):
        warn("IncrementalLinearRegression is experimental and might be changed or removed in future.")
        module = self._get_backend("linear_model", "regression")
```

Alexsandruss (commented):
I guess it's fine to merge if all fails are fixed.

onedal/linear_model/linear_model.cpp
```diff
@@ -80,75 +78,110 @@ auto get_onedal_result_options(const py::dict& params) {
...
     auto operator()(const py::dict& params) {
         using namespace dal::linear_regression;
 
         const auto intercept = params["intercept"].cast<bool>();
 
-        auto desc = descriptor_creator<Float, Method, Task>::get(intercept).set_result_options(
-            get_onedal_result_options(params));
+        auto desc = linear_regression::descriptor<Float,
+                                             linear_regression::method::norm_eq,
+                                             linear_regression::task::regression>(intercept)
```
ethanglaser, line 90:
```suggestion
                                             Method,
                                             Task>(intercept)

```

onedal/linear_model/incremental_linear_model.py
```diff
@@ -0,0 +1,150 @@
...
+        super().__init__(fit_intercept=fit_intercept, copy_X=copy_X, algorithm=algorithm)
+        self._partial_result = module.partial_train_result()
+
+    def _reset(self):
+        module = self._get_backend("linear_model", "regression")
+        self._partial_result = module.partial_train_result()
+
+    def partial_fit(self, X, y, queue=None):
+        """
+        Computes partial data for basic statistics
```
ethanglaser, line 56:
```suggestion
        Computes partial data for linear regression
```

onedal/linear_model/incremental_linear_model.py
```diff
@@ -0,0 +1,150 @@
...
+    def finalize_fit(self, queue=None):
+        """
+        Finalizes linear regression computation and obtains coefficients
+        from the current `_partial_result`.
+
+        Parameters
+        ----------
+        queue : dpctl.SyclQueue
+            Not used here, added for API conformance
+
```
ethanglaser, line 121: do we only have fit() implemented for GPU in OneDAL c++?

sklearnex/tests/test_n_jobs_support.py
```diff
@@ -32,7 +32,7 @@
     )
 )
 
-X, Y = make_classification(n_samples=40, n_features=4, random_state=42)
+X, y = make_classification(n_samples=40, n_features=4, random_state=42)
```
Alexsandruss, line 35: `Y` was capitalized because it was constant and in main scope of script.
Code style changes are not subject to this PR, so it should remain without code change.

examples/sklearnex/incremental_linear_regression.py
ethanglaser, line 1: Good demonstration of usage here - I recommend being a little more explicit in the comments for differentiating the 2 use cases. For example in line 23: "One use case of incremental linear regression would be manually designated chunks of data" and in line 36: "For a second use case, we use the same data but the algorithm splits chunks using the `batch_size` parameter"

sklearnex/tests/test_memory_usage.py
```diff
@@ -96,7 +96,10 @@ def remove_duplicated_estimators(estimators_list):
     return estimators_map.values()
 
 
-BANNED_ESTIMATORS = ("TSNE",)  # too slow for using in testing on common data size
+BANNED_ESTIMATORS = (
+    "IncrementalLinearRegression",  # TODO fix memory leak issue in private CI for data_shape = (1000, 100), data_transform_function = dataframe_f
```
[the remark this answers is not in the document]
olegkkruglov: no, the task for investigation is created

onedal/linear_model/incremental_linear_model.py
olegkkruglov: no, we have for all devices. in finalize_fit queue is not necessary because the policy is taken from previous call of partial_fit (currently there should be data in method's args to get the policy, that's why it is impossible to get policy directly in finalize_fit).

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,387 @@
...
+        else:
+            raise AttributeError(
+                f"'{self.__class__.__name__}' object has no attribute 'coef_'"
+            )
+
+    def set_coef_(self, value):
+        self.__dict__["coef_"] = value
+        if hasattr(self, "_onedal_estimator"):
+            self._onedal_estimator.coef_ = value
+            del self._onedal_estimator._onedal_model
```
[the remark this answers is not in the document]
olegkkruglov: it is necessary for explicit coefs setting. predict would not work with explicitly updated coefs without this deleltion

sklearnex/tests/_utils.py
```diff
@@ -135,8 +135,10 @@ def gen_dataset(estimator, queue=None, target_df=None, dtype=np.float64):
     else:
         raise ValueError("Unknown dataset type")
 
-    X = _convert_to_dataframe(X, sycl_queue=queue, target_df=target_df, dtype=dtype)
-    y = _convert_to_dataframe(y, sycl_queue=queue, target_df=target_df, dtype=dtype)
+    if queue is not None:
+        X = _convert_to_dataframe(X, sycl_queue=queue, target_df=target_df, dtype=dtype)
+        y = _convert_to_dataframe(y, sycl_queue=queue, target_df=target_df, dtype=dtype)
+
```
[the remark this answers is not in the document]
olegkkruglov: I'm pretty sure that the test was red without this edit. I'll recheck and create the task for investigation if it is still true

sklearnex/tests/test_n_jobs_support.py
```diff
@@ -79,15 +79,27 @@ def check_methods_decoration(estimator):
     check_estimator_doc(estimator_instance)
     # check `n_jobs` log entry for supported methods
     # `fit` call is required before other methods
-    check_method(X, Y, method=estimator_instance.fit, caplog=caplog)
+    # `partial_fit` is checked last if necessary
+    if "y" in inspect.signature(estimator_instance.fit).parameters:
+        check_method(X, y, method=estimator_instance.fit, caplog=caplog)
+    else:
+        check_method(X, method=estimator_instance.fit, caplog=caplog)
```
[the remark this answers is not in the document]
olegkkruglov: because fit method does not necessarily contain y (for example, it's true for covariance estimator)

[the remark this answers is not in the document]
olegkkruglov: https://github.com/intel/scikit-learn-intelex/pull/1656/files/32fc7bba122b7d86bcef2c56cd20bcd1ddb27d75#r1556676027

sklearnex/linear_model/tests/test_incremental_linear.py
```diff
@@ -0,0 +1,192 @@
...
+from sklearnex._device_offload import _transfer_to_host
+from sklearnex.linear_model import IncrementalLinearRegression
+
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+@pytest.mark.parametrize("fit_intercept", [True, False])
+@pytest.mark.parametrize("macro_block", [None, 1024])
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_sklearnex_fit_on_gold_data(dataframe, queue, fit_intercept, macro_block, dtype):
+    X = np.array([[1], [2]])
```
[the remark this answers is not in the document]
olegkkruglov: why do you think it is not?

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,387 @@
...
+@register_hyperparameters(
+    {
+        "fit": get_hyperparameters("linear_regression", "train"),
+        "partial_fit": get_hyperparameters("linear_regression", "train"),
+    }
+)
+@control_n_jobs(
+    decorated_methods=["fit", "partial_fit", "predict", "_onedal_finalize_fit"]
+)
+class IncrementalLinearRegression(MultiOutputMixin, RegressorMixin, BaseEstimator):
```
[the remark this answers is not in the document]
olegkkruglov: I think you mean removal of sklearn BaseEstimator on onedal4py side. Here it definitely should be, otherwise it would be impossible to run sklearn conformance tests for this estimator.

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,387 @@
...
+        else:
+            X = check_array(
+                X,
+                dtype=[np.float64, np.float32],
+                copy=self.copy_X,
+            )
+
+        assert hasattr(self, "_onedal_estimator")
+        if self._need_to_finalize:
+            self._onedal_finalize_fit()
```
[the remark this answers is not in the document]
olegkkruglov: no, there is not

sklearnex/tests/test_n_jobs_support.py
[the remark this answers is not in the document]
Alexsandruss: image
It's API mistake in IncCov estimator. All sklearn-like estimators should list `y` as parameter.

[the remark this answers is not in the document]
Alexsandruss: image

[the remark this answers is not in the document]
olegkkruglov: you are right, thanks for pointing out. your suggestion does not work though because not all estimators have partial_fit method. but I removed possibility of fit call without y. please check if the logic is fine now.

sklearnex/tests/_utils.py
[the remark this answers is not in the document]
olegkkruglov: image
the problem still exists. _convert_to_dataframe changes data here if called with queue=None and it lead to failure. I will add TODO comment and try to do shorter reproducer for the future task. I guess it would be better to leave the logic here as is for now.

sklearnex/tests/test_memory_usage.py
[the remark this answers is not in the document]
ethanglaser: can you mention during pre-planning if it hasn't already been brought up

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,387 @@
...
+            self.n_samples_seen_ += X.shape[0]
+        onedal_params = {"fit_intercept": self.fit_intercept, "copy_X": self.copy_X}
+        if not hasattr(self, "_onedal_estimator"):
+            self._onedal_estimator = self._onedal_incremental_linear(**onedal_params)
+        self._onedal_estimator.partial_fit(X, y, queue)
+        self._need_to_finalize = True
+
+    def _onedal_finalize_fit(self):
+        assert hasattr(self, "_onedal_estimator")
+        is_good_for_onedal = self.n_samples_seen_ >= self.n_features_in_ + int(
```
icfaust, line 191:
```suggestion
        is_underdetermined = self.n_samples_seen_ < self.n_features_in_ + int(
```

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,387 @@
...
+            self.fit_intercept
+        )
+        if not is_good_for_onedal:
```
icfaust, line 194:
```suggestion
        if is_underdetermined:
```

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,387 @@
...
+            y = check_array(
+                y,
+                dtype=[np.float64, np.float32],
+                copy=False,
+                ensure_2d=False,
+            )
+
+        n_samples, n_features = X.shape
+
+        is_good_for_onedal = X.shape[0] >= X.shape[1] + int(self.fit_intercept)
```
icfaust, line 222:
```suggestion
        is_underdetermined = X.shape[0] < X.shape[1] + int(self.fit_intercept)
```

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,387 @@
...
+        if not is_good_for_onedal:
```
icfaust, line 223:
```suggestion
        if is_underdetermined:
```

onedal/linear_model/incremental_linear_model.py
```diff
@@ -0,0 +1,156 @@
...
+import warnings
+
+import numpy as np
+
+from daal4py.sklearn._utils import get_dtype
+
+from ..common.hyperparameters import get_hyperparameters
+from ..datatypes import _convert_to_supported, from_table, to_table
+from ..utils import _check_X_y, _num_features
+from .linear_model import BaseLinearRegression
```
icfaust, line 26:
```suggestion
from daal4py.sklearn._utils import get_dtype
from onedal.common.hyperparameters import get_hyperparameters
from onedal.datatypes import _convert_to_supported, from_table, to_table
from onedal.utils import _check_X_y, _num_features
from onedal.linear_model import BaseLinearRegression
```
I would prefer we avoid relative imports if possible

icfaust (commented):
is_good_for_onedal had been giving me issues when reviewing the LinearRegression PR and then this one. Having now looked into the oneDAL code and the NormEq solver, its requiring that the system be not underdetermined (doing gaussian elimination). We should switch the variable names to make this clearer to future maintainers and people who look into the code.  Second I added a small request about relative imports.

sklearnex/linear_model/incremental_linear.py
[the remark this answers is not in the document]
ethanglaser: I do not see much use of BaseEstimator in sklearnex/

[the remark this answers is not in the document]
ethanglaser: And if so - why is it not present in other incremental algos?

sklearnex/tests/test_n_jobs_support.py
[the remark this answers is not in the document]
icfaust: This issue will be corrected in this PR as well, #1737 which does it based on the number of arguments without default values.

sklearnex/linear_model/incremental_linear.py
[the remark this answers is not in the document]
olegkkruglov: yes, you are right. as we decided with @icfaust, sklearn conformance tests should be done for all algos in patch map. BasicStatistics is not in patch map currently (this should be discussed on arch meeting and updated later if necessary). For IncCov it is done in https://github.com/intel/scikit-learn-intelex/pull/1674 and for IncPCA it is unnecessary since it has corresponding class in sklearn

sklearnex/linear_model/incremental_linear.py
olegkkruglov: the naming was inherited from linear regression implementation which had been done before my PRs. but I can rename if you insist. then it is better to update it in LinearRegression as well. and I have to add that we have created task to align C++ side behavior with sklearn on undetermined cases. when it will be done this flag should be deleted at all.

onedal/linear_model/incremental_linear_model.py
olegkkruglov: it contradicts with sklearn developer guide

sklearnex/linear_model/incremental_linear.py
olegkkruglov: I would leave it as is for now to be consistent with LinearRegression and update both cases when status of mentioned task become clear.

onedal/linear_model/incremental_linear_model.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+# TODO: make arch. decision about Incremental APIs
```
Alexsandruss, line 29: I mentioned it in wrong place: TODO and warning should be in sklearnex

onedal/linear_model/linear_model.cpp
```diff
@@ -80,75 +78,108 @@ auto get_onedal_result_options(const py::dict& params) {
...
-            const table& responses) {
-                using namespace dal::linear_regression;
-                using input_t = train_input<Task>;
-                train_ops ops(policy, input_t{ data, responses }, params2desc{});
-                return fptype2t{ method2t{ Task{}, ops } }(params);
-            }
-        );
-    }
+    m.def("train", [](
+        const Policy& policy,
```
samir-nasibli, line 113: I think here should be `else` in case of onedal version < 20240000

onedal/linear_model/tests/test_incremental_linear_regression.py
samir-nasibli: Please remove TODO and ticket to the backlog

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,392 @@
...
+        self : object
+            Returns the instance itself.
+        """
+
+        dispatch(
+            self,
+            "partial_fit",
+            {
+                "onedal": self.__class__._onedal_partial_fit,
+                "sklearn": None,
```
samir-nasibli, line 322: What will happen in case if we are not supporting it via onedal branch?
It is undefined for me. This not the scope of current PR, but we should raise common error for such estimators, and some common test shoul check this error raised for such estimators.

sklearnex/linear_model/tests/test_incremental_linear.py
```diff
@@ -0,0 +1,192 @@
...
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
+from onedal.tests.utils._dataframes_support import (
+    _as_numpy,
+    _convert_to_dataframe,
+    get_dataframes_and_queues,
+)
+from sklearnex._device_offload import _transfer_to_host
```
samir-nasibli, line 26:
```suggestion
```

onedal/linear_model/linear_model.cpp
olegkkruglov: I guess it was done like that to keep ability to call methods without hyperparameters from C++ side. probably they act differently. here it is done exactly like in all other algorithms with hyperparameters. I think if we want to change such behavior then it is better to do in separate PR and for all algos.

examples/sklearnex/incremental_linear_regression_dpctl.py
```diff
@@ -0,0 +1,51 @@
...
+import dpctl.tensor as dpt
+
+from sklearnex.linear_model import IncrementalLinearRegression
+
+# We create GPU SyclQueue and then put data to dpctl tensors using
+# the queue. It allows us to do computation on GPU.
+
+queue = dpctl.SyclQueue("gpu")
+
+inclin = IncrementalLinearRegression()
```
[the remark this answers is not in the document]
olegkkruglov: I would say it contradicts codestyle. variable are not named with capital letters almost nowhere in the library. also naming here is the same now as in other incremental algos tests. I would keep it as is.

onedal/linear_model/incremental_linear_model.py
olegkkruglov: I have no preferrable way. Let's choose one of them and finally merge this PR.

sklearnex/linear_model/incremental_linear.py
olegkkruglov: I guess it fails in dispatch in such cases. I will create task to handle it later.

sklearnex/linear_model/incremental_linear.py
icfaust: #1804 should be the changes needed for LinearRegression. Let me know what you think.

Alexandr-Solovev (approved):
Looks good to me! Nice work!

sklearnex/linear_model/tests/test_incremental_linear.py
```diff
@@ -0,0 +1,200 @@
...
+    assert inclin.n_features_in_ == 2
+    tol = 7e-6 if dtype == np.float32 else 1e-7
+    assert_allclose(inclin.coef_, [1.0, 2.0], atol=tol)
+    if fit_intercept:
+        assert_allclose(inclin.intercept_, 3.0, atol=tol)
+
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+    y_pred = inclin.predict(X_df)
+
+    assert_allclose(_as_numpy(y_pred), y, rtol=1e-6, atol=1e-6)
```
ethanglaser, line 134: are both rtol and atol needed here? probably just atol makes sense

sklearnex/linear_model/incremental_linear.py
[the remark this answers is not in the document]
ethanglaser: thanks for clarification

sklearnex/linear_model/incremental_linear.py
```diff
@@ -0,0 +1,391 @@
...
+if sklearn_check_version("1.2"):
+    from sklearn.utils._param_validation import Interval
+
+from onedal.common.hyperparameters import get_hyperparameters
+
+from .._device_offload import dispatch, wrap_output_data
+from .._utils import PatchingConditionsChain, register_hyperparameters
+
+
+# TODO: make arch. decision about Incremental APIs
```
ethanglaser, line 40:
```suggestion
```
probably not necessary to have this in a comment

ethanglaser (approved):
Assuming one final green internal CI run

PR #1660, opened 2024-01-22
feat: SHAP value support for XGBoost's binary classification models

ahuber21:
Add SHAP value calculation for XGBoost binary classification tasks.


razdoburdin (approved):
Looks good.

icfaust:
This hopefully should work,  they try exception-catching style matches other implemenations in the file, but I think a ```daal_check_version``` at instantiation implementation of this functionality would be more straightforward and waste fewer cycles with older versions.

razdoburdin (approved):
looks good for me.
We can merge it, if CI is ok.

icfaust:
manually run since the bot seems to be down: http://intel-ci.intel.com/ef6c54a3-946a-f152-a360-a4bf010d0e2e

PR #1662, opened 2024-01-22
[bug,testing] correct n_jobs testing for aliased class methods

icfaust:
# Description
The current logic will check the name of the method via the ```__name__``` parameter of the attribute. For aliased methods, this will give the incorrect name for the attribute (it will provide the original rather than new name). If a new attribute is created from one with n_jobs_support with an additional wrap, this will cause an erroneous failure in the test. The following would fail.

```python
@control_n_jobs(decorated_methods=["hello"])
class example():
    def hello():
        pass
    hello2 = wrap_output_data(hello)
    
print(getattr(example, "hello").__name__)
print(getattr(example,"hello2").__name__) # will print hello instead of hello2
```
Such a class above will test the ```hello2``` and ```hello``` methods as the name ```hello```. This issue was observed in the LocalOutlierFactor refactor (#1640), where certain methods need to be accessed without wrap_output_data, and overall needs to have n_jobs support (see _kneighbors and kneighbors methods).
 
- fix the error by using a dict to store names, use those for testing
- use hasattr to handle available_if rather than try catch (fixes issues for lof, NuSVC, etc)

### DIFF
--- a/sklearnex/tests/test_n_jobs_support.py
+++ b/sklearnex/tests/test_n_jobs_support.py
@@ -65,7 +65,7 @@ def test_n_jobs_support(caplog, estimator_class, n_jobs):
 
         for i in funcs:
             assert hasattr(funcs[i], "__onedal_n_jobs_decorated__") == (
-                i in estimator._n_jobs_suppored_onedal_methods
+                i in estimator._n_jobs_supported_onedal_methods
             ), f"{estimator}.{i} n_jobs decoration does not match {estimator} n_job supported methods"
 
     caplog.set_level(logging.DEBUG, logger="sklearnex")
--- a/sklearnex/tests/test_n_jobs_support.py
+++ b/sklearnex/tests/test_n_jobs_support.py
@@ -64,10 +64,9 @@ def test_n_jobs_support(caplog, estimator_class, n_jobs):
         }
 
         for i in funcs:
-            if func in estimator._n_jobs_supported_onedal_methods:
-                assert hasattr(funcs[i], "__onedal_n_jobs_decorated__")
-            else:
-                assert not hasattr(funcs[i], "__onedal_n_jobs_decorated__")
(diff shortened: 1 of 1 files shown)


sklearnex/tests/test_n_jobs_support.py
```diff
@@ -58,19 +57,16 @@ def check_method(*args, method, caplog):
...
+        funcs = {
+            i: getattr(estimator, i)
+            for i in dir(estimator)
+            if hasattr(estimator, i) and callable(getattr(estimator, i))
+        }
+
+        for i in funcs:
+            assert hasattr(funcs[i], "__onedal_n_jobs_decorated__") == (
+                i in estimator._n_jobs_supported_onedal_methods
+            ), f"{estimator}.{i} n_jobs decoration does not match {estimator} n_job supported methods"
```
Alexsandruss, line 69:
```suggestion
        for func_name, func in funcs.items():
            assert hasattr(func, "__onedal_n_jobs_decorated__") == (
                func_name in estimator._n_jobs_supported_onedal_methods
            ), f"{estimator}.{func_name} n_jobs decoration does not match {estimator} n_jobs supported methods"
```

icfaust:
private CI re-run http://intel-ci.intel.com/eeb9db7d-3dea-f167-8390-a4bf010d0e2e

samir-nasibli (commented):
@icfaust Thank you for the fix!
having detailed refer to the cause is just better to review and understanding of the issue/bug.
Otherwise it looks like just a small refactoring :)

icfaust:
Thanks! @samir-nasibli  added a referral to #1640 where this issue was first seen

samir-nasibli (commented):
Looks good to me!
Please address some fails in intel CI.

icfaust:
CI issues are related to 1.3.1 sklearn conformance, and unrelated to sklearnex tests. (Looks to be a scipy version issue for python
3.12)

Alexsandruss (approved):

PR #1663, opened 2024-01-23
[bug] adding _parameter_constraints to additional algos

ethanglaser:
# Description
Avoid stock sklearn conformance fail by adding _parameter_constraints to modules that don't have it


samir-nasibli (approved):
Looks good to me.
waiting CI

Alexsandruss (approved):
Assuming CI will be green.

ethanglaser:
Based on CI, intended fix is working. There are sporadic issues that I see in other jobs as well - will create a ticket to address separately.

PR #1666, opened 2024-01-24
[enhancement] prevent double-wrapping of run_with_n_jobs

icfaust:
Seen in #1649, when an object aliases a method which has been wrapped with n_jobs support, it will fail testing unless it is included in decorated_methods. When it is in decorated_methods, it will wrap the method again (which is unnecessary).  This adds a check to see if it already has an n_jobs wrap, and sets getattr to fail if the method doesn't exist (currently sets to None, but an exception is preferrable)

- Add if statement to prevent double wrap
- Remove getattr to set to None in case method doesn't exist (exception preferrable)


icfaust:
Rerun to see if omp issues are sporadic

http://intel-ci.intel.com/eebab51f-093c-f1b5-932a-a4bf010d0e2e

icfaust:
Please see second run for green CI

Alexsandruss (approved):

PR #1668, opened 2024-01-24
[enhancement] isolate underlying fit call in daal4py LogisticRegression

icfaust:
# Description
Re-use of daal4py methods directly can cause issue with various wrappers, and currently the implementation of fit doesn't match that of predict and predict_proba. This change introduces an indirection using daal4py_fit. This will help solve issues observed in #1666 and #1649

- daal4py_fit

### DIFF
--- a/daal4py/sklearn/linear_model/logistic_path.py
+++ b/daal4py/sklearn/linear_model/logistic_path.py
@@ -1020,7 +1020,7 @@ if sklearn_check_version("0.24"):
             -----
             The SAGA solver supports both float64 and float32 bit arrays.
             """
-            return self.daal4py_fit(self, X, y, sample_weight)
+            return self.daal4py_fit(X, y, sample_weight)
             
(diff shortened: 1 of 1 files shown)


avolkov-intel (approved):

daal4py/sklearn/linear_model/logistic_path.py
```diff
@@ -776,6 +776,22 @@ def _func_(x, *args):
     return np.array(coefs), np.array(Cs), n_iter
 
 
+def daal4py_fit(self, X, y, sample_weight=None):
+    if sklearn_check_version("1.0"):
+        self._check_feature_names(X, reset=True)
+    if sklearn_check_version("1.2"):
+        self._validate_params()
```
Alexsandruss, line 783: Should be moved to `fit` because sklearnex estimator will duplicate these lines

Alexsandruss (changes requested):

samir-nasibli (approved):

PR #1671, opened 2024-01-26
Apply black 24.1 formatting

Alexsandruss:
New rules in black 24.1 version failed CI Lint check.
This PR applies black 24.1.0 formatting to code base.


icfaust (approved):
Pull when CI is complete (assuming good CI)

napetrov:
@Alexsandruss  @icfaust  - should we pin black versioning so it will be controlled within sources?  Can be done as separate change

icfaust:
I think thats a good idea

PR #1674, opened 2024-01-26
[enhancement] Add IncrementalEmpiricalCovariance to patch_map

icfaust:
# Description
IncrementalEmpiricalCovariance update to match sklearnex estimator characteristics

- Add mahalanobis method using sklearnex pairwise_distances to IncrementalEmpiricalCovariance
- Add error_norm
- Add onedal_cpu_support and onedal_gpu_support to IncrementalEmpiricalCovariance
- Add BaseEstimator to IncrementalEmpiricalCovariance inheritance
- Add wrap_output_data where necessary in IncrementalEmpiricalCovariance
- Fix covariance_ and location_ attributes to report as not existing before fit
- Add n_features_in_ attribute for sklearn conformity
- Add n_samples_seen_ attribute to match style of IncrementalPCA
- Add init keywords to match sklearn EmpiricalCovariance
- Add store_precision functionality
- Add assume_centered functionality
- Add relevant sklearn tests to sklearnex tests via a separate monkeypatch

Tasks
- [x] Implement changes for sklearnex/sklearn conformance
- [x] Add to patchmap to access sklearn tests
- [x] Modify patchmap to work with new estimators
- [x] Add EmpiricalCovariance tests from sklearn to test_covariance for IncrementalEmpiricalCovariance
- [x] Pass Tests
- [x] Split off changes to patchmap and tests into separate PR


sklearnex/covariance/incremental_covariance.py
```diff
@@ -14,18 +14,30 @@
...
+
+if sklearn_check_version("1.2"):
+    from sklearn.utils._param_validation import Interval
 
 
-@control_n_jobs(decorated_methods=["partial_fit"])
-class IncrementalEmpiricalCovariance:
+# TODO: consult with others whether this should support store_precision and assume_centered
+@control_n_jobs(decorated_methods=["partial_fit", "fit"])
+class IncrementalEmpiricalCovariance(BaseEstimator):
```
samir-nasibli, line 40: add baseEstimator import

sklearnex/covariance/incremental_covariance.py
```diff
@@ -117,14 +184,63 @@ def fit(self, X, queue=None):
...
             X_batch = X[batch]
-            self.partial_fit(X_batch, queue=queue)
+            self._onedal_partial_fit(X_batch, queue=queue)
+            self.n_samples_seen_ += X.shape[0]
 
         self._onedal_finalize_fit()
         return self
+
+    get_precision = sklearn_EmpiricalCovariance.get_precision
+    error_norm = wrap_output_data(sklearn_EmpericalCovariance.error_norm)
```
samir-nasibli, line 210: Why are you wrapping output data here?

sklearnex/covariance/incremental_covariance.py
```diff
@@ -14,18 +14,30 @@
 # limitations under the License.
 # ===============================================================================
 
+import numbers
+
 import numpy as np
+from scipy import linalg
```
samir-nasibli, line 20: please remove unnecessary/unused imports

sklearnex/covariance/incremental_covariance.py
```diff
@@ -14,18 +14,30 @@
...
+from sklearnex.metrics import pairwise_distances
+
+if sklearn_check_version("1.2"):
+    from sklearn.utils._param_validation import Interval
 
 
-@control_n_jobs(decorated_methods=["partial_fit"])
-class IncrementalEmpiricalCovariance:
+# TODO: consult with others whether this should support store_precision and assume_centered
+@control_n_jobs(decorated_methods=["partial_fit", "fit"])
```
olegkkruglov, line 39: decorated_methods should be ["partial_fit", "_onedal_finalize_fit"]. there is no need to decorate fit since it does not call onedal directly, and _onedal_finalize_fit does

samir-nasibli (commented):
From a arch I recomend to investiaget how structurally it looks like for PCA and IincrementalPCA: 

https://github.com/scikit-learn/scikit-learn/blob/9e38cd00d032f777312e639477f1f52f3ea4b3b7/sklearn/decomposition/_pca.py#L122

https://github.com/scikit-learn/scikit-learn/blob/9e38cd00d032f777312e639477f1f52f3ea4b3b7/sklearn/decomposition/_incremental_pca.py#L19

For both estimators it is better to have baseEmpricalCovarience to inheritance.
With it we will have better conformity and less code duplicartion

sklearnex/covariance/incremental_covariance.py
```diff
@@ -14,18 +14,30 @@
 # limitations under the License.
 # ===============================================================================
 
+import numbers
+
 import numpy as np
+from scipy import linalg
+from sklearn.covariance import EmpiricalCovariance as sklearn_EmpericalCovariance
```
olegkkruglov, line 21: sklearn_EmpiricalCovariance

sklearnex/covariance/incremental_covariance.py
```diff
@@ -50,10 +62,23 @@ class IncrementalEmpiricalCovariance:
...
+
+    def __init__(self, batch_size=None, copy=False):
         self._need_to_finalize = False  # If True then finalize compute should
         #      be called to obtain covariance_ or location_ from partial compute data
         self.batch_size = batch_size
+        self.copy = copy
+
+    def _onedal_supported(self, method_name, *data):
+        patching_status = PatchingConditionsChain(
+            "sklearn.covariance.{self.__class__.__name__}.{method_name}"
```
olegkkruglov, line 79: f"sklearn.covariance.{self.__class__.__name__}.{method_name}"

sklearnex/dispatcher.py
```diff
@@ -64,6 +64,9 @@ def get_patch_map():
             from .utils.parallel import _FuncWrapperOld as _FuncWrapper_sklearnex
 
         from .cluster import DBSCAN as DBSCAN_sklearnex
+        from .covariance import (
+            IncrementalEmpiricalCovariance as IncrementalEmpericalCovariance_sklearnex,
```
olegkkruglov, line 68: IncrementalEmpiricalCovariance_sklearnex

sklearnex/covariance/incremental_covariance.py
```diff
@@ -117,14 +184,63 @@ def fit(self, X, queue=None):
...
+            self._validate_params()
+            X = self._validate_data(X, dtype=[np.float64, np.float32], copy=self.copy)
+        else:
+            X = check_array(X, dtype=[np.float64, np.float32], copy=self.copy)
+
+        n_samples, self.n_features_in_ = X.shape
+
         if self.batch_size is None:
-            batch_size_ = 5 * n_features
+            self.batch_size_ = 5 * self.n_features_in_
```
olegkkruglov, line 198: what is the purpose to put it to estimator's attribute?

icfaust: https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep010/proposal.html

olegkkruglov: i mean batch_size_

icfaust: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/decomposition/_incremental_pca.py#L236

icfaust:
Will write tickets associated with dataframe0 and test memory leak, it is gpu and cpu side meaning its likely in the glue portion of the code. Also, it only occurs on one of the two meaning something else fishy is going on.  This will have to be revisited in the future.  I will now slim this PR down and remove the content from the other 3.

icfaust:
Code has now been reduced to just changes in IncrementalEmpiricalCovariance

icfaust:
Unrelated issue in private CI (logistic regression convergence on cpu)

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -141,3 +147,19 @@ def test_sklearnex_fit_on_random_data(
...
+
+# Monkeypatch IncrementalEmpiricalCovariance into relevant sklearn.covariance tests
+@pytest.mark.parametrize(
+    "sklearn_test",
+    [
+        test_covariance,
+        test_EmpiricalCovariance_validates_mahalanobis,
+    ],
+)
+def test_IncrementalEmpiricalCovariance_against_sklearn(monkeypatch, sklearn_test):
```
olegkkruglov, line 160: what is monkeypatch here?

icfaust: https://docs.pytest.org/en/latest/how-to/monkeypatch.html

sklearnex/covariance/incremental_covariance.py
```diff
@@ -14,76 +14,157 @@
...
+            f"sklearn.covariance.{self.__class__.__name__}.{method_name}"
+        )
+        return patching_status
 
     def _onedal_finalize_fit(self):
         assert hasattr(self, "_onedal_estimator")
         self._onedal_estimator.finalize_fit()
         self._need_to_finalize = False
 
+        if self.assume_centered:
```
olegkkruglov, line 120: I don't think it makes sense in this way. in sklearn.EmpiricalCovariance this flag is used to simplify computations when data are assumed to be centered. but here it causes additional computations. IMO it is better to remove it until we will add this feature on oneDAL side.

sklearnex/covariance/incremental_covariance.py
```diff
@@ -14,76 +14,157 @@
...
     def _onedal_partial_fit(self, X, queue):
         onedal_params = {
             "method": "dense",
             "bias": True,
         }
         if not hasattr(self, "_onedal_estimator"):
             self._onedal_estimator = self._onedal_incremental_covariance(**onedal_params)
-        self._onedal_estimator.partial_fit(X, queue)
-        self._need_to_finalize = True
+        try:
```
olegkkruglov, line 140: what is the purpose of this change?

sklearnex/covariance/incremental_covariance.py
```diff
@@ -117,14 +232,71 @@ def fit(self, X, queue=None):
...
+            X = self._validate_data(
+                X, dtype=[np.float64, np.float32], copy=self.copy, force_all_finite=False
+            )
         else:
-            batch_size_ = self.batch_size
-        for batch in gen_batches(n_samples, batch_size_):
+            X = check_array(
+                X, dtype=[np.float64, np.float32], copy=self.copy, force_all_finite=False
+            )
+            self.n_features_in_ = X.shape[1]
```
olegkkruglov, line 262:
```suggestion
        self.n_features_in_ = X.shape[1]
```

sklearnex/covariance/incremental_covariance.py
```diff
@@ -117,14 +232,71 @@ def fit(self, X, queue=None):
...
+                "sklearn": None,
+            },
+            X,
+        )
+        return self
+
+    def _onedal_fit(self, X, queue=None):
+        self.n_samples_seen_ = 0
+
+        if sklearn_check_version("1.2"):
```
olegkkruglov, line 250: does it makes sense to move such stuff to fit method for consistency with other _onedal_* methods?

sklearnex/covariance/incremental_covariance.py
icfaust: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/base.py#L572 When reset=True for validate_data, it will set that parameter

sklearnex/covariance/incremental_covariance.py
icfaust: There exists a logic issue here then. Either it is like IncrementalPCA, where all the methods of PCA are supported (in this case for EmpiricalCovariance) or the class should be named differently.  I believe this class should have the same inputs as EmpiricalCovariance. If a user tries to use this class thinking it can be a drop-in replacement for EmpiricalCovariance, they will face issues. There is a warning in the class documentation not to use it, but is available as necessary to the user. It is mathematically valid, just a poor choice to use it.

sklearnex/covariance/incremental_covariance.py
icfaust: Though this may come down to your understanding of the onedal side, maybe this is a bad assumption? let me know

olegkkruglov: but if it fails on partial_fit stage then it doesn't make sense to do finalize after, so, setting _need_to_finalize equal to True after that is quite weird for me

sklearnex/covariance/incremental_covariance.py
olegkkruglov: i feel that it is not good choice to add something which shouldn't be used. then it is better to do nothing and just add it to args for API conformance (with highlighting that fact in the docstring). have to notice that if data is really centered then output for both cases is the same, so it is weird to do redundant computations. i highly doubt that someone uses covariance estimator to compute scalar product of vectors. more likely, the user wants to compute covariance faster setting assume_centered=True and doing nothing special is more preferrable in this case

olegkkruglov: also have to add that IMO the API for this class shouldn't be exactly the same as for EmpiricalCovariance, it is just assumption which can be ignored at this point if it creates some problems.

olegkkruglov: ok, probably removing the parameter will cause issues with passing tests for sklearn.EmpiricalCovariance, so, doing nothing must be the best way

sklearnex/covariance/incremental_covariance.py
olegkkruglov: it makes more sense but i would say that the best way in this case just to let it fail as is and let user decide what to do with it depending on the failure.

sklearnex/covariance/incremental_covariance.py
icfaust: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/covariance/_empirical_covariance.py#L107  its a rather simple thing that I would expect wouldn't be worth the work (and I am not sure why they implemented it in the first place since its rather trivial). I guess I am just hesitant about missing out on additional tests because it no longer conforms to sklearn's empirical covariance.  As a compromise, I'd suggest we add a warning just before the math is applied.

let me know what you think

sklearnex/covariance/incremental_covariance.py
icfaust: if check_array is used before dispatch, it could strip away the queue in cases of dpnp and dpctl.  Since we don't test fallbacks to CPU yet, this could be silently masking issues with the underlying SYCL support.

olegkkruglov: but you use check_array before dispatch in partial_fit. moreover, it is used like that in sklearnex.EmpiricalCovariance. it looks like it is my bad because i was not able to find such usage in other estimators. probably should be fixed everywhere then. it might be worth to avoid check_array usage in sklearnex part and move it to onedal package.

sklearnex/covariance/incremental_covariance.py
olegkkruglov: > i feel that it is not good choice to add something which shouldn't be used. then it is better to do nothing and just add it to args for API conformance (with highlighting that fact in the docstring). have to notice that if data is really centered then output for both cases is the same, so it is weird to do redundant computations. i highly doubt that someone uses covariance estimator to compute scalar product of vectors. more likely, the user wants to compute covariance faster setting assume_centered=True and doing nothing special is more preferrable in this case

this argument is still valid for me. i think that the best way is to just ignore this parameter with marking it as added for API conformance, otherwise, it will force the user to remove assume_centered=True from estimator calls in their external code to reach the best performance (that's probably exactly what they wanted setting this parameter)

sklearnex/covariance/incremental_covariance.py
olegkkruglov: but i don't really like adding scikit dependency to onedal package

sklearnex/covariance/incremental_covariance.py
```diff
@@ -14,76 +14,157 @@
...
+        if hasattr(self, "_onedal_estimator"):
+            if self._need_to_finalize:
+                self._onedal_finalize_fit()
+            return self._onedal_estimator.location_
+        else:
+            raise AttributeError(
+                f"'{self.__class__.__name__}' object has no attribute 'location_'"
+            )
 
-    @support_usm_ndarray()
```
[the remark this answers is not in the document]
samir-nasibli: dispatch is used for patching sklearn methods.
Currently this is useless for non-patched methods, estimators.

`Support_usm_ndarray` is better.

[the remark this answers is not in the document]
olegkkruglov: as far as I know, after Ian's changes discussed on the last architectural meeting (https://github.com/intel/scikit-learn-intelex/pull/1678) it makes sense to use dispatch for such methods (after adding it to patching map). am I missing something?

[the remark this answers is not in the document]
samir-nasibli: Actually, support_usm_ndarray also should be a little bit updated. It were designed for onedal4py estimators, that have queue param for methods.

sklearnex/tests/test_memory_usage.py
```diff
@@ -94,7 +94,10 @@ def remove_duplicated_estimators(estimators_list):
     return estimators_map.values()
 
 
-BANNED_ESTIMATORS = ("TSNE",)  # too slow for using in testing on common data size
+BANNED_ESTIMATORS = (
+    "IncrementalEmpiricalCovariance",  # dataframe_f issues
+    "TSNE",
+)  # too slow for using in testing on common data size
```
Alexsandruss, line 100: Probably, comment was moved during reformatting
```suggestion
    "TSNE",  # too slow for using in testing on common data size
)  
```

sklearnex/covariance/incremental_covariance.py
Alexandr-Solovev: I would like to say that its not extremely hard to add this support on cpp(oneDAL) side firstly, instead of using additional computations, due to user will face with degradation, when speedup is expected

Alexandr-Solovev:
If its possible, please wait this pr, for avoid python computation for `assume_centered` flag support https://github.com/oneapi-src/oneDAL/pull/2658

sklearnex/covariance/incremental_covariance.py
icfaust: Maybe I misunderstand, this is in the sklearnex/covariance section and not in the onedal package.

icfaust:
Only Issue is a errant PCA test case failure.

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -41,24 +46,32 @@ def test_sklearnex_partial_fit_on_gold_data(dataframe, queue, dtype):
         result = inccov.partial_fit(X_split_df)
 
     expected_covariance = np.array([[0, 0], [0, 0]])
-    expected_means = np.array([0, 1])
+    expected_means = np.array([0, 0])
+
+    if assume_centered:
+        expected_covariance = np.array([[0, 0], [0, 1]])
+    else:
+        expected_means = np.array([0, 1])
```
ethanglaser, line 54: this is a bit confusing to read - can lines 48-49 be added to the conditionals does that cause error with variable not being initialized?

ethanglaser: if all expected_means and expected_covariance cant be added inside conditionals then I'd set them for assume_centered False outside the condition and update if assume_centered is True

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -117,9 +130,9 @@ def test_sklearnex_partial_fit_on_random_data(
 
 
 @pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
-@pytest.mark.parametrize("num_batches", [2, 4, 6, 8, 10])
-@pytest.mark.parametrize("row_count", [100, 1000, 2000])
-@pytest.mark.parametrize("column_count", [10, 100, 200])
+@pytest.mark.parametrize("num_batches", [2, 10])
+@pytest.mark.parametrize("row_count", [10, 100])
+@pytest.mark.parametrize("column_count", [10, 100])
```
olegkkruglov, line 135: why don't you add assume_centered parametrize here?

onedal/covariance/incremental_covariance.py
```diff
@@ -46,8 +53,8 @@ class IncrementalEmpiricalCovariance(BaseEmpiricalCovariance):
         Estimated covariance matrix
     """
 
-    def __init__(self, method="dense", bias=False):
-        super().__init__(method, bias)
+    def __init__(self, method="dense", bias=False, assume_centered=False):
```
olegkkruglov, line 56: could you please add _reset method to this class and its usage in sklearnex fit method like it is done in IncrementalLinearRegression? otherwise two consecutive fit calls or fit call after partial_fit call give wrong results.

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -132,12 +146,35 @@ def test_sklearnex_fit_on_random_data(
...
+# Monkeypatch IncrementalEmpiricalCovariance into relevant sklearn.covariance tests
+@pytest.mark.allow_sklearn_fallback
+@pytest.mark.parametrize(
+    "sklearn_test",
+    [
+        test_covariance,
+        test_EmpiricalCovariance_validates_mahalanobis,
+    ],
+)
+def test_IncrementalEmpiricalCovariance_against_sklearn(monkeypatch, sklearn_test):
```
ethanglaser, line 175: so does it run sklearn tests for covariance on our incremental cov estimator?

icfaust: thats right, though it doesn't automatically discover new tests in sklearn going forward, it only includes tests we know about now

icfaust:
Rerun due to openblas issues: http://intel-ci.intel.com/ef02ba38-9fbd-f127-9fb6-a4bf010d0e2e

ethanglaser (approved):
LGTM, @olegkkruglov please confirm changes are sklearnex/covariance/incremental_covariance.py look good.

olegkkruglov (approved):
LGTM, thanks for the update!

PR #1677, opened 2024-01-29
[enhancement] Enable oneDAL support for EmpiricalCovariance with default parameters

icfaust:
# Description
EmpiricalCovariance is missing certain attributes necessary to run oneDAL algorithms for default parameters. This limits its utility. By applying sklearn EmpiricalCovariance native functions, the default output can be supported. This will enhance user experience, and improve testability of the preview algorithm.

- add store_precision to EmpiricalCovariance so that default parameters will be supported
- changes to patching conditions change to reflect new capabilities
- wrap error_norm and score methods for dpnp and dpctl support
- Add mahalanobis using sklearnex's pairwise_distances (if metric implemented in the future)


sklearnex/preview/covariance/covariance.py
```diff
@@ -39,7 +42,7 @@ class EmpiricalCovariance(sklearn_EmpiricalCovariance):
 
     def _save_attributes(self):
         assert hasattr(self, "_onedal_estimator")
-        self.covariance_ = self._onedal_estimator.covariance_
+        self._set_covariance(self._onedal_estimator.covariance_)
```
olegkkruglov, line 45: where is definition of this method?

icfaust: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/covariance/_empirical_covariance.py#L195

sklearnex/preview/covariance/covariance.py
```diff
@@ -100,4 +106,27 @@ def fit(self, X, y=None):
 
         return self
 
+    # This needs to be included to guarantee that we use the sklearnex pairwise_distances
```
olegkkruglov, line 109: why is this necessary? we don't support this metric in sklearnex anyway, so, the default one still will be called.

sklearnex/preview/covariance/covariance.py
```diff
@@ -100,4 +106,27 @@ def fit(self, X, y=None):
...
+    @wrap_output_data
+    def mahalanobis(self, X, queue=None):
```
samir-nasibli, line 111: Could you please clarify, why queue is here?

sklearnex/preview/covariance/covariance.py
```diff
@@ -100,4 +106,27 @@ def fit(self, X, y=None):
...
+        with config_context(assume_finite=True):
+            # compute mahalanobis distances
+            dist = pairwise_distances(
+                X, self.location_[np.newaxis, :], metric="mahalanobis", VI=precision
+            )
+
+        return np.reshape(dist, (len(X),)) ** 2
+
+    error_norm = wrap_output_data(sklearn_EmpiricalCovariance.error_norm)
+    score = wrap_output_data(sklearn_EmpiricalCovariance.score)
```
samir-nasibli, line 127: It make sense generally have a mechanism, that wraps output data for all non-patched method of the estimator.

sklearnex/preview/covariance/covariance.py
samir-nasibli: I don't like this API inconsistency for `fit` and `mahalanobis`

sklearnex/preview/covariance/covariance.py
icfaust: Agreed, this should be a follow up PR which we will have to add to the backlog (I imagine it will be a long undertaking)

sklearnex/preview/covariance/covariance.py
icfaust: I have removed the queue. pairwise_distances in daal4py has input checks within it following daal4py style.

olegkkruglov (approved):
LGTM

PR #1679, opened 2024-01-29
MAINT: adding common BaseEstimatorSPMD for spmd ifaces

samir-nasibli:
# Description
Just common estimator for spmd estimators, for avoiding code duplication


olegkkruglov (approved):
LGTM

ethanglaser (approved):
Good idea

olegkkruglov:
Does it make sense to do the same for non-SPMD algos?

PR #1568, opened 2023-11-06
WIP: DataManagement update

KulikovNikita:

PR #1683, opened 2024-01-30
ENH: update of onedal4py SVM fit method for sparse support

samir-nasibli:
# Description
Part of the #1568 splitting work
follow up changes after Datamanagement update for sparsity data ( #1568 )

TBD tests and other requirements

NOTE: the need for this PR will be determined after merging common primitives for Data Management from #1568 

---------
Co-authored-by: Nikita Kulikov <nikita.kulikov@intel.com>


samir-nasibli:
Approach is outdated. Python coverage of #1568 will be deeply refactored

PR #1688, opened 2024-01-31
[bug] fix preview get_patch_map bug in sklearnex

icfaust:
# Description
Any assert in test_monkeypatch will leave the patch in place. This is very much an undesired behavior. Using a try finally statement will guarantee the use of unpatch_sklearn in the case of a failure (thereby limiting the scope of the patch to the test). Secondarily, test_patch_by_list_many_estimators was missing an unpatch call, which means it was impacting all tests following it. Thirdly, there is an issue where the get_patch_map lru_cache was only checking _is_preview_enabled on first cache, as lru_cache looks at *args and **kwargs for similarity to previous inputs. unpatch_sklearn clears that cache, which fixed the issue previously in the test. A wrapper was added to require checking for preview when trying to recover the lru_cache. Finally test_parallel is modified so that it properly unloads patch_sklearn at the end of the test.

- Introduce try finally for every sklearn_patch call in all tests of test_monkeypatch.py
- Add unpatch_sklearn to test_patch_by_list_many_estimators
- Reintroduce test_monkeypatch into standard CI run call
- wrap get_patch_map to properly check for preview
- modify test_parallel.py to also respect global scope for patch_sklearn


sklearnex/tests/test_monkeypatch.py
```diff
@@ -14,10 +14,37 @@
...
+# this file must be decorated with reset_sklearn_on_completion
+def reset_sklearn_on_completion(test):
+    @wraps(test)
+    def test_wrapper(*args, **kwargs):
+        try:
+            result = test(*args, **kwargs)
+        finally:
+            for i in sys.modules.copy():
+                if i.startswith("sklearn."):
+                    importlib.reload(sys.modules[i])
```
samir-nasibli, line 40:
```suggestion
            for module_name in sys.modules.copy():
                if module_name.startswith("sklearn."):
                    importlib.reload(sys.modules[module_name])

```

sklearnex/tests/test_monkeypatch.py
```diff
@@ -262,6 +269,8 @@ def get_estimators():
 
     finally:
         sklearnex.unpatch_sklearn()
+        if not flag:
+            del os.environ["SKLEARNEX_PREVIEW"]
```
samir-nasibli, line 273: What if key is not exists?

sklearnex/tests/test_monkeypatch.py
```diff
@@ -27,142 +27,166 @@ def test_monkey_patching():
...
-
-        sklearnex.patch_sklearn(t)
-
-        class_module = getattr(p, n).__module__
-        assert class_module.startswith("daal4py") or class_module.startswith(
-            "sklearnex"
-        ), "Patching has completed with error."
-
-    sklearnex.unpatch_sklearn()
+    try:
```
olegkkruglov, line 30: is it good to add it to test? will we catch any exceptions and errors this way?

icfaust: https://docs.python.org/3/reference/compound_stmts.html#finally

icfaust: It essentially insured that the sklearn patch will be reverted even if an exception/error is raised

icfaust:
private CI rerun to see if linear regression fail on GPU is related http://intel-ci.intel.com/eec0d542-000a-f1a2-b8f0-a4bf010d0e2e

sklearnex/tests/test_monkeypatch.py
```diff
@@ -27,150 +27,170 @@ def test_monkey_patching():
...
+        sklearnex.patch_sklearn(["LogisticRegression"])
 
-    assert RandomForestRegressor.__module__.startswith("sklearn")
-    assert KNeighborsRegressor.__module__.startswith("sklearn")
-    if daal_check_version((2024, "P", 1)):
-        assert LogisticRegression.__module__.startswith("sklearnex")
-    else:
-        assert LogisticRegression.__module__.startswith("daal4py")
-    assert SVC.__module__.startswith("sklearn")
+        from sklearn.ensemble import RandomForestRegressor
```
samir-nasibli, line 95: I think it make sense decorate all this kind of tests with a decorator that tries test suit and on finally does sklearnex unpatch.

icfaust:
CI failure comes from unrelated daal4py example failure.

icfaust:
CI rerun due to private CI issue: http://intel-ci.intel.com/eec9ad1f-d7de-f13f-a11c-a4bf010d0e2e

ahuber21 (approved):

samir-nasibli (approved):

icfaust:
Turns out that whoever programmed the patch_map was abusing lru_cache to make get_patch_map into a state machine, rather than setting a module attribute or making a class they went with this. I can't dump the lru_cache without doing a much larger refactor, but I think I can devise a way to at least make it not be buggy while also allowing for preview. This absolutely needs a set of tickets for a rewrite, and is a ticking timebomb for anyone who does anything other than just vanilla patch_sklearn and unpatch_sklearn when combined with preview.

.ci/pipeline/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
   - script: |
       python -m pip install --upgrade pip setuptools
       pip install isort black[jupyter]==24.1.1
-      isort --check . && black --check .
+      isort --profile black --check . && black --check .
```
samir-nasibli, line 56: Why it is in bug fix scope?)

PR #1696, opened 2024-02-05
[bug] fix ensemble algo _onedal_gpu_supported logic

icfaust:
# Description
The gpu support check looks to see if sample_weights are included, it should not allow sample weights, but the logic is opposite of desired

- "swap is not None" to "is None"
- add to deselected_tests as necessary


ethanglaser:
Could it be related to some of the issues in these files? https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/ensemble/tests/test_forest.py#L48 or https://github.com/intel/scikit-learn-intelex/blob/main/onedal/ensemble/tests/test_random_forest.py#L41

icfaust:
@ethanglaser I have re-enabled those tests you mentioned, but given the different design between GPU and CPU they will definitely fail. I guess it'll be interesting to see what comes out. I am leaning towards leaving those sklearnex test changes out of this PR and will make a ticket for it.

ethanglaser:
@icfaust yeah no pressure to fix those in this PR, was just curious if there were any updates to those tests. Looks like et_classifier is passing at least, so we could leave that one if we want some et testing on our side. Looks like `test_min_samples_split` and all `oob` still have some fails (and maybe some new conformance fails?) as well.

sklearnex/ensemble/_forest.py
```diff
@@ -747,7 +747,7 @@ def _onedal_gpu_supported(self, method_name, *data):
                         or self.estimator.__class__ == DecisionTreeClassifier,
                         "ExtraTrees only supported starting from oneDAL version 2023.1",
                     ),
-                    (sample_weight is not None, "sample_weight is not supported."),
+                    (sample_weight is None, "sample_weight is not supported."),
```
samir-nasibli, line 750: good catch

icfaust:
If tickets don't exist already, the differences between GPU and CPU output need to be investigated.

icfaust:
rerun of CI http://intel-ci.intel.com/eec9abd4-7baf-f10e-ab02-a4bf010d0e2e

ethanglaser:
Are we expecting the same results on GPU and CPU? If so, we may still want to make these as TODOs. But its good to re-enable the test and set expected values for our current implementation at least

ethanglaser (approved):

Alexsandruss (approved):

ethanglaser:
https://github.com/Mergifyio backport rls/2024.2.0-rls

PR #1697, opened 2024-02-05
enh: Add SPMD interfaces for Covariance

ethanglaser:
# Description
Addition of spmd interfaces for covariance with example usage. 

Reference https://github.com/intel/scikit-learn-intelex/pull/1673 for most recent similar change.


samir-nasibli (approved):
Looks good to me. Very consistent with other estimators.
Thank you!

PR #1699, opened 2024-02-06
Correct DPCPP detection in setup

Alexsandruss:
Update dpcpp compiler detection to last changes on Windows platform:
- Watch for `icpx` executable instead of `DPCPPROOT` variable (it's not set in last versions)
- Disable interprocedural optimization on Windows due to icx and pybind11 linker flags incompatibility

### DIFF
--- a/daal4py/oneapi/__init__.py
+++ b/daal4py/oneapi/__init__.py
@@ -18,6 +18,7 @@ import platform
 
 if "Windows" in platform.system():
     import os
+    import shutil
     import sys
     import sysconfig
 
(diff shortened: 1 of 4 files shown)


Alexsandruss:
Combined CI job with https://github.com/oneapi-src/oneDAL/pull/2659: http://intel-ci.intel.com/eec606f5-57f6-f156-956b-a4bf010d0e2e

ethanglaser (approved):
Updated build logs look better and changes look good. A couple comments that we may want to explore moving forward:
1. Complete removal of DPCPPROOT usage, as this is no longer supported by compiler
2. Adding a CI check that would fail in this case to better detect this in the future

Assuming there were no issues with linux then LGTM

Alexsandruss:
@Mergifyio backport rls/2024.2.0-rls

PR #1700, opened 2024-02-07
enh: re-enabling spmd rf interfaces

ethanglaser:
# Description
When paired with https://github.com/intel/scikit-learn-intelex/pull/1696, onedal distributed random forest interfaces will be reached
 
This is a draft and very much open to discussion as far as how we want to integrate it and any revisions, but its functional


ethanglaser:
Leads to CI fail: Estimator RandomForestRegressor should not set any attribute apart from parameters during init. Found attributes ['_onedal_factory'].

icfaust:
I know its still in draft, but I am not sure why a class attribute change isn't overwriting properly, and making an instance attribute is necessary

here is a toy example (can be copied into a python terminal):

```python
class example():
    text = "hello world"
    def print(self):
         print(self.text)

class child(example):
     text = "overwrite"

temp = example()
temp.print()
temp = child()
temp.print()
```

ethanglaser:
> I know its still in draft, but I am not sure why a class attribute change isn't overwriting properly and making an instance attribute is necessary

Yeah I think will need to change this up. It's confusing to me why this is happening, because to me the example is a clear simplified replication of the situation. Will keep poking around to see about a resolution.

ethanglaser:
The issue seems to be resolved by adding in __class__ to the usage...to be honest not a fan of the way python handles this

But with __class__ added it is once again functional (without adding _onedal_factory to self)

sklearnex/ensemble/_forest.py
```diff
@@ -843,14 +841,12 @@ def __init__(
...
+            self._onedal_factory, onedal_RandomForestRegressor
         ):
             self._onedal_factory = onedal_RandomForestRegressor
-        elif (
-            self.estimator.__class__ == ExtraTreeRegressor
-            and self._onedal_factory != onedal_ExtraTreesRegressor
+        elif self.estimator.__class__ == ExtraTreeRegressor and not isinstance(
+            self._onedal_factory, onedal_ExtraTreesRegressor
         ):
             self._onedal_factory = onedal_ExtraTreesRegressor
```
ethanglaser, line 851: Getting conformance fails `check_no_attributes_set_in_init` even though this already exists in master, I guess will want to explore options for replacing self._onedal_factory with _onedal_factory in these places

sklearnex/ensemble/_forest.py
```diff
@@ -453,16 +453,14 @@ def __init__(
...
-            self.estimator.__class__ == DecisionTreeClassifier
-            and self._onedal_factory != onedal_RandomForestClassifier
+        if self.estimator.__class__ == DecisionTreeClassifier and not isinstance(
+            self._onedal_factory, onedal_RandomForestClassifier
         ):
-            self._onedal_factory = onedal_RandomForestClassifier
-        elif (
-            self.estimator.__class__ == ExtraTreeClassifier
-            and self._onedal_factory != onedal_ExtraTreesClassifier
+            _onedal_factory = onedal_RandomForestClassifier
```
icfaust, line 459: Maybe we switch these to a warning? It looks like _onedal_factory would be with these changes only set during the operation of ```__init__```.  Essentially the purpose of these checks is to make sure someone doesn't use a random splitter with random forest, and a decision tree splitter with extra trees if they were to spin their own class based off of ForestRegressor or ForestClassifier

icfaust (approved):

icfaust:
Merge when ready

ethanglaser:
@Mergifyio backport rls/2024.2.0-rls

PR #1727, opened 2024-02-19
DEP: deprecate to use sklearn estimators in onedal4py SVM

samir-nasibli:
# Description
Trying to remove sklearn's BaseEstimator from OneDAL4py's SVM.
OneDAL4py estimators should be sklearn compatible but sklearn like iface.

After dispatching to onedal4py backend, no any base/non-base sklearn estimators should be used.

### DIFF
--- a/onedal/svm/svm.py
+++ b/onedal/svm/svm.py
@@ -20,7 +20,6 @@ from numbers import Number, Real
 
 import numpy as np
 from scipy import sparse as sp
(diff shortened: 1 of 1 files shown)


samir-nasibli:
It fails all onedal4py tests, so tests have to be updated as well for svm then.
sklearn tests passed.

icfaust:
Changes in #1835 complete what is in this PR, therefore this PR can be closed.

PR #1737, opened 2024-02-21
[enhancement] enable dpnp/dpctl testing in test_patching

icfaust:
# Description
So far test_patching has been limited to cpu algorithms, this PR will enable get_dataframes_and_queues to test gpu support if available.

- remove _numpy only in pytest.mark.parameterize("dataframe,queue")
- Modify _copy_to_usm to be recursive for iterable outputs (failures related to the kneighbors method)
- disable SYCL testing with fp16 when not hardware supported
- Add score method which avoids wrap_output_data in classifier estimators (will break sklearn's type_of_target without array_api support)
- add array_api-like checks to predict_log_proba to access the namespaces' logarithm call
- Introduce get_namespace to sklearnex.utils (likely to be changed soon)
- add support_usm_ndarray to Kmeans' score (enable host evaluation for dpnp/dpctl)
- fix issue in PCA (just out of preview), which doesn't properly handle fit_transform by using _onedal_transform directly (doesn't convert to numpy)
- Add LogisticRegression(solver="newton-cg") to support testing on GPU onedal LogisticRegression implementation
- Fix missing type of target check for LogisticRegression on GPU, otherwise causes ValueErrors
- Fix issue with LocalOutlierFactor predict, which was missing a proper array indexing for dpnp/dpctl

NOTE: so far dpnp and dpctl are only tested in public CI on linux with Python3.9 and Python3.10. These new tests run rather slow and adds 10 minutes to testing. This means there are some serious performance limitations in many of these methods, likely due to direct conversion to numpy arrays which are then used. (this means it is a numpy array with dtype=object). Follow on tickets will need to be written to be involved in array_api support, because this represents what is likely to happen there.

Tasks
- [x] Remove numpy check
- [x] Find failure points in public CI
- [x] Find failure points in private CI
- [x] Correct issues
- [x] Pass public CI
- [x] Pass private CI


samir-nasibli (approved):
Looks good to me. Assuming CI checks

icfaust:
625 test case failures in sklearn1.2 :(

icfaust:
423 fails remaining

icfaust:
Looks like pairwise distances and kmeans on gpu have issues, ontop of the score method on classification (as shown on private CI).

sklearnex/_device_offload.py
```diff
@@ -200,9 +209,14 @@ def _copy_to_usm(queue, array):
...
-    mem.copy_from_host(array.tobytes())
-    return usm_ndarray(array.shape, array.dtype, buffer=mem)
+    if hasattr(array, "nbytes"):
+        mem = MemoryUSMDevice(array.nbytes, queue=queue)
+        mem.copy_from_host(array.tobytes())
+        return usm_ndarray(array.shape, array.dtype, buffer=mem)
+    else:
+        if isinstance(array, Iterable):
+            array = [_copy_to_usm(queue, i) for i in array]
+        return array
```
icfaust, line 219: This solves issues with the kneighbors method, which can return a list

icfaust:
8 remaining

sklearnex/_device_offload.py
```diff
@@ -200,9 +201,25 @@ def _copy_to_usm(queue, array):
...
+
+if dpnp_available:
+
+    def _convert_to_dpnp(array):
+        if isinstance(array, usm_ndarray):
+            return dpnp.array(array, copy=False)
+        elif isinstance(array, Iterable):
+            for i in range(len(array)):
+                array[i] = _covert_to_dpnp(array[i])
+        return array
```
icfaust, line 222: This solves an issue with PCA where a list containing a None can be returned, which cannot be converted to a list of dpnp arrays.

icfaust:
I am going to continue looking at why Kmeans is passing sklearn conformance testing on gpu, but otherwise this PR is ready to go

icfaust:
Currently investigating KMeans gpu passing private CI, I have forced an assert for gpu offloading in daal4py's _device_offload when the object is KMeans: https://github.com/intel/scikit-learn-intelex/compare/main...icfaust:scikit-learn-intelex:test/test_kmeans?expand=1 and I am running private CI against this repo: http://intel-ci.intel.com/eeeaa574-a9c3-f1c6-a2b8-a4bf010d0e2e

icfaust:
> Currently investigating KMeans gpu passing private CI, I have forced an assert for gpu offloading in daal4py's _device_offload when the object is KMeans: https://github.com/intel/scikit-learn-intelex/compare/main...icfaust:scikit-learn-intelex:test/test_kmeans?expand=1 and I am running private CI against this repo: http://intel-ci.intel.com/eeeaa574-a9c3-f1c6-a2b8-a4bf010d0e2e

All GPU testing for kmeans is currently deactivated. This PR will continue that, but extra care will need to go into the KMeans out of preview because of this. If this PR is merged before KMeans out of preview, it will become a prerequisite to reintroduce this testing.

sklearnex/tests/test_patching.py
```diff
@@ -129,6 +133,8 @@ def test_standard_estimator_patching(caplog, dataframe, queue, dtype, estimator,
             and dtype in [np.uint32, np.uint64]
         ):
             pytest.skip("Windows segmentation fault for Ridge.predict for unsigned ints")
+        elif estimator == "KMeans" and queue and queue.sycl_device.is_gpu:
+            pytest.skip("KMeans does not support GPU queues")
```
ethanglaser, line 137: This may change when out of preview is finalized

sklearnex/tests/test_patching.py
```diff
@@ -56,18 +56,24 @@
...
 @pytest.mark.parametrize("dtype", DTYPES)
-@pytest.mark.parametrize(
-    "dataframe, queue", get_dataframes_and_queues(dataframe_filter_="numpy")
-)
+@pytest.mark.parametrize("dataframe, queue", get_dataframes_and_queues())
 @pytest.mark.parametrize("metric", ["cosine", "correlation"])
 def test_pairwise_distances_patching(caplog, dataframe, queue, dtype, metric):
     with caplog.at_level(logging.WARNING, logger="sklearnex"):
+        if dtype == np.float16 and queue and not queue.sycl_device.has_aspect_fp16:
+            pytest.skip("Hardware does not support fp16 SYCL testing")
```
ethanglaser, line 64: Would we want a similar check for fp64 support?

daal4py/sklearn/cluster/_k_means_0_23.py
```diff
@@ -684,29 +684,5 @@ def predict(
...
-            are assigned equal weight.
-
-        Returns
-        -------
-        labels : ndarray of shape (n_samples,)
-            Index of the cluster each sample belongs to.
-        """
-        return super().fit_predict(X, y, sample_weight)
+    fit_predict = support_usm_ndarray()(KMeans_original.fit_predict)
+    score = support_usm_ndarray()(KMeans_original.score)
```
ethanglaser, line 688: Its more concise but unless we are doing this elsewhere I'd lean towards the more verbose option. but open for discussion

sklearnex/decomposition/pca.py
ethanglaser, line 1: This looks like it might resolve the datatype/fp64 issue I've encountered in windows testing...checking it now..looks very promising!

sklearnex/linear_model/logistic_regression.py
```diff
@@ -132,6 +132,8 @@ def predict(self, X):
                 X,
             )
 
+        predict = wrap_output_data(_predict)
+
```
ethanglaser, line 136: Can you elaborate on what impact this has? just curious

sklearnex/svm/nusvc.py
```diff
@@ -78,39 +86,6 @@ def __init__(
...
-            Fitted estimator.
-
-        Notes
-        -----
-        If X and y are not C-ordered and contiguous arrays of np.float64 and
-        X is not a scipy.sparse.csr_matrix, X and/or y may be copied.
-
-        If X is a dense array, then the other methods will not support sparse
-        matrices as input.
-        """
```
ethanglaser, line 113: what's reason for removing? because its exact copy of sklearn?

sklearnex/svm/nusvc.py
```diff
@@ -191,13 +165,61 @@ def predict_proba(self, X):
...
+            return log_ufunc(self.predict_proba(X))
+
     else:
 
         @property
         def predict_proba(self):
             self._check_proba()
             return self._predict_proba
 
+        def _predict_log_proba(self, X):
```
ethanglaser, line 213: sklearn version 0.* only has _predict_log_proba? (and not predict_log_proba)?

ethanglaser:
CI looks close but some test_stacking fail - could be related to predict/_predict in rf classifier?

sklearnex/linear_model/logistic_regression.py
icfaust: the score method needs to access numpy array versions of data in order to properly calculate using sklearn. accessing the result of predict without having turning it into a dpctl/dpnp tensor is required, which means having access to predict without wrap_output_data. I make a new method called _predict which then can be used in score, and then expose predict by directly wrapping with wrap_output_data.

sklearnex/svm/nusvc.py
icfaust: Thats right, it can be generated by copying the docs from sklearn's version (done below). The direct copy of the docs is only necessary when the attribute is wrapped with sklearn's available_if wrapper (which is horrible). That occurs in SVM algos and in LocalOutlierFactor

sklearnex/svm/nusvc.py
icfaust: There is some weirdness in the older versions on how predict_log_proba is exposed to the user. You are right.

sklearnex/tests/test_patching.py
icfaust: So far CI testing has shown no issues with fp64 support, but if this is something popping up on windows we can look at running special CI jobs against this PR and make requisite changes for it.

sklearnex/tests/test_patching.py
icfaust: Definitely correct. I will have to leave a note in the other PR to specifically remove these lines.

sklearnex/linear_model/logistic_regression.py
```diff
@@ -160,6 +162,16 @@ def predict_log_proba(self, X):
                 X,
             )
 
+        @wrap_output_data
+        def score(self, X, y, sample_weight=None):
+            if hasattr(y, "__sycl_usm_array_interface__"):
+                if hasattr(y, "__array_namespace__"):
+                    y = y.__array_namespace__().asnumpy(y)
+                else:
+                    y = y.asnumpy()
```
samir-nasibli, line 171: I think it is better to generlize wrap_output_data or use usm_ndrarray_support instead of such conversions here

sklearnex/_device_offload.py
```diff
@@ -200,9 +201,25 @@ def _copy_to_usm(queue, array):
...
-    mem.copy_from_host(array.tobytes())
-    return usm_ndarray(array.shape, array.dtype, buffer=mem)
+    if hasattr(array, "nbytes"):
+        mem = MemoryUSMDevice(array.nbytes, queue=queue)
+        mem.copy_from_host(array.tobytes())
+        return usm_ndarray(array.shape, array.dtype, buffer=mem)
+    else:
+        if isinstance(array, Iterable):
+            array = [_copy_to_usm(queue, i) for i in array]
+        return array
```
samir-nasibli, line 211: I would like to avoid this recursion here. It seems to me more safe

```suggestion
    def copy_to_usm(queue, array): #better name is needed
        mem = MemoryUSMDevice(array.nbytes, queue=queue)
        mem.copy_from_host(array.tobytes())
        return usm_ndarray(array.shape, array.dtype, buffer=mem)
   
    if hasattr(array, "nbytes"):
        return copy_to_usm(queue, array)
    else:
        if isinstance(array, Iterable):
            array = [copy_to_usm(queue, i) for i in array]
        return array
```

sklearnex/tests/test_patching.py
ethanglaser: There wouldnt be an issue on PVC since its supporting, but I am guessing there may be on DG1 and DG2. Will keep an eye out in the future and can add as needed.

sklearnex/_device_offload.py
```diff
@@ -200,9 +201,25 @@ def _copy_to_usm(queue, array):
...
+
+if dpnp_available:
+
+    def _convert_to_dpnp(array):
+        if isinstance(array, usm_ndarray):
+            return dpnp.array(array, copy=False)
+        elif isinstance(array, Iterable):
+            for i in range(len(array)):
+                array[i] = _convert_to_dpnp(array[i])
+        return array
```
samir-nasibli, line 222: I would like to move such stuf to separate module for dp tensors/ndarrays support

icfaust: perfect thing to tie into the next array_api PR

icfaust:
I have slightly modified CI to see where long duration tests exist in sklearnex with dpnp/dpctl support on linux. This will be reverted later.

icfaust:
Things to fix in the future: 

LogisticRegression's decision_function is costing ~1 minute of CI time, but will run through. Why isn't this the case also for linear regression?
All regression estimators score method, which uses r2 score and is likely not working well with array_api namespaces for sklearn < 1.2 (will run through but is slow).

sklearnex/svm/nusvc.py
ethanglaser, line 1: only comment on svm stuff is looks like all 3 added functions are the exact same so theoretically could place them in BaseSVC class - however would then need to add any functions they call to base with NotImplementedError. Plus it looks like we already have overlapping functions so I think for this PR its not a concern

sklearnex/linear_model/logistic_regression.py
```diff
@@ -198,6 +227,10 @@ def _onedal_gpu_fit_supported(self, method_name, *data):
                     (self.warm_start == False, "Warm start is not supported."),
                     (self.l1_ratio is None, "l1 ratio is not supported."),
                     (sample_weight is None, "Sample weight is not supported."),
+                    (
+                        type_of_target(y) == "binary",
+                        "Only binary classification is supported",
+                    ),
```
ethanglaser, line 233: nice additional logreg support w this and test_patching

sklearnex/utils/_namespace.py
ethanglaser, line 1: no objections but let's go through it in tomorrow's discussion

icfaust:
Rerun: http://intel-ci.intel.com/eef59e06-84fa-f17f-b1e8-a4bf010d0e2e

icfaust:
http://intel-ci.intel.com/eef62d12-5eae-f129-821d-a4bf010d0e2e

ethanglaser:
With lower-precision GPUs, just for fun: http://intel-ci.intel.com/eef680ea-4ebf-f124-8ebc-a4bf010d0e2e

icfaust:
> With lower-precision GPUs, just for fun: http://intel-ci.intel.com/eef680ea-4ebf-f124-8ebc-a4bf010d0e2e

Good call, it looks like there should be a lot of fails on GPU that aren't happening in the new tests. Will need to investigate.

icfaust:
@ethanglaser Thanks for the private CI run, it now makes a little bit more sense. If implemented correctly in the onedal folder, data will use the function _convert_to_supported (https://github.com/intel/scikit-learn-intelex/blob/main/onedal/datatypes/_data_conversion.py#L80), which has a check for fp64 support in it. It will down-convert to fp32 if not supported.  The failed IncrementalCovariance tests are because it lacks _convert_to_supported. If we pull in this PR, update #1674, and rerun the tests with lower precision, these new tests should fail also for IncrementalCovariance.

The second issue is that the test_patching seg faulted for Ridge, and probably will for Lasso and ElasticNet.  These 3 daal4py estimators do not have the same guard protection for gpu offloading like daal4py's KMeans.  I will need to add some pytest skips for these and will need to rerun.

The final question is: should we add an fp16 support check to _convert_to_supported? I'd like to do it in a follow-up PR just to limit the growth in this one.

sklearnex/tests/test_patching.py
```diff
@@ -122,8 +122,17 @@ def test_standard_estimator_patching(caplog, dataframe, queue, dtype, estimator,
...
+        if queue:
+            if dtype == np.float16 and not queue.sycl_device.has_aspect_fp16:
+                pytest.skip("Hardware does not support fp16 SYCL testing")
+            elif queue.sycl_device.is_gpu and estimator in [
+                "KMeans",
+                "ElasticNet",
+                "Lasso",
+                "Ridge",
+            ]:
+                pytest.skip(f"{estimator} does not support GPU queues")
```
samir-nasibli, line 134: Why we are removing them from GPU testing?

icfaust: https://github.com/intel/scikit-learn-intelex/pull/1737#issuecomment-2046586614 written here

icfaust:
Another run after fixing the introduced error: http://intel-ci.intel.com/eef72327-da06-f16a-b38e-a4bf010d0e2e

icfaust:
Retry with changes to cover fp64 support on copy_to_usm, skip float64 tests if aspect is missing. Internal conversions of other datatypes to float64 via check_array will cover float64 inputs to the oneDAL backend, meaning test coverage loss is minimal (i.e. circumstances where float64 data is sent to gpu when using config_context). http://intel-ci.intel.com/eef731b4-3c1b-f127-a4cc-a4bf010d0e2e

ethanglaser:
Looks like it adds roughly 4000 tests and 10 mins of runtime to PVC tests step in internal CI, just FYI

ethanglaser (approved):

icfaust:
rerun of lower precision GPUs to verify merge of main speeds new tests: http://intel-ci.intel.com/eef7c87b-710b-f19b-a4cc-a4bf010d0e2e

icfaust:
rerun of lower precision GPUs to verify merge of main speeds new tests: http://intel-ci.intel.com/eef7d9d9-fe4e-f13c-a2b8-a4bf010d0e2e

icfaust:
rerun of lower precision GPUs to verify merge of main speeds new tests: http://intel-ci.intel.com/eef7ec24-ca63-f164-9fcc-a4bf010d0e2e

icfaust:
http://intel-ci.intel.com/eef83a27-3091-f194-bf04-a4bf010d0e2e

icfaust:
Full green and ready to go.

samir-nasibli:
@icfaust thank you for the integration!
There are several fixes and important initial features for further development.

PR #1738, opened 2024-02-21
MAINT: aligning gpu interfaces y param with sklearn

ethanglaser:
# Description
scikit-learn provides dummy y variable in cluster/compute ([example](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/decomposition/_incremental_pca.py#L201)) interfaces for general alignment - we mostly have this implemented in our daal4py/sklearnex cpu interfaces already. It's not used in sklearn or in sklearnex but provides uniform interfaces and will be useful in resolving common sklbench errors.

also adds queue as kwarg when passing in cases where its not already done.

### DIFF
--- a/onedal/cluster/kmeans.py
+++ b/onedal/cluster/kmeans.py
@@ -387,7 +387,7 @@ class KMeans(_BaseKMeans):
         self.algorithm = algorithm
         assert self.algorithm == "lloyd"
 
-    def fit(self, X, queue=None):
+    def fit(self, X, y=None, queue=None):
         return super()._fit(X, self._get_backend("kmeans", "clustering", None), queue)
 
     def predict(self, X, queue=None):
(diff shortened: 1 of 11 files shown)


samir-nasibli:
@ethanglaser could you please reference sklearn_bench where this fix is required?

samir-nasibli:
> @ethanglaser could you please reference sklearn_bench where this fix is required?

Is it required for spmd ifaces?

samir-nasibli (approved):

ethanglaser:
> > @ethanglaser could you please reference sklearn_bench where this fix is required?
> 
> Is it required for spmd ifaces?

It mostly depends on config files setup, but in cases where same datasets are used for multiple algos (ie higgs for kmeans, linreg) it will pass dummy y data and breaks spmd interfaces that dont have dummy data in this case. Since this would align with our d4p/cpu and stock sklearn interfaces, and avoids this issue, it probably would be good do here.

ethanglaser:
Based on internal CI there is just an issue with empty result options on windows

ethanglaser:
Any objections on this? Is it okay with API or is it considered breaking?

PR #1739, opened 2024-02-22
Cache results of daal/sklearn version check

Alexsandruss:
Change of caching scheme for daal/sklearn version check:
- Replace dictionary of stored sklearn version checks with `lru_cache`
- Add `lru_cache` for daal version check
- Exact daal version value in `check_daal_version` instead of function as optional argument

### DIFF
--- a/daal4py/sklearn/_utils.py
+++ b/daal4py/sklearn/_utils.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
+import functools
 import os
 import sys
 import warnings
@@ -75,23 +76,24 @@ def get_daal_version() -> DaalVersionTuple:
     return int(dv()[0:4]), str(dv()[10:11]), int(dv()[4:8])
 
 
+@functools.cache
 def daal_check_version(
     required_version: Tuple[Any, ...],
-    _get_daal_version: Callable[[], DaalVersionTuple] = get_daal_version,
+    daal_version: Tuple[Any, ...] = get_daal_version(),
 ) -> bool:
     """Check daal version provided as (MAJOR, STATUS, MINOR+PATCH)
 
     This function also accepts a list or tuple of daal versions. It will return true if
-    any version in the list/tuple is <= `_get_daal_version()`.
+    any version in the list/tuple is <= `daal_version`.
     """
     if isinstance(required_version[0], (list, tuple)):
-        # a list of version candidates was provided, recursively check if any is <= _get_daal_version
+        # a list of version candidates was provided, recursively check if any is <= daal_version
         return any(
-            map(lambda ver: daal_check_version(ver, _get_daal_version), required_version)
+            map(lambda ver: daal_check_version(ver, daal_version), required_version)
         )
 
     major_required, status_required, patch_required = required_version
-    major, status, patch = _get_daal_version()
+    major, status, patch = daal_version
 
     if status != status_required:
         return False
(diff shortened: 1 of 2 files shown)


icfaust (approved):
Looks fine to me, good call.

daal4py/sklearn/_utils.py
```diff
@@ -75,23 +76,24 @@ def get_daal_version() -> DaalVersionTuple:
     return int(dv()[0:4]), str(dv()[10:11]), int(dv()[4:8])
 
 
+@functools.lru_cache(maxsize=256, typed=False)
 def daal_check_version(
     required_version: Tuple[Any, ...],
-    _get_daal_version: Callable[[], DaalVersionTuple] = get_daal_version,
+    daal_version: Tuple[Any, ...] = get_daal_version(),
```
ahuber21, line 82: There's a subtle change here. Previously we called `get_daal_version()` on runtime, i.e. when the function `daal_check_version()` is called. With your change this is changed to load time. In particular, `get_daal_version()` is only executed once, and that is when the module is loaded.
Right now, I don't see how this is an issue, as `get_daal_version()` normally produces a constant result. But just wanted to double check that you considered this subtlety.

ahuber21:
General question: Why is this necessary. Caching itself has some overhead and to me the operations within the function look very cheap. I don't see the point right away.

Alexsandruss:
@ahuber21 

> General question: Why is this necessary. Caching itself has some overhead and to me the operations within the function look very cheap. I don't see the point right away.

At least, sklearn version check without caching has relatively high computation cost for cases where oneDAL computation costs few milliseconds.

Alexsandruss:
Script to measure computation cost of checks:
```python
from daal4py.sklearn._utils import sklearn_check_version, daal_check_version
from timeit import default_timer as timer


def measure_func(func, *args, n_times=3, **kwargs):
    for _ in range(n_times):
        t0 = timer()
        res = func(*args, **kwargs)
        t1 = timer()
        print(round((t1 - t0) * 1e9), "ns")
    print("result:", res)


print("sklearn_check_version - 1.3")
measure_func(sklearn_check_version, "1.3")

print("sklearn_check_version - 1.1")
measure_func(sklearn_check_version, "1.1")

print("sklearn_check_version - 1.5")
measure_func(sklearn_check_version, "1.5")

print("daal_check_version - (2023, \"P\", 200)")
measure_func(daal_check_version, (2023, "P", 200))

print("daal_check_version - (2021, \"P\", 205)")
measure_func(daal_check_version, (2021, "P", 205))

print("daal_check_version - (2026, \"P\", 205)")
measure_func(daal_check_version, (2026, "P", 205))
```
Output:
```
sklearn_check_version - 1.3
682 ns
296 ns
155 ns
result: True
sklearn_check_version - 1.1
208 ns
149 ns
115 ns
result: True
sklearn_check_version - 1.5
43207 ns
338 ns
145 ns
result: False
daal_check_version - (2023, "P", 200)
595 ns
185 ns
151 ns
result: True
daal_check_version - (2021, "P", 205)
1455 ns
193 ns
123 ns
result: True
daal_check_version - (2026, "P", 205)
690 ns
194 ns
142 ns
result: False
```

daal4py/sklearn/_utils.py
Alexsandruss: daal version is not expected to change during process life, so `get_daal_version` has static result.

ahuber21: Agree, just wanted to mention it :)

PR #1740, opened 2024-02-22
Logistic Regression SPMD timeout temporary resolution

md-shafiul-alam:
- Bump required version to 2024.4.0 to skip logistic regression spmd example during CI runs to avoid time out, will be reverted to 2024.1.0 once the issue is resolved

### DIFF
--- a/tests/run_examples.py
+++ b/tests/run_examples.py
@@ -143,7 +143,7 @@ req_version["kmeans_spmd.py"] = (2024, "P", 200)
 req_version["knn_bf_classification_spmd.py"] = (2023, "P", 100)
 req_version["knn_bf_regression_spmd.py"] = (2023, "P", 100)
 req_version["linear_regression_spmd.py"] = (2023, "P", 100)
-req_version["logistic_regression_spmd.py"] = (2024, "P", 100)
+req_version["logistic_regression_spmd.py"] = (2024, "P", 400)
 
 req_device = defaultdict(lambda: [])
 req_device["basic_statistics_spmd.py"] = ["gpu"]


ethanglaser:
Before merging it may be worth syncing with @avolkov-intel on this issue. For me I haven't encountered problems running spmd logreg via sklbench (contrary to previous kmeans issue)

samir-nasibli (commented):
This issue is not related with backend version. So maybe we have another mechanism to skip this with a meaningful comment?

md-shafiul-alam:
> This issue is not related with backend version. So maybe we have another mechanism to skip this with a meaningful comment?

It is true that the issue is not related to backend version, but this forces us to look back after the next release. A ticket has been created to actually investigate and resolve this. If there are no other issues except CI, we can even ignore it for now.

ethanglaser:
If we don't want to complete "deselect" the example, we could see if small number of rows yields the same issue, and if not use that temporarily

avolkov-intel:
This is a sporadic failure as it doesn't occur in 100% runs and now we are expecting difficulties with reproducing it. Therefore, I think it is better to keep this test enabled to see the timeout in nightly when it happens.

md-shafiul-alam:
> This is a sporadic failure as it doesn't occur in 100% runs and now we are expecting difficulties with reproducing it. Therefore, I think it is better to keep this test enabled to see the timeout in nightly when it happens.

Agree, however, we have already observed the failures 4/5 times in last 7/8 nightlies. I think we should investigate this in the coming sprint.

ethanglaser (approved):

PR #1742, opened 2024-02-23
[CI] fix clashing between isort and black that can occur with long imports

icfaust:
# Description
isort wants to newline imports using \\, where black wants to use ( ) . isort can be made to follow black's rules with this change.

- add --profile black to isort


ahuber21 (approved):
Great, that should have been there from the beginning

Issue #1329, opened 2023-06-13
Unable to use sklearnex after building daal4py from source


rakshithgb-fujitsu:
I built oneDAL from [this PR](https://github.com/oneapi-src/oneDAL/pull/2396), after successfull build I tried to build daal4py and use it. Though daal4py built without any issues, I'm getting error during import:

```
>>> import daal4py
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/.local/lib/python3.8/site-packages/daal4py/__init__.py", line 38, in <module>
    from daal4py._daal4py import *
ImportError: /home/ubuntu/sandbox/trial_pr/daal/__release_lnx_clang_ref/daal/latest/lib/intel64/libonedal_core.so.1: undefined symbol: sdot_
```

daal4py build:
```
$ export LD_LIBRARY_PATH=/home/ubuntu/sandbox/trial_pr/daal/__release_lnx_clang_ref/daal/latest/lib/intel64:/home/ubuntu/sandbox/trial_pr/daal/__release_lnx_clang_ref/tbb/latest/lib/intel64:/opt/OpenBLAS/lib:$LD_LIBRARY_PATH
$ export DALROOT=/home/ubuntu/sandbox/trial_pr/daal/__release_lnx_clang_ref/daal/latest
$ export NO_DIST=1
$ export NO_STREAM=1
$ export OFF_ONEDAL_IFACE=1
$ python3 setup.py install --single-version-externally-managed --record=record.txt
```

More details on the oneDAL build:
Compiler: clang (10.0.0-4ubuntu1)
Backend (new feature in the PR): ref (uses openblas v0.3.23)

oneDAL build command used:
`make -f makefile daal oneapi_c BACKEND_CONFIG=ref PLAT=lnx32e COMPILER=clang `

I understand this PR brings in a lot of changes, but any clue on how to resolve this would be a great help.

icfaust:
Does compiling oneapi_c without BACKEND_CONFIG=ref and setting that version to DALROOT cause the same undefined symbol?

rakgb-fujitsu:
@icfaust resolved this issue by making few changes in the make file of the PR, it was expecting openblas libs. Daal4py now works without any issues, however building sklearnex I'm now getting runtime errors during patching:

```
Traceback (most recent call last):
[...]

PR #1744, opened 2024-02-23
Fix: Do not import onedal when OFF_ONEDAL_IFACE=1

ajay-fuji:
# Description
- sklearnex can be built/installed without **onedal** interface by setting `OFF_ONEDAL_IFACE=1` during installation.
    - Due to which onedal interface is not built/installed and when we try to `import sklearnex` we get import error because `sklearnex/__init__.py` has `import onedal` statement without checking `OFF_ONEDAL_IFACE` flag.
  
## Changes proposed in this pull request:
- This PR fixes import error by adding conditional check of `OFF_ONEDAL_IFACE` flag when importing **onedal**

Reference - [Github issue](https://github.com/intel/scikit-learn-intelex/issues/1329)

### DIFF
--- a/sklearnex/__init__.py
+++ b/sklearnex/__init__.py
@@ -1,6 +1,6 @@
 # ==============================================================================
 # Copyright 2021 Intel Corporation
-# Copyright contributors to the oneDAL project
+# Copyright 2024 Fujitsu Limited
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/sklearnex/__init__.py
+++ b/sklearnex/__init__.py
@@ -51,7 +51,7 @@ __all__ = [
     "unpatch_sklearn",
     "utils",
 ]
-onedal_iface_flag = os.getenv("OFF_ONEDAL_IFACE", "0")
+onedal_iface_flag = os.environ.get("OFF_ONEDAL_IFACE", "0")
 if onedal_iface_flag == "0":
     from onedal import _is_spmd_backend
     from onedal.common.hyperparameters import get_hyperparameters
--- a/sklearnex/__init__.py
+++ b/sklearnex/__init__.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
-from onedal.common.hyperparameters import get_hyperparameters
+import os
 
 from . import utils
 from ._config import config_context, get_config, set_config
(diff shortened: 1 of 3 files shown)


sklearnex/__init__.py
```diff
@@ -50,12 +51,13 @@
     "unpatch_sklearn",
     "utils",
 ]
+onedal_iface_flag = os.getenv("OFF_ONEDAL_IFACE", "0")
+if onedal_iface_flag == "0":
```
icfaust, line 55:
```suggestion
onedal_iface_flag = os.getenv("OFF_ONEDAL_IFACE", None)
if onedal_iface_flag:
```

icfaust (commented):
I'll wait for others to weigh in, just what I noticed on first glance.

sklearnex/__init__.py
ajay-fuji: Then In this case, if `OFF_ONEDAL_IFACE=0` is set then still it will try to import `onedal` which again can cause error.

icfaust: Then I would also recommend changing here: https://github.com/intel/scikit-learn-intelex/blob/main/setup.py#L75

samir-nasibli:
@ajay-fuji please rebase with main your branch, some parts looks outdated.

sklearnex/__init__.py
```diff
@@ -1,5 +1,6 @@
 # ==============================================================================
 # Copyright 2021 Intel Corporation
+# Copyright contributors to the oneDAL project
```
napetrov, line 3: this brings interesting question. sklearnex is not part of UXL or oneDAL project - just opensource on it's own

Topic to discuss potentially within WG/UXL

ajay-fuji: Thanks for your comment @napetrov. Will plan this discussion. 
Just wanted to know if Scikit-learn-intelex has any CLA process?

Till we reach a final conclusion on copyright, shall we mark this PR as draft?

ajay-fuji: Update - 
Based on [LICENE](https://github.com/intel/scikit-learn-intelex/blob/main/LICENSE) file, I have modified the copyright text. 
@napetrov please review and suggest any changes.

napetrov: Looks good for now. I've started discussion on if we would like to change this in longerterm process. 

With one of the options to also give scikit to UXL or keeping things as is.  Will update on results of those discussions

napetrov:
License wise this looks good for me - @icfaust  @Alexsandruss  - appreciate if you can look on changes as well

sklearnex/__init__.py
```diff
@@ -50,12 +51,13 @@
     "unpatch_sklearn",
     "utils",
 ]
+onedal_iface_flag = os.getenv("OFF_ONEDAL_IFACE", "0")
```
ethanglaser, line 54: any particular reason os.getenv is used instead of os.environ.get?

ethanglaser: if this is preferred for always returning string then we can also use it in setup.py and dispatcher.py

ajay-fuji: Both are technically same. It's just choice of usage. Sure let me change to `os.getenv` in all other places for uniformity.

napetrov (approved):

PR #1749, opened 2024-02-29
[enhancement] add check for sklearn estimator or Mixin inheritance

icfaust:
# Description
Simple but obvious test, does the sklearnex estimator inherit from the sklearn estimator it patches out. If not, does it contain BaseEstimator and at least one sklearn mixin? This will be useful for testing sklearnex-only estimators for methods like "score" and "fit_transform" which are missed by other tests (they rely on Mixin inheritance to dictate what to test).

- Add test_if_estimator_inherits_sklearn

### DIFF
--- a/sklearnex/tests/test_patching.py
+++ b/sklearnex/tests/test_patching.py
@@ -42,6 +42,14 @@ from onedal.tests.utils._dataframes_support import (
     _convert_to_dataframe,
     get_dataframes_and_queues,
 )
+from sklearn.base import (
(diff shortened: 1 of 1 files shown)


icfaust:
Rerun due to infrastructure issue (though not necessary as it was on an sklearn test, and this change should only show up on sklearnex testing) http://intel-ci.intel.com/eed700cd-d8e7-f134-932a-a4bf010d0e2e

Alexsandruss (approved):

PR #1750, opened 2024-02-29
[enhancement] add cpu guard to sklearnex testing

icfaust:
# Description
Tests which use dpnp, dpctl avoid config_context allow_fallback_to_host limitations in testing. This means that dpnp dpctl may be using the cpu code when we want to test the viability of the SYCL code. This adds another generator which will force all tests to run with a config_context(allow_fallback_to_host=False).

- Refactor logger into separate generator called sklearn_guard
- add cpu_guard
- add mark to allow for toggling cpu_guard


icfaust:
This PR is unnecessary due to defaults

PR #1751, opened 2024-02-29
FEAT(forest): hyperparameter support for RF clf inference

ahuber21:
## Description

Add hyperparameters to Random Forest classification inference, corresponding oneDAL changes in https://github.com/oneapi-src/oneDAL/pull/2638

* Add hyperparameters to `forest.py` & `forest.cpp` where supported by oneDAL
* Add tests that set a range of possible values and run inference
---

**Testing**

**Performance**

_does not apply_

### DIFF
--- a/onedal/common/dispatch_utils.hpp
+++ b/onedal/common/dispatch_utils.hpp
@@ -195,7 +195,7 @@ struct infer_ops_with_hyperparams {
     Hyperparams hyperparams;
 };
 
-#endif // defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240000
+#endif // defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240300
 
 template <typename Policy, typename Input, typename Ops>
 struct partial_compute_ops {
--- a/onedal/common/dispatch_utils.hpp
+++ b/onedal/common/dispatch_utils.hpp
@@ -169,6 +169,34 @@ struct infer_ops {
     Ops ops;
 };
 
+#if defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240300
+
+template <typename Policy, typename Input, typename Ops, typename Hyperparams>
+struct infer_ops_with_hyperparams {
+    using Task = typename Input::task_t;
+
+    infer_ops_with_hyperparams(
+        const Policy& policy, const Input& input,
+        const Ops& ops, const Hyperparams& hyperparams)
+        : policy(policy),
+          input(input),
+          ops(ops),
+          hyperparams(hyperparams) {}
+
+    template <typename Float, typename Method, typename... Args>
+    auto operator()(const pybind11::dict& params) {
+        auto desc = ops.template operator()<Float, Method, Task, Args...>(params);
+        return dal::infer(policy, desc, hyperparams, input);
+    }
+
+    Policy policy;
+    Input input;
+    Ops ops;
+    Hyperparams hyperparams;
+};
+
+#endif // defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240000
+
 template <typename Policy, typename Input, typename Ops>
 struct partial_compute_ops {
     using Task = typename Input::task_t;
--- a/onedal/common/hyperparameters.py
+++ b/onedal/common/hyperparameters.py
@@ -89,6 +89,7 @@ if daal_check_version((2024, "P", 0)):
             "train",
         ): _backend.linear_model.regression.train_hyperparameters(),
         ("covariance", "compute"): _backend.covariance.compute_hyperparameters(),
+        ("decision_forest", "infer"): _backend.decision_forest.infer_hyperparameters(),
     }
     hyperparameters_map = {}
 
--- a/onedal/common/hyperparameters.py
+++ b/onedal/common/hyperparameters.py
@@ -91,10 +91,10 @@ if daal_check_version((2024, "P", 0)):
         ): _backend.linear_model.regression.train_hyperparameters(),
         ("covariance", "compute"): _backend.covariance.compute_hyperparameters(),
     }
-    if daal_check_version((2024, "P", 3)):
-        hyperparameters_backend[("decision_forest", "infer")] = (
-            _backend.decision_forest.infer_hyperparameters(),
-        )
+    if daal_check_version((2024, "P", 300)):
+        hyperparameters_backend[
+            ("decision_forest", "infer")
+        ] = _backend.decision_forest.infer_hyperparameters()
     hyperparameters_map = {}
 
     for (algorithm, op), hyperparameters in hyperparameters_backend.items():
--- a/onedal/common/hyperparameters.py
+++ b/onedal/common/hyperparameters.py
@@ -98,8 +98,8 @@ else:
         ("covariance", "compute"): _backend.covariance.compute_hyperparameters(),
     }
     if daal_check_version((2024, "P", 300)):
-        infer_hp = _backend.decision_forest.infer_hyperparameters
-        hyperparameters_backend[("decision_forest", "infer")] = infer_hp()
+        df_infer_hp = _backend.decision_forest.infer_hyperparameters
+        hyperparameters_backend[("decision_forest", "infer")] = df_infer_hp()
     hyperparameters_map = {}
 
     for (algorithm, op), hyperparameters in hyperparameters_backend.items():
--- a/onedal/common/hyperparameters.py
+++ b/onedal/common/hyperparameters.py
@@ -111,7 +111,7 @@ if daal_check_version((2024, "P", 0)):
         )
 
     def get_hyperparameters(algorithm, op):
-        return hyperparameters_map[(algorithm, op)]
+        return hyperparameters_map.get((algorithm, op), None)
 
 else:
 
--- a/onedal/common/hyperparameters.py
+++ b/onedal/common/hyperparameters.py
@@ -84,10 +84,7 @@ if daal_check_version((2024, "P", 0)):
         }
 
     hyperparameters_backend = {
-        (
-            "linear_regression",
-            "train",
-        ): _backend.linear_model.regression.train_hyperparameters(),
+        ("linear_regression", "train"): _backend.linear_model.regression.train_hyperparameters(),
         ("covariance", "compute"): _backend.covariance.compute_hyperparameters(),
         ("decision_forest", "infer"): _backend.decision_forest.infer_hyperparameters(),
     }
--- a/onedal/common/hyperparameters.py
+++ b/onedal/common/hyperparameters.py
@@ -21,7 +21,10 @@ from warnings import warn
 from daal4py.sklearn._utils import daal_check_version
 from onedal import _backend
 
-if daal_check_version((2024, "P", 0)):
+if not daal_check_version((2024, "P", 0)):
+    warn("Hyperparameters are supported in oneDAL starting from 2024.0.0 version.")
+    hyperparameters_map = {}
+else:
     _hparams_reserved_words = [
         "algorithm",
         "op",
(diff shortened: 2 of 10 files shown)


ahuber21:
CI with branch from [oneDAL#2638](https://github.com/oneapi-src/oneDAL/pull/2638)

http://intel-ci.intel.com/eee09489-921d-f165-8b50-a4bf010d0e2e

ahuber21:
Updated run with changes from 2638 http://intel-ci.intel.com/eee0cbf3-f8dd-f171-9c4f-a4bf010d0e2e

ahuber21:
Benchmarks: http://intel-ci.intel.com/eee0cf89-b051-f1ed-a729-a4bf010d0e2e

onedal/common/dispatch_utils.hpp
```diff
@@ -168,6 +168,34 @@ struct infer_ops {
...
+        return dal::infer(policy, desc, hyperparams, input);
+    }
+
+    Policy policy;
+    Input input;
+    Ops ops;
+    Hyperparams hyperparams;
+};
+
+#endif // defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240000
```
Vika-F, line 197:
```suggestion
#endif // defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240300
```

onedal/ensemble/forest.cpp
```diff
@@ -228,13 +228,32 @@ void init_train_ops(py::module_& m) {
...
+              using namespace decision_forest;
+              using input_t = infer_input<Task>;
+
+              infer_ops_with_hyperparams ops(policy,
+                                             input_t{ model, data },
+                                             params2desc{},
+                                             hyperparams);
+              return fptype2t{ method2t{ Task{}, ops } }(params);
+          });
+#endif
```
Vika-F, line 255:
```suggestion
#endif // defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240300
```

Vika-F (commented):
Thank you for adding Python API to hyperparameters in RF inference.
My only concern is that no test is provided for the functionality.
Here is the example of the test for similar functionality in linear regression:
https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/linear_model/tests/test_linear.py#L31

ahuber21:
Updated job with oneDAL + scikit-learn-intelex changes http://intel-ci.intel.com/eee23804-7d24-f117-86db-a4bf010d0e2e

KulikovNikita:
Are hyperparameters just another name for performance tuning parameters?

onedal/ensemble/forest.cpp
```diff
@@ -310,6 +329,57 @@ void init_infer_result(py::module_& m) {
...
+            .def("set_block_size",
+                 [](infer_hyperparams_t& self, int64_t block_size) {
+                     self.set_block_size(block_size);
+                 })
+            .def("get_block_size",
+                 [](const infer_hyperparams_t& self) {
+                     return self.get_block_size();
+                 })
+            .def("set_min_trees_for_threading",
+                 [](infer_hyperparams_t& self, int64_t trees) {
```
KulikovNikita, line 358:
```suggestion
                 [](infer_hyperparams_t& self, std::int64_t trees) {
```

Please always use 'std::' for the sake of consistency

onedal/ensemble/forest.cpp
```diff
@@ -310,6 +329,57 @@ void init_infer_result(py::module_& m) {
...
+                     self.set_min_trees_for_threading(trees);
+                 })
+            .def("get_min_trees_for_threading",
+                 [](const infer_hyperparams_t& self) {
```
KulikovNikita, line 362: I would highly recommend to provide the explicit output type for the compiler:

```suggestion
                 [](const infer_hyperparams_t& self) -> std::int64_t {
```

ethanglaser:
@Vika-F are we able to finalize this one?

onedal/ensemble/forest.py
```diff
@@ -40,6 +44,11 @@
 )
 
 
+def get_hyperparameters_ver(*args):
+    if daal_check_version((2024, "P", 300)):
+        return get_hyperparameters(*args)
```
Alexsandruss, line 49: This version check overcomplicates code, while you can just modify original `get_hyperparameters` to return None:
```python
def get_hyperparameters(algorithm, op):
    return hyperparameters_map.get((algorithm, op), None)
```

Alexsandruss (changes requested):

samir-nasibli:
@Vika-F @ethanglaser @icfaust @avolkov-intel 
Who is expected to be next assigned to this job done?

onedal/ensemble/forest.cpp
ahuber21: I was being consistent, `std::` isn't used anywhere in this part of the code. But I can add it everywhere

onedal/ensemble/forest.cpp
ahuber21: Done

samir-nasibli:
@ahuber21 please use PR template in your description for Check PR Checklist / Close all checkboxes before moving from draft (pull_request) 
Please mark it ready for review, moving to draft now

ahuber21:
Intel CI successful except for an environment error and a timeout in unrelated test cases: http://intel-ci.intel.com/ef8c9e4c-c4c4-f1f1-8c83-a4bf010d0e2d

ahuber21:
@Vika-F would be great if you could take another look. Two questions in particular

- Does the code reflect the (updated) hyperparameters from https://github.com/oneapi-src/oneDAL/pull/2692
- Do you agree with how I handle classification and regression separately (since the hyperparameters are only available in classification)

onedal/ensemble/forest.py
```diff
@@ -357,7 +367,12 @@ def _predict(self, X, module, queue):
         model = self._onedal_model
         X = _convert_to_supported(policy, X)
         params = self._get_onedal_params(X)
-        result = module.infer(policy, params, model, to_table(X))
+        hparams = self._get_hyperparameters("infer")
```
Alexsandruss, line 370:
```suggestion
        if hasattr(self, "get_hyperparameters"):
            hparams = self.get_hyperparameters("infer")
        else:
            hparams = None
```

onedal/ensemble/forest.py
```diff
@@ -191,6 +196,11 @@ def _get_onedal_params(self, data):
             onedal_params["splitter_mode"] = self.splitter_mode
         return onedal_params
 
+    # base version of _get_hyperparameters that returns None
+    # overloaded versions in derived classes may return HyperParameters, if supported
+    def _get_hyperparameters(self, _: OperationType) -> Optional[HyperParameters]:
+        return None
+
```
Alexsandruss, line 203:
```suggestion
```

onedal/ensemble/forest.py
```diff
@@ -378,6 +393,15 @@ def _predict_proba(self, X, module, queue):
...
 
+# Version of register_hyperparameters that supports versioning
+register_hyperparameters_ver = daal_require_version_wrapper((2024, "P", 300))(
+    register_hyperparameters
+)
+
+
+@register_hyperparameters_ver(
+    {"predict": get_hyperparameters("decision_forest", "infer")}
+)
```
Alexsandruss, line 404:
```suggestion
@register_hyperparameters(
    {"predict": get_hyperparameters("decision_forest", "infer")}
)
```

onedal/ensemble/forest.py
```diff
@@ -448,6 +472,9 @@ def _validate_targets(self, y, dtype):
         #    self.n_classes_ = self.classes_
         return y
 
+    def _get_hyperparameters(self, op: OperationType) -> Optional[HyperParameters]:
+        return get_hyperparameters("decision_forest", op)
+
```
Alexsandruss, line 477:
```suggestion
```

Alexsandruss (commented):
Why this simpler approach is not used?

onedal/common/hyperparameters.py
```diff
@@ -56,9 +60,12 @@ def __getattribute__(self, __name):
                 return super().__getattribute__(__name)
             elif __name in self.getters.keys():
                 return self.getters[__name]()
-            else:
-                raise ValueError(
-                    f"Unknown '{__name}' name in "
+            try:
+                # return all variables that are in super() class normally
+                return super().__getattribute__(__name)
```
Alexsandruss, line 65: Not clear why is it needed for class without any parent. Comment with justification would be great

ahuber21: It is still a Python object, so there are builtin attributes, like `__class__`, __sizeof__`, etc. 
They are used in debuggers and making them inaccessible just leads to confusion. I will update the comment to better explain it.

onedal/ensemble/forest.py
ahuber21: Could you comment on why the version check is not necessary? I forgot why I added it back in March. I probably followed examples in other classes. Are you sure it's safe to remove?

samir-nasibli (commented):
@ahuber21 many thanks for working on it!
please address CI fails before making it ready for the review, moving to draft

onedal/ensemble/forest.py
```diff
@@ -378,6 +387,7 @@ def _predict_proba(self, X, module, queue):
         return y
 
 
+@register_hyperparameters({"infer": get_hyperparameters("decision_forest", "infer")})
```
samir-nasibli, line 390: Shouldn't this depend on onedal backend version?

samir-nasibli (commented):
@ahuber21 I don't have any specific comments. Looks good to me and it seems consistent with other estimators.
Dependent on @Alexsandruss @icfaust concerns being properly addressed.

icfaust (commented):
So far the codebase has very little typing outside of daal4py.sklearn._utils, model builders or test_daal4py_examples.  Could you comment on the need for having them in the hyperparameter file in onedal.common? I guess, do we expect an input outside of the types, is there some discussion we should have about static typing through the onedal package? This isn't a blocker for merging.

ahuber21:
Typing tools in Python and common IDEs have become very mature and are an immense help when debugging code. I strongly favor adding as much type information as possible. It is sometimes difficult in this project when objects and classes are coming from cython, but limited typing information is still better than none.
It is simply common and best practice nowadays and we have to start somewhere. Typing is never enforced in Python, so even if we receive input outside of the types, nothing bad happens (except maybe for undefined behavior, because our understanding of the code is wrong).
I would argue against static typing checks for now because I would expect it to be highly non-trivial for this particular repo. Moreover, we should probably first create all necessary type stubs so that we can actually use type hints consistently. If you look into my changes, I annotated the `hyperparameters_backend` dictionary as `Dict[Tuple[str, str], Any]`, simply because the type of the `*_hyperparameters()` getters from backend isn't readily available.
For now all contributors should feel encouraged but not forced to use type hints. If you want to bring this discussion to a dedicated issue go for it :)

onedal/ensemble/forest.py
```diff
@@ -378,6 +387,7 @@ def _predict_proba(self, X, module, queue):
...
 class RandomForestClassifier(ClassifierMixin, BaseForest, metaclass=ABCMeta):
```
Alexsandruss, line 391: Registration is attached in wrong place: it should be sklearnex, not onedal class.

onedal/ensemble/tests/test_random_forest.py
```diff
@@ -22,9 +22,19 @@
...
+    (32, 100, 32, 0.3),
+    (64, 10, 32, 0.1),
+    (128, 100, 1000, 1.0),
+]
+
 
 @pytest.mark.parametrize("queue", get_queues())
-def test_rf_classifier(queue):
+@pytest.mark.parametrize("block, trees, rows, scale", hparam_values)
+def test_rf_classifier(queue, block, trees, rows, scale):
```
Alexsandruss, line 37: Same with test: it should be for sklearnex module.

Alexsandruss (changes requested):

ahuber21:
Thank you @Alexsandruss. It is updated

Alexsandruss (approved):

PR #1756, opened 2024-03-08
Add _n_inner_iter attribute for Logistic Regression

avolkov-intel:
# Description
If the solver was set to newton-cg and training was done on GPU, it is possible to get the total number of CG-solver iterations done via _onedal_estimator._n_inner_iter attribute. This attribute can be used to estimate the time per iteration. Note that this attribute will be available if build with oneDAL of version 2024.4.0 or higher

### DIFF
--- a/onedal/linear_model/logistic_regression.cpp
+++ b/onedal/linear_model/logistic_regression.cpp
@@ -95,9 +95,12 @@ auto get_onedal_result_options(const py::dict& params) {
             else if (match.str() == "iterations_count") {
                 onedal_options = onedal_options | result_options::iterations_count;
             }
+#if ONEDAL_VERSION >= 20240400
             else if (match.str() == "inner_iterations_count") {
                 onedal_options = onedal_options | result_options::inner_iterations_count;
-            } else {
+            }
+#endif 
+            else {
                 ONEDAL_PARAM_DISPATCH_THROW_INVALID_VALUE(result_option);
             }
         }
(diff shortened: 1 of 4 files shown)


onedal/linear_model/logistic_regression.cpp
```diff
@@ -95,8 +95,14 @@ auto get_onedal_result_options(const py::dict& params) {
             else if (match.str() == "iterations_count") {
                 onedal_options = onedal_options | result_options::iterations_count;
             }
-            else
+#if ONEDAL_VERSION >= 20240400
+            else if (match.str() == "inner_iterations_count") {
+                onedal_options = onedal_options | result_options::inner_iterations_count;
+            }
+#endif
```
Alexsandruss, line 102: No way to test it until oneDAL version is updated to 2024.4

avolkov-intel: yes, so I suggest merging https://github.com/oneapi-src/oneDAL/pull/2671

avolkov-intel: UPD: I changed required onedal version to 2024.3, since it seems that this functionality will be available in earlier version

avolkov-intel:
@Alexsandruss @ethanglaser please review PR

onedal/linear_model/logistic_regression.py
```diff
@@ -54,7 +54,11 @@ def _get_onedal_params(self, dtype=np.float32):
             "max_iter": self.max_iter,
             "C": self.C,
             "optimizer": self.solver,
-            "result_option": (intercept + "coefficients|iterations_count"),
+            "result_option": (
+                intercept + "coefficients|iterations_count" + "|inner_iterations_count"
+                if self.solver == "newton-cg"
+                else ""
+            ),
         }
```
ethanglaser, line 62: Some parentheses may be needed, its hard to tell what is intended by the logic here.
My assumption is that we only want to append "|inner_iterations_count" if solver is newton cg, however I think what actually happens is result_option will be empty string if solver is not newton cg

samir-nasibli (commented):
I would like to add pytests for the validation if onedal 2024.3 in the env

avolkov-intel:
http://intel-ci.intel.com/eef59312-18b2-f110-bcc8-a4bf010d0e2e CI rerun

ethanglaser:
please rebase

ethanglaser:
ci rerun (its green): http://intel-ci.intel.com/eef818d4-ee89-f141-8546-a4bf010d0e2e

ethanglaser (approved):

PR #1759, opened 2024-03-11
[testing] add _assert_all_finite tests

icfaust:
# Description
Insanely, daal4py doesn't have any tests for _assert_all_finite. The testing of _assert_all_finite in sklearn is also minimal. These tests allow for making sure that _assert_all_finite works in the way we expect. It includes a time dependent RNG seed, such that general support is guaranteed.

- Add tests


sklearnex/utils/tests/test_finite.py
```diff
@@ -0,0 +1,95 @@
...
+        else:
+            assert_raises(ValueError, _assert_all_finite, X, allow_nan=allow_nan)
+    finally:
+        print(f"SEED: {seed}")
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+@pytest.mark.parametrize("allow_nan", [False, True])
+@pytest.mark.parametrize("check", ["inf", "NaN", None])
+@pytest.mark.parametrize("seed", [0, int(time.time())])
```
samir-nasibli, line 79: How this will change logs?
Will we able to deselect this test suits/cases by name adding to deselect list?

sklearnex/utils/tests/test_finite.py
```diff
@@ -0,0 +1,95 @@
...
+        loc = rand.randint(0, X.size - 1)
+        X.reshape((-1,))[loc] = float(check)
+
+    try:
+        if check is None or (allow_nan and check == "NaN"):
+            _assert_all_finite(X, allow_nan=allow_nan)
+        else:
+            assert_raises(ValueError, _assert_all_finite, X, allow_nan=allow_nan)
+    finally:
+        print(f"SEED: {seed}")
```
Alexsandruss, line 73: Use logging instead of print: https://docs.pytest.org/en/7.1.x/how-to/logging.html

sklearnex/utils/tests/test_finite.py
```diff
@@ -0,0 +1,95 @@
...
+        loc = rand.randint(0, X.size - 1)
+        X[loc] = float(check)
+
+    try:
+        if check is None or (allow_nan and check == "NaN"):
+            _assert_all_finite(X, allow_nan=allow_nan)
+        else:
+            assert_raises(ValueError, _assert_all_finite, X, allow_nan=allow_nan)
+    finally:
+        print(f"SEED: {seed}")
```
Alexsandruss, line 95: Same: print->logging

sklearnex/utils/tests/test_finite.py
```diff
@@ -0,0 +1,95 @@
...
+        ],
+        [1000, 1000],
+    ],
+)
+@pytest.mark.parametrize("allow_nan", [False, True])
+@pytest.mark.parametrize("check", ["inf", "NaN", None])
+@pytest.mark.parametrize("seed", [0, int(time.time())])
+def test_assert_finite_random_location(dtype, shape, allow_nan, check, seed):
+    rand.seed(seed)
+    X = (np.finfo(dtype).max * rand.random_sample(shape)).astype(dtype)
```
Alexsandruss, line 61:
```suggestion
    X = rand.uniform(low=0, high=np.finfo(dtype).max, size=shape).astype(dtype)
```

sklearnex/utils/tests/test_finite.py
icfaust: The time dependent input for seed will show up in the title of the test. This can be specifically deselected via list, but since it is time-dependent on unix timestamps, it would only be a problem if the test occurs to the exact second ~68 years later (assuming unix time is still a 32 bit integer).

samir-nasibli: I think we misunderstood each other a little :)
I mean to say that we have different identifications for the test suites. In case we want to add this test to deselected, there will be problems with correctly identifying the cases.
From our public CI logs I see different names for the exact test cases on different launches.
Input seed id: **1711142958** ([link](https://dev.azure.com/daal/daal4py/_build/results?buildId=32836&view=logs&j=9a864fd9-6c8f-52ca-79ce-2aa6dca1a1de&t=a545531a-4e35-5d11-8855-7f2aae021e99&l=2988)):
```
utils/tests/test_finite.py::test_assert_finite_random_shape_and_location[1711142958-inf-False-float32] PASSED [ 99%]
utils/tests/test_finite.py::test_assert_finite_random_shape_and_location[1711142958-inf-False-float64] PASSED [ 99%]
```
Input seed id: **1711143452** ([link](https://dev.azure.com/daal/daal4py/_build/results?buildId=32836&view=logs&j=517fe804-fa30-5dc2-1413-330699242c05&t=6200c9a9-fd17-566a-4b28-c9cb31ff00c5&l=3402)):
```
utils/tests/test_finite.py::test_assert_finite_random_shape_and_location[1711143452-inf-False-float32] PASSED [ 99%]
utils/tests/test_finite.py::test_assert_finite_random_shape_and_location[1711143452-inf-False-float64] PASSED [ 99%]
```

icfaust: I guess a higher level question I have is:

What is our procedure for adding randomized tests to sklearnex? When utilizing a fixed seed, that is deterministic, but how will we standardize continually random inputs?

Btw, I am against a deselected_tests yaml for sklearnex. Unlike sklearn, we have control of the code base, therefore manual use of pytest.skip is possible and makes issues more apparent.

sklearnex/utils/tests/test_finite.py
```diff
@@ -0,0 +1,89 @@
...
+from sklearnex.utils import _assert_all_finite
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+@pytest.mark.parametrize(
+    "shape",
+    [
+        [16, 2048],
+        [
+            2**16 + 3,
```
KulikovNikita, line 33: Are you sure that it should be a 1 dimensional array?

sklearnex/utils/tests/test_finite.py
```diff
@@ -0,0 +1,89 @@
...
+    else:
+        assert_raises(ValueError, _assert_all_finite, X, allow_nan=allow_nan)
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+@pytest.mark.parametrize("allow_nan", [False, True])
+@pytest.mark.parametrize("check", ["inf", "NaN", None])
+@pytest.mark.parametrize("seed", [0, int(time.time())])
+def test_assert_finite_random_shape_and_location(dtype, allow_nan, check, seed):
+    lb, ub = 32768, 1048576  # lb is a patching condition, ub 2^20
```
KulikovNikita, line 78: Here is the place to use "2**x" I suppose

KulikovNikita (approved):
LGTM

sklearnex/utils/tests/test_finite.py
Alexsandruss: If fixed and time-based seeds are both presented, it's fine to parameterize test in this way for now.
Seed is recorded in pytest output anyway, and test name will also contain it on failure.

Alexsandruss (approved):

PR #1767, opened 2024-03-18
FIX: no dpc if windows debug

ethanglaser:
# Description
Do not build sklearnex dpc/spmd libs if in using windows debug - onedal dpc libs not built. Should resolve fail in build step of sklearnex windows in debug jobs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,8 +84,9 @@ d4p_version = (
 trues = ["true", "True", "TRUE", "1", "t", "T", "y", "Y", "Yes", "yes", "YES"]
 no_dist = True if "NO_DIST" in os.environ and os.environ["NO_DIST"] in trues else False
 no_stream = "NO_STREAM" in os.environ and os.environ["NO_STREAM"] in trues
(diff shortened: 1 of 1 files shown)


ethanglaser:
Debug job with changes: http://intel-ci.intel.com/eee55e4f-8bea-f143-8cf1-a4bf010d0e2e

md-shafiul-alam:
looks good to me. Interesting that we build daal4py on nightly debug jobs but do not run any sklearnex tests. Are they out of scope for these debug jobs? Or were they disabled because of failures in the build steps?

ethanglaser:
> looks good to me. Interesting that we build daal4py on nightly debug jobs but do not run any sklearnex tests. Are they out of scope for these debug jobs? Or were they disabled because of failures in the build steps?

Good question - no idea about the answer but it does seem odd, my only thought is that debug build is only useful for C++ testing. @homksei is there any benefit to running the sklearnex tests with debug build?

md-shafiul-alam (approved):

PR #1763, opened 2024-03-13
[testing] introduce Pandas testing to onedal/sklearnex

icfaust:
# Description
daal4py has significant pandas testing, where in onedal/sklearnex does not.

Changes proposed in this pull request:
- add pandas to _get_dataframes_and_queues (actually the only dataframe in the list)
- add pandas to _convert_to_dataframe

PR #1776, opened 2024-03-26
[testing] refactor test_memory_usage.py

icfaust:
# Description
test_memory_usage is another test suite  which evaluates all scikit-learn-intelex algorithms. This one tests for memory leaks and can be useful for general memory management understanding.  This has unnecessary duplication of code, and should follow other standards to minimize work. Secondarily, with more and more GPU algorithm support, this test suite provides no useful information.  A onedal function is created to query the queue's sycl device for free global memory, and works in place of tracemalloc's get_traced_memory (albeit in a much more crude way).

- rewrite test_memory_usage to be useful for gpu-only (zero copy) algorithms
- generalize and centralize to minimize maintenance

Tasks
- [x] Implement Intel GPU memory monitor
- [x] rewrite functions tests using partial
- [x] take estimators from sklearnex/tests/_utils.py
- [x] evaluate special estimators
- [x] rewrite memory evaluation to use get_dataframes_and_queues (re-parameterize fortran and c contiguous separately)
- [x] pass public CI
- [ ] pass private CI


icfaust:
NOTE: to self, will need to manipulate ZES_ENABLE_SYSMAN with a level zero driver to make this work, otherwise will raise a RuntimeError that will need to be caught (and probably require a pytest.skip)

sklearnex/tests/test_memory_usage.py
```diff
@@ -143,19 +106,28 @@ def gen_clsf_data(n_samples, n_features):
...
             x_train, x_test = x.iloc[train_index], x.iloc[test_index]
             y_train, y_test = y.iloc[train_index], y.iloc[test_index]
-        # TODO: add parameters for all estimators to prevent
-        # fallback to stock scikit-learn with default parameters
 
-        alg = estimator()
+        try:
+            alg = estimator()
+        except:
+            alg = clone(estimator)
```
samir-nasibli, line 129: Please follow codefactor suggestions here

samir-nasibli (commented):
This PR mostly looks like enhancement rather than just refactoring :)
I again voting for splitting out refactoring and enhancements from one PR

icfaust:
> This PR mostly looks like enhancement rather than just refactoring :) I again voting for splitting out refactoring and enhancements from one PR

Please review #1763, this will greatly simplify this PR.

icfaust:
kneighbors calls via sycl are causing CI timeouts. Also, it looks like the private CI drivers aren't level-zero, both of these need to be investigated.

onedal/linear_model/linear_model.py
```diff
@@ -73,8 +73,11 @@ def _create_model(self, policy):
         coefficients, intercept = make2d(coefficients), make2d(intercept)
         coefficients = coefficients.T if n_targets_in == 1 else coefficients
 
-        assert coefficients.shape == (n_targets_in, n_features_in)
-        assert intercept.shape == (n_targets_in, 1)
+        assert coefficients.shape == (
+            n_targets_in,
+            n_features_in,
+        ), f"{coefficients.shape}, {n_targets_in}, {n_features_in}"
```
KulikovNikita, line 79:
```suggestion
        desired_shape = (n_targets_in, n_features_in)
        assert coefficients.shape == desired_shape, f"{coefficients.shape}, {str(desired_shape)}"
```

sklearnex/linear_model/tests/test_linear.py
```diff
@@ -52,9 +52,9 @@ def test_sklearnex_import_linear(dataframe, queue, dtype, macro_block):
     assert "sklearnex" in linreg.__module__
     assert linreg.n_features_in_ == 2
 
-    tol = 1e-5 if X.dtype == np.float32 else 1e-7
+    tol = 1e-5 if dtype == np.float32 else 1e-7
```
KulikovNikita, line 55: Btw you probably can estimate the upper bound of imperfection for calculations as 'np.finfo(dtype).eps * n_operations', where 'n_operations' is intrinsic to an algorithm and generally should be equal to 'n_cols' in case of LinearRegression, PCA & Covariance; 'n_rows' for knn, kmeans

icfaust: @olegkkruglov what do you think?

icfaust:
While CI is green, changes in private CI will need to occur before pulling in this PR. There will likely be a lot of testing needed.

icfaust:
Private CI has been updated to support ZES_ENABLE_SYSMAN, will see if it runs the new tests.

icfaust:
Seg fault for forest on GPU, will need to create a reproducer script for it, and disable it from this testing.

icfaust:
#1784 will need to be integrated to solve issues with LocalOutlierFactor(novelty=True) on GPU

icfaust:
The initial fit of the initial estimator in the for loop causes a memory leak in most of the gpu codes. I will need to ask someone with more GPU oneDAL knowledge to ask how things are stored on GPU after something is run.

icfaust:
image

I am going to swap orders of things to see if these failures are stable.

icfaust:
image
Will swap the special instances and standards and rerun with fortran first. This case brought fewer failures. This looks problematic.

icfaust:
image

icfaust:
There are some algorithms which consistently pass. I want to replicate these on a local machine to see how deterministic the results are.

icfaust:
Rerun for determinacy: http://intel-ci.intel.com/eef59a23-df27-f1f4-8546-a4bf010d0e2e

icfaust:
IncrementalLinearRegression shows some oddities that require further analysis.

icfaust:
It looks like there might be an issue in potrf in oneMKL/sycl

sklearnex/tests/test_memory_usage.py
```diff
@@ -14,127 +14,112 @@
...
+    from onedal import _backend
 
 
-BANNED_ESTIMATORS = (
+CPU_BANNED_LIST = (
+    "TSNE",  # too slow for using in testing on common data size
+    "config_context",  # does not malloc
+    "get_config",  # does not malloc
+    "set_config",  # does not malloc
+    "SVC(probability=True)",  # F numpy (investigate _fit_proba)
```
ethanglaser, line 51: What is meant by "F numpy"

sklearnex/tests/test_memory_usage.py
```diff
@@ -14,127 +14,112 @@
...
-    CosineDistancesEstimator,
-    CorrelationDistancesEstimator,
-    RocAucEstimator,
-]
-get_patched_estimators(BANNED_ESTIMATORS, estimators)
-estimators = remove_duplicated_estimators(estimators)
 
+GPU_BANNED_LIST = (
+    "TSNE",  # too slow for using in testing on common data size
+    "RandomForestRegressor",  # too slow for using in testing on common data size
```
ethanglaser, line 61: Can it be tested on smaller case?

sklearnex/tests/test_memory_usage.py
```diff
@@ -14,127 +14,112 @@
...
+    "config_context",  # does not malloc
+    "get_config",  # does not malloc
+    "set_config",  # does not malloc
+    "Ridge",  # does not support GPU offloading (fails silently)
+    "ElasticNet",  # does not support GPU offloading (fails silently)
+    "Lasso",  # does not support GPU offloading (fails silently)
+    "SVR",  # does not support GPU offloading (fails silently)
+    "NuSVR",  # does not support GPU offloading (fails silently)
+    "NuSVC",  # does not support GPU offloading (fails silently)
+    "LogisticRegression",  # default parameters not supported, see solver=newton-cg
```
ethanglaser, line 72: Is it possible to add any check for it?

sklearnex/tests/test_memory_usage.py
```diff
@@ -148,41 +133,69 @@ def gen_clsf_data(n_samples, n_features):
...
+    else:
+        return tracemalloc.get_traced_memory()[0]
+
+
+def take(x, index, axis=0, queue=None):
+    xp, array_api = get_namespace(x)
+    if array_api:
+        return xp.take(x, xp.asarray(index, device=queue), axis=axis)
+    else:
+        return x.take(index, axis=axis)
```
ethanglaser, line 148: Can you clarify what this function is doing

sklearnex/tests/test_memory_usage.py
```diff
@@ -212,21 +230,51 @@ def _kfold_function_template(estimator, data_transform_function, data_shape):
...
+    os.getenv("ZES_ENABLE_SYSMAN") is None or not is_dpctl_available("gpu"),
+    reason="SYCL device memory leak check requires the level zero sysman",
+)
+@pytest.mark.parametrize("queue", get_queues("gpu"))
+@pytest.mark.parametrize("estimator", GPU_ESTIMATORS.keys())
+@pytest.mark.parametrize("order", ["F", "C"])
 @pytest.mark.parametrize("data_shape", data_shapes)
-def test_memory_leaks(estimator, data_transform_function, data_shape):
-    _kfold_function_template(estimator, data_transform_function, data_shape)
+def test_gpu_memory_leaks(estimator, queue, order, data_shape):
```
ethanglaser, line 274: Where are these functions called? test_gpu_memory_leaks and test_memory_leaks

ethanglaser (commented):
Mostly just clarification questions for now

sklearnex/tests/test_memory_usage.py
icfaust: when we use fancy indices in numpy like
```python
example = np.mgrid[:20]
index = [0,5,4,3,10]
print(example[index]) # uses fancy indexing
```
its actually replicating numpy.take.  However, for array api, take will only work if the index array is also located on the same device. So this special function is necessary so that array_api inputs (like dpctl tensors) will work properly with train_test_split.  Its likely we need to look into our version of that function for proper dpnp and dpctl support. Also a good candidate for array_api rollout.

sklearnex/tests/test_memory_usage.py
icfaust: they are automatically picked up by pytest because of test in the name.  The gpu memory leaks require some specific steps to run, namely a level_zero driver, SYCL support in oneDAL and sklearnex, and dpctl.

sklearnex/tests/test_memory_usage.py
icfaust: LogisticRegression(solver='newton-cg') has been added to SPECIAL_INSTANCES in sklearnex/tests/_utils.py a follow up PR (after I clear out my current backlog) will add a test to make sure that all of our defined estimators are either in SPECIAL_INSTANCES or PATCHED_MODELS. For example, so far we are not testing the BasicStatistics estimator in any of sklearnex/tests/

sklearnex/tests/test_memory_usage.py
icfaust: Good question, but I have to say there are probably some alarming performance issues we need to look into on oneDAL side...  Ideally we could try to reduce it some. Let me know what you think.

sklearnex/tests/test_memory_usage.py
icfaust: Ah, its referencing some of the parameterized inputs into test_memory_leaks, but I will change it to be more descriptive. Its a fortran-order numpy array issue on SVC(probabilty=True).  Something funky is going on in _fit_proba, which may open a can of worms.

sklearnex/tests/test_memory_usage.py
ethanglaser: Yeah might be best to worry about it after general rf issues are resolved

ethanglaser:
Can you resolve warning from CI: `ConstantInputWarning: An input array is constant; the correlation coefficient is not defined. mem_iter_corr, _ = pearsonr(mem_tracks, list(range(len(mem_tracks))))`

Internal CI is yellow on GPU tests but no error/fail message but maybe this warning causes it

icfaust:
Fully green CI

sklearnex/tests/test_memory_usage.py
```diff
@@ -14,127 +14,113 @@
...
-    TrainTestSplitEstimator,
-    FiniteCheckEstimator,
-    CosineDistancesEstimator,
-    CorrelationDistancesEstimator,
-    RocAucEstimator,
-]
-get_patched_estimators(BANNED_ESTIMATORS, estimators)
-estimators = remove_duplicated_estimators(estimators)
 
+GPU_BANNED_LIST = (
```
ethanglaser, line 60: Do we have tasks created for these or some sort of tracking of known problems?

sklearnex/tests/test_memory_usage.py
```diff
@@ -14,127 +14,113 @@
...
+    if not k in GPU_BANNED_LIST
+}
 
-data_shapes = [(1000, 100), (2000, 50)]
+data_shapes = [
+    pytest.param((1000, 100), id="(1000, 100)"),
+    pytest.param((2000, 50), id="(2000, 50)"),
+]
 
 EXTRA_MEMORY_THRESHOLD = 0.15
```
ethanglaser, line 121: is this value still valid?

sklearnex/tests/test_memory_usage.py
```diff
@@ -212,21 +240,51 @@ def _kfold_function_template(estimator, data_transform_function, data_shape):
...
+@pytest.mark.parametrize("queue", get_queues("gpu"))
+@pytest.mark.parametrize("estimator", GPU_ESTIMATORS.keys())
+@pytest.mark.parametrize("order", ["F", "C"])
 @pytest.mark.parametrize("data_shape", data_shapes)
-def test_memory_leaks(estimator, data_transform_function, data_shape):
-    _kfold_function_template(estimator, data_transform_function, data_shape)
+def test_gpu_memory_leaks(estimator, queue, order, data_shape):
+    func = ORDER_DICT[order]
+    if "ExtraTrees" in estimator and data_shape == (2000, 50):
+        pytest.skip("Avoid a segmentation fault in Extra Trees algorithms")
```
ethanglaser, line 287: any idea why it would seg fault for 2000x50 and not 1000x100?

ethanglaser (approved):
No objections from my side - ambiguity of EXTRA_MEMORY_THRESHOLD is a little sketchy but had this before as well, nice work with GPU part

sklearnex/tests/test_memory_usage.py
icfaust: There is definitely some issues in the ExtraTrees code on oneDAL side we need to address. Unfortunately it will take some time, as I suspect accuracy issues are probably tied in there as well.

sklearnex/tests/test_memory_usage.py
icfaust: You are right, I have not made tickets for these issues yet, and they need to be done.  I will do that before merge, and will send verification to you.

sklearnex/tests/test_memory_usage.py
icfaust: I have no idea on why it was set to 15%, maybe also worth a discussion.

sklearnex/tests/test_memory_usage.py
```diff
@@ -212,21 +240,51 @@ def _kfold_function_template(estimator, data_transform_function, data_shape):
...
-    mem_after, _ = tracemalloc.get_traced_memory()
+    mem_after = get_traced_memory(queue)
     tracemalloc.stop()
     mem_diff = mem_after - mem_before
 
-    assert mem_diff < EXTRA_MEMORY_THRESHOLD * data_memory_size, message.format(
-        "after", mem_diff, round((mem_diff) / data_memory_size * 100, 2)
-    )
+    # GPU offloading with SYCL contains a program/kernel cache which should
+    # be controllable via a KernelPorgramCache object in the SYCL context.
```
Alexsandruss, line 248:
```suggestion
    # be controllable via a KernelProgramCache object in the SYCL context.
```

PR #1781, opened 2024-03-29
Remove code branches for sklearn versions older than 0.24

Alexsandruss:
Changes:
- Remove all code branches related to sklearn versions older than 0.24
- Replace extra docstrings with `__doc__` attr. assignment


icfaust:
@napetrov thoughts?

Alexsandruss:
> @napetrov thoughts? Realistically we should update the codebase to show support for 1.1, 1.2, 1.3 and 1.4 due to the sklearn update. Then we should not only drop anything older than .24, but anything older than 1.1 (given the logic of the PR).

0.24.* still has at least [10% of 1.* downloads](https://www.pepy.tech/projects/scikit-learn?versions=0.24.*&versions=0.23.*&versions=0.22.*&versions=1.*). Also, it's last to support py36.

icfaust:
> > @napetrov thoughts? Realistically we should update the codebase to show support for 1.1, 1.2, 1.3 and 1.4 due to the sklearn update. Then we should not only drop anything older than .24, but anything older than 1.1 (given the logic of the PR).
> 
> 0.24.* still has [~10% of 1.* downloads](https://www.pepy.tech/projects/scikit-learn?versions=0.24.*&versions=0.23.*&versions=0.22.*&versions=1.*). Also, it's last to support py36.

the logic in this argument/PR is an example of a larger question, namely: what is the definition for sklearn support for this repository? I find the argument sufficient, but it needs a larger discussion out of just a PR.

napetrov:
We normally support 4 last versions, but the good question is if we should just crawl through all - not all versions are equally important, so potentially 1.0 might be more important than 1.1 

Also supported Python versions are important consideration - Python versions are simple enough- supporting active Python versions is what we are currently doing. This would influence supported scikit versions

And third point is cleaning up code

napetrov:
Interesting comparison 0.24 have more downloads than 1.1 and 1.2 versions - https://www.pepy.tech/projects/scikit-learn?versions=0.24.*&versions=1.0*&versions=1.1*&versions=1.2*&versions=1.3*&versions=1.4*

So keeping 0.24 supported make some sense. for scikit 0.24 pythons 3.6-3.9 are supported.

another interesting data - scikit have 3.7 and 3.8 downloads in top 
image

napetrov (approved):

icfaust (changes requested):
I would like a fixed design rule, or rule of thumb associated with dropping support. This will be important in reducing developer oversight. I feel like we are on our way there with the previous discussion.

icfaust:
@Alexsandruss I disagree with your choice. The dismissal of my review was unwarranted.

PR #1783, opened 2024-04-02
[bug] fix dpnp/dpctl slowdown in fit method of neighbors algorithms

icfaust:
# Description
There is a 400x performance difference in running fit using dpnp arrays/dpctl tensors vs config_context with a device offload. This is not due to the movement off of GPU, but due to numpy calls on dpnp/dpctl in _fit_validation.  By moving that call within the onedal_cpu_supported and onedal_gpu_supported, the input data is guaranteed to by numpy arrays, and therefore gets the equivalent performance of config_context.

There issues were discovered in #1737 and #1776 and should speed up CI in those cases substantially for both CPU and GPU offloading via dpnp/dpctl.

- move _fit_validation for LocalOutlierFactor
- move _fit_validation for NearestNeighbors
- move _fit_validation for KNeighborsClassifier
- move _fit_validation for KNeighborsRegressor

### DIFF
--- a/sklearnex/neighbors/_lof.py
+++ b/sklearnex/neighbors/_lof.py
@@ -100,7 +100,6 @@ class LocalOutlierFactor(KNeighborsDispatchingBase, sklearn_LocalOutlierFactor):
         return self
 
     def fit(self, X, y=None):
-        self._fit_validation(X, y)
         result = dispatch(
             self,
             "fit",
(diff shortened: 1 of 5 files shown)


ethanglaser (approved):
LGTM but 2 questions - are there any other algos where this inefficiency is present? and 400x speedup on fit? because for knn (at least brute force) fit should be minimal in terms of runtime

icfaust:
lof has a slow method score_samples which also will need to be added

icfaust:
> LGTM but 2 questions - are there any other algos where this inefficiency is present? and 400x speedup on fit? because for knn fit should be minimal in terms of runtime

yeah it was weird, it took 160s to run when normally it was .4s. its mainly because some math was being done outside the dispatch call, which is rather unusual

samir-nasibli (approved):
I vote for this changes. Validations for onedal backend should be done on/after dispatching.
Thank you!
Assuming green CI.

PR #1794, opened 2024-04-12
Update sklearnex estimators to support sklearn 1.5

Alexsandruss:
Changes related to sklearn 1.5:
 - Apply renaming of `n_iter` parameter to `max_iter` in TSNE
 - Apply array API support in PCA internal functions
 - Apply `covariance_eigh` solver support in PCA
 - Change PCA "auto" solver choice to complete compliance with sklearn

### DIFF
--- a/sklearnex/decomposition/pca.py
+++ b/sklearnex/decomposition/pca.py
@@ -35,6 +35,9 @@ if daal_check_version((2024, "P", 100)):
     if sklearn_check_version("1.1") and not sklearn_check_version("1.2"):
         from sklearn.utils import check_scalar
 
+    if sklearn_check_version("1.2"):
+        from sklearn.utils._param_validation import StrOptions
+
     from sklearn.decomposition import PCA as sklearn_PCA
 
     from onedal.decomposition import PCA as onedal_PCA
@@ -46,6 +49,10 @@ if daal_check_version((2024, "P", 100)):
 
         if sklearn_check_version("1.2"):
             _parameter_constraints: dict = {**sklearn_PCA._parameter_constraints}
+            if not sklearn_check_version("1.5"):
+                _parameter_constraints["svd_solver"] = StrOptions(
+                    {"auto", "full", "covariance_eigh", "arpack", "randomized"}
+                )
 
         if sklearn_check_version("1.1"):
 
--- a/sklearnex/decomposition/pca.py
+++ b/sklearnex/decomposition/pca.py
@@ -162,20 +162,20 @@ if daal_check_version((2024, "P", 100)):
 
         def _onedal_transform(self, X, queue=None):
             check_is_fitted(self)
+            if sklearn_check_version("1.0"):
+                self._check_feature_names(X, reset=False)
             X = self._validate_data(
                 X,
                 dtype=[np.float64, np.float32],
                 reset=False,
             )
             self._validate_n_features_in_after_fitting(X)
-            if sklearn_check_version("1.0"):
-                self._check_feature_names(X, reset=False)
 
             return self._onedal_estimator.predict(X, queue=queue)
 
         def fit_transform(self, X, y=None):
             if sklearn_check_version("1.5"):
-                U, S, Vt, X, x_is_centered, xp = self._fit(X)
+                U, S, Vt, X_fit, x_is_centered, xp = self._fit(X)
             else:
                 U, S, Vt = self._fit(X)
             if hasattr(self, "_onedal_estimator"):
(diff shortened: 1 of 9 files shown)


sklearnex/decomposition/pca.py
```diff
@@ -35,6 +35,9 @@
     if sklearn_check_version("1.1") and not sklearn_check_version("1.2"):
         from sklearn.utils import check_scalar
 
+    if sklearn_check_version("1.4"):
+        from sklearn.utils._array_api import get_namespace
```
icfaust, line 39:
```suggestion
    from sklearnex.utils import get_namespace

```

sklearnex/tests/test_n_jobs_support.py
```diff
@@ -73,6 +74,9 @@ def check_methods_decoration(estimator):
     # by default, [Nu]SVC.predict_proba is restricted by @available_if decorator
     if estimator_class in [SVC, NuSVC]:
         estimator_kwargs["probability"] = True
+    # by default, sklearn 1.5 PCA selects non-patched method
+    if estimator_class == PCA:
+        estimator_kwargs["svd_solver"] = "full"
```
icfaust, line 79: Ideally test_n_jobs_support should utilize sklearnex.tests._utils.SPECIAL_INSTANCES to make sure that the loss of a default check is covered by all of the test suite (and should be added there). This will save us headaches in the future.  Have we talked with @md-shafiul-alam about what we would need to do to support the default method?

icfaust: Btw, I don't think its necessary for this PR to do it, but something we should plan for.

Alexsandruss: It should be separate discussion/PR.
`SPECIAL_INSTANCES` has a specific behaviour: it initializes estimators once and will reuse them in different places. While estimator API theoretically allows this without any troubles, usage of same estimator instance in different tests might cause unexpected behaviour in future.

icfaust: Thats absolutely true Maybe I should try to insert clone() at various places in the other tests using it.

icfaust: Please look at #1801, which should address that concern.

sklearnex/decomposition/pca.py
```diff
@@ -156,32 +162,38 @@ def transform(self, X):
...
                 U = U[:, : self.n_components_]
 
                 if self.whiten:
-                    U *= sqrt(X.shape[0] - 1)
+                    U *= sqrt(X_fit.shape[0] - 1)
                 else:
                     U *= S[: self.n_components_]
 
                 return U
+            else:
```
md-shafiul-alam, line 194: The current default method of sklearnex PCA is "covariance_eigh" although we call it "full". We don't have to fallback to sklearn. The _is_solver_compatible method can be updated to return true if the method=="covariance_eigh".

sklearnex/decomposition/pca.py
```diff
@@ -129,7 +129,7 @@ def _onedal_fit(self, X, queue=None):
             onedal_params = {
                 "n_components": self.n_components,
                 "is_deterministic": True,
-                "method": "cov",
+                "method": "cov" if self._fit_svd_solver == "covariance_eigh" else "svd",
```
md-shafiul-alam, line 132: "svd" is slower than "cov" at the moment,  better to keep "cov" as default for now until "svd" is optimized. a separate investigation will be done to tune svd_solver for different data size

Alexsandruss: When svd and covariance solvers are simultaneously available in sklearn API, always defaulting to covariance is direct inconsistency with sklearn and disinforms users about used implementation.

aepanchi:
I'm good with the doc changes

icfaust (approved):
This approval is contingent on the PCA discussion being completed (i.e. approval by @md-shafiul-alam or someone else with deeper knowledge on the PCA implementation). I think the other changes look good or have been discussed elsewhere (like the ensemble deselections).

sklearnex/decomposition/pca.py
md-shafiul-alam: Agree on the inconsistency but the svd has been implemented very recently. Would be nice to have a perf measurement and decide on the trade-off.

md-shafiul-alam: BTW, "svd" is not an option for user input.

Alexsandruss: Considering oneDAL's SVD perf., I decided to choose next approach for svd_solver dispatching:
- Leave previous behaviour for sklearn<1.5
- If `full` svd_solver is explicitly requested fallback to stock sklearn for versions>=1.5
- If `auto` svd_solver selection of stock sklearn dispatches to `full` or `covariance_eigh`, use oneDAL PCA-Covariance with warning for `full` solver

md-shafiul-alam: The idea of explicit "full" dispatched to use SVD sounds good to me. Only thing is, oneDAL SVD solver should still outperform sklearn_svd_solver, unless you noticed otherwise. In that case, when explicit "full" is called, we can dispatch to onedal_svd.

Alexsandruss: oneDAL SVD performance improvement is not consistent on all cases/configuration, so it is better to leave it out of patch until SVD perf. is fixed/improved.

md-shafiul-alam (approved):
Looks good to me, the CI failures are addressed in https://github.com/intel/scikit-learn-intelex/pull/1812

PR #1795, opened 2024-04-12
[testing] Remove sklearn mixin dependence from test_patching

icfaust:
# Description
Some sklearn estimators do not have mixins associated with them, with a large example being EmpiricalCovariance. Without a sklearn mixin, it is still important to test the 'fit' method for operation. This adds the possibility in test_patching. When EmpiricalCovariance is taken out of preview, standardized testing should still be made available for it.  This will also be the case when IncrementalEmpiricalCovariance is added to the patch map. This PR is a snippet from #1674.

- Modify sklearnex/tests/_utils.py::gen_models_info to return the estimator and None if no mixin is associated with the estimator
- Remove mixin assert from test_if_estimator_inherits_sklearn
- Add checks for a method in test_standard_estimator_patching and test_special_estimator_patching

### DIFF
--- a/sklearnex/tests/_utils.py
+++ b/sklearnex/tests/_utils.py
@@ -119,6 +119,12 @@ def gen_models_info(algorithms):
                 methods |= candidates & set(method)
 
         output += [[i, j] for j in methods]
+
+    # In the case that no methods are available, set method to None.
+    # This will allow estimators without mixins to still test the fit
+    # method in various tests.
+    if not output:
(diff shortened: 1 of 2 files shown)


icfaust:
rerun for sporadic check: http://intel-ci.intel.com/eefbed7f-5ff5-f130-afb1-a4bf010d0e2e

Alexsandruss:
Assuming CI has sporadic non-related failures

PR #1799, opened 2024-04-15
[bug] fix sample_weight check for IncrementalBasicStatistics

icfaust:
# Description
Uncovered in #1763, IncrementalBasicStatistics assumes a numpy, dpnp or dpctl array. Passing other elements (i.e. pandas dataframes, pandas series etc) will not work properly, also doesn't seem like there are element size/shape checks vs X either. This will also add finite checks as well.

- Add _check_sample_weight from sklearn.utils.validation (used commonly in sklearn)
- switch weights keyword to sample_weights to match sklearn convention
- add y=None to fit to follow sklearn convention
- remove call to self.copy_X which doesn't exist in any instance.

### DIFF
--- a/sklearnex/basic_statistics/incremental_basic_statistics.py
+++ b/sklearnex/basic_statistics/incremental_basic_statistics.py
@@ -251,7 +251,7 @@ class IncrementalBasicStatistics(BaseEstimator):
                 "sklearn": None,
             },
             X,
-            weights,
+            sample_weights,
         )
         return self
 
--- a/sklearnex/basic_statistics/incremental_basic_statistics.py
+++ b/sklearnex/basic_statistics/incremental_basic_statistics.py
@@ -17,6 +17,7 @@
 import numpy as np
 from sklearn.base import BaseEstimator
 from sklearn.utils import check_array, gen_batches
+from sklearn.utils.validation import _check_sample_weights
 
 from daal4py.sklearn._n_jobs_support import control_n_jobs
 from daal4py.sklearn._utils import sklearn_check_version
@@ -139,7 +140,7 @@ class IncrementalBasicStatistics(BaseEstimator):
         self._onedal_estimator.finalize_fit()
         self._need_to_finalize = False
 
-    def _onedal_partial_fit(self, X, weights, queue):
+    def _onedal_partial_fit(self, X, sample_weight=None, queue=None):
         first_pass = not hasattr(self, "n_samples_seen_") or self.n_samples_seen_ == 0
 
         if sklearn_check_version("1.0"):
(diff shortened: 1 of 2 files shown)


sklearnex/basic_statistics/tests/test_incremental_basic_statistics.py
```diff
@@ -272,15 +272,15 @@ def test_fit_single_option_on_random_data(
     X = X.astype(dtype=dtype)
     X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
     if weighted:
-        weights = gen.uniform(low=-0.5, high=+1.0, size=row_count)
+        weights = gen.uniform(low=0.0, high=1.0, size=row_count)
```
olegkkruglov, line 275: what is the reason of this change? does it not work with negative weights?

sklearnex/basic_statistics/incremental_basic_statistics.py
```diff
@@ -152,9 +153,11 @@ def _onedal_partial_fit(self, X, weights, queue):
             X = check_array(
                 X,
                 dtype=[np.float64, np.float32],
-                copy=self.copy_X,
             )
 
+        if sample_weight is not None:
+            sample_weight = _check_sample_weight(sample_weight, X)
```
samir-nasibli, line 159: If I am not wrong we are already doing such sample weight checks on onedal side

icfaust: Unfortunately it was missed when IncrementalBasicStatistics was pulled in recently. This check guarantees that the supported datatype is some flavor of numpy for dtype checks, which was uncovered in the Pandas dataframe support PR. This is also causing issues with fp32 support for GPU,

sklearnex/basic_statistics/tests/test_incremental_basic_statistics.py
icfaust: Other than being very very unusual. I guess what we should do is add a test which inverts the weights and tests to see if it matches min to max, etc etc.

icfaust: Yep, exactly, I guess in my experience before I had only interacted with circumstances where positive weight values were necessary. I guess if I wanted to normalize against another parameter/ calculate the moment of data using the various statistics which weren't positively constrained.  Btw, why was it using +1.0 vs just 1.0? Just seems stylistically different.

icfaust: I'll revert those changes, just let me know.

olegkkruglov: I would keep negative weights in tests. I'm pretty sure that such applications exist (maybe they are not as widespread). And if our implementation supports negative weights, then imo better to test with negatives as well. 

Regarding +1.0, I probably just copied that from batch basic stat test, I agree that it's better to remove.

samir-nasibli (approved):
Looks good to me. Waiting your last decision on weights generation.

olegkkruglov (approved):
LGTM, thanks for fix. please wait private CI before merge.

PR #1800, opened 2024-04-15
[enhancement, testing] add initial array_api testing capabilities using array-api-strict

icfaust:
# Description
Array-api-strict is a module which exposes only the bare essentials defined by the array-api standard. This wraps over top numpy.  This will allow for testing of array_api compatability in sklearnex (only on CPU).  This will be enabled by either setting get_dataframes_and_queues specifically with "array_api" in the string, or by setting a sklearn context (via set_context or config_context) with array_api_dispatch=True. As this context is available only starting in sklearn1.2, the two options are necessary.  Ideally the first modules will be able to use specific new tests, and later then a context can be modified for global testing, similar to how GPU testing is done.

- add array-api-strict to test requirements
- add import checks in _dataframes_support
- extend get_dataframes_and_queues
- extend convert_to_supported


KulikovNikita (commented):
How does it work with the data management update PR?

requirements-test.txt
```diff
@@ -12,3 +12,4 @@ lightgbm==4.3.0
 catboost==1.2.5
 shap==0.44.1 ; python_version == '3.8'
 shap==0.45.1 ; python_version >= '3.9'
+array-api-strict ; python_version >= '3.9'
```
ethanglaser, line 15: generally a version is required for requirements.txt/dependencies-dev

onedal/tests/utils/_dataframes_support.py
```diff
@@ -31,6 +32,24 @@
...
+try:
+    # This should be lazy imported in the
+    # future along with other popular
+    # array_api libraries when testing
+    # GPU-no-copy.
+    import array_api_strict
+
+    # Run check if "array_api_dispatch" is configurable
+    array_api_enabled = lambda: get_config()["array_api_dispatch"]
+    _ = array_api_enabled()
```
ethanglaser, line 44:
```suggestion
    array_api_enabled()
```

ethanglaser (approved):
No objections from my end, just minor suggestionss - @samir-nasibli thoughts on this?

samir-nasibli (approved):
@icfaust thank you very much for the great work done. I am merging it. In case if some updates will required, let's fix/update it on follow up PRs.

PR #1801, opened 2024-04-16
[bug, testing] remove estimator state indeterminacy from SPECIAL INSTANCES

icfaust:
# Description
SPECIAL_INSTANCES only works with a single instance of sklearnex estimators. This dict is needed for specifically testing oneDAL supported sklearnex features.  By using a single instance, tests may impact one another. By using clone, a new estimator is created using the same parameters without state. Clone depends on the ```__init__``` parameters of the estimator only.

- create a special dictionary object which will clone the special estimator instances for use
- use sklearn clone in ```__getitem__```

### DIFF
--- a/sklearnex/tests/_utils.py
+++ b/sklearnex/tests/_utils.py
@@ -17,6 +17,7 @@
 from inspect import isclass
 
 import numpy as np
+from sklearn import clone
 from sklearn.base import (
     BaseEstimator,
     ClassifierMixin,
@@ -25,7 +26,6 @@ from sklearn.base import (
     RegressorMixin,
     TransformerMixin,
 )
-from sklearn import clone
 from sklearn.datasets import load_diabetes, load_iris
 from sklearn.neighbors._base import KNeighborsMixin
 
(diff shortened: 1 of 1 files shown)


sklearnex/tests/_utils.py
```diff
@@ -87,18 +88,26 @@ def _load_all_models(with_sklearnex=True, estimator=True):
...
-        LogisticRegression(solver="newton-cg"),
-    ]
-}
+class sklearn_clone_dict(dict):
+
+    def __getitem__(self, key):
+        return clone(super().__getitem__(key))
+
+
+SPECIAL_INSTANCES = sklearn_clone_dict(
```
samir-nasibli:
```suggestion
class _sklearn_clone_dict(dict):

    def __getitem__(self, key):
        return clone(super().__getitem__(key))


SPECIAL_INSTANCES = _sklearn_clone_dict(
```

icfaust:
private CI failures due to unrelated PCA issues (not in special_instances)

PR #1803, opened 2024-04-17
MAINT: remove debug print from BS tests

samir-nasibli:
# Description
remove debug print from BS tests


olegkkruglov (approved):
LGTM, thanks for the fix

PR #1804, opened 2024-04-17
[enhancement] clarify math in LinearRegression

icfaust:
# Description
Previous variable name is_good_for_onedal is based off of the naming scheme in daal4py.  This naming covers-up an important characteristic of the input data that should be known to the developer and the user, namely that the input data should not be underdetermined (https://en.wikipedia.org/wiki/Underdetermined_system).  This PR changes that to simplify understanding going forward.

- is_good_for_onedal -> is_underdetermined


olegkkruglov (approved):
LGTM, I'll change it in https://github.com/intel/scikit-learn-intelex/pull/1656

samir-nasibli (approved):

PR #1808, opened 2024-04-22
MAINT: Fix test name for IncrementalBasicStatistics

olegkkruglov:
- Fixed duplicated testname in sklearnex test for IncrementalBasicStatistics which caused test skipping
- Fixed test tolerance

### DIFF
--- a/sklearnex/basic_statistics/tests/test_incremental_basic_statistics.py
+++ b/sklearnex/basic_statistics/tests/test_incremental_basic_statistics.py
@@ -300,7 +300,7 @@ def test_fit_single_option_on_random_data(
 @pytest.mark.parametrize("column_count", [10, 100])
 @pytest.mark.parametrize("weighted", [True, False])
 @pytest.mark.parametrize("dtype", [np.float32, np.float64])
-def test_partial_fit_multiple_options_on_random_data(
+def test_fit_multiple_options_on_random_data(
     dataframe, queue, num_batches, row_count, column_count, weighted, dtype
 ):
     seed = 77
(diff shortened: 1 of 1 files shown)


samir-nasibli (approved):

sklearnex/basic_statistics/tests/test_incremental_basic_statistics.py
```diff
@@ -165,7 +165,7 @@ def test_partial_fit_multiple_options_on_random_data(
             expected_sum(X),
         )
 
-    tol = 1e-5 if res_mean.dtype == np.float32 else 1e-7
+    tol = 3e-4 if res_mean.dtype == np.float32 else 1e-7
```
KulikovNikita, line 168: What is the reason of such an inaccuracy

Alexsandruss: There should be comment explaining this dramatic difference between fp64 and fp32 tolerance

olegkkruglov: I don't know what explanation you want. it just takes numbers from C++ side and does nothing with them. the question should be addressed to authors of corresponding C++ code. in this PR I recognized that one of tests doesn't run because of incorrect name, fixed it and fixed tolerance not properly aligned before to pass this test. the tolerance here now is the same as it already is in main for corresponding partial_fit test (https://github.com/intel/scikit-learn-intelex/blob/000dc26ce87a235ac13bb0f632ca57e7554fb096/sklearnex/basic_statistics/tests/test_incremental_basic_statistics.py#L340). and both tests would fail if tolerance is set more precisely.

olegkkruglov: regarding tolerance on C++ side
https://github.com/oneapi-src/oneDAL/blob/a6e75ea734e8520a051164a211895c62ae0767f4/cpp/oneapi/dal/algo/basic_statistics/test/fixture.hpp#L220

Alexsandruss (approved):

PR #1817, opened 2024-04-29
DEPS: numpy 2.0 support

ethanglaser:
# Description
General migration to numpy 2.0
- some issues with dependencies (numba via shap, and catboost) - https://github.com/catboost/catboost/issues/2671, temporarily skipping this until those issues are resolved respectively
- a few updated import paths and deprecations addressed


ethanglaser:
Seems pre-release package is not available via conda - will close and re-open when official drop occurs

dependencies-dev
```diff
@@ -3,6 +3,6 @@ Jinja2==3.1.3
 numpy==1.19.5 ; python_version <= '3.9'
 numpy==1.21.6 ; python_version == '3.10'
 numpy==1.23.5 ; python_version == '3.11'
-numpy==1.26.4 ; python_version >= '3.12'
+numpy==2.0.0b1 ; python_version >= '3.12'
```
Alexsandruss, line 6: There is fresher version (rc2) already: https://pypi.org/project/numpy/2.0.0rc2/

requirements-test.txt
```diff
@@ -2,14 +2,16 @@ pytest==7.4.4 ; python_version == '3.8'
...
 scikit-learn==1.2.2 ; python_version == '3.8'
 scikit-learn==1.4.2 ; python_version >= '3.9'
 pandas==2.0.3 ; python_version == '3.8'
-pandas==2.1.3 ; python_version >= '3.9'
+pandas==2.1.3 ; python_version >= '3.9' and python_version < '3.11'
+pandas==2.2.2 ; python_version >= '3.11'
 xgboost==2.0.3
 lightgbm==4.3.0
-catboost==1.2.5
+catboost==1.2.5 ; python_version < '3.12' # TODO: Remove 3.12 condition when catboost supports numpy 2.0
```
napetrov, line 14: Can/should we split this? CatBoost required only for FTI testing, so it's not required dependency for scikit

requirements-test.txt
```diff
@@ -2,14 +2,16 @@ pytest==7.4.4 ; python_version == '3.8'
 pytest==8.2.1 ; python_version >= '3.9'
 numpy>=1.19.5 ; python_version <= '3.9'
 numpy>=1.21.6 ; python_version == '3.10'
-numpy>=1.23.5 ; python_version >= '3.11'
+numpy>=1.23.5 ; python_version == '3.11'
+numpy==2.0.0rc2 ; python_version >= '3.12'
```
napetrov, line 6: We can even keep previous line and do not force this dependency. Might be instead just using separate steps?  But it's fine to keep as is.

One thing is that once 2.0 would be released it would be taken by most Python versions. So we have to put upper bound to control what will be tested with 1.x numpy

ethanglaser: In general yes - with some restrictions based on pandas and scikit-learn versions as well. I think its worth having broader discussion on how this is handled, but probably best for separate PR

requirements-test.txt
ethanglaser: Currently requirements-test contains all dependencies for public and internal CI. It could be split into "deps required for using the library after building" (ie scikit-learn) and "extra deps required only for testing" (ie catboost) if the original purpose of the file is for users to reference

requirements-test.txt
ethanglaser: For this situation specifically, numpy 2.0.0rc2 is not installed if >=1.23.5 is used (1.26 is instead)

napetrov:
@oleksandr-pavlyk @xaleryb - FYI

requirements-test.txt
napetrov: Should it be >= 2.0 to allow updates there?

napetrov (approved):

daal4py/sklearn/utils/validation.py
```diff
@@ -19,7 +19,14 @@
...
 import scipy.sparse as sp
-from numpy.core.numeric import ComplexWarning
+
+try:
+    # numpy_version >= 2.0
+    from numpy.exceptions import ComplexWarning
+except ModuleNotFoundError:
+    # numpy_version < 2.0
+    from numpy.core.numeric import ComplexWarning
+
```
oleksandr-pavlyk, line 29: I think it is better, and faster to do 

```python
if np.lib.NumpyVersion(np.__version__) >= "2.0.0a0":
    from numpy.exceptions import ComplexWarning
else:
    from numpy.core.numeric import ComplexWarning
```

onedal/utils/validation.py
```diff
@@ -20,6 +20,14 @@
...
 import numpy as np
 from scipy import sparse as sp
+
+try:
+    # numpy_version >= 2.0
+    from numpy.exceptions import VisibleDeprecationWarning
+except ModuleNotFoundError:
+    # numpy_version < 2.0
+    from numpy import VisibleDeprecationWarning
+
```
oleksandr-pavlyk, line 30: Same comment as above

requirements-test.txt
ethanglaser: Good idea, added it

ethanglaser:
will merge tomorrow if no objections

PR #1825, opened 2024-05-07
Fix DPNP/DPCTL F-contiguous data processing

Vika-F:
More correct processing of F-contiguous data was added into _device_offload.py.


samir-nasibli (approved):
Assuming green CI.
Thank you!

Alexsandruss (approved):

PR #1829, opened 2024-05-08
[enhancement] block use of sklearn_check_version in `onedal/`

icfaust:
### Description 
Add a design rule which forces all sklearn-related interfacing to occur in its proper package, namely sklearnex.  This prevents bad estimator architectures which place too much of the sklearn-related aspects in the wrong area. This test crawls through the onedal folder and greps for "sklearn_check_version".  Ideally this test would have no exceptions going forward. This adds the onedal/tests/test_common.py file, which will likely be expanded in the future with more onedal folder design checks.

Tasks -

- [x] Initial compile
- [x] correct public CI failures
- [x] Split into separate PRs as necessary (and merge them)
- [ ] pass private CI

### DIFF
--- a/dependencies-dev
+++ b/dependencies-dev
@@ -5,4 +5,4 @@ numpy==1.21.6 ; python_version == '3.10'
 numpy==1.23.5 ; python_version == '3.11'
 numpy==1.26.4 ; python_version >= '3.12'
 pybind11==2.12.0
-cmake==3.29.2
+cmake==3.29.3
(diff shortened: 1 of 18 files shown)


samir-nasibli (commented):
Maybe it make sense just ban or monitor this during review?

Refactoring is needed for:
https://github.com/intel/scikit-learn-intelex/blob/ea79e8f212e3a9431277023e2d54d5dc8e629abc/onedal/svm/svm.py#L25

https://github.com/intel/scikit-learn-intelex/blob/ea79e8f212e3a9431277023e2d54d5dc8e629abc/onedal/ensemble/forest.py#L27

samir-nasibli (approved):
Thank you! Assuming green CI

PR #1834, opened 2024-05-14
[enhancement] remove sklearn_check_version dependence from `onedal/ensemble`

icfaust:
### Description 
Add a design rule which forces all sklearn-related interfacing to occur in its proper package, namely sklearnex.  This prevents bad estimator architectures which place too much of the sklearn-related aspects in the wrong area.   Ideally this test would have no exceptions going forward.

This is a subsegment of #1829 which handles just the ensemble algorithm so that sklearn_check_version is not in onedal/ensemble

Tasks -

- [ ] Initial compile
- [ ] correct public CI failures
- [ ] Split into separate PRs as necessary (and merge them)
- [ ] pass private CI

### DIFF
--- a/onedal/ensemble/forest.py
+++ b/onedal/ensemble/forest.py
@@ -24,7 +24,7 @@ import numpy as np
 from sklearn.ensemble import BaseEnsemble
 from sklearn.utils import check_random_state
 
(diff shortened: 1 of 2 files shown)


samir-nasibli (approved):
Assuming green CI

PR #1361, opened 2023-07-13
Implemented proper work with multiple threads

olegkkruglov:
Changes proposed in this pull request:
- Add patching for _FuncWrapper class from sklearn.utils.parallel to properly call patched config_context
- Fix patching for get_config to allow run patched version in sklearn.utils.parallel
- Remove unnecessary queue propagation via config in BaseSVC._fit_proba which causes test failure after get_config patch fix

PR #1835, opened 2024-05-14
[enhancement] remove sklearn_check_version dependence from `onedal/svm`

icfaust:
### Description 
Add a design rule which forces all sklearn-related interfacing to occur in its proper package, namely sklearnex.  This prevents bad estimator architectures which place too much of the sklearn-related aspects in the wrong area.   Ideally this test would have no exceptions going forward.

This is a subsegment of #1829 which handles just the svm algorithms so that sklearn_check_version is not in onedal/svm

This also solves the PR #1727 by removing sklearn inheritance from onedal/svm classes

This also solves issues with gpu offloading with fit_proba that had been removed errantly in #1361 

This also fixes an issue where sample_weights are always generated, which may impact performance.

Tasks -

- [x] Initial compile
- [x] correct public CI failures
- [x] Split into separate PRs as necessary (and merge them)
- [ ] pass private CI

### DIFF
--- a/onedal/svm/svm.py
+++ b/onedal/svm/svm.py
@@ -130,6 +130,24 @@ class BaseSVM(metaclass=ABCMeta):
             accept_sparse="csr",
         )
         y = self._validate_targets(y, X.dtype)
+        if sample_weight is not None and len(sample_weight) > 0:
+            sample_weight = _check_array(
+                sample_weight,
+                accept_sparse=False,
+                ensure_2d=False,
+                dtype=X.dtype,
+                order="C",
+            )
+        elif self.class_weight is not None:
+            sample_weight = np.ones(X.shape[0], dtype=X.dtype)
+
+        if sample_weight is not None:
+            if self.class_weight_ is not None:
+                for i, v in enumerate(self.class_weight_):
+                    sample_weight[y == i] *= v
+            data = (X, y, sample_weight)
+        else:
+            data = (X, y)
         self._sparse = sp.issparse(X)
 
         if self.kernel == "linear":
@@ -155,9 +173,9 @@ class BaseSVM(metaclass=ABCMeta):
                 _gamma = self.gamma
             self._scale_, self._sigma_ = _gamma, np.sqrt(0.5 / _gamma)
 
-        policy = _get_policy(queue, X, y, sample_weight)
+        policy = _get_policy(queue, *data)
         params = self._get_onedal_params(X)
-        result = module.train(policy, params, *to_table(X, y, sample_weight))
+        result = module.train(policy, params, *to_table(*data))
 
         if self._sparse:
             self.dual_coef_ = sp.csr_matrix(from_table(result.coeffs).T)
--- a/onedal/svm/svm.py
+++ b/onedal/svm/svm.py
@@ -16,13 +16,10 @@
 
 from abc import ABCMeta, abstractmethod
 from enum import Enum
-from numbers import Number, Real
 
 import numpy as np
 from scipy import sparse as sp
-from sklearn.base import BaseEstimator
 
-from daal4py.sklearn._utils import sklearn_check_version
 from onedal import _backend
 
 from ..common._estimator_checks import _check_is_fitted
@@ -45,7 +42,7 @@ class SVMtype(Enum):
     nu_svr = 3
 
 
-class BaseSVM(BaseEstimator, metaclass=ABCMeta):
+class BaseSVM(metaclass=ABCMeta):
     @abstractmethod
     def __init__(
         self,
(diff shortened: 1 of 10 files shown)


onedal/svm/svm.py
```diff
@@ -45,7 +42,7 @@ class SVMtype(Enum):
     nu_svr = 3
 
 
-class BaseSVM(BaseEstimator, metaclass=ABCMeta):
```
samir-nasibli, line 48: Please close https://github.com/intel/scikit-learn-intelex/pull/1727 in case of green ci checks in this PR

sklearnex/svm/_common.py
```diff
@@ -100,27 +224,38 @@ def _compute_balanced_class_weight(self, y):
         return recip_freq[le.transform(classes)]
 
     def _fit_proba(self, X, y, sample_weight=None, queue=None):
+        from .._config import config_context, get_config
```
Alexsandruss, line 227: Runtime import is not necessary overhead, especially, when `.._config` is not cached.

Alexsandruss (changes requested):

samir-nasibli:
@icfaust 
onedal4py depends on sklearn. Could you please clarify more why we can not use sklearn versioning here?

sklearnex/svm/_common.py
icfaust: I guess that was a mistake from the original design (taken directly from #1361), will change it.

icfaust:
> @icfaust onedal4py depends on sklearn. Could you please clarify more why we can not use sklearn versioning here?

The onedal folder installs via setup.py, requirements for the setup.py do not include a sklearn version (as per INSTALL.md). Therefore, if the user were install onedal without having run setup_sklearnex.py, they are in for a nasty surprise. 

Secondly, sklearn_check_version used only for input verification and conformance on raising errors in sklearn (in practice in `onedal/`). By design, sklearn conformance should be done in the sklearnex directory, and not in the onedal directory. Why would a user care about sklearn if they were using onedal classes directly? For example, why else have these functions defined at all? https://github.com/intel/scikit-learn-intelex/blob/main/onedal/utils/validation.py

Thirdly, operational maintenance of the onedal folder should be minimized (i.e. . By using sklearn_check_version, it is increasing our burden in observation and knowledge into the various onedal classes).

If we want to remove sklearn entirely, we need to fix issues in (not counting testsing):
`onedal/cluster/kmeans.py`
`onedal/cluster/dbscan.py`
`onedal/ensemble/forest.py`
`onedal/common/_mixin.py`
`onedal/utils/validation.py` (This one would be difficult)

samir-nasibli:
> > @icfaust onedal4py depends on sklearn. Could you please clarify more why we can not use sklearn versioning here?
> 
> The onedal folder installs via setup.py, requirements for the setup.py do not include a sklearn version (as per INSTALL.md). Therefore, if the user were install onedal without having run setup_sklearnex.py, they are in for a nasty surprise.
> 
> Secondly, sklearn_check_version is used only for input verification and conformance on raising errors in sklearn (in practice in `onedal/`). By design, sklearn conformance should be done in the sklearnex directory, and not in the onedal directory. Why would a user care about sklearn if they were using onedal classes directly? For example, why else have these functions defined at all? https://github.com/intel/scikit-learn-intelex/blob/main/onedal/utils/validation.py
> 
> Thirdly, operational maintenance of the onedal folder should be minimized (i.e. . By using sklearn_check_version, it is increasing our burden in observation and knowledge into the various onedal classes).
> 
> If we want to remove sklearn entirely, we need to fix issues in (not counting testing): `onedal/cluster/kmeans.py` `onedal/cluster/dbscan.py` (solved in #1837) `onedal/ensemble/forest.py` `onedal/common/_mixin.py` `onedal/utils/validation.py` (This one would be difficult)

> > The onedal folder installs via setup.py, requirements for the setup.py do not include a sklearn version (as per INSTALL.md). Therefore, if the user were install onedal without having run setup_sklearnex.py, they are in for a nasty surprise.

`onedal4py` is already depends on `daal4py`, and it is actually part of  current `daal4py` package. `sklearn` is a runtime dependency, we are versioning it on `daal4py`, so why we can not do it on `onedal4py`.
Probably make sense design `onedal4py` as a independent, but currently it is not `sklearn` independent. And since for some primitives it is used in `onedal4py`, `sklearn` provide different ifaces/deprecate smth, so make sense also having versioning there.

sklearnex/svm/_common.py
```diff
@@ -85,6 +91,125 @@ def _onedal_cpu_supported(self, method_name, *data):
...
+
+            if sample_weight.shape != (n_samples,):
+                raise ValueError(
+                    "sample_weight.shape == {}, expected {}!".format(
+                        sample_weight.shape, (n_samples,)
+                    )
+                )
+
+        if np.all(sample_weight <= 0):
+            raise ValueError(self._err_msg)
```
Alexsandruss, line 209: There is no need to store `_err_msg` in `BaseSVM` class namespace
```suggestion
        if np.all(sample_weight <= 0):
            if "nusvc" in self.__module__:
                raise ValueError("negative dimensions are not allowed")
            else:
                raise ValueError("Invalid input - all samples have zero or negative weights.")
```

icfaust:
> > > @icfaust onedal4py depends on sklearn. Could you please clarify more why we can not use sklearn versioning here?
> > 
> > 
> > The onedal folder installs via setup.py, requirements for the setup.py do not include a sklearn version (as per INSTALL.md). Therefore, if the user were install onedal without having run setup_sklearnex.py, they are in for a nasty surprise.
> > Secondly, sklearn_check_version is used only for input verification and conformance on raising errors in sklearn (in practice in `onedal/`). By design, sklearn conformance should be done in the sklearnex directory, and not in the onedal directory. Why would a user care about sklearn if they were using onedal classes directly? For example, why else have these functions defined at all? https://github.com/intel/scikit-learn-intelex/blob/main/onedal/utils/validation.py
> > Thirdly, operational maintenance of the onedal folder should be minimized (i.e. . By using sklearn_check_version, it is increasing our burden in observation and knowledge into the various onedal classes).
> > If we want to remove sklearn entirely, we need to fix issues in (not counting testing): `onedal/cluster/kmeans.py` `onedal/cluster/dbscan.py` (solved in #1837) `onedal/ensemble/forest.py` `onedal/common/_mixin.py` `onedal/utils/validation.py` (This one would be difficult)
> 
> > > The onedal folder installs via setup.py, requirements for the setup.py do not include a sklearn version (as per INSTALL.md). Therefore, if the user were install onedal without having run setup_sklearnex.py, they are in for a nasty surprise.
> 
> `onedal4py` already depends on `daal4py`, and it is actually part of current `daal4py` package. `sklearn` is a runtime dependency, we are versioning it on `daal4py`, so why we can not do it on `onedal4py`? Probably make sense design `onedal4py` as a independent, but currently it is not `sklearn`-independent. And since for some primitives it is used in `onedal4py` and `sklearn` provide different ifaces/deprecate smth based on version, so make sense also having versioning there in my opinion for a while.

There is a key distinction here: if daal4py depends on sklearn versions, then problems with sklearn versioning will be solved in daal4py. If we add sklearn versioning to onedal4py, we increase our maintenance burden.  If you list the primitives that you mention I'd like to review, from my perspective sklearn has a stable API for what we use in onedal4py. If this becomes a problem in a future sklearn version, then this test will specifically trigger and highlight a discussion about the architecture which we should have.

Alexsandruss (approved):

samir-nasibli (approved):

PR #1836, opened 2024-05-15
[CI] Increase tolerance for IncrementalPCA gold data tests with whiten and float32 datatypes

icfaust:
### Description 
Increase tol in `onedal/decomposition/tests/test_incremental_pca.py` and `sklearnex/preview/decomposition/tests/test_incremental_pca.py` gold data checks

Fixes # - some change in incrementalPCA is causing larger error than before with whiten and float32 data, but is within acceptable range. Add if statements and increase the tolerance in this circumstance.


olegkkruglov (approved):
LGTM

PR #1839, opened 2024-05-21
[enhancement] block use of target_offload in `/sklearnex`

icfaust:
### Description 
Add a design rule which forces all estimators to properly offload to GPU not using config_context via target_offload.  This prevents bad estimator architectures which can hide queues. This test crawls through the sklearnex folder and greps for "target_offload".  Ideally this test would have no exceptions going forward (currently an exception is grandfathered for nuSVC and SVC). This adds the sklearnex/tests/test_common.py file, which will likely be expanded in the future with more sklearnex folder design checks.

Tasks -

- [ ] Initial compile
- [ ] correct public CI failures
- [ ] Split into separate PRs as necessary (and merge them)
- [ ] pass private CI

### DIFF
--- a/sklearnex/tests/test_common.py
+++ b/sklearnex/tests/test_common.py
@@ -19,7 +19,7 @@ from glob import glob
 
 import pytest
 
-ALLOWED_LOCATIONS = ["_config.py", "_device_offload.py", "test"]
+ALLOWED_LOCATIONS = ["_config.py", "_device_offload.py", "test", "svc.py"]
 
 
 def test_target_offload_ban():
--- a/sklearnex/tests/test_common.py
+++ b/sklearnex/tests/test_common.py
@@ -1,0 +1,48 @@
+# ==============================================================================
+# Copyright 2024 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import os
+from glob import glob
+
+import pytest
+
+ALLOWED_LOCATIONS = ["_config.py", "_device_offload.py", "test"]
+
+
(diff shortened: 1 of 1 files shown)


samir-nasibli:
@icfaust at #1829 we have similar test case. Maybe it make sense create common test suit for these banned primitives in certain failes/forlders/modules?

icfaust:
> @icfaust at #1829 we have similar test case. Maybe it make sense create common test suit for these banned primitives in certain files/forlders/modules?

Depends on how this would be integrated into the various test suites. It would seem to me that having design tests local to the package is the best. Though I am open to suggestion.

samir-nasibli:
> > @icfaust at #1829 we have similar test case. Maybe it make sense create common test suit for these banned primitives in certain files/forlders/modules?
> 
> Depends on how this would be integrated into the various test suites. It would seem to me that having design tests local to the package is the best. Though I am open to suggestion.

We can have the common part in onedal4py, could be reusable in sklearnex as well.

samir-nasibli:
> > > @icfaust at #1829 we have similar test case. Maybe it make sense create common test suit for these banned primitives in certain files/forlders/modules?
> > 
> > 
> > Depends on how this would be integrated into the various test suites. It would seem to me that having design tests local to the package is the best. Though I am open to suggestion.
> 
> We can have the common part in onedal4py, could be reusable in sklearnex as well.

This could be done in follow up to #1839 and #1829

samir-nasibli (approved):
LGTM. Assuming green CI
Thank you!

samir-nasibli:
> > > > @icfaust at #1829 we have similar test case. Maybe it make sense create common test suit for these banned primitives in certain files/forlders/modules?
> > > 
> > > 
> > > Depends on how this would be integrated into the various test suites. It would seem to me that having design tests local to the package is the best. Though I am open to suggestion.
> > 
> > 
> > We can have the common part in onedal4py, could be reusable in sklearnex as well.
> 
> This could be done in follow up to #1839 and #1829

@icfaust 
I think the best place for checking all banned primitives is [scikit-learn-intelex/tests](https://github.com/intel/scikit-learn-intelex/tree/main/tests). Generally this check related not with packages functionality, I think it is better them to be located in [scikit-learn-intelex/tests](https://github.com/intel/scikit-learn-intelex/tree/main/tests), and I would not want the testing to get into the packages, since this is not functional testing, but design testing.
We can as a follow up move all https://github.com/intel/scikit-learn-intelex/pull/1839, https://github.com/intel/scikit-learn-intelex/pull/1829 and https://github.com/intel/scikit-learn-intelex/pull/1849 in one place and generalize them.

PR #1770, opened 2024-03-22
KMeans OOP

md-shafiul-alam:
This PR intends to take KMeans algorithm out of preview namespace. The changes proposed:
 - KMeans out of preview
 - add sparsity support for KMeans
 - reduce number of deselected tests for KMeans

Nuances: KMeans Init does not have GPU support for sparse data. Hence CPU implementation is used using Host policy.

Based on https://github.com/intel/scikit-learn-intelex/pull/1634.
Needs to merge https://github.com/oneapi-src/oneDAL/pull/2815 first.

Fixes # - issue number(s) if exists
- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes, if necessary (updated in # - add PR number)
- [x] The unit tests pass successfully.
- [x] I have run it locally and tested the changes extensively.
- [x] I have resolved any merge conflicts that might occur with the base branch.
- [x] Git commit message contains an appropriate signed-off-by string (see [CONTRIBUTING.md](https://github.com/intel/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)
- [x] I have added a respective label(s) to PR if I have a permission for that.

PR #1843, opened 2024-05-23
ENH: Adding Ridge Regression support into `sklearnex.preview`

DDJHB:
### Description 
Current sklearnex Ridge uses daal4py Ridge implementation. This pr is part of moving it to onedal4py and enabling GPU support. Initial changes introduce onedal4py based Ridge to sklearnex preview.


sklearnex/preview/linear_model/tests/test_ridge.py
```diff
@@ -0,0 +1,51 @@
...
+
+import numpy
+import pytest
+
+from onedal.tests.utils._dataframes_support import (
+    _convert_to_dataframe,
+    get_dataframes_and_queues,
+)
+
+
```
icfaust, line 25: As written elsewhere, try testing with a sparse input too.

icfaust (commented):
Just my initial thoughts.  I guess I really dislike some aspects of the LinearRegression implementation, so that is most of the things I've commented on (or at least the source of it).

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,343 @@
...
+                ),
+            ]
+        )
+
+        if device == "gpu":
+            alpha_is_scalar = isinstance(self.alpha, (int, float))
+            dal_ready = patching_status.and_condition(
+                alpha_is_scalar,
+                "Non-scalar alpha is not supported for GPU.",
+            )
```
icfaust, line 185: Instead of passing the device (which is the purpose of _onedal_gpu_supported/_onedal_cpu_supported in the first place), start and pass the patchingconditionschain instead and move this check into _onedal_gpu_supported after the chain is initialized. You can remove device from the args, and replace it with the chain.  This will make things a little easier to follow.

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,343 @@
...
+                estimator_name=self.__class__.__name__,
+            )
+
+        self._initialize_onedal_estimator()
+        try:
+            # Falling back to daal4py if the device is CPU since
+            # onedal does not support non-scalars for alpha, thus
+            # should only be used for GPU to not limit the functionality
+            cpu_device = queue is None or queue.sycl_device.is_cpu
+            if cpu_device:
```
icfaust, line 288: I know its all going to the same place, but this causes the cpu branch of the oneDAL LinearRegression never to run via sklearnex. I sorta thought there would be an alpha check in the if statement...

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,343 @@
...
+
+    def set_intercept_(self, value):
+        self.__dict__["intercept_"] = value
+        if hasattr(self, "_onedal_estimator"):
+            self._onedal_estimator.intercept_ = value
+            del self._onedal_estimator._onedal_model
+
+    def _save_attributes(self):
+        self.coef_ = property(self.get_coef_, self.set_coef_)
+        self.intercept_ = property(self.get_intercept_, self.set_intercept_)
```
icfaust, line 336: Use the trick elsewhere to solve this (raising a specific attribute error will false out hasattr).

icfaust:
Somehow TSNE is seg faulting????

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,390 @@
...
+            if hasattr(self._onedal_estimator, "_onedal_model") and hasattr(
+                self, "_intercept"
+            ):
+                del self._onedal_estimator._onedal_model
+        self._intercept = value
+
+    def _save_attributes(self):
+        self.n_features_in_ = self._onedal_estimator.n_features_in_
+        self._sparse = False
+        self.coef_ = self._onedal_estimator.coef_
```
icfaust, line 385: I thought about it some, You can drop the and hasattr(self, "_coef") and hasattr(self, "_intercept") if you set self._coef and self._intercept in _save_attributes instead of self.coef_ and self.intercept_.

sklearnex/preview/linear_model/tests/test_ridge.py
```diff
@@ -0,0 +1,54 @@
...
+    X, y = numpy.random.rand(sample_count, feature_count), numpy.random.rand(sample_count)
+    X_c = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+    y_c = _convert_to_dataframe(y, sycl_queue=queue, target_df=dataframe)
+
+    sklearnex_model.fit(X_c, y_c)
+    sklearn_model.fit(X, y)
+    numpy.testing.assert_allclose(sklearnex_model.coef_, sklearn_model.coef_)
+
+    prediction_sklearnex = numpy.asarray(
```
samir-nasibli, line 46: please use `_to_numpy` instead https://github.com/intel/scikit-learn-intelex/blob/15eb80ee3d6fb688c701e18c250fd5bba84a11e6/onedal/tests/utils/_dataframes_support.py#L86

sklearnex/preview/linear_model/tests/test_ridge.py
```diff
@@ -0,0 +1,54 @@
...
+
+from onedal.tests.utils._dataframes_support import (
+    _convert_to_dataframe,
+    get_dataframes_and_queues,
+)
+
+
+@pytest.mark.parametrize("alpha", [0.5, 1.0, 5.0])
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+def test_ridge_sklearnex_sklearn_conformance(alpha, dataframe, queue):
```
samir-nasibli, line 28: We already doing this via sklearn testing

samir-nasibli (commented):
@DDJHB please provide more details even for the draft PR.
It make sense high level highlight your suggestions

sklearnex/preview/linear_model/tests/test_ridge.py
ethanglaser: Agreed, we don't need to define our own testing to validate on scikit-learn. Can get creative with devising tests here. I recommend basing off of existing linear regression coverage: https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/linear_model/tests/test_linear.py. With linear regression there is simple solution, ridge is a bit more complicated but can be done using numpy linalg operations, could be a good place to start

ethanglaser (commented):
Partially reviewed (haven't fully looked at the actual ridge.py). Need to add test coverage on onedal4py side, can use https://github.com/intel/scikit-learn-intelex/blob/main/onedal/linear_model/tests/test_linear_regression.py as a reference (not all tests will be applicable)

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,422 @@
...
+
+            res = self._onedal_estimator.predict(X, queue=queue)
+            return res
+
+        def _onedal_score(self, X, y, sample_weight=None, queue=None):
+            return r2_score(
+                y, self._onedal_predict(X, queue=queue), sample_weight=sample_weight
+            )
+
+        @property
```
icfaust, line 381: This is the way I would like it done in the KMeans OOP #1770

sklearnex/dispatcher.py
```diff
@@ -90,6 +91,15 @@ def get_patch_map_core(preview=False):
                     None,
                 ]
             ]
+
+            # Ridge
+            linear_model_module, _, _ = mapping["ridge"][0][0]
+            sklearn_obj = mapping["ridge"][0][1]
```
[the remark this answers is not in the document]
DDJHB: To my best understanding so far - it is used as fallback object when sklearnex functionality fails. daal4py dispatcher fills it with stock sklearn object when patching happens and replaces sklearnex one when unpatch is needed. Since we are handling previews "manually" in sklearnex dispatcher, need to put the stock sklearn object there so during possible unpatch it knows what to fallback to.

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,422 @@
...
+        def _test_type_and_finiteness(self, X_in):
+            xp, _ = get_namespace(X_in)
+            X = xp.asarray(X_in)
+
+            if np.iscomplexobj(X):
+                return False
+
+            try:
+                _assert_all_finite(X)
+            except BaseException:
```
samir-nasibli, line 167: shouldn't be ValueError here?

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,422 @@
...
+                {
+                    "onedal": self.__class__._onedal_score,
+                    "sklearn": sklearn_RidgeRegression.score,
+                },
+                X,
+                y,
+                sample_weight=sample_weight,
+            )
+
+        def _test_type_and_finiteness(self, X_in):
```
samir-nasibli, line 158:
```suggestion
        def _check_type_and_finiteness(self, X_in):
```

sklearnex/preview/linear_model/ridge.py
DDJHB: not sure about this one, took it from other algos. Since the function only returns True/False I would assume it does not matter what this except catches, any exception means the "test" failed but we really want the parent function test_type_and_finiteness to proceed and the "error" (False) to be handled within patching_status that wraps around it. Another reason might be that the implementation of _assert_all_finite function uses methods from numpy/sklearn, so to be safe against the exceptions those could raise BaseException is used. Overall, I agree that we lose some granularity of potential errors, so if you think the use of BaseException is still unjustifiable, we can agree on the change.

sklearnex/preview/linear_model/ridge.py
DDJHB: using prefix "_test" to conform to sklearnex.linear_model.logistic_regression.py/linear.py style using the same function definition

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,434 @@
...
+
+    from onedal.linear_model import LinearRegression as onedal_RidgeRegression
+    from onedal.utils import _num_features, _num_samples
+
+    from ..._device_offload import dispatch, wrap_output_data
+    from ..._utils import PatchingConditionsChain, get_patch_message
+    from ...utils import get_namespace
+    from ...utils.validation import _assert_all_finite
+
+    def is_numeric_scalar(value):
```
samir-nasibli, line 43:
```suggestion
    def _is_numeric_scalar(value):
```
Generally, make sense maybe add this into onedal.utils._validation, could be re-usable.

Another question? doesn't `numpy.isscalar` cover this?

samir-nasibli: Honestly I don't see any needity for this func. it is called twice only for alpha param. Make sense just use isinstance(value, (int, float)) inplace. No need for additional function at least in ridge module. If it is required in several places in sklearnex, make sense to add in onedal.utils._validation and reuse it.

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,434 @@
...
+                self._save_attributes()
+
+            except RuntimeError:
+                logging.getLogger("sklearnex").info(
+                    f"{self.__class__.__name__}.fit "
+                    + get_patch_message("sklearn_after_onedal")
+                )
+
+                del self._onedal_estimator
+                super().fit(X, y)
```
samir-nasibli, line 374: Not in all our estimators we have this fallback after dispatching. I think we should avoid it. I think all fallbacks should occur on dispatching.

I will escalate removal of such fallback after fail, but right now I am curios how much conformance tests we are failing without this?

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,434 @@
...
+
+        def _onedal_fit_supported(self, patching_status, method_name, *data):
+            assert method_name == "fit"
+            assert len(data) == 3
+            X, y, sample_weight = data
+
+            if not patching_status:
+                patching_status = PatchingConditionsChain(
+                    f"sklearn.linear_model.{self.__class__.__name__}.fit"
+                )
```
samir-nasibli, line 192: I don't think that this is required

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,422 @@
...
+
+    from onedal.common.hyperparameters import get_hyperparameters
+    from onedal.linear_model import LinearRegression as onedal_RidgeRegression
+    from onedal.utils import _num_features, _num_samples
+
+    from ..._device_offload import dispatch, wrap_output_data
+    from ..._utils import (
+        PatchingConditionsChain,
+        get_patch_message,
+        register_hyperparameters,
```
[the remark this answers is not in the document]
samir-nasibli: If it is required please don't forget to cover this enablement with tickets

sklearnex/preview/linear_model/ridge.py
samir-nasibli: It is strange, that they using "_test" prefix. But I think we can go with you proposal then. Thank you!

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,434 @@
...
+                self._save_attributes()
+
+            except RuntimeError:
+                logging.getLogger("sklearnex").info(
+                    f"{self.__class__.__name__}.fit "
+                    + get_patch_message("sklearn_after_onedal")
+                )
+
+                del self._onedal_estimator
+                super().fit(X, y)
```
samir-nasibli, line 374: I have a strong opinion about this kind of fallbacks.

It should be removed. If something doesn't work on our side according to the given parameters, then it should be determined at the moment of dispatch call and within the framework of _onedal_cpu_support/_onedal_gpu_support checks.

I suggest remove this at all and try to address fails on `_onedal_cpu_support`/`_onedal_gpu_support` patching checks. I would like see how much test fails do we have in this case and want to see conformance test report.

samir-nasibli (dismissed):

sklearnex/preview/linear_model/ridge.py
DDJHB: I agree, thanks for emphasizing this, inherited it from other algos as well. Will try your suggestion and report back.

samir-nasibli:
@DDJHB 
Thank you for the very good updates!

I propose to have common class for the LinearRegression and the Ridge, and have separate interface for the Ridge and LinearRegression on onedal4py side, since we would like to have SPMD iface for the Ridge as well as a next step.

onedal/linear_model/tests/test_ridge.py
```diff
@@ -0,0 +1,95 @@
...
+    def test_diabetes(queue, dtype):
+        X, y = load_diabetes(return_X_y=True)
+        X, y = X.astype(dtype), y.astype(dtype)
+        X_train, X_test, y_train, y_test = train_test_split(
+            X, y, train_size=0.8, random_state=777
+        )
+        model = LinearRegression(fit_intercept=True, alpha=0.1)
+        model.fit(X_train, y_train, queue=queue)
+        y_pred = model.predict(X_test, queue=queue)
+        assert mean_squared_error(y_test, y_pred) < 2396
```
md-shafiul-alam, line 41: Is it possible to be more deterministic here? Even if dropping MSE is a good thing, might be good to throw an error.

onedal/linear_model/tests/test_ridge.py
```diff
@@ -0,0 +1,95 @@
...
+    @pytest.mark.parametrize("queue", get_queues())
+    @pytest.mark.parametrize("dtype", [np.float32, np.float64])
+    def test_no_intercept_results(queue, dtype):
+        seed = 42
+        f_count, r_count = 19, 7
+        s_count, t_count = 3500, 1999
+
+        gen = np.random.default_rng(seed)
+
+        X = gen.random(size=(s_count, f_count), dtype=dtype)
```
md-shafiul-alam, line 71: better naming is advisable for the sake of developers after you :). Might consider n_sample, n_feature etc.

onedal/linear_model/tests/test_ridge.py
DDJHB: this is copy paste from [lin reg test](https://github.com/intel/scikit-learn-intelex/blob/2cd3be447f80da891c3275c9509d7b66ec2ab339/onedal/linear_model/tests/test_linear_regression.py#L30), not sure if modifying it makes sense. If that is not what you meant, could you please clarify exactly what part of this is not deterministic enough or where exactly an error should be thrown?

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,412 @@
...
+
+        if np.iscomplexobj(X):
+            return False
+
+        try:
+            _assert_all_finite(X)
+        except BaseException:
+            return False
+
+        return True
```
samir-nasibli, line 75: shouldn't this be Ridge class method? Is it required to be free func?

sklearnex/preview/linear_model/tests/test_ridge.py
```diff
@@ -0,0 +1,97 @@
...
+    y_c = _convert_to_dataframe(y, sycl_queue=queue, target_df=dataframe)
+    ridgereg = Ridge(alpha=0.5).fit(X_c, y_c)
+
+    if daal_check_version((2024, "P", 600)):
+        assert "preview" in ridgereg.__module__
+    else:
+        assert "daal4py" in ridgereg.__module__
+
+    numpy.testing.assert_allclose(ridgereg.intercept_, 3.86, rtol=1e-2)
+    numpy.testing.assert_allclose(ridgereg.coef_, [0.91, 1.64], rtol=1e-2)
```
samir-nasibli, line 45: Please do import numpy.testing.assert_allclose on top and use it

icfaust (commented):
Please merge in main, run private CI and then ping me when it completes, I'll give it a look.

sklearnex/preview/linear_model/ridge.py
DDJHB: does not use anything related to class object, why make it a method? i would consider moving it to general utils that can be reused across different modules

samir-nasibli: My comment based on your previous replies. https://github.com/intel/scikit-learn-intelex/pull/1843#discussion_r1656944833. And here it is a method. 
Previously my comment on moving this into utils module was ignored, so I decided to clarify.

If this is going to be common utility than I am totally ok with this.

sklearnex/preview/linear_model/tests/test_ridge.py
```diff
@@ -0,0 +1,97 @@
...
+        assert "preview" in ridgereg.__module__
+    else:
+        assert "daal4py" in ridgereg.__module__
+
+    assert_allclose(ridgereg.intercept_, 3.86, rtol=1e-2)
+    assert_allclose(ridgereg.coef_, [0.91, 1.64], rtol=1e-2)
+
+
+@pytest.mark.parametrize("alpha", [0.1, 0.5, 1.0])
+def test_ridge_coefficients(alpha):
```
samir-nasibli, line 49: Why this is not extended for different dataframes?

sklearnex/preview/linear_model/ridge.py
ethanglaser, line 75: scikit-learn should have functions that cover this (ie _validate_params, check_xy etc.) - @icfaust can you confirm this?

icfaust: This is a carry-over from [LinearRegression](https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/linear_model/linear.py#L141) (and has found its way into [LogisticRegression](https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/linear_model/logistic_regression.py#L180) too). Ideally we would remove this bad design pattern from all three.  The finite assert occurs in sklearn [LinearRegression](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/linear_model/_base.py#L600), meaning it is duplicated here unnecessarily and is causing an unnecessary performance loss in LinearRegression and LogisticRegression as well (O(n)).  This whole function can be converted into the use of a single np.iscomplexobj call, since that will work with values, lists, arrays and array like objects (no conversion to numpy necessary).  I would remove it here, and we make another PR for LogisticRegression and LinearRegression separately.  Thanks @ethanglaser for making me dive in, I should have done this on the LogisticRegression out of preview PR.

icfaust: Now implemented in #1923

samir-nasibli: I see that discussion the productive discussion about this code continued at https://github.com/intel/scikit-learn-intelex/pull/1843#discussion_r1670156726. Closing this one as resolved while.

onedal/linear_model/tests/test_ridge.py
```diff
@@ -0,0 +1,95 @@
...
+
+if daal_check_version((2024, "P", 600)):
+    import numpy as np
+    import pytest
+    from numpy.testing import assert_allclose, assert_array_equal
+    from sklearn.datasets import load_diabetes
+    from sklearn.metrics import mean_squared_error
+    from sklearn.model_selection import train_test_split
+
+    from onedal.linear_model import LinearRegression
```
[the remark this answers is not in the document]
DDJHB: good catch, this was written before i separated Ridge into separate class, thanks

sklearnex/preview/linear_model/tests/test_ridge.py
```diff
@@ -0,0 +1,97 @@
...
+    # using the normal equation formula: (X^T * X + lambda * I)^-1 * X^T * y
+    lambda_identity = alpha * numpy.eye(X.shape[1])
+    inverse_term = numpy.linalg.inv(numpy.dot(X.T, X) + lambda_identity)
+    xt_y = numpy.dot(X.T, y)
+    coefficients_manual = numpy.dot(inverse_term, xt_y)
+
+    assert_allclose(ridgereg.coef_, coefficients_manual, rtol=1e-2)
+
+
+if daal_check_version((2024, "P", 600)):
```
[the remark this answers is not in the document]
DDJHB: if onedal version is not new enough, the preview Ridge falls back to daal4py Ridge. score method is not implemented in daal4py ridge so it falls back to stock sklearn which fails the test automatically*. for predict, daal4py implementation of patching status checks writes to log output that it now falls back to stock sklearn predict to handle the NotFittedError which also automatically* fails.

\* - if any fallback to stock sklearn is properly recorded in logs, the test fails

onedal/linear_model/tests/test_ridge.py
md-shafiul-alam: If the MSE is 2395.567, could we test if it is 2395.567 with some tolerance instead of testing <2396? If the responses stays the same MSE also should stay the same apart from some numerical instability. Applies for lin reg as well.

sklearnex/preview/linear_model/ridge.py
samir-nasibli: outdated

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,412 @@
...
+                "multi_output": True,
+                "force_all_finite": False,
+            }
+            if sklearn_check_version("1.2"):
+                X, y = self._validate_data(**check_params)
+                self._validate_params()
+            else:
+                X, y = check_X_y(**check_params)
+
+            if sklearn_check_version("1.0") and not sklearn_check_version("1.2"):
```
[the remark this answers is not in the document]
samir-nasibli: Please ignore this. Out of the scope of the current PR.

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,375 @@
...
+            if hasattr(self, "_onedal_estimator"):
+                self._onedal_estimator.intercept_ = value
+                # checking if the model is already fitted and if so, deleting the model
+                if hasattr(self._onedal_estimator, "_onedal_model"):
+                    del self._onedal_estimator._onedal_model
+            self._intercept = value
+
+        def _save_attributes(self):
+            self.n_features_in_ = self._onedal_estimator.n_features_in_
+            self._sparse = False
```
samir-nasibli, line 357: Where this is used?

sklearnex/preview/linear_model/tests/test_ridge.py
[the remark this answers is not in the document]
samir-nasibli: Do I understand correctly that stock sklearn checks for NotFittedError?
It’s strange for me to know that stock sklearn fails raising NotFittedError. What could be the reason? Which versions fails it?

For all mentioned methods in stock sklearn check_is_fitted https://github.com/scikit-learn/scikit-learn/blob/a95203b249c1cf392f86d001ad999e29b2392739/sklearn/linear_model/base.py#L136C9-L136C39 is called with a check for "coeff_" attr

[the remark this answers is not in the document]
DDJHB: might be a misunderstanding here, stock sklearn does not fail to raise NotFittedError but rather fallback to stock sklearn automatically fails our tests, and such fallback occurs in both predict and score methods under certain conditions described above

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,375 @@
...
+                    "onedal": self.__class__._onedal_score,
+                    "sklearn": sklearn_Ridge.score,
+                },
+                X,
+                y,
+                sample_weight=sample_weight,
+            )
+
+        def _onedal_fit_supported(self, patching_status, method_name, *data):
+            assert method_name == "fit"
```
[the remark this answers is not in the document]
DDJHB: agree but it is following sklearnex/lin_reg style, so kept it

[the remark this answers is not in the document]
samir-nasibli: What is this style and why is it needed here?

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,372 @@
...
+            self._onedal_estimator = onedal_Ridge(**onedal_params)
+
+        def _daal_fit(self, X, y, sample_weight=None):
+            daal4py_fit_ridge(self, X, y, sample_weight)
+            self._onedal_estimator.n_features_in_ = _num_features(X, fallback_1d=True)
+            self._onedal_estimator.coef_ = self.coef_
+            self._onedal_estimator.intercept_ = self.intercept_
+
+        def _onedal_fit(self, X, y, sample_weight, queue=None):
+            assert sample_weight is None
```
samir-nasibli, line 273:
```suggestion
            # `Sample weight` is not supported. Expected to be None value.
            assert sample_weight is None
```

sklearnex/preview/linear_model/ridge.py
samir-nasibli: Probably misunderstanding. The question where `_sparse` attr is used?

samir-nasibli (commented):
Generally looks good to me. Please address minor comments and I think we are ready to integration. Thank you!

sklearnex/preview/linear_model/ridge.py
icfaust: I like this suggestion.

icfaust:
I know its annoying, but I think the issues in CI have been sorted out. You may have to rebase this again for CI runs.

sklearnex/preview/linear_model/ridge.py
DDJHB: good point, nowhere in particular, inherited it from sklearnex linear regression implementation with an assumption of possible usage elsewhere. removing it if no objections

sklearnex/preview/linear_model/ridge.py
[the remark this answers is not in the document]
DDJHB: my ideal understanding is: if i remove it here, it should be removed everywhere where it can be considered an "overkill", which is preview kmeans and sklearnex linear regression. changing it only here would increase variance and i believe methods with similar purpose across project should implicitly tend towards similar implementation/style; so i would either change it everywhere or not touch at all. let me know your thoughts on this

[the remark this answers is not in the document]
md-shafiul-alam: Probably Linreg and KMeans are the only ones that have this style. My suggestion would be to refactor this on a separate PR.

onedal/linear_model/linear_model.py
```diff
@@ -224,3 +234,92 @@ def fit(self, X, y, queue=None):
...
+        improves the conditioning of the problem and reduces the variance of
+        the estimates. Larger values specify stronger regularization.
+
+    fit_intercept : bool, default=True
+        Whether to calculate the intercept for this model. If set
+        to False, no intercept will be used in calculations
+        (i.e. data is expected to be centered).
+
+    copy_X : bool, default=True
+        If True, X will be copied; else, it may be overwritten.
```
icfaust, line 256: copy_X is used weirdly in LinearRegression.fit, but goes unused in your Ridge. You may have to change the documentation for this

sklearnex/dispatcher.py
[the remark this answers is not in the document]
icfaust: > To my best understanding so far - it is used as fallback object when sklearnex functionality fails. daal4py dispatcher fills it with stock sklearn object when patching happens and replaces sklearnex one when unpatch is needed. Since we are handling previews "manually" in sklearnex dispatcher, need to put the stock sklearn object there so during possible unpatch it knows what to fallback to. @icfaust please correct me if I am wrong.

You are right, to be honest, when we patch we probably should be more careful, there is definitely some concurrency bugs which maybe occurring there that we aren't testing.  At least this reasonably handles preview when another patch may be in place.

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,372 @@
...
+        """
+        Determines if the provided value is an instance of either int or float.
+
+        Args:
+        value: The value to be checked.
+
+        Returns:
+        bool: True if the value is either an int or a float, False otherwise.
+        """
+        return isinstance(value, (int, float))
```
icfaust, line 49: https://docs.python.org/3/library/numbers.html#numbers.Real ?

sklearnex/preview/linear_model/ridge.py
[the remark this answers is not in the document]
icfaust: +1 for a design document, too many styles tbh

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,372 @@
...
+            patching_status = PatchingConditionsChain(
+                f"sklearn.linear_model.{self.__class__.__name__}.fit"
+            )
+
+            if method_name == "fit":
+                alpha_is_scalar = _is_numeric_scalar(self.alpha)
+                patching_status.and_condition(
+                    alpha_is_scalar,
+                    "Non-scalar alpha is not supported for GPU.",
+                )
```
icfaust, line 232:
```suggestion
                patching_status.and_condition(
                    isinstance(self.alpha, numbers.Real),
                    "Non-scalar alpha is not supported for GPU.",
                )
```

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,372 @@
...
+                    estimator_name=self.__class__.__name__,
+                )
+
+            self._initialize_onedal_estimator()
+
+            # Falling back to daal4py if the device is CPU since
+            # onedal does not support non-scalars for alpha, thus
+            # should only be used for GPU/CPU with scalar alpha to not limit the functionality
+            cpu_device = queue is None or queue.sycl_device.is_cpu
+            if cpu_device and not _is_numeric_scalar(self.alpha):
```
icfaust, line 303:
```suggestion
            if cpu_device and not isinstance(self.alpha, numbers.Real):
```

icfaust (commented):
Small changes before approval

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,372 @@
...
+                self._daal_fit(X, y)
+            else:
+                self._onedal_estimator.fit(X, y, queue=queue)
+
+            self._save_attributes()
+
+        def _onedal_predict(self, X, queue=None):
+            if sklearn_check_version("1.0"):
+                X = self._validate_data(X, accept_sparse=False, reset=False)
+
```
ethanglaser, line 313: is it meant to match https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/linear_model/linear.py#L257-L260 ? If so may require slight revision

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,372 @@
...
+                "X": X,
+                "y": y,
+                "dtype": [np.float64, np.float32],
+                "accept_sparse": ["csr", "csc", "coo"],
+                "y_numeric": True,
+                "multi_output": True,
+            }
+            if sklearn_check_version("1.2"):
+                X, y = self._validate_data(**check_params)
+                self._validate_params()
```
ethanglaser, line 286: move to before dispatch (in fit()) to match other estimators

sklearnex/preview/linear_model/ridge.py
```diff
@@ -0,0 +1,372 @@
...
+    if sklearn_check_version("1.0") and not sklearn_check_version("1.2"):
+        from sklearn.linear_model._base import _deprecate_normalize
+
+    from onedal.linear_model import Ridge as onedal_Ridge
+    from onedal.utils import _num_features, _num_samples
+
+    from ..._device_offload import dispatch, wrap_output_data
+    from ..._utils import PatchingConditionsChain
+
+    def _is_numeric_scalar(value):
```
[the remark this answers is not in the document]
ethanglaser: nvm I see that you have added it - you should be able to remove definition and use of _is_numeric_scalar() now

sklearnex/preview/linear_model/ridge.py
ethanglaser, line 303: I don't really understand this - can you explain a scenario this would occur? When would alpha not be a scalar

ethanglaser (commented):
Overall a few relatively minor comments - this is very close to being ready

sklearnex/preview/linear_model/ridge.py
icfaust: Its possible to pass an array-like input into alpha, which can be handled by sklearn and daal, but not yet in oneDAL. https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.Ridge.html

onedal/linear_model/linear_model.py
DDJHB: thanks for catch, now using copy_X in Ridge

sklearnex/preview/linear_model/ridge.py
DDJHB: changed implementation of _is_numeric_scalar to suggested isinstance(self.alpha, numbers.Real). please resolve if okay

sklearnex/preview/linear_model/ridge.py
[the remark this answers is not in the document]
DDJHB: used to validate that alpha is numeric since it can also be array-like for multitarget regression but our oneDAL implementation lacks this functionality yet. please resolve if this makes sense

sklearnex/preview/linear_model/ridge.py
DDJHB: https://github.com/intel/scikit-learn-intelex/pull/1843#discussion_r1645036767
https://github.com/intel/scikit-learn-intelex/pull/1843#discussion_r1645012243
have a look please, nice explanation from @icfaust

sklearnex/preview/linear_model/ridge.py
DDJHB: @ethanglaser please resolve if this addresses your question

sklearnex/preview/linear_model/ridge.py
DDJHB: https://github.com/intel/scikit-learn-intelex/pull/1843#discussion_r1645012243
valid point from @icfaust (if i understood it correctly), let me know what you think

ethanglaser: Thanks

sklearnex/preview/linear_model/ridge.py
ethanglaser: Thanks for clarification

ethanglaser:
No objections from my end at this point - just would like to see green internal CI before approval

icfaust (approved):

DDJHB:
latest private ci runs:
https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/ef451330-b666-f193-9381-a4bf010d0e2e
https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/ef455961-8cad-f1bd-978b-a4bf010d0e2e
https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/ef4571ed-4297-f1f8-9e46-a4bf010d0e2e

@ethanglaser please review

ethanglaser (approved):

PR #1844, opened 2024-05-24
[CI, Enhancement] copy a nightly oneDAL build for other sklearnex GitHub actions

icfaust:
### Description 
This PR introduces a nightly Github action CI job which copies the nightly builds from the oneapi-src/oneDAL nightly build workflow. This is necessary as pull_request Github Action workflows cannot access secrets (i.e. the tokens) due to it leading to security vulnerabilities.  This PR creates a scheduled/workflow_dispatch watchdog workflow which can use secrets, which downloads the completed artifacts.  This can be used by pull_request workflows without having to use secrets.  This workflow is timed to start 5 minutes after the daily scheduled oneDAL nightly build run, and will wait for completion of that run before downloading and uploading the artifacts.


ethanglaser:
Any idea how the cache works? Is it cleaned up automatically/is there any limit on storage? Just curious how storing the builds works logistically

icfaust:
This PR needs to wait until oneDAL windows builds using visual studio are fixed.

icfaust:
> Any idea how the cache works? Is it cleaned up automatically/is there any limit on storage? Just curious how storing the builds works logistically

Artifacts have a limited default lifetime of 90 days, but we can shorten that if we want. We should be able to modify things under settings and/or actions on the repo (given the right permissions) for them. I couldn't find the limit in size, but in good news, assuming someone has access, they can download the built binary for local use if they so desire.

ethanglaser (approved):

icfaust:
Temporarily modified to run on PRs to main in order to show it in operation for Windows and Linux. Will be removed after a green CI run (will be used as evidence).

.github/workflows/oneDAL.yml
```diff
@@ -0,0 +1,116 @@
+name: oneDAL-nightly
```
napetrov, line 1: should we have this running in oneDAL repo and just deploy packages to https://github.com/oneapi-src/oneDAL/packages ?

icfaust: Good question, I am not sure which registry to use...  I mainly did it this way to simplify the next step (the testing CI), but I am open to changing it.

icfaust:
This PR has been transitioned to the oneDAL repo in this PR: oneapi-src/oneDAL#2863

icfaust:
Will now re-open this to modify as necessary to pull oneDAL build from oneapi-src/oneDAL

icfaust:
Due to the use of secrets, validation of this PR cannot be done via the pull_request workflow.  I will make meetings to cover this in depth.

PR #1855, opened 2024-06-05
[bug] fix issues with dpnp/dpctl regressor score method

icfaust:
### Description 
This will solve sporadic private CI failures related to r2_score and dpnp/dpctl power, as well as correct issues observed in sklearn 1.5

Add score method which properly works with numpy arrays for:
Ridge
Lasso
ElasticNet
SVR
NuSVR
LinearRegression
IncrementalLinearRegression
KNeighborsRegressor
ExtraTreesRegressor
RandomForestRegressor

There are definitely simpler ways to implement this, but this is the most straightforward, and will need to be rewritten for zero-copy array_api support

### DIFF
--- a/conda-recipe/run_test.sh
+++ b/conda-recipe/run_test.sh
@@ -58,7 +58,7 @@ pytest --verbose --pyargs sklearnex
 return_code=$(($return_code + $?))
 
 echo "Pytest of onedal running ..."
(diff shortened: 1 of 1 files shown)


ethanglaser (approved):
LGTM, great work on getting this together

ethanglaser:
@Alexsandruss check this out and then can re-open sklearn 1.5 PR - it should resolve the previous errors

PR #1856, opened 2024-06-06
[testing] enable get_queue for `/onedal` testing without dpctl support

icfaust:
### Description 
A valid value for queue is None (this is the default value for onedal class methods like 'fit' and 'predict').  Without a value, tests which parameterize via get_queue will be skipped as the list is empty.  This initializes the output with None if cpu is in the filter.  This is currently a silent fail in the sklearnex testing step on public CI (see last main run) where most tests are now skipped.  In some windows cases, only 1 test of 1000 will run without this change.


onedal/common/tests/test_policy.py
```diff
@@ -47,6 +47,9 @@ def test_with_numpy_data(queue):
 @pytest.mark.parametrize("queue", get_queues("cpu,gpu"))
 @pytest.mark.parametrize("memtype", get_memory_usm())
 def test_with_usm_ndarray_data(queue, memtype):
+    if queue is None:
+        pytest.skip("queue=None not supported for usm_ndarray")
+
```
icfaust, line 52: There may be a bug here we need to investigate.

samir-nasibli: I think disabling this not good solution. What kind of bug here?
It should work with queue=None and/or None data_queue for `_get_policy`

icfaust: This yields a mismatch and fails, if you would like to change this PR to a bugfix, let me know.

icfaust: The issue is on DPCTL side, a None queue will look for a cached queue, which will often be the GPU causing the failure.  The get_policy works fine, test is malformed for None. I.e. people should be very careful with direct usm_ndarray buffer allocations with memory

samir-nasibli: I don't think that this is bug.
`dpctl` just differently processes None input. I am good with the changes here. Thank you!

samir-nasibli (approved):
Thank you @icfaust 
These changes will enable correct onedal testing even without dpctl.
Assuming Green CI

ethanglaser:
Nice work here, looks like 1279 skipped in main branch on public 3.11 sklearnex onedal step compared to just 18 with these changes

ethanglaser:
Is there any functional difference between using no queue and using dpctl CPU queue?

icfaust:
> Is there any functional difference between using no queue and using dpctl CPU queue?

Nope! https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/_device_offload.py#L155 sets it and sends it back.

PR #1864, opened 2024-06-12
MAINT: d4p sycl interface removal

ethanglaser:
### Description 
Accompanies https://github.com/oneapi-src/oneDAL/pull/2816
Dependent on: https://github.com/intel/scikit-learn-intelex/pull/1770, https://github.com/intel/scikit-learn-intelex/pull/1832

Removals include:
 - Removal of daal4py/sycl examples
 - Removal of src/oneapi and daal4py/oneapi due to dependence on removed DAAL interfaces (especially execution contexts)
 - Removal of any other references to execution context and GPU in daal4py (generator/, daal4py/sklearn/_utils.py)
 - Removal of build_oneapi_backend and build of daal4py/oneapi in setup scripts
 - Corresponding refactoring of run_examples test script

- [ ] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/intel/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_


samir-nasibli:
Please also take care all places where daal4py.oneapi import we have.
For example: https://github.com/intel/scikit-learn-intelex/blob/772f678e53c501059cf3c83c403ff05c4e2983f7/onedal/common/_policy.py#L23
Partially some work in case of daal4py.oneapi sycl_context suage remove done on #1832

ethanglaser:
> Please also take care all places where daal4py.oneapi import we have. For example:

Thanks for the inputs, this will be very helpful moving forward. Currently mostly focusing on OneDAL but will take care of this soon.

scripts/build_backend.py
```diff
@@ -40,87 +40,6 @@
     IS_WIN = True
 
 
-def build_cpp(
```
icfaust, line 43: Man, I am glad this is deleted. The naming and lack of documentation makes it really hard for me to see how this is sycl related.  Who knows this code the best here?  As it is I would need to really sit down and take time here to give this deletion a faithful review.

icfaust (commented):
Generally an unoffensive PR for a 3.6k line change.  There is one section that it would be best to have someone parse/review, but I could do it if requested and given time.  Let me know how to proceed, until then I will leave this as a comment rather than an approval.  Side question: Does this improve build times noticeably on linux 3.9/3.11 (those with dpc enabled)?

scripts/build_backend.py
ethanglaser: The reason I removed is because it was only referenced from build_oneapi_backend function in setup.py which is also removed

ethanglaser:
> Also please flesh out the description for future readers. Maybe even include an impacted algorithm list.

Added list summarizing all changes/removals in PR description

tests/run_examples.py
```diff
@@ -313,7 +287,7 @@ def run(exdir, logdir, args):
 
 
 def run_all(args):
-    if args.assert_gpu and "gpu" not in available_devices:
+    if args.assert_gpu and dpctl_available and "gpu" not in available_devices:
```
samir-nasibli, line 290: Generally, this is wrong, because we have an option to offload GPU without dpctl.
See `config_context(target_offload=True)`

samir-nasibli: Check of the GPU availability depends on GPU, there should be another mechanism for that I think.

samir-nasibli: At least here check for `dpctl_available` is unnecessary.

ethanglaser: This was originally added because there is concern when dpctl is present but gpu not in available devices, since dpctl is only installed for GPU testing

icfaust (approved):
Approval with caveat of simultaneous merge of other related PRs.

tests/run_examples.py
samir-nasibli: I think, here check for `dpctl_available` is unnecessary. In the top of the file it was done for `available_devices`.
We can leave it as is here.
```suggestion
    if args.assert_gpu and "gpu" not in available_devices:
```

ethanglaser: Its because of assert_gpu logic. But it makes sense to do this and revise infra logic, done.

samir-nasibli (approved):
Thank you for this done!
I have launched private CI, shouldn't be any issue there. Just to be sure.
Assuming green CI

PR #1861, opened 2024-06-12
ENH: functional support for Array API

samir-nasibli:
### Description 
Array API functional support without zero-copy in scikit-learn-intelex.
Depends on #1999 and #1998 

Brings extension for the device_offload logic and extends support_usm_ndarray decorator and make this work also for the array-api-strict inputs. New support_array_api decorator mainly impl Array API functional support in scikit-learn-intelex.

Proposed changes shown on the dataflow diagram below. Yellow plates is extensions in the sklearnex dataflow diagram and show the first iteration of the scikit-learn-intelex Array API support (functional support).

image

Works fine with array-api-strict v1 and v2 as well. Some test suits requires updates for the v2, that will be solved on #2012 

Changes includes:
* updates on _device_offload modules both of onedal4py and sklearnex.
  * extensions for `sklearnex._device_offload.wrap_output_data` decorator and `sklearnex._device_offload.dispatch` function.
  * `onedal._device_offload._transfer_to_host` extended for the Array API inputs.
  * `onedal._device_offload.support_usm_ndarray` renamed to `onedal._device_offload.support_array_api` extend for Array API inputs handling. (several files changed for this this renaming)
* onedal4py's testing utility `get_dataframes_and_queues` extended for the array_api inputs
* `_array_api` modules added in the both onedal4py and sklearnex.\
  * `onedal.utils._array_api` module includes several array api utilities and `_get_sycl_namespace` that is used in the  `get_namespace` (from `sklearnex.utils._arrray_api`), that returns supported array api namespace.
  
Changes are documented in the files. General documentation for Array API support will be able at #1918 

### Test statistics
**More than 2834 new test cases (+ 46%) added for Array API inputs.** On `sklearnex` testing you may observe the rise of test pass statistics
from 
```5943 passed, 898 skipped, 42083 warnings in 394.72s (0:06:34)```
to 
```8674 passed, 1001 skipped, 44350 warnings in 538.83s (0:08:58)```

- [X] I have reviewed my changes thoroughly before submitting this pull request.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes, if necessary (updated in #1918)
- [X] The unit tests pass successfully.
- [X] I have run it locally and tested the changes extensively.
- [X] I have resolved any merge conflicts that might occur with the base branch.
[...]

PR #1813, opened 2024-04-26
ENH: initial `_array_api` module + for `dpnp`/`dpctl` extension

samir-nasibli:
# Description
Please do not merge/review yet.

# TODO
* remove utils._namespace module and move some funcs from there to _array_api
* check for dpnp/dpctl and branches where it is required
* move dpnp_converter to array_api module

PR #1865, opened 2024-06-13
[testing] Expand sklearnex testing to all methods taking "X" or "y" as input

icfaust:
### Description 
All testing under the `sklearnex/tests` folder use centralized estimator and method lists which were previously limited to methods based off of the various inherited sklearn Mixins.  This tested all 'core' functionality like 'fit', 'predict', 'transform', etc. which are directly patched out by oneDAL functions.  Some next level methods (those which use these core functionality with a small number of additional calculations) like `score` were included.  However, this list was incomplete, as some estimators (like PCA and KNeighbors) implement special additional methods like `inverse_transform` and `kneighbors_graph`, which require special inputs and/or outputs.  Up until now, these methods have not been tested using the centralized testing suite (and may have not been tested at all).  This PR dynamically checks and collects all methods to all patched estimators which contain "X" or "y" in the input into testing, increasing flexibility and reducing maintenance. 

The additional code comes from having to fix issues related to taking in dpnp and dptcl inputs since they were previously untested.

Rather than using ```_device_offload.dispatch``` which is code heavy, cumbersome and with significant overhead, a limited use of `get_namespace` was added to these failing methods which provide the necessary pre- or post-processing of 'core' oneDAL methods.

With these additions, dpnp and dpctl inputs are almost fully covered in the codebase.  One deselection was made in `radius_neighbors` and `radius_neighbors_graph` in the NearestNeighbors algorithm, as these would require significant work (i.e. the implementation of a new estimator) in order to properly support dpnp/dpctl inputs (unless we knowingly only allow CPU computation).

The listed changes are as follows:
- use sklearn native validation `_check_sample_weight` instead of a custom ensemble algorithms `check_sample_weight` method to `_check_sample_weight` to avoid new checks
- Change IncrementalEmpiricalCovariance `mahalanobis` and `score` methods to accept dpnp/dpctl inputs (`score` was previously missed as this estimator has no Mixins)
- Provide error to NearestNeighbors algorithm for use of `radius_neighbors` and `radius_neighbors_graph` for non-numpy/pandas inputs, which are unsupported.
- Remove radius_neighbors from KNeighborsRegressor and KNeighorsClassifier, as they do not support it from sklearn, and it was purely using sklearn functionality
- Fix `kneighbors_graph` method of `NearestNeighbors`, `KNeighborsRegressor`, and `KNeighborsClassifier` to take in dpnp/dpctl inputs, and return a scipy.csr_matrix (removed wrap output data, etc.)
- Disable `score_samples` and `score` methods in stability testing for PCA (will add tickets to address these issues).
- Introduce PCA.inverse_transform using get_namespace to insure components_ and mean_ are on the same device as X

NOTE: current private CI fails for PCA in the `inverse_transform` method in dpnp testing, this is due to an out-of-date version of dpnp, and is currently getting addressed. Private CI will be re-run as soon as it is updated.


sklearnex/covariance/incremental_covariance.py
```diff
@@ -293,21 +294,22 @@ def _onedal_fit(self, X, queue=None):
...
     # expose sklearnex pairwise_distances if mahalanobis distance eventually supported
-    @wrap_output_data
     def mahalanobis(self, X):
         if sklearn_check_version("1.0"):
-            self._validate_data(X, reset=False, copy=self.copy)
-        else:
-            check_array(X, copy=self.copy)
+            self._check_feature_names(X, reset=False)
 
+        xp, _ = get_namespace(X)
```
samir-nasibli, line 301: I think it make sense move out this the scopoe of the PR.
If this needed to be addressed related Array API, please mention it on #1861, #1813

samir-nasibli (changes requested):
Comments above

icfaust:
sklearnex test fails now limited to test_stability

onedal/tests/utils/_dataframes_support.py
```diff
@@ -90,6 +91,10 @@ def _as_numpy(obj, *args, **kwargs):
         return dpt.to_numpy(obj, *args, **kwargs)
     if isinstance(obj, pd.DataFrame) or isinstance(obj, pd.Series):
         return obj.to_array(*args, **kwargs)
+    if hasattr(obj, "__array_namespace__"):
```
samir-nasibli, line 94: I will take care of it on #1861 or #1813
It make sense enable it on inital functional support of Array API

sklearnex/covariance/incremental_covariance.py
```diff
@@ -197,6 +199,23 @@ def _onedal_partial_fit(self, X, queue=None, check_input=True):
...
+        est = clone(self)
+        est.set_params(**{"assume_centered": True})
+
+        xp, _ = get_namespace(X_test)
+
+        location = self.location_
+        if "numpy" not in str(xp).lower():
+            location = xp.asarray(location, device=X_test.device)
+
+        # test_cov is a numpy array, but calculated on device
```
samir-nasibli, line 213: Let's skip integration of this part for now.
For not duplication the work, it make sense to us first finalize #1813 and #1861

sklearnex/covariance/incremental_covariance.py
```diff
@@ -293,21 +312,34 @@ def _onedal_fit(self, X, queue=None):
...
-            # compute mahalanobis distances
-            dist = pairwise_distances(
-                X, self.location_[np.newaxis, :], metric="mahalanobis", VI=precision
-            )
+        # compute mahalanobis distances
+        # pairwise_distances will check n_features (via n_feature matching with
+        # self.location_) , and will check for finiteness via check array
+        # check_feature_names will match _validate_data functionally
+        location = self.location_[np.newaxis, :]
+        if "numpy" not in str(xp).lower():
```
samir-nasibli, line 326: Could be replaced by https://github.com/samir-nasibli/scikit-learn-intelex/blob/4179f22b5671b9ff83baa4d2588556e791459c58/onedal/_device_offload.py#L210
But again it make sense to do after #1861 or #1813

sklearnex/neighbors/common.py
```diff
@@ -264,3 +266,41 @@ def _onedal_gpu_supported(self, method_name, *data):
...
+        if mode == "connectivity":
+            A_ind = self.kneighbors(X, n_neighbors, return_distance=False)
+            xp, _ = get_namespace(A_ind)
+            n_queries = A_ind.shape[0]
+            A_data = xp.ones(n_queries * n_neighbors)
+
+        elif mode == "distance":
+            A_data, A_ind = self.kneighbors(X, n_neighbors, return_distance=True)
+            xp, _ = get_namespace(A_ind)
+            A_data = xp.reshape(A_data, (-1,))
```
samir-nasibli, line 287: The same comment as above.

sklearnex/neighbors/knn_unsupervised.py
```diff
@@ -88,10 +89,14 @@ def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
...
         self, X=None, radius=None, return_distance=True, sort_results=False
     ):
-        _onedal_estimator = getattr(self, "_onedal_estimator", None)
-
+        if X is not None and "numpy" not in str(get_namespace(X)[0]).lower():
+            raise TypeError(
+                f"{self.__class__.__name__} does not support {type(X)} inputs for radius_neighbors"
+            )
+        # This definitely doesn't cover all cases, and doesn't raise an error to the user
+        # It must be updated.
```
samir-nasibli, line 97: Make sense update after #1813 and #1861

samir-nasibli (commented):
@icfaust  Thank you very much for your work done here.
I know this pr is draft but let me share some my points of view.

Some of the development is already being done on #1813, #1861 and  I encourage you for help with reviews and suggestions there at the moment.

It will be good if we limit the scope of this PR by enabling DPNP/dPCTL coverage. 
For example we have problems https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/cluster/tests/test_dbscan.py#L22:L27 which I think at the moment is more priority than just adding branches for future array api enablement .

sklearnex/covariance/incremental_covariance.py
icfaust: We discussed this elsewhere. will mark as resolved.

sklearnex/neighbors/common.py
```diff
@@ -264,3 +266,41 @@ def _onedal_gpu_supported(self, method_name, *data):
...
+        check_is_fitted(self)
+        if n_neighbors is None:
+            n_neighbors = self.n_neighbors
+
+        # check the input only in self.kneighbors
+
+        # construct CSR matrix representation of the k-NN graph
+        if mode == "connectivity":
+            A_ind = self.kneighbors(X, n_neighbors, return_distance=False)
+            xp, _ = get_namespace(A_ind)
```
samir-nasibli, line 280: In case of Array API style coding, I think it make sense early call at once get namespace for the input.
Wouldn't be it better to move on line 272 get namespace for input X.

samir-nasibli (commented):
initially reviewed

sklearnex/ensemble/_forest.py
```diff
@@ -78,7 +78,7 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
         )
 
         if sample_weight is not None:
-            sample_weight = self.check_sample_weight(sample_weight, X)
+            sample_weight = self._check_sample_weight(sample_weight, X)
```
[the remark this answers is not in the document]
icfaust: this was necessary to avoid the callable method discovery introduced in the new test coverage, this is shown in the changes in _utils.py. it is not a method of the sklearn estimator, but does take x as an input. it is used privately and should be made private.

sklearnex/covariance/incremental_covariance.py
```diff
@@ -197,6 +199,39 @@ def _onedal_partial_fit(self, X, queue=None, check_input=True):
...
+            )
+
+        if "numpy" not in xp.__name__:
+            location = xp.asarray(location, device=X_test.device)
+            if isinstance(X, np.ndarray):
+                X = X_test
+
+        est = clone(self)
+        est.set_params(**{"assume_centered": True})
+
```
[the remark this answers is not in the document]
icfaust: The purpose of this PR is to verify all estimators with "X" or "y" will properly evaluate using the various dataframes and queues in all of our various tests (datatypes and stability currently, possibly memory leaks later).  Score is implemented in the way it is in order to guarantee dpnp/dpctl tensor conformance.

sklearnex/covariance/incremental_covariance.py
```diff
@@ -293,21 +328,34 @@ def _onedal_fit(self, X, queue=None):
         return self
 
     # expose sklearnex pairwise_distances if mahalanobis distance eventually supported
-    @wrap_output_data
     def mahalanobis(self, X):
         if sklearn_check_version("1.0"):
-            self._validate_data(X, reset=False, copy=self.copy)
-        else:
-            check_array(X, copy=self.copy)
+            self._check_feature_names(X, reset=False)
```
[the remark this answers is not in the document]
icfaust: This is the case throughout the LogisticRegression implementation:
It has been included before the dispatch, meaning it operates on the sycl usm ndarrays when provided.

https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/linear_model/logistic_regression.py#L111
Testing of fit in LogisticRegression:
https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/linear_model/tests/test_logreg.py#L50

_check_feature_names is in sklearn's BaseEstimator validate_data: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/base.py#L654 It purposesfully works with the data which hasn't been touched by check_array, meaning it is working on the raw input.

The purpose of this method is to validate aspects of pandas, and pandas-like dataframes, which could have names to features: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/base.py#L479 which the most important part is https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/validation.py#L2226

sklearnex/neighbors/common.py
icfaust: What is difficult here is that X is a kwarg with a default value of None.  In that case, it would depend on array type of the stored values in onedal_estimator: https://github.com/intel/scikit-learn-intelex/blob/main/onedal/neighbors/neighbors.py#L298  It is currently numpy arrays, but that's not guaranteed.  The most dependable way is to check the output of kneighbors,

sklearnex/covariance/incremental_covariance.py
```diff
@@ -197,6 +199,39 @@ def _onedal_partial_fit(self, X, queue=None, check_input=True):
...
+            X = check_array(
+                X_test,
+                dtype=[np.float64, np.float32],
+                copy=self.copy,
+            )
+
+        if "numpy" not in xp.__name__:
+            location = xp.asarray(location, device=X_test.device)
+            if isinstance(X, np.ndarray):
+                X = X_test
```
md-shafiul-alam, line 224: X=X_test is basically bypassing the check_array or _validate_data, but still the checks will be performed. Should the checks be conditional?

sklearnex/covariance/incremental_covariance.py
```diff
@@ -293,21 +328,34 @@ def _onedal_fit(self, X, queue=None):
...
+        location = self.location_[np.newaxis, :]
+        if "numpy" not in xp.__name__:
+            location = xp.asarray(location, device=X.device)
+
+        try:
+            dist = pairwise_distances(X, location, metric="mahalanobis", VI=precision)
+        except ValueError as e:
+            # Throw the expected sklearn error in an n_feature length violation
+            if "Incompatible dimension for X and Y matrices: X.shape[1] ==" in str(e):
+                n_features = re.findall(r"\s\d+", str(e))[0].lstrip()
```
md-shafiul-alam, line 350: Is this really necessary? :)

icfaust: Unfortunately yes, its catching a specific error and value from a raised error in pairwise_distances and throwing an error like _validate_data. This is necessary for sklearn test_conformance.

sklearnex/covariance/incremental_covariance.py
icfaust: I want the checks to always occur, but I want to prevent casting the dpnp/dpctl tensor to a numpy array using asarray() in them. Do you think I should add a comment in the code explaining this weirdness?

md-shafiul-alam: I see, probably a comment will be better.

sklearnex/covariance/incremental_covariance.py
icfaust: Swapped to _num_features from sklearn instead. Good call.

sklearnex/covariance/incremental_covariance.py
icfaust: Comment added

sklearnex/covariance/incremental_covariance.py
```diff
@@ -197,6 +199,39 @@ def _onedal_partial_fit(self, X, queue=None, check_input=True):
 
         return self
 
+    @wrap_output_data
+    def score(self, X_test, y=None):
```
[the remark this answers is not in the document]
icfaust: I have now added a test which validates numerical conformance.  However, the purpose of this PR was to extend checking with dpnp/dpctl to everything possible.

sklearnex/decomposition/pca.py
```diff
@@ -210,6 +210,24 @@ def fit_transform(self, X, y=None):
                 # Scikit-learn PCA["covariance_eigh"] was fit
                 return self._transform(X_fit, xp, x_is_centered=x_is_centered)
 
+        @wrap_output_data
+        def inverse_transform(X):
+            xp, _ = get_namespace(X)
```
icfaust, line 215: weakness in input checking matches issues in sklearn, currently a PR to fix it is in: scikit-learn/scikit-learn#29310

icfaust:
http://intel-ci.intel.com/ef345679-7188-f16b-a099-a4bf010d0e2e

icfaust:
http://intel-ci.intel.com/ef346297-5405-f1e1-ad44-a4bf010d0e2e

sklearnex/decomposition/pca.py
```diff
@@ -210,6 +210,24 @@ def fit_transform(self, X, y=None):
...
+                    xp.sqrt(self.explained_variance_[:, np.newaxis]) * self.components_
+                )
+            else:
+                components = self.components_
+
+            if "numpy" not in xp.__name__:
+                components = xp.asarray(components, device=X.device)
+                mean = xp.asarray(mean, device=X.device)
+
+            return X @ components + mean
```
Vika-F, line 229: Please add a comment why the check and asarray conversion is required here.

sklearnex/neighbors/common.py
```diff
@@ -264,3 +266,41 @@ def _onedal_gpu_supported(self, method_name, *data):
...
+                'Unsupported mode, must be one of "connectivity", '
+                f'or "distance" but got "{mode}" instead'
+            )
+
+        n_queries = A_ind.shape[0]
+        n_samples_fit = self.n_samples_fit_
+        n_nonzero = n_queries * n_neighbors
+        A_indptr = xp.arange(0, n_nonzero + 1, n_neighbors)
+
+        kneighbors_graph = sp.csr_matrix(
```
Vika-F, line 300: I think it would be better to use `sp.csr_array` if possible.
Because scipy is in the process of switching to array interface. See the note here: https://docs.scipy.org/doc/scipy/reference/sparse.html

sklearnex/tests/test_run_to_run_stability.py
```diff
@@ -73,23 +73,41 @@ def eval_method(X, y, est, method):
...
+        if method == "inverse_transform":
+            # PCA's inverse_transform takes (n_samples, n_components)
+            data = (
+                (X[:, : est.n_components_],) if X.shape[1] != est.n_components_ else (X,)
+            )
+        elif method not in ["score", "partial_fit", "path"]:
+            data = (X,)
         else:
-            res = est.score(X, y)
+            data = (X, y)
```
Vika-F, line 85: This code is duplicated here and in `sklearnex/tests/test_patching.py`.
Is it possible to have it in some common place?

Vika-F (commented):
Thank you for improving dpnp/dpctl support and increasing the product's code coverage.
See a couple of my comments below.

sklearnex/neighbors/common.py
icfaust: Totally agree, unfortunately we are stuck with csr_matrix for conformance reasons to sklearn: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/neighbors/_base.py#L1010

sklearnex/tests/test_run_to_run_stability.py
icfaust: Good idea, will change that.

onedal/tests/utils/_dataframes_support.py
```diff
@@ -119,6 +120,8 @@ def _as_numpy(obj, *args, **kwargs):
         return dpt.to_numpy(obj, *args, **kwargs)
     if isinstance(obj, pd.DataFrame) or isinstance(obj, pd.Series):
         return obj.to_array(*args, **kwargs)
+    if sp.issparse(obj):
+        return obj.toarray(*args, **kwargs)
```
samir-nasibli, line 124: Is it really needed in the scope of the PR? Please remove not used or not needed code from the PR.

sklearnex/covariance/incremental_covariance.py
```diff
@@ -196,6 +198,45 @@ def _onedal_partial_fit(self, X, queue=None, check_input=True):
...
```
samir-nasibli, line 202:
```suggestion
    def score(self, X, y=None):
```

sklearnex/covariance/incremental_covariance.py
```diff
@@ -196,6 +198,45 @@ def _onedal_partial_fit(self, X, queue=None, check_input=True):
...
+                dtype=[np.float64, np.float32],
+                reset=False,
+                copy=self.copy,
+            )
+        else:
+            X = check_array(
+                X_test,
+                dtype=[np.float64, np.float32],
+                copy=self.copy,
+            )
```
samir-nasibli, line 218: We are not supporting skl <1.0

sklearnex/covariance/incremental_covariance.py
samir-nasibli, line 202: I don't really understand mixing `wrap_output_data` and using `get_namespace`? could you please clarify?

samir-nasibli (commented):
Partially reviewed. Will take a look more later

sklearnex/covariance/incremental_covariance.py
icfaust: Lets try and keep that discussion to #1906

sklearnex/covariance/incremental_covariance.py
icfaust: I agree, but it was done for conformance reasons: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/covariance/_empirical_covariance.py#L255

onedal/tests/utils/_dataframes_support.py
icfaust: Please see the results of kneighbors_graph, there are circumstances where a scipy csr_matrix is returned. This is a necessary addition.

sklearnex/covariance/incremental_covariance.py
samir-nasibli: Oh, yeah. Good catch. I thought, exactly for the conformance it should be `X`. Thank you for providing the link

sklearnex/neighbors/common.py
samir-nasibli, line 300: I am curious does it really work with dpnp/dpctl array API inputs?

sklearnex/tests/_utils.py
```diff
@@ -123,14 +123,24 @@ def gen_models_info(algorithms):
...
-
-        for mixin, method, _ in mixin_map:
-            if issubclass(est, mixin):
-                methods |= candidates & set(method)
+        # remove BaseEstimator methods (get_params, set_params)
+        candidates = set(dir(est)) - set(dir(BaseEstimator))
+        # remove private methods
+        candidates = set([i for i in candidates if not i.startswith("_")])
+        # required to enable other methods
+        candidates = candidates - {"fit"}
```
ethanglaser, line 131: Why no fit?

sklearnex/tests/_utils.py
```diff
@@ -112,7 +112,7 @@ def __getitem__(self, key):
 )
 
 
-def gen_models_info(algorithms):
+def gen_models_info(algorithms, required_inputs=["X", "y"]):
```
ethanglaser, line 115: this is a not straightforward function and given that we are refactoring it would be a good opportunity to improve variable names (specifically i) - would recommend making i,j,k slightly more informative

sklearnex/tests/test_run_to_run_stability.py
```diff
@@ -73,23 +74,31 @@ def eval_method(X, y, est, method):
...
 
     if not isinstance(res, Iterable):
-        res = [res]
+        results = [_as_numpy(res)] if res is not est else []
+    else:
+        results = []
+        for i in res:
+            # Nearest Neighbors radius_neighbors causes this if statement
+            # It is unique in returning a numpy array of numpy arrays in
+            # a list
```
ethanglaser, line 86: if radius neighbors is being skipped would we still need this?

sklearnex/neighbors/knn_unsupervised.py
```diff
@@ -22,11 +22,12 @@
 from onedal.neighbors import NearestNeighbors as onedal_NearestNeighbors
 
 from .._device_offload import dispatch, wrap_output_data
+from ..utils import get_namespace
 from .common import KNeighborsDispatchingBase
 
 
 @control_n_jobs(decorated_methods=["fit", "kneighbors"])
-class NearestNeighbors(sklearn_NearestNeighbors, KNeighborsDispatchingBase):
+class NearestNeighbors(KNeighborsDispatchingBase, sklearn_NearestNeighbors):
```
ethanglaser, line 30: does order make a difference?

sklearnex/neighbors/knn_unsupervised.py
```diff
@@ -88,10 +89,14 @@ def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
...
         self, X=None, radius=None, return_distance=True, sort_results=False
     ):
-        _onedal_estimator = getattr(self, "_onedal_estimator", None)
-
+        if X is not None and "numpy" not in get_namespace(X)[0].__name__:
+            raise TypeError(
+                f"{self.__class__.__name__} does not support {type(X)} inputs for radius_neighbors"
+            )
+        # This definitely doesn't cover all cases, and doesn't raise an error to the user
+        # It must be updated.
```
ethanglaser, line 97: to be removed

sklearnex/neighbors/knn_unsupervised.py
```diff
@@ -22,11 +22,12 @@
 from onedal.neighbors import NearestNeighbors as onedal_NearestNeighbors
 
 from .._device_offload import dispatch, wrap_output_data
+from ..utils import get_namespace
```
ethanglaser, line 25: can remove this assuming below also removed

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -163,6 +168,36 @@ def test_sklearnex_fit_on_random_data(
     assert_allclose(expected_means, result.location_, atol=1e-6)
 
 
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+def test_whitened_toy_score(dataframe, queue):
+    from sklearnex.covariance import IncrementalEmpiricalCovariance
```
ethanglaser, line 173: just to clarify, this is for validation of addition of score()?

sklearnex/neighbors/knn_unsupervised.py
icfaust: Yeah, the inheritance of kneighbors_graph will otherwise default to sklearn's version due to the MRO.

sklearnex/covariance/tests/test_incremental_covariance.py
icfaust: Added because of: https://github.com/intel/scikit-learn-intelex/pull/1865#discussion_r1654991351

sklearnex/tests/test_run_to_run_stability.py
icfaust: Good catch.  Let me know and I will delete it.

sklearnex/tests/test_patching.py
```diff
@@ -139,6 +140,9 @@ def test_standard_estimator_patching(caplog, dataframe, queue, dtype, estimator,
             ]:
                 pytest.skip(f"{estimator} does not support GPU queues")
 
+        if "NearestNeighbors" in estimator and "radius" in method:
+            pytest.skip(f"RadiusNeighbors estimator not implemented in sklearnex")
```
samir-nasibli, line 144: I wouldn't turn off testing. On our side, sklearn provides functional support for dpnp/dpctl interop. Tests should not fail

icfaust: Thank you for the suggestion.  In this case, we discussed how we wanted to give the user an informative message when using radius_neighbors and radius_neighbors_graph when setting the sklearnex logger to debug, which should tell them that they are using sklearn functionality.  This is done here: https://github.com/icfaust/scikit-learn-intelex/blob/dev/n_jobs_refactor/sklearnex/neighbors/knn_unsupervised.py#L96 . In so doing, this will trigger the sklearnex testing fallback guard here: https://github.com/icfaust/scikit-learn-intelex/blob/dev/n_jobs_refactor/sklearnex/conftest.py#L52 Therefore causing the tests with radius_neighbors and radius_neighbors_graph to fail.  We could turn off the fallback guard, but then that would impact the entirety of the patching tests that this is a part of. @ethanglaser and I discussed and decided that this was the best approach. @samir-nasibli let me know what you think.

sklearnex/tests/_utils.py
icfaust: Good question, in most cases it is necessary to call fit in order to access further functionality, and it is assumed to have been used.

sklearnex/covariance/incremental_covariance.py
```diff
@@ -196,6 +198,45 @@ def _onedal_partial_fit(self, X, queue=None, check_input=True):
...
+        if "numpy" not in xp.__name__:
+            location = xp.asarray(location, device=X_test.device)
+            # depending on the sklearn version, check_array
+            # and validate_data will return only numpy arrays
+            # which will break dpnp/dpctl support. If the
+            # array namespace isn't from numpy and the data
+            # is now a numpy array, it has been validated and
+            # the original can be used.
+            if isinstance(X, np.ndarray):
+                X = X_test
```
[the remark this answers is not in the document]
icfaust: The copy parameter follows the guidelines set by sklearns' IncrementalPCA, https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.IncrementalPCA.html.  This activates the copy keyword of check_array, which will only not copy when it is in the numpy namespace (see here https://github.com/scikit-learn/scikit-learn/blob/70fdc843a/sklearn/utils/validation.py#L1071). In this case the status of the copy parameters is independent of what occurs.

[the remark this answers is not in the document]
icfaust: @samir-nasibli actually you are right, the subtraction will make the copy I need, so its unnecessary to toggle it.

sklearnex/tests/test_run_to_run_stability.py
ethanglaser: Based on the comment it seems like yes - if there are problems with CI after removing then can re-add it.

sklearnex/covariance/incremental_covariance.py
```diff
@@ -196,6 +198,43 @@ def _onedal_partial_fit(self, X, queue=None, check_input=True):
...
+
+        est = clone(self)
+        est.set_params(**{"assume_centered": True})
```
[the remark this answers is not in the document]
icfaust: In sklearn, the calculation of the empirical covariance is a function, where the assumed_centered is a flag. It is a deeply hidden function which is re-used in this case to calculate the score with assumed_centered=True.   Since the oneDAL functionality isn't like this (an object with state), the only way to support non-numpy inputs and to change the assumed centered flag without overwriting the fit values was to use a sklearn copy of the estimator. Sort of a hack I'll admit.

sklearnex/tests/_utils.py
icfaust: I have made the iterables names more informative and added documentation to the functions to explain their use.

sklearnex/tests/test_run_to_run_stability.py
icfaust: Removed.

icfaust:
Private CI run: http://intel-ci.intel.com/ef44e378-76c8-f185-b210-a4bf010d0e2e

sklearnex/tests/test_run_to_run_stability.py
```diff
@@ -73,23 +74,25 @@ def eval_method(X, y, est, method):
...
 
     if not isinstance(res, Iterable):
-        res = [res]
+        results = [_as_numpy(res)] if res is not est else []
+    else:
+        results = []
+        for i in res:
+            results += [_as_numpy(i)]
+
+    attributes = [method for i in results]
```
ethanglaser, line 86:
```suggestion
    attributes = [method] * len(results)
```

sklearnex/tests/test_run_to_run_stability.py
```diff
@@ -73,23 +74,25 @@ def eval_method(X, y, est, method):
...
-            res = est.score(X, y)
+        res = call_method(est, method, X, y)
 
     if not isinstance(res, Iterable):
-        res = [res]
+        results = [_as_numpy(res)] if res is not est else []
+    else:
+        results = []
+        for i in res:
+            results += [_as_numpy(i)]
```
ethanglaser, line 84: list comprehension preferred

sklearnex/tests/test_run_to_run_stability.py
```diff
@@ -148,8 +151,13 @@ def _run_test(estimator, method, datasets):
 def test_standard_estimator_stability(estimator, method, dataframe, queue):
     if estimator in ["LogisticRegression", "TSNE"]:
         pytest.skip(f"stability not guaranteed for {estimator}")
-    if estimator in ["KMeans", "PCA"] and method == "score" and queue == None:
+    if estimator in ["KMeans", "PCA"] and "score" in method and queue == None:
         pytest.skip(f"variation observed in {estimator}.score")
+    if estimator in ["IncrementalEmpiricalCovariance"] and method == "mahalanobis":
```
ethanglaser, line 156:
```suggestion
    if estimator is "IncrementalEmpiricalCovariance" and method == "mahalanobis":
```
Unless its not functionally equivalent - is regular EmpiricalCovariance not part of this?

sklearnex/tests/test_patching.py
ethanglaser, line 1: Quite a bit of code duplication in these test files - possibly something that could be cleaned up in the future

sklearnex/tests/test_patching.py
```diff
@@ -183,11 +185,15 @@ def test_special_estimator_patching(caplog, dataframe, queue, dtype, estimator,
...
-        if dtype == np.float16 and queue and not queue.sycl_device.has_aspect_fp16:
-            pytest.skip("Hardware does not support fp16 SYCL testing")
-        elif dtype == np.float64 and queue and not queue.sycl_device.has_aspect_fp64:
-            pytest.skip("Hardware does not support fp64 SYCL testing")
+        if queue:
+            # Its not possible to get the dpnp/dpctl arrays to be in the proper dtype
+            if dtype == np.float16 and not queue.sycl_device.has_aspect_fp16:
+                pytest.skip("Hardware does not support fp16 SYCL testing")
+            elif dtype == np.float64 and not queue.sycl_device.has_aspect_fp64:
+                pytest.skip("Hardware does not support fp64 SYCL testing")
```
ethanglaser, line 193: I can see the practical reason for this adjustment, but from a readability perspective I think the original makes more sense

sklearnex/tests/_utils.py
```diff
@@ -112,38 +134,103 @@ def __getitem__(self, key):
...
+                attribute = getattr_static(est, attr)
+                if callable(attribute):
+                    params = signature(attribute).parameters
+                    if any([inp in params for inp in required_inputs]):
+                        methods += [attr]
+        else:
+            methods = candidates
 
-        output += [[i, j] for j in methods] if methods else [[i, None]]
+        output += [(estimator, j) for j in methods] if methods else [(estimator, None)]
```
ethanglaser, line 184:
```suggestion
        output += [(estimator, method) for method in methods] if methods else [(estimator, None)]
```

sklearnex/neighbors/common.py
```diff
@@ -264,3 +270,41 @@ def _onedal_gpu_supported(self, method_name, *data):
 
     def _onedal_cpu_supported(self, method_name, *data):
         return self._onedal_supported("cpu", method_name, *data)
+
+    def kneighbors_graph(self, X=None, n_neighbors=None, mode="connectivity"):
```
ethanglaser, line 274: Should X be optional here?

ethanglaser (approved):
I don't have any objections at this point - a bit of minor polishing on some pythonics would be good. I can't access internal CI but would want to make sure it's green. Any other reviews also appreciated as I am not very experienced in this part of code.

sklearnex/tests/test_run_to_run_stability.py
icfaust: good catch. EmpiricalCovariance will have the same issue when it comes out of preview

sklearnex/neighbors/common.py
icfaust: would fail patching/ sklearn conformance tests otherwise, here is the sklearn source code https://github.com/scikit-learn/scikit-learn/blob/70fdc843a/sklearn/neighbors/_base.py#L949

sklearnex/tests/test_run_to_run_stability.py
ethanglaser: Sounds good, that makes sense. Or could be super sneaky and use `if estimator in "IncrementalEmpiricalCovariance"` to cover both 😏 (don't actually do this)

icfaust:
Failure unrelated to PR

PR #1678, opened 2024-01-29
[enhancement] Add intelex-only sklearnex estimators to sklearn via monkeypatch

icfaust:
# Description
Sklearnex contains estimators which are not available in sklearn.  Our unique estimators will now be made available via sklearn import when available in the patch map. This PR enables this style of sklearn patching, and should simplify our users code for users which use patch_sklearn and want to use our unique estimators. It also allows for improved testability of intel-only estimators (making it available to internal sklearnex testing via the patch_map, makes available for testing via sklearn's test_common.py).

Changes proposed in this pull request:
- Modify do_patch and do_unpatch (really should be named undo_patch tbh)
- Modify sklearnex/tests/test_patching.py
- Modify sklearnex/tests/test_monkeypatch.py

PR #1875, opened 2024-06-18
[bug] fix KMeans preview patching

icfaust:
### Description 
I introduced a bug in #1678 which improperly patched KMeans in preview. The sklearn.cluster namespace doesn't have "kmeans" but instead "KMeans". Found by @DDJHB


samir-nasibli (approved):
Assuming green CI.

ethanglaser:
Good catch!

Issue #621, opened 2021-04-14
SVC performance drop if probability=True


Alexsandruss:
**Describe the bug**
daal4py SVC performance drops in first run and becomes slower than sklearn if probability=True passed to params.

**To Reproduce**
Code to reproduce the behavior:
```python
from sklearn.datasets import make_classification
from sklearn.model_selection import train_test_split
from sklearn.svm import SVC as SKLSVC
from daal4py.sklearn.svm import SVC as D4PSVC
from timeit import default_timer as timer


params = {
    'probability': True,
    'C': 1.0,
    'kernel': 'rbf',
    'random_state': 42
}

x, y = make_classification(n_samples=4000, n_features=32, n_classes=2, random_state=42)
X_train, X_test, Y_train, Y_test = train_test_split(x, y, test_size=0.2, random_state=42)

for i in range(5):
    print(f'{i} run:')
    for estimator_class in [SKLSVC, D4PSVC]:
        t0 = timer()
        clsf = estimator_class(**params)
        clsf.fit(X_train, Y_train)
        t1 = timer()
        print(
            str(estimator_class).split("'")[1].split(".")[0], '-',
            round(t1 - t0, 6), 's', '-',
            clsf.score(X_test, Y_test), 'acc'
        )
```

**Expected behavior**
daal4py SVC is faster than from sklearn 

**Output**
```
0 run:
sklearn - 1.721927 s - 0.94 acc
daal4py - 18.420483 s - 0.94375 acc
1 run:
sklearn - 2.12422 s - 0.94 acc
daal4py - 0.207263 s - 0.94375 acc
2 run:
sklearn - 2.122007 s - 0.94 acc
daal4py - 0.202505 s - 0.94375 acc
3 run:
sklearn - 2.090547 s - 0.94 acc
daal4py - 0.210052 s - 0.94375 acc
4 run:
sklearn - 2.111722 s - 0.94 acc
daal4py - 0.204351 s - 0.94375 acc
```

**Environment:**
 - OS: Red Hat Enterprise Linux Server release 7.9
 - SW: scikit-learn 0.24.1, Python 3.8.8
 - scikit-learn-intelex version: 2021.2.2

gorogm:
Similar here, CV+PCA+SVC runs 3-4 times slower with intelex.
Ubuntu, Python 3.8, intelex 2021.2.2.

owerbat:
@gorogm Thank you for reporting. We'll investigate it

[...]

PR #1879, opened 2024-06-19
Change SVM._fit_proba implementation

Alexsandruss:
### Description 

- [ ] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/intel/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_


samir-nasibli:
Thank you @Alexsandruss !
For better review please provide some inputs in the description:
* Justification why it is needed?
* What do we achieve with these changes?
* What will we change for the end user?

ethanglaser (approved):

samir-nasibli (changes requested):
Please address comment above

icfaust:
I think there is a causation issue with switching to 'prefit' on CalibratedClassifierCV (https://stats.stackexchange.com/a/263411), which can explain possible overfitting. Sklearn's SVM is based on the libsvm package (https://www.csie.ntu.edu.tw/~cjlin/libsvm/faq.html#/Q06:_Probability_outputs) which also uses a cross-validation strategy.  Looking at the libsvm source code, the probability metrics are calculated via a default 5 fold stratified Kfold strategy, where the training data fit normally. and then the data is split for the probability calculations into the stratified 5-folds (https://github.com/cjlin1/libsvm/blob/master/svm.cpp#L2067 (regression) https://github.com/cjlin1/libsvm/blob/master/svm.cpp#L1903 (classification)). Our current implementation faithfully reproduces this, but the use of the scikit-learn calibrated cross-validators always seemed like a patch-over to the real solution: implementing a cross-validation strategy in oneDAL.

icfaust (commented):
I won't put a changes requested on the PR, but I'd like to discuss my comment above before any action on the PR is taken (i.e. merge or close).

ethanglaser:
> I won't put a changes requested on the PR, but I'd like to discuss my comment above before any action on the PR is taken (i.e. merge or close).

Yeah I maybe prematurely approved this as I didn't realize offline discussion would be necessary, let's make sure we are all aligned on this before merge

Alexsandruss:
@icfaust @ethanglaser

There are three options about acceleration of SVM classification with probabilities:

- Enable `n_jobs` in `CalibratedClassifierCV` to parallelize training of SVMs, but this way is not stable across all cases due to parallelization backend specifics: `loky` backend has significant startup penalty resulting in long first run while `multiprocessing` and `threading` don't give consistent speedup.
- Leave current approach because it's mathematically correct.
- **Switch to prefit estimator in `CalibratedClassifierCV` for now and optimize calibration on oneDAL side later (gives 2-3x speedup consistently across all cases)**. While this perf. fix is not mathematically correct as calibration data should be separate from fitting data, real impact to quality metrics of models trained in benchmarks is low (except logloss which is already known to be different from stock sklearn).

I suggest to go with last one.

Aggregated benchmark results for this PR:
image

ethanglaser:
Thank you for providing additional context and for encouraging discussion on this. I don’t have objections to the last suggested option. Let's make sure a ticket is created for OneDAL implementation, rebase and re-run CI. But I stand by my initial approval at this point.

samir-nasibli:
@Alexsandruss does this fix original issue on #621
Is it observed yet?
I didn't get what was the root cause of the original issue. Could you please bring clarification in description?

Alexsandruss:
@samir-nasibli, description already covers it: prefit estimator is faster than cross-validation.

samir-nasibli:
> @samir-nasibli, description already covers it: prefit estimator is faster than cross-validation.

@Alexsandruss I found a partial answer not in the description of the PR, but in this comment https://github.com/intel/scikit-learn-intelex/pull/1879#issuecomment-2196646328

As I understand it, this is not a fix, but rather a workaround, taking into account mathematical correctness.

But I have to ask, why in this case are we worse than standard sklearn? What is the real root cause? bottle neck? I still don’t understand what it is so that fixes can be addressed specifically against it.

Usually in such cases, if there is no perf, we fallback to the original sklearn.

I also see a valid comment from @icfaust, and assumed that the fix would be approved and merge after the quorum agreed. I didn't understand the rush.

But anyway, thanks for the updates.

icfaust:
@Alexsandruss please send me the ticket for the follow-up development based off of this PR.

PR #1894, opened 2024-06-25
[enhancement] address new warning in LocalOutlierFactor in sklearn1.6

icfaust:
Adds conformance for the new test: 
test_lof.py::test_lof_duplicate_samples

### DIFF
--- a/sklearnex/neighbors/_lof.py
+++ b/sklearnex/neighbors/_lof.py
@@ -97,6 +97,13 @@ class LocalOutlierFactor(KNeighborsDispatchingBase, sklearn_LocalOutlierFactor):
                 self.negative_outlier_factor_, 100.0 * self.contamination
             )
 
+        if sklearn_check_version("1.6"):
+            if np.min(self.negative_outlier_factor_) < -1e7 and not self.novelty:
+                warnings.warn(
+                    "Duplicate values are leading to incorrect results. "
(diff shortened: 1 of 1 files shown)


sklearnex/neighbors/_lof.py
```diff
@@ -97,6 +97,13 @@ def _onedal_fit(self, X, y, queue=None):
                 self.negative_outlier_factor_, 100.0 * self.contamination
             )
 
+        if sklearn_check_version("1.6"):
```
samir-nasibli, line 100: Could you please reference where and why is it needed?

samir-nasibli (commented):
General question. 1.6 is not stable release version. Should we address this cases, even if we don't have this on our matrices?

sklearnex/neighbors/_lof.py
icfaust: scikit-learn#28773 which is already in planned inclusion in sklearn 1.6: https://scikit-learn.org/dev//whats_new/v1.6.html

Alexsandruss: I guess Samir meant this sklearn issue/PR should be referenced in comment for better documented code. Anyway, I think it is good to start doing it in sklearnex for future explainability of code.

Alexsandruss (approved):

PR #1907, opened 2024-06-29
ENH: fallback to stock `scikit-learn` after `onedal` fail

samir-nasibli:
### Description 
In several estimators there are second level fallback flow, that is not obvious and controlled by the user.
Proposed to add control on fallback computation to stock scikit-learn after `onedal` backend in case of runtime error on `onedal` backend computations. `config_context` extended for `allow_sklearn_after_onedal` param, that control this behavior. 
By default it is allowed to fallback to stock scikit-learn after `onedal` backend in case of runtime error. In `sklearnex` testing and sklearn conformance testing this is blocked, just to check onedal backend.

### Done
* new sklearnex config paramter to control fallback to stock scikit-learn if onedal fails
* added new test case to check config_contex proper work: `test_config_context_works`
* refactor and fixed `test_set_config_works` test suit.
* added comments and docstrings.

- [x] I have updated the documentation to reflect the changes, if necessary (updated in # - _add PR number_)


icfaust:
@samir-nasibli do you need clarification on the background of this oddity in the codebase before we close this PR?  It comes from these steps in the DAAL codebase: https://github.com/oneapi-src/oneDAL/blob/main/cpp/daal/src/algorithms/service_kernel_math.h#L721

samir-nasibli:
> @samir-nasibli do you need clarification on the background of this oddity in the codebase before we close this PR? It comes from these steps in the DAAL codebase: https://github.com/oneapi-src/oneDAL/blob/main/cpp/daal/src/algorithms/service_kernel_math.h#L721

@icfaust thank you! I am definitely not going to close this PR, some case needed to be addressed.

icfaust:
> > @samir-nasibli do you need clarification on the background of this oddity in the codebase before we close this PR? It comes from these steps in the DAAL codebase: https://github.com/oneapi-src/oneDAL/blob/main/cpp/daal/src/algorithms/service_kernel_math.h#L721
> 
> @icfaust thank you! I am definitely not going to close this PR, some case needed to be addressed.

I'll wait to review until you mark it ready, I'm curious to see if you plan to refactor the fallback mechanism or fix a bug in the PLU factorization on oneDAL-side first.

samir-nasibli:
> > > @samir-nasibli do you need clarification on the background of this oddity in the codebase before we close this PR? It comes from these steps in the DAAL codebase: https://github.com/oneapi-src/oneDAL/blob/main/cpp/daal/src/algorithms/service_kernel_math.h#L721
> > 
> > 
> > @icfaust thank you! I am definitely not going to close this PR, some case needed to be addressed.
> 
> I'll wait to review until you mark it ready, I'm curious to see if you plan to refactor the fallback mechanism or fix a bug in the PLU factorization on oneDAL-side first.



> I'll wait to review until you mark it ready, I'm curious to see if you plan to refactor the fallback mechanism or fix a bug in the PLU factorization on oneDAL-side first.

Please read the short description to the PR more carefully.
At the moment there is no ready-made solution for a particular case, there is a draft PR, in the future it is planned to manage this through config_context.

When the proposal is expanded and ready for discussion/review, the PR will be marked ready for review
Thank you!

samir-nasibli:
Linear regression fails all tests on both GPU and CPU, that requires additional investigation.

I will separate deprecation and changes for the LogReg into the separate PR.

samir-nasibli:
The statistics of the onedal4py backend usage and sklearn after onedal4py backend should be investigate.
It seems LinReg has some defects, that should be addressed.

The mechanism itself for falling back sklearn_after_onedal will be updated via config_context.

Follow up tickets will be created.

onedal/_config.py
```diff
@@ -21,6 +21,7 @@
 _default_global_config = {
     "target_offload": "auto",
     "allow_fallback_to_host": False,
+    "allow_sklearn_after_onedal": True,
```
samir-nasibli, line 24: This will be true by default, but in conformance testing will be switched to False just to check onedal backend full.

sklearnex/tests/test_config.py
```diff
@@ -30,13 +30,17 @@ def test_get_config_contains_sklearn_params():
 def test_set_config_works():
     default_config = sklearnex.get_config()
     sklearnex.set_config(
-        assume_finite=True, target_offload="cpu:0", allow_fallback_to_host=True
+        assume_finite=True,
+        target_offload="cpu:0",
+        allow_fallback_to_host=True,
+        allow_sklearn_after_onedal=False,
```
samir-nasibli, line 36: Check generally this test, how is it safe for the config_context work

sklearnex/_config.py
```diff
@@ -51,6 +56,10 @@ def set_config(target_offload=None, allow_fallback_to_host=None, **sklearn_confi
         If True, allows to fallback computation to host device
         in case particular estimator does not support the selected one.
         Global default: False.
+    allow_sklearn_after_onedal : bool, default=True
+        If True, allows to fallback computation to sklearn after onedal
+        backend in case of runtime error on onedal backend computations.
+        Global default: False.
```
samir-nasibli, line 62:
```suggestion
        Global default: True.
```

icfaust (commented):
This is impacted by #2083 , that would mean only LinearRegression GPU, and LogisticRegression GPU estimators are influenced.  Is it worth having a new config keyword for these specific cases? @Alexsandruss should weigh in.  I'm still in favor of closing this PR.  I would say the removal can occur naturally when these issues are addressed in the individual estimators.

samir-nasibli:
> This is impacted by #2083 , that would mean only LinearRegression GPU, and LogisticRegression GPU estimators are influenced. Is it worth having a new config keyword for these specific cases? @Alexsandruss should weigh in. I'm still in favor of closing this PR. I would say the removal can occur naturally when these issues are addressed in the individual estimators.

This PR doesn't change py default anything. Only on testing we are removing these.
We are already have flag for example allowing do computations on onedal cpu if gpu not supported by default, but we are disabling this for testing. The same idea here. If this cases will be addressed then we can remove totally this flag from config_context.

.circleci/run_xpu_tests.py
```diff
@@ -79,10 +79,14 @@
...
 
-    if args.device == "gpu":
-        from sklearnex._config import config_context
+    from sklearnex._config import config_context
 
-        with config_context(target_offload=args.device, allow_fallback_to_host=False):
+    with config_context(allow_sklearn_after_onedal=False):
+
+        if args.device == "gpu":
+            with config_context(target_offload=args.device, allow_fallback_to_host=False):
```
Alexsandruss, line 87: Nested `config_context` doesn't stack - lower context overwrites upper, see `allow_sklearn_after_onedal`:
```python
>>> from sklearnex import config_context, get_config
>>> with config_context(allow_sklearn_after_onedal=False):
...   with config_context(target_offload="gpu", allow_fallback_to_host=False):
...     c = get_config()
...     print(c)
...
```
`{'assume_finite': False, 'working_memory': 1024, 'print_changed_only': True, 'display': 'diagram', 'pairwise_dist_chunk_size': 256, 'enable_cython_pairwise_dist': True, 'array_api_dispatch': False, 'transform_output': 'default', 'enable_metadata_routing': False, 'skip_parameter_validation': False, 'target_offload': 'gpu', 'allow_fallback_to_host': False, 'allow_sklearn_after_onedal': True}`


Is it expected?

Alexsandruss (changes requested):

sklearnex/tests/test_config.py
samir-nasibli, line 1: Needed to add test for a nested `config_context` call

.circleci/run_xpu_tests.py
samir-nasibli: Very good catch.
It is expected that lower context overwrites upper, but only in case if it explicitly updated on the lower context with config_context param. The logic was broken for `allow_sklearn_after_onedal`, update it. Will cover this case with test for nested config_context check for our scikit-learn-intelex config params

samir-nasibli: Fixed at https://github.com/intel/scikit-learn-intelex/pull/1907/commits/50d569fd642a463a45b44bee297bff037e3e3aad and covered with new tests see https://github.com/intel/scikit-learn-intelex/pull/1907/commits/205289e8758b2cc0f4e1364b348c4f36d6484f99

sklearnex/tests/test_config.py
samir-nasibli: done at https://github.com/intel/scikit-learn-intelex/pull/1907/commits/205289e8758b2cc0f4e1364b348c4f36d6484f99

sklearnex/tests/test_config.py
```diff
@@ -28,16 +28,96 @@ def test_get_config_contains_sklearn_params():
 
 
 def test_set_config_works():
+    """Test validates that the config settings were applied correctly by
+    set_config.
+    """
```
Alexsandruss, line 33: These tests will require generalization and parameterization in future.

Alexsandruss (approved):

sklearnex/tests/test_config.py
samir-nasibli: Thank you @Alexsandruss . Please suggest inputs I will cover with a ticket

PR #1924, opened 2024-07-09
ENH: Renamed layout check macro for np.array, changed default layout for oneDAL table

olegkkruglov:
### Description 
* Renamed the layout check macro from `array_is_behaved` to `array_is_behaved_C` to be aligned with the opposite macro `array_is_behaved_F`.
* Changed layout choice in the case if source np.array is both C-contagious and F-contagious. Now C-style layout is chosen which is considered as default in our library.

- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes, if necessary.
- [x] The unit tests pass successfully.
- [x] I have run it locally and tested the changes extensively.
- [x] I have resolved any merge conflicts that might occur with the base branch.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/intel/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_
- [x] I have added a respective label(s) to PR if I have a permission for that.

PR #1925, opened 2024-07-09
ENH: Added serialization for IncrementalEmpiricalCovariance

olegkkruglov:
### Description 
* Added `__getstate__` method to `IncrementalEmpiricalCovariance`
* Finalization is called with serialization during `__getstate__` method
* `_need_to_finalize` flag is added to avoid unnecessary call of finalization backend.
* Added test for checking if it works well
* Updated deselected_tests.yaml

- [x] I have updated the documentation to reflect the changes, if necessary.

### DIFF
--- a/onedal/covariance/covariance.cpp
+++ b/onedal/covariance/covariance.cpp
@@ -18,6 +18,11 @@
 
 #include "oneapi/dal/algo/covariance.hpp"
 
+#define NO_IMPORT_ARRAY // import_array called in table.cpp
+#define PYBIND11_DETAILED_ERROR_MESSAGES
+#include "onedal/datatypes/data_conversion.hpp"
+
+
 #include "onedal/common.hpp"
 #include "onedal/version.hpp"
 
@@ -137,7 +142,25 @@ inline void init_partial_compute_result(pybind11::module_& m) {
         .def(pybind11::init())
         .def_property("partial_n_rows", &result_t::get_partial_n_rows, &result_t::set_partial_n_rows)
         .def_property("partial_crossproduct", &result_t::get_partial_crossproduct, &result_t::set_partial_crossproduct)
-        .def_property("partial_sums", &result_t::get_partial_sum, &result_t::set_partial_sum);
+        .def_property("partial_sums", &result_t::get_partial_sum, &result_t::set_partial_sum)
+        .def(py::pickle(
+            [](const result_t& res) {
+                return py::make_tuple(
+                    convert_to_pyobject(res.get_partial_n_rows()),
+                    convert_to_pyobject(res.get_partial_crossproduct()),
+                    convert_to_pyobject(res.get_partial_sum())
+                );
+            },
+            [](py::tuple t) {
+                if (t.size() != 3)
+                    throw std::runtime_error("Invalid state!");
+                result_t res;
+                res.set_partial_n_rows(convert_to_table(t[0].ptr()));
+                res.set_partial_crossproduct(convert_to_table(t[1].ptr()));
+                res.set_partial_sum(convert_to_table(t[2].ptr()));
+                return res;
+            }
+        ));
 }
 
 #if defined(ONEDAL_VERSION) && ONEDAL_VERSION >= 20240000
--- a/onedal/covariance/covariance.cpp
+++ b/onedal/covariance/covariance.cpp
@@ -19,7 +19,6 @@
 #include "oneapi/dal/algo/covariance.hpp"
 
 #define NO_IMPORT_ARRAY // import_array called in table.cpp
-#define PYBIND11_DETAILED_ERROR_MESSAGES
 #include "onedal/datatypes/data_conversion.hpp"
 
 
@@ -155,9 +154,9 @@ inline void init_partial_compute_result(pybind11::module_& m) {
                 if (t.size() != 3)
                     throw std::runtime_error("Invalid state!");
                 result_t res;
-                res.set_partial_n_rows(convert_to_table(t[0].ptr()));
-                res.set_partial_crossproduct(convert_to_table(t[1].ptr()));
-                res.set_partial_sum(convert_to_table(t[2].ptr()));
+                if (py::cast<int>(t[0].attr("size")) != 0) res.set_partial_n_rows(convert_to_table(t[0].ptr()));
+                if (py::cast<int>(t[1].attr("size")) != 0) res.set_partial_crossproduct(convert_to_table(t[1].ptr()));
+                if (py::cast<int>(t[2].attr("size")) != 0) res.set_partial_sum(convert_to_table(t[2].ptr()));
                 return res;
             }
         ));
--- a/onedal/covariance/covariance.cpp
+++ b/onedal/covariance/covariance.cpp
@@ -146,9 +146,9 @@ inline void init_partial_compute_result(pybind11::module_& m) {
         .def(py::pickle(
             [](const result_t& res) {
                 return py::make_tuple(
-                    py::object(convert_to_pyobject(res.get_partial_n_rows())),
-                    py::object(convert_to_pyobject(res.get_partial_crossproduct())),
-                    py::object(convert_to_pyobject(res.get_partial_sum()))
+                    py::cast<py::object>(convert_to_pyobject(res.get_partial_n_rows())),
+                    py::cast<py::object>(convert_to_pyobject(res.get_partial_crossproduct())),
+                    py::cast<py::object>(convert_to_pyobject(res.get_partial_sum()))
                 );
             },
             [](py::tuple t) {
--- a/onedal/covariance/covariance.cpp
+++ b/onedal/covariance/covariance.cpp
@@ -146,9 +146,9 @@ inline void init_partial_compute_result(pybind11::module_& m) {
         .def(py::pickle(
             [](const result_t& res) {
                 return py::make_tuple(
-                    convert_to_pyobject(res.get_partial_n_rows()),
-                    convert_to_pyobject(res.get_partial_crossproduct()),
-                    convert_to_pyobject(res.get_partial_sum())
+                    py::object(convert_to_pyobject(res.get_partial_n_rows())),
+                    py::object(convert_to_pyobject(res.get_partial_crossproduct())),
+                    py::object(convert_to_pyobject(res.get_partial_sum()))
                 );
             },
             [](py::tuple t) {
--- a/onedal/covariance/incremental_covariance.py
+++ b/onedal/covariance/incremental_covariance.py
@@ -63,37 +63,38 @@ class IncrementalEmpiricalCovariance(BaseEmpiricalCovariance):
         )
 
     def __getstate__(self):
+        """
+        Converts estimator's data to serializable format.
+        All tables contained in partial result are converted to np.arrays.
+
+        Notes
+        -----
+        Since finalize_fit can't be dispatched without directly provided queue
+        and the dispatching policy can't be serialized, the computation is finalized
+        here and the policy is not saved in serialized data.
+        """
         self.finalize_fit()
         data = self.__dict__.copy()
-        partial_result_data = dict()
-        partial_result_data["partial_n_rows"] = from_table(
-            data["_partial_result"].partial_n_rows
-        )
-        partial_result_data["partial_crossproduct"] = from_table(
-            data["_partial_result"].partial_crossproduct
-        )
-        partial_result_data["partial_sums"] = from_table(
-            data["_partial_result"].partial_sums
-        )
+
+        tables_to_save = ["partial_n_rows", "partial_crossproduct", "partial_sums"]
+        partial_result_data = {
+            table_name: from_table(getattr(data["_partial_result"], table_name))
+            for table_name in tables_to_save
+        }
         data["_partial_result"] = partial_result_data
         data.pop("_policy", None)
 
         return data
 
     def __setstate__(self, data):
+        """
+        Restores estimator from serializable data.
+        """
         partial_result = self._get_backend("covariance", None, "partial_compute_result")
-        if data["_partial_result"]["partial_n_rows"].size > 0:
-            partial_result.partial_n_rows = to_table(
-                data["_partial_result"]["partial_n_rows"]
-            )
-        if data["_partial_result"]["partial_crossproduct"].size > 0:
-            partial_result.partial_crossproduct = to_table(
-                data["_partial_result"]["partial_crossproduct"]
-            )
-        if data["_partial_result"]["partial_sums"].size > 0:
-            partial_result.partial_sums = to_table(
-                data["_partial_result"]["partial_sums"]
-            )
+        saved_tables = data["_partial_result"]
+        for table_name, table_data in saved_tables.items():
+            if table_data.size > 0:
+                setattr(partial_result, table_name, to_table(table_data))
 
         data["_partial_result"] = partial_result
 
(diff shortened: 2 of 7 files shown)


onedal/covariance/tests/test_incremental_covariance.py
```diff
@@ -27,8 +27,8 @@ def test_on_gold_data_unbiased(queue, dtype):
     from onedal.covariance import IncrementalEmpiricalCovariance
 
     X = np.array([[0, 1], [0, 1]])
-    X = X.astype(dtype)
```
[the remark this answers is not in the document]
olegkkruglov: it is correct here. check again please.

icfaust (commented):
Good work. I'd like to see getstate and setstate tidied up a bit, and overall some documentation since this is not a straightforward problem. Otherwise good to go.

onedal/datatypes/data_conversion.cpp
```diff
@@ -81,7 +81,7 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
         column_count = static_cast<std::int64_t>(array_size(np_data, 1));
     }
     const auto layout =
-        array_is_behaved_F(np_data) ? dal::data_layout::column_major : dal::data_layout::row_major;
+        array_is_behaved_C(np_data) ? dal::data_layout::row_major : dal::data_layout::column_major;
```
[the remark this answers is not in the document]
olegkkruglov: Here default layout is changed for the cases where converted numpy array is both C-contagious and F-contagious. It can be happen if the array is of shape (1, n) or (n, 1) (it can be seen in `__setstate__` method here). So it means that from_table/to_table cycle changes layout on oneDAL side. and since currently only C-style onedal tables can be converted to Python (I plan to fix it soon), it prevents multiple pickling. Please see https://github.com/intel/scikit-learn-intelex/pull/1924.

onedal/covariance/incremental_covariance.py
```diff
@@ -58,10 +58,48 @@ def __init__(self, method="dense", bias=False, assume_centered=False):
...
+        partial_result_data["partial_n_rows"] = from_table(
+            data["_partial_result"].partial_n_rows
+        )
+        partial_result_data["partial_crossproduct"] = from_table(
+            data["_partial_result"].partial_crossproduct
+        )
+        partial_result_data["partial_sums"] = from_table(
+            data["_partial_result"].partial_sums
+        )
+        data["_partial_result"] = partial_result_data
```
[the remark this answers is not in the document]
icfaust: It looks like the partial result object only contains these three attributes: https://github.com/intel/scikit-learn-intelex/blob/main/onedal/covariance/covariance.cpp#L138
```suggestion
        p_res = data["_partial_result"]
        data["_partial_result"] = {key:from_table(value) for key, value in p_res.__dict__.items()}
```

onedal/covariance/incremental_covariance.py
```diff
@@ -58,10 +58,48 @@ def __init__(self, method="dense", bias=False, assume_centered=False):
...
+        if data["_partial_result"]["partial_crossproduct"].size > 0:
+            partial_result.partial_crossproduct = to_table(
+                data["_partial_result"]["partial_crossproduct"]
+            )
+        if data["_partial_result"]["partial_sums"].size > 0:
+            partial_result.partial_sums = to_table(
+                data["_partial_result"]["partial_sums"]
+            )
+
+        data["_partial_result"] = partial_result
```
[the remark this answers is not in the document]
icfaust: I worry about checking against .size. Is it always initialized to 0? Or is hasattr a better solution? Not sure.

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -163,6 +163,39 @@ def test_sklearnex_fit_on_random_data(
...
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+@pytest.mark.parametrize("dtype", [np.float32])
+def test_pickle(dataframe, queue, dtype):
+    import pickle
+
+    from sklearnex.covariance import IncrementalEmpiricalCovariance
+
+    inccov = IncrementalEmpiricalCovariance()
+    dump = pickle.dumps(inccov)
+    inccov_loaded = pickle.loads(dump)
```
ethanglaser, line 175: I see with our existing pickle tests we are only pickling after fitting, is there a reason to validate pickling on just the object before fitting? is this something that should be extended to other algos?

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -163,6 +163,39 @@ def test_sklearnex_fit_on_random_data(
...
+    dump = pickle.dumps(inccov_loaded)
+    inccov_loaded = pickle.loads(dump)
+    X_split_df = _convert_to_dataframe(X_split[1], sycl_queue=queue, target_df=dataframe)
+    inccov.partial_fit(X_split_df)
+    inccov_loaded.partial_fit(X_split_df)
+    dump = pickle.dumps(inccov)
+    inccov_loaded = pickle.loads(dump)
+
+    dump = pickle.dumps(inccov_loaded)
+    inccov_loaded = pickle.loads(dump)
```
ethanglaser, line 193: is this necessary if we already do it in lines 189-190?

onedal/covariance/tests/test_incremental_covariance.py
ethanglaser, line 1: Is it necessary to validate this in both onedal4py and sklearnex?

onedal/covariance/incremental_covariance.py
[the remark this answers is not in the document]
olegkkruglov: since it is not class with dynamic attributes (see https://pybind11.readthedocs.io/en/stable/classes.html) it has no `__dict__` attribute. so the second option is not valid.

onedal/covariance/incremental_covariance.py
```diff
@@ -58,10 +58,48 @@ def __init__(self, method="dense", bias=False, assume_centered=False):
         self._reset()
 
     def _reset(self):
+        self._need_to_finalize = False
         self._partial_result = self._get_backend(
             "covariance", None, "partial_compute_result"
         )
 
+    def __getstate__(self):
```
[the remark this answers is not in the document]
olegkkruglov: I'll add some comments but I guess documentation will be updated in separate PR since currently we don't even have any mention of incremental algos in our docs.

sklearnex/covariance/tests/test_incremental_covariance.py
olegkkruglov: Just in case and to check we don't get fail after such call. for example it's theoretically possible that there might be errors in __getstate__ and __setstate__ because not all attributes are initialized

onedal/covariance/tests/test_incremental_covariance.py
olegkkruglov: onedal4py IncrementalEmpiricalCovariance and sklearnex IncrementalEmpiricalCovariance are different classes with different attributes and it's not completely fair to assume that they are either both serializable or both not. i think it doesn't make things worse to recheck again if everything works well. it might make sense to remove onedal4py part because onedal4py estimator is an attribute of sklearnex estimator. but as for me it is better to have independent and as full as possible test coverage on both sides

sklearnex/covariance/tests/test_incremental_covariance.py
olegkkruglov: initially this pickling didn't work, so it allowed me to fix a bug here. might make sense to check if other estimators works in this case, especially such that have C++ attributes (not sure if we have examples other than incremental algos).

onedal/covariance/incremental_covariance.py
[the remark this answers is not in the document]
olegkkruglov: could you clarify please what is wrong with size checking? if the tables from getstate are empty, they are converted to empty list (not sure it is good practice but anyway) and here I try to check if getstate returned empty list to avoid providing it into to_table.

sklearnex/covariance/tests/test_incremental_covariance.py
olegkkruglov: it is checked here that finalized estimator is serialized without issues.

olegkkruglov: oh, my bad, for sklearnex estimator the finalization occurs after result check. fixed.

icfaust (commented):
Since the `__dict__` thing I looked into the pybind11 documentation a little bit more.  I found this: https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support  Should we do it on the cpp side for the _partial_result object only?  It would be closer to the source in this case rather than doing it on the python side.  You should be able to use `convert_to_pyobject` and `convert_to_table` from the data_conversion header file.

onedal/covariance/incremental_covariance.py
```diff
@@ -58,10 +58,49 @@ def __init__(self, method="dense", bias=False, assume_centered=False):
...
+            table_name: from_table(getattr(data["_partial_result"], table_name))
+            for table_name in tables_to_save
+        }
+        data["_partial_result"] = partial_result_data
+        data.pop("_policy", None)
+
+        return data
+
+    def __setstate__(self, data):
+        """
```
icfaust, line 91: Just so I understand, what happens when the _policy attribute doesn't exist?  What does that default to in the computation, or will it get auto-generated with a _get_policy call somewhere later on (after unpickling)?

olegkkruglov: yes, it will be autogenerated using provided data. because we finalized computation before serialization, the case when we have to call finalize_fit without data is not possible.

olegkkruglov:
> Since the `__dict__` thing I looked into the pybind11 documentation a little bit more. I found this: https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support Should we do it on the cpp side for the _partial_result object only? It would be closer to the source in this case rather than doing it on the python side. You should be able to use `convert_to_pyobject` and `convert_to_table` from the data_conversion header file.
thanks for pointing out! to be honest I see no serious advantages of this approach over already implemented one. `partial_result` object is used only here by design so no need to implement its serialization separately. If you see something I'm missing please let me know.

icfaust:
> > Since the `__dict__` thing I looked into the pybind11 documentation a little bit more. I found this: https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support Should we do it on the cpp side for the _partial_result object only? It would be closer to the source in this case rather than doing it on the python side. You should be able to use `convert_to_pyobject` and `convert_to_table` from the data_conversion header file.
> 
> thanks for pointing out! to be honest I see no serious advantages of this approach over already implemented one. `partial_result` object is used only here by design so no need to implement its serialization separately. If you see something I'm missing please let me know.

@samir-nasibli @ethanglaser I'd like one/both of you all to weigh in on this.  I think actions of an object should lie with the object itself if possible.  From a maintenance standpoint, any changes to this pybind11 object will likely be propagated elsewhere to a hard-coded list.  From a usability standpoint, while unexpected, pickling the `_partial_fit` object itself doesn't seem possible?

If you guys like it as is, I will accept it.

onedal/covariance/incremental_covariance.py
```diff
@@ -121,21 +161,24 @@ def finalize_fit(self, queue=None):
...
-            self._partial_result,
-        )
-        if daal_check_version((2024, "P", 1)) or (not self.bias):
-            self.covariance_ = from_table(result.cov_matrix)
-        else:
-            n_rows = self._partial_result.partial_n_rows
-            self.covariance_ = from_table(result.cov_matrix) * (n_rows - 1) / n_rows
-
-        self.location_ = from_table(result.means).ravel()
+        if self._need_to_finalize:
```
samir-nasibli, line 164: This comment is out of the scope of the current PR.
I think make sense to do refactoring in follow up. It seems to me that better to have some kind of common BaseIncremenatalEstimator class with such attribute documented well and base abstract methods stored there

onedal/covariance/incremental_covariance.py
```diff
@@ -121,21 +161,24 @@ def finalize_fit(self, queue=None):
...
+            params = self._get_onedal_params(self._dtype)
+            result = self._get_backend(
+                "covariance",
+                None,
+                "finalize_compute",
+                self._policy,
```
samir-nasibli, line 170: Another attr for BaseIncremental estimators. Handling it should be on there as well.
https://github.com/intel/scikit-learn-intelex/pull/1925/files#r1677806644

icfaust:
> > > Since the `__dict__` thing I looked into the pybind11 documentation a little bit more. I found this: https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support Should we do it on the cpp side for the _partial_result object only? It would be closer to the source in this case rather than doing it on the python side. You should be able to use `convert_to_pyobject` and `convert_to_table` from the data_conversion header file.
> > 
> > 
> > thanks for pointing out! to be honest I see no serious advantages of this approach over already implemented one. `partial_result` object is used only here by design so no need to implement its serialization separately. If you see something I'm missing please let me know.
> 
> @samir-nasibli @ethanglaser I'd like one/both of you all to weigh in on this. I think actions of an object should lie with the object itself if possible. From a maintenance standpoint, any changes to this pybind11 object will likely be propagated elsewhere to a hard-coded list. From a usability standpoint, while unexpected, pickling the `_partial_fit` object itself doesn't seem possible?
> 
> If you guys like it as is, I will accept it.

@olegkkruglov I am waiting for a response from them before the next review cycle.

icfaust:
@olegkkruglov any update on this PR? It will need to be rebased for 2025.0 (and what about moving the serialization to pybind side?)

icfaust:
I'm going to wait on private CI to verify the re-enabled test, when it goes through we can merge this.

icfaust (approved):

onedal/covariance/tests/test_incremental_covariance.py
```diff
@@ -120,3 +120,56 @@ def test_partial_fit_on_random_data(
...
+    X = gen.uniform(low=-0.3, high=+0.7, size=(10, 10))
+    X = X.astype(dtype)
+    X_split = np.array_split(X, 2)
+    inccov.partial_fit(X_split[0], queue=queue)
+    inccov_loaded.partial_fit(X_split[0], queue=queue)
+
+    assert inccov._need_to_finalize == True
+    assert inccov_loaded._need_to_finalize == True
+
+    # Check that estmator can be serialized after partial_fit call.
```
icfaust, line 148:
```suggestion
    # Check that estimator can be serialized after partial_fit call.
```

icfaust (commented):
please check and make changes as per what is requested in #2180 before merging.

onedal/covariance/covariance.cpp
```diff
@@ -18,6 +18,10 @@
 
 #include "oneapi/dal/algo/covariance.hpp"
 
+#define NO_IMPORT_ARRAY // import_array called in table.cpp
```
samir-nasibli, line 21: Why is this needed?

onedal/covariance/incremental_covariance.py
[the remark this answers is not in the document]
samir-nasibli: We now have this section. I hope this activity already covered by ticket. I am ok to do it in separate PR

samir-nasibli (approved):

onedal/covariance/covariance.cpp
olegkkruglov: https://github.com/numpy/numpy/issues/9309#issuecomment-311320497
without this define it fails

PR #1909, opened 2024-07-02
TEST: using `get_dataframes_and_queues` instead of `get_queues` in onedal4py testing

samir-nasibli:
### Description 
using `get_dataframes_and_queues` instead of `get_queues` in onedal4py testing.
Enabled only for covariance, incremental covarience, incremental pca (just for verifying the primitive work)
### Justification
onedal4py API and SPMD APIs have ability to work with `numpy` and `sycl_queue` provided. Currently we are testing them with numpy with None queue or with dpctl/dpnp with None queue param. This PR updates get_dataframes_and_queues testing primitive, enhancement includes cases `numpy` + `sycl_queue`.

### TODO
Separately enable it for all cases where `get_queues` used + SPMD iface (#1777 )

Should be 
- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes, if necessary (updated in # - _add PR number_)
- [x] The unit tests pass successfully.
- [x] I have run it locally and tested the changes extensively.
- [x] I have resolved any merge conflicts that might occur with the base branch.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/intel/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_
- [x] I have added a respective label(s) to PR if I have a permission for that.

PR #1777, opened 2024-03-26
TEST: test coverage for sklearnex SPMD ifaces

ethanglaser:
# Description
Scope:
 - Introduces initial validation of spmd algos via mpi-pytest
 - Identify and create follow-up tasks for issues identified in initial testing

For each sklearnex spmd algorithm, individual manually-created tests and parametrized synthetic tests validate model attributes and prediction results against batch implementations

Known issues to address in follow-up:
 - KMeans centers not aligned after init
 - Number of KMeans iterations do not align
 - PCA falls back if mle or n_components integer
 - PCA fails if n_rows_rank < n_cols < n_rows
 - Forest results are not aligned
 - LinReg fails if n_rows_rank < n_cols < n_rows
 - LogReg model coefficients do not align
 - LogReg fails if n_rows_rank < n_cols < n_rows
 - kNN needs all classes represented on each process initially

PR #1940, opened 2024-07-16
FIX: prevent `support_usm_ndarray` from changing queue if explicitly provided.

olegkkruglov:
### Description 

* Changed logic of `support_usm_ndarray`. Now it does not change the queue if it was explicitly provided to args. Previous logic does not allow to use `np.array` + `SyclQueue` interface with decorated method.
* `queue_param` is removed from `support_usm_ndarray` since the check can be done via kwargs checking.
* Docstring is added to `support_usm_ndarray` with highlighting the change

- [x] I have updated the documentation to reflect the changes, if necessary

### DIFF
--- a/onedal/_device_offload.py
+++ b/onedal/_device_offload.py
@@ -178,11 +178,23 @@ if dpnp_available:
 
 
 def support_usm_ndarray(freefunc=False, queue_param=True):
+    """
+    Handles USMArray input. Puts SYCLQueue from data to method arguments.
+
+    Parameters
+    ----------
+    freefunc (bool) : Set to True if decorates free function.
+    queue_param (bool) : Set to True if queue from data should be provided to the method args.
+                         Queue will not be changed if provided explicitly.
+    """
+
     def decorator(func):
         def wrapper_impl(obj, *args, **kwargs):
             usm_iface = _extract_usm_iface(*args, **kwargs)
             data_queue, hostargs, hostkwargs = _get_host_inputs(*args, **kwargs)
-            if queue_param:
+            if queue_param and not (
+                "queue" in hostkwargs and hostkwargs["queue"] is not None
+            ):
                 hostkwargs["queue"] = data_queue
             result = _run_on_device(func, obj, *hostargs, **hostkwargs)
             if usm_iface is not None and hasattr(result, "__array_interface__"):
--- a/onedal/_device_offload.py
+++ b/onedal/_device_offload.py
@@ -179,12 +179,13 @@ if dpnp_available:
 
 def support_usm_ndarray(freefunc=False, queue_param=True):
     """
-    Handles USMArray input. Puts SYCLQueue from data to method arguments.
+    Handles USMArray input. Puts SYCLQueue from data to decorated function arguments.
+    Converts output of decorated function to dpnp.ndarray if input was of this type.
 
     Parameters
     ----------
     freefunc (bool) : Set to True if decorates free function.
-    queue_param (bool) : Set to True if queue from data should be provided to the method args.
+    queue_param (bool) : Set to True if queue from data should be provided to the decorated function args.
                          Queue will not be changed if provided explicitly.
     """
 
--- a/onedal/_device_offload.py
+++ b/onedal/_device_offload.py
@@ -180,13 +180,16 @@ if dpnp_available:
 def support_usm_ndarray(freefunc=False, queue_param=True):
     """
     Handles USMArray input. Puts SYCLQueue from data to decorated function arguments.
-    Converts output of decorated function to dpnp.ndarray if input was of this type.
+    Converts output of decorated function to dpctl.tensor/dpnp.ndarray if input was of this type.
 
     Parameters
     ----------
     freefunc (bool) : Set to True if decorates free function.
-    queue_param (bool) : Set to True if queue from data should be provided to the decorated function args.
-                         Queue will not be changed if provided explicitly.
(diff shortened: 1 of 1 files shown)


onedal/_device_offload.py
```diff
@@ -178,11 +178,23 @@ def _convert_to_dpnp(array):
...
+                         Queue will not be changed if provided explicitly.
+    """
+
     def decorator(func):
         def wrapper_impl(obj, *args, **kwargs):
             usm_iface = _extract_usm_iface(*args, **kwargs)
             data_queue, hostargs, hostkwargs = _get_host_inputs(*args, **kwargs)
-            if queue_param:
+            if queue_param and not (
+                "queue" in hostkwargs and hostkwargs["queue"] is not None
```
samir-nasibli, line 196:
```suggestion
                "queue" in hostkwargs and hostkwargs["queue"] is None
```
queue is prior than data_queue

samir-nasibli (commented):
Thank you for founding out this issue.
This is critical for sklearnex.spmd ifaces with numpy + sycl_queue. Unfortunately we don't have properly testing yet for this API. Only some examples with dpnp/dpctl input, that not the case of this bug. Hopefully development on #1777 and #1909 will cover this cases as well.

I would like to have some tests in case. I can follow this add it here.

onedal/_device_offload.py
```diff
@@ -178,11 +178,23 @@ def _convert_to_dpnp(array):
...
 
 def support_usm_ndarray(freefunc=False, queue_param=True):
+    """
+    Handles USMArray input. Puts SYCLQueue from data to method arguments.
+
+    Parameters
+    ----------
+    freefunc (bool) : Set to True if decorates free function.
+    queue_param (bool) : Set to True if queue from data should be provided to the method args.
+                         Queue will not be changed if provided explicitly.
```
samir-nasibli, line 188:
```suggestion
    queue_param (bool) : Set to True if decorated function has a `queue` parameter.
                         Queue will not be changed if provided explicitly.
```

onedal/_device_offload.py
samir-nasibli, line 196:
```suggestion
                "queue" in hostkwargs and hostkwargs["queue"] is None
```
I think provided queue param value is more prior that input data's context. But maybe some checks are required

olegkkruglov: I can't get your suggested change. Currently the condition in this row means that queue is provided and not equal to None, so shouldn't be changed

onedal/_device_offload.py
olegkkruglov: I would say that my comment makes more sense. It might be possible that decorated function has `queue` parameter but user doesn't want to put queue from data there.

onedal/_device_offload.py
samir-nasibli: Ah, sorry, didn't see not before the braces.

onedal/_device_offload.py
```diff
@@ -178,11 +178,24 @@ def _convert_to_dpnp(array):
...
+    """
+
     def decorator(func):
         def wrapper_impl(obj, *args, **kwargs):
             usm_iface = _extract_usm_iface(*args, **kwargs)
             data_queue, hostargs, hostkwargs = _get_host_inputs(*args, **kwargs)
-            if queue_param:
+            if queue_param and not (
+                "queue" in hostkwargs and hostkwargs["queue"] is not None
+            ):
```
samir-nasibli, line 198:
```suggestion
            if not ("queue" in hostkwargs and hostkwargs["queue"] is not None):
```
This param were added on #1832, now it looks overkill for this decorator. Maybe it make sense just remove `queue_param` usage and param from `support_usm_ndarray`and just leave this condition checking.

Just needed to remove all `queue_param=False` from everywhere.

onedal/_device_offload.py
samir-nasibli: outdated

onedal/_device_offload.py
```diff
@@ -178,11 +178,25 @@ def _convert_to_dpnp(array):
...
 def support_usm_ndarray(freefunc=False, queue_param=True):
+    """
+    Handles USMArray input. Puts SYCLQueue from data to decorated function arguments.
+    Converts output of decorated function to dpctl.tensor/dpnp.ndarray if input was of this type.
+
+    Parameters
+    ----------
+    freefunc (bool) : Set to True if decorates free function.
+    queue_param (bool) : Set to True if queue from data should be provided to the decorated function args
+                        (recommended to set to True if decorated function has `queue` parameter).
```
samir-nasibli, line 189:
```suggestion
                        Note: set to True if decorated function has `queue` parameter.
```

samir-nasibli (changes requested):

samir-nasibli (commented):
We went a little bit wrong in our thoughts, but it seems like it should be normal now.

Again without testing this primitive we could possible have some other issues. It would be great to add tests to check for this estimator expected behavior for the dummy function. Could be done in separate PR as follow up, but make sure that we have ticket covered. Thank you!

samir-nasibli (approved):
LGTM. Assuming green CI

onedal/_device_offload.py
samir-nasibli: Outdated.

ethanglaser:
Please run black formatting on the file

ethanglaser (approved):

samir-nasibli:
@mergify backport rls/2024.6.0-rls

PR #1961, opened 2024-07-25
ENH: SPMD interface for IncrementalBasicStatistics

olegkkruglov:
### Description 

* Added SPMD interface for IncrementalBasicStatistics
* Changed policy saving workflow, now queue is saved to attributes instead of policy. It is necessary because `finalize_fit` requires `spmd_policy`, but `partial_fit` requires `data_parallel_policy` on oneDAL side
* `finalize_fit` now uses provided queue for computations on onedal4py side. 
* Contains some content from https://github.com/intel/scikit-learn-intelex/pull/1777 for test implementation

- [ ] I have updated the documentation to reflect the changes, if necessary.

### DIFF
--- a/onedal/spmd/basic_statistics/__init__.py
+++ b/onedal/spmd/basic_statistics/__init__.py
@@ -15,5 +15,6 @@
 # ==============================================================================
 
 from .basic_statistics import BasicStatistics
+from .incremental_basic_statistics import IncrementalBasicStatistics
 
 __all__ = ["BasicStatistics"]
--- a/onedal/spmd/basic_statistics/incremental_basic_statistics.py
+++ b/onedal/spmd/basic_statistics/incremental_basic_statistics.py
@@ -1,0 +1,76 @@
+# ==============================================================================
+# Copyright 2024 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from daal4py.sklearn._utils import get_dtype
+
+from ..._device_offload import support_usm_ndarray
+from ...basic_statistics import (
+    IncrementalBasicStatistics as IncrementalBasicStatistics_Batch,
+)
+from ...datatypes import _convert_to_supported, to_table
+from .._base import BaseEstimatorSPMD
+
+
+class IncrementalBasicStatistics(BaseEstimatorSPMD, IncrementalBasicStatistics_Batch):
+    def _reset(self):
+        self._partial_result = super(IncrementalBasicStatistics_Batch, self)._get_backend(
+            "basic_statistics", None, "partial_compute_result"
+        )
+
+    @support_usm_ndarray
+    def partial_fit(self, X, weights=None, queue=None):
+        """
+        Computes partial data for basic statistics
+        from data batch X and saves it to `_partial_result`.
+
+        Parameters
+        ----------
+        X : array-like of shape (n_samples, n_features)
+            Training data batch, where `n_samples` is the number of samples
+            in the batch, and `n_features` is the number of features.
+
+        queue : dpctl.SyclQueue
+            If not None, use this queue for computations.
+
+        Returns
+        -------
+        self : object
+            Returns the instance itself.
+        """
+        if not hasattr(self, "_queue"):
+            self._queue = queue
+        policy = self._get_policy(queue, X)
+        X, weights = _convert_to_supported(policy, X, weights)
+
+        if not hasattr(self, "_onedal_params"):
+            dtype = get_dtype(X)
+            self._onedal_params = self._get_onedal_params(dtype)
+
+        X_table, weights_table = to_table(X, weights)
+        self._partial_result = super(IncrementalBasicStatistics_Batch, self)._get_backend(
+            "basic_statistics",
+            None,
+            "partial_compute",
+            policy,
+            self._onedal_params,
+            self._partial_result,
+            X_table,
+            weights_table,
+        )
+
+    @support_usm_ndarray()
+    def finalize_fit(self, queue=None):
+        return super().finalize_fit(queue=queue)
--- a/sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
+++ b/sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
@@ -1,0 +1,306 @@
+# ==============================================================================
+# Copyright 2024 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+
+from onedal.basic_statistics.tests.test_basic_statistics import options_and_tests
+from onedal.tests.utils._dataframes_support import (
+    _convert_to_dataframe,
+    get_dataframes_and_queues,
+)
+from sklearnex.tests._utils_spmd import (
+    _generate_statistic_data,
+    _generate_weights,
+    _get_local_tensor,
+    _mpi_libs_and_gpu_available,
+)
+
+
+@pytest.mark.skipif(
+    not _mpi_libs_and_gpu_available,
+    reason="GPU device and MPI libs required for test",
+)
+@pytest.mark.parametrize(
+    "dataframe,queue",
+    get_dataframes_and_queues(dataframe_filter_="dpnp,dpctl", device_filter_="gpu"),
+)
+@pytest.mark.parametrize("weighted", [True, False])
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+@pytest.mark.mpi
+def test_incremental_basic_statistics_fit_spmd_gold(dataframe, queue, weighted, dtype):
+    # Import spmd and batch algo
+    from sklearnex.basic_statistics import IncrementalBasicStatistics
+    from sklearnex.spmd.basic_statistics import (
+        IncrementalBasicStatistics as IncrementalBasicStatistics_SPMD,
+    )
+
+    # Create gold data and process into dpt
+    data = np.array(
+        [
+            [0.0, 0.0, 0.0],
+            [0.0, 1.0, 2.0],
+            [0.0, 2.0, 4.0],
+            [0.0, 3.0, 8.0],
+            [0.0, 4.0, 16.0],
+            [0.0, 5.0, 32.0],
+            [0.0, 6.0, 64.0],
+        ]
+    ).astype(dtype=dtype)
+
+    dpt_data = _convert_to_dataframe(data, sycl_queue=queue, target_df=dataframe)
+
+    local_dpt_data = _convert_to_dataframe(
+        _get_local_tensor(data), sycl_queue=queue, target_df=dataframe
+    )
+
+    if weighted:
+        weights = np.array([0, 1, 2, 3, 4, 5, 6]).astype(dtype=dtype)
+        dpt_weights = _convert_to_dataframe(
+            weights, sycl_queue=queue, target_df=dataframe
+        )
+        local_dpt_weights = _convert_to_dataframe(
+            _get_local_tensor(weights), sycl_queue=queue, target_df=dataframe
+        )
+
+    # ensure results of batch algo match spmd
+
+    if weighted:
+        incbs_spmd = IncrementalBasicStatistics_SPMD().fit(
+            local_dpt_data, sample_weight=local_dpt_weights
+        )
+        incbs = IncrementalBasicStatistics().fit(dpt_data, sample_weight=dpt_weights)
+
+    else:
+        incbs_spmd = IncrementalBasicStatistics_SPMD().fit(local_dpt_data)
+        incbs = IncrementalBasicStatistics().fit(dpt_data)
+
+    for option in options_and_tests:
+        assert_allclose(getattr(incbs_spmd, option[0]), getattr(incbs, option[0]))
+
+
+@pytest.mark.skipif(
+    not _mpi_libs_and_gpu_available,
+    reason="GPU device and MPI libs required for test",
+)
+@pytest.mark.parametrize(
+    "dataframe,queue",
+    get_dataframes_and_queues(dataframe_filter_="dpnp,dpctl", device_filter_="gpu"),
+)
+@pytest.mark.parametrize("num_blocks", [1, 2])
+@pytest.mark.parametrize("weighted", [True, False])
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+@pytest.mark.mpi
+def test_incremental_basic_statistics_partial_fit_spmd_gold(
+    dataframe, queue, num_blocks, weighted, dtype
+):
+    # Import spmd and batch algo
+    from sklearnex.basic_statistics import IncrementalBasicStatistics
+    from sklearnex.spmd.basic_statistics import (
+        IncrementalBasicStatistics as IncrementalBasicStatistics_SPMD,
+    )
+
+    # Create gold data and process into dpt
+    data = np.array(
+        [
+            [0.0, 0.0, 0.0],
+            [0.0, 1.0, 2.0],
+            [0.0, 2.0, 4.0],
+            [0.0, 3.0, 8.0],
+            [0.0, 4.0, 16.0],
+            [0.0, 5.0, 32.0],
+            [0.0, 6.0, 64.0],
+        ]
+    )
+    data = data.astype(dtype=dtype)
+    dpt_data = _convert_to_dataframe(data, sycl_queue=queue, target_df=dataframe)
+    local_data = _get_local_tensor(data)
+    split_local_data = np.array_split(local_data, num_blocks)
+
+    if weighted:
+        weights = np.array([0, 1, 2, 3, 4, 5, 6])
+        dpt_weights = _convert_to_dataframe(
+            weights, sycl_queue=queue, target_df=dataframe
+        )
+        local_weights = _get_local_tensor(weights)
+        split_local_weights = np.array_split(local_weights, num_blocks)
+
+    incbs_spmd = IncrementalBasicStatistics_SPMD()
+    incbs = IncrementalBasicStatistics()
+
+    for i in range(num_blocks):
+        local_dpt_data = _convert_to_dataframe(
+            split_local_data[i], sycl_queue=queue, target_df=dataframe
+        )
+        if weighted:
+            local_dpt_weights = _convert_to_dataframe(
+                split_local_weights[i], sycl_queue=queue, target_df=dataframe
+            )
+            incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights)
+        else:
+            incbs_spmd.partial_fit(local_dpt_data)
+
+    if weighted:
+        incbs.fit(dpt_data, sample_weight=dpt_weights)
+    else:
+        incbs.fit(dpt_data)
+
+    for option in options_and_tests:
+        assert_allclose(getattr(incbs_spmd, option[0]), getattr(incbs, option[0]))
+
+
+@pytest.mark.skipif(
+    not _mpi_libs_and_gpu_available,
+    reason="GPU device and MPI libs required for test",
+)
+@pytest.mark.parametrize(
+    "dataframe,queue",
+    get_dataframes_and_queues(dataframe_filter_="dpnp,dpctl", device_filter_="gpu"),
+)
+@pytest.mark.parametrize("num_blocks", [1, 2])
+@pytest.mark.parametrize("weighted", [True, False])
+@pytest.mark.parametrize("result_option", options_and_tests)
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+@pytest.mark.mpi
+def test_incremental_basic_statistics_single_option_partial_fit_spmd_gold(
+    dataframe, queue, num_blocks, weighted, result_option, dtype
+):
+    # Import spmd and batch algo
+    from sklearnex.basic_statistics import IncrementalBasicStatistics
+    from sklearnex.spmd.basic_statistics import (
+        IncrementalBasicStatistics as IncrementalBasicStatistics_SPMD,
+    )
+
+    # Create gold data and process into dpt
+    data = np.array(
+        [
+            [0.0, 0.0, 0.0],
+            [0.0, 1.0, 2.0],
+            [0.0, 2.0, 4.0],
+            [0.0, 3.0, 8.0],
+            [0.0, 4.0, 16.0],
+            [0.0, 5.0, 32.0],
+            [0.0, 6.0, 64.0],
+        ]
+    )
+    data = data.astype(dtype=dtype)
+    dpt_data = _convert_to_dataframe(data, sycl_queue=queue, target_df=dataframe)
+    local_data = _get_local_tensor(data)
+    split_local_data = np.array_split(local_data, num_blocks)
+
+    if weighted:
+        weights = np.array([0, 1, 2, 3, 4, 5, 6])
+        dpt_weights = _convert_to_dataframe(
+            weights, sycl_queue=queue, target_df=dataframe
+        )
+        local_weights = _get_local_tensor(weights)
+        split_local_weights = np.array_split(local_weights, num_blocks)
+
+    option, _, _ = result_option
+    incbs_spmd = IncrementalBasicStatistics_SPMD(result_options=option)
+    incbs = IncrementalBasicStatistics(result_options=option)
+
+    for i in range(num_blocks):
+        local_dpt_data = _convert_to_dataframe(
+            split_local_data[i], sycl_queue=queue, target_df=dataframe
+        )
+        if weighted:
+            local_dpt_weights = _convert_to_dataframe(
+                split_local_weights[i], sycl_queue=queue, target_df=dataframe
+            )
+            incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights)
+        else:
+            incbs_spmd.partial_fit(local_dpt_data)
+
+    if weighted:
+        incbs.fit(dpt_data, sample_weight=dpt_weights)
+    else:
+        incbs.fit(dpt_data)
+
+    assert_allclose(getattr(incbs_spmd, option), getattr(incbs, option))
+
+
+@pytest.mark.skipif(
+    not _mpi_libs_and_gpu_available,
+    reason="GPU device and MPI libs required for test",
+)
+@pytest.mark.parametrize(
+    "dataframe,queue",
+    get_dataframes_and_queues(dataframe_filter_="dpnp,dpctl", device_filter_="gpu"),
+)
+@pytest.mark.parametrize("num_blocks", [1, 2])
+@pytest.mark.parametrize("weighted", [True, False])
+@pytest.mark.parametrize("n_samples", [100, 10000])
+@pytest.mark.parametrize("n_features", [10, 100])
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+@pytest.mark.mpi
+def test_incremental_basic_statistics_partial_fit_spmd_synthetic(
+    dataframe, queue, num_blocks, weighted, n_samples, n_features, dtype
+):
+    # Import spmd and batch algo
+    from sklearnex.basic_statistics import IncrementalBasicStatistics
+    from sklearnex.spmd.basic_statistics import (
+        IncrementalBasicStatistics as IncrementalBasicStatistics_SPMD,
+    )
+
+    # Create gold data and process into dpt
+    data = _generate_statistic_data(n_samples, n_features).astype(dtype=dtype)
+    local_data = _get_local_tensor(data)
+    split_local_data = np.array_split(local_data, num_blocks)
+    split_data = np.array_split(data, num_blocks)
+
+    if weighted:
+        weights = _generate_weights(n_samples).astype(dtype=dtype)
+        local_weights = _get_local_tensor(weights)
+        split_local_weights = np.array_split(local_weights, num_blocks)
+        split_weights = np.array_split(weights, num_blocks)
+
+    incbs_spmd = IncrementalBasicStatistics_SPMD()
+    incbs = IncrementalBasicStatistics()
+
+    for i in range(num_blocks):
+        local_dpt_data = _convert_to_dataframe(
+            split_local_data[i], sycl_queue=queue, target_df=dataframe
+        )
+        dpt_data = _convert_to_dataframe(
+            split_data[i], sycl_queue=queue, target_df=dataframe
+        )
+        if weighted:
+            local_dpt_weights = _convert_to_dataframe(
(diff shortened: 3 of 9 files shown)


onedal/spmd/basic_statistics/__init__.py
```diff
@@ -15,5 +15,6 @@
 # ==============================================================================
 
 from .basic_statistics import BasicStatistics
+from .incremental_basic_statistics import IncrementalBasicStatistics
 
 __all__ = ["BasicStatistics"]
```
ethanglaser, line 20:
```suggestion
__all__ = ["BasicStatistics", "IncrementalBasicStatistics"]
```

onedal/spmd/basic_statistics/incremental_basic_statistics.py
```diff
@@ -0,0 +1,76 @@
...
+            If not None, use this queue for computations.
+
+        Returns
+        -------
+        self : object
+            Returns the instance itself.
+        """
+        if not hasattr(self, "_queue"):
+            self._queue = queue
+        policy = self._get_policy(queue, X)
```
ethanglaser, line 55: Why is super policy not needed here? I see in other incremental algos it is - what makes it be one way or the other?

olegkkruglov: it was a bug, fixed now. thanks for pointing out.

olegkkruglov:
https://ecmd.jf.intel.com/commander/link/jobDetails/jobs/ef5f1a8a-fd29-f1cd-b7c8-a4bf010d0e2e

ethanglaser:
https://intel-ci.intel.com/ef5f1a8a-fd29-f1cd-b7c8-a4bf010d0e2e

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+            [0.0, 7.0, 128.0],
+        ]
+    )
+    data = data.astype(dtype=dtype)
+    dpt_data = _convert_to_dataframe(data, sycl_queue=queue, target_df=dataframe)
+    local_data = _get_local_tensor(data)
+    split_local_data = np.array_split(local_data, num_blocks)
+
+    if weighted:
+        weights = np.array([0, 1, 2, 3, 4, 5, 6, 7])
```
icfaust, line 137: See other comment

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+            incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights)
+        else:
+            incbs_spmd.partial_fit(local_dpt_data)
+
+    if weighted:
+        incbs.fit(dpt_data, sample_weight=dpt_weights)
+    else:
+        incbs.fit(dpt_data)
+
+    assert_allclose(getattr(incbs_spmd, option), getattr(incbs, option))
```
icfaust, line 237: why use getattr?

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+            incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights)
+            incbs.partial_fit(dpt_data, sample_weight=dpt_weights)
+        else:
+            incbs_spmd.partial_fit(local_dpt_data)
+            incbs.partial_fit(dpt_data)
+
+    for option in options_and_tests:
+        assert_allclose(
+            getattr(incbs_spmd, option[0]), getattr(incbs, option[0]), atol=tol
+        )
```
icfaust, line 303:
```suggestion
    for option, _, _ in options_and_tests:
        assert_allclose(
            getattr(incbs_spmd, option), getattr(incbs, option), atol=tol
        )
```

for consistency with line 216

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+                split_local_weights[i], sycl_queue=queue, target_df=dataframe
+            )
+            incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights)
+        else:
+            incbs_spmd.partial_fit(local_dpt_data)
+
+    if weighted:
+        incbs.fit(dpt_data, sample_weight=dpt_weights)
+    else:
+        incbs.fit(dpt_data)
```
icfaust, line 235:
```suggestion
   incbs.fit(dpt_data, sample_weight=dpt_weights if weighted else None)
```
can be made a one-liner

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+                split_local_weights[i], sycl_queue=queue, target_df=dataframe
+            )
+            dpt_weights = _convert_to_dataframe(
+                split_weights[i], sycl_queue=queue, target_df=dataframe
+            )
+            incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights)
+            incbs.partial_fit(dpt_data, sample_weight=dpt_weights)
+        else:
+            incbs_spmd.partial_fit(local_dpt_data)
+            incbs.partial_fit(dpt_data)
```
icfaust, line 298:
```suggestion
        incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights if weighted else None)
        incbs.partial_fit(dpt_data, sample_weight=dpt_weights if weighted else None)
```

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+    incbs_spmd = IncrementalBasicStatistics_SPMD()
+    incbs = IncrementalBasicStatistics()
+
+    for i in range(num_blocks):
+        local_dpt_data = _convert_to_dataframe(
+            split_local_data[i], sycl_queue=queue, target_df=dataframe
+        )
+        dpt_data = _convert_to_dataframe(
+            split_data[i], sycl_queue=queue, target_df=dataframe
+        )
```
icfaust, line 286: could you add ``` assert(dpt_data.dtype == dtype) ``` somewhere later in your tests, since dtype isn't being passed to _convert_to_dataframe, I would like to verify that the conversion hasn't corrupted the dtype. Either that or explicitly pass the dtype as a kwarg.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+            [0.0, 2.0, 4.0],
+            [0.0, 3.0, 8.0],
+            [0.0, 4.0, 16.0],
+            [0.0, 5.0, 32.0],
+            [0.0, 6.0, 64.0],
+            [0.0, 7.0, 128.0],
+        ]
+    )
+    data = data.astype(dtype=dtype)
+    dpt_data = _convert_to_dataframe(data, sycl_queue=queue, target_df=dataframe)
```
icfaust, line 204: see comment about dtype verification

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+                split_local_weights[i], sycl_queue=queue, target_df=dataframe
+            )
+            incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights)
+        else:
+            incbs_spmd.partial_fit(local_dpt_data)
+
+    if weighted:
+        incbs.fit(dpt_data, sample_weight=dpt_weights)
+    else:
+        incbs.fit(dpt_data)
```
icfaust, line 162:
```suggestion
    incbs.fit(dpt_data, sample_weight=dpt_weights if weighted else None)
```

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+        local_dpt_data = _convert_to_dataframe(
+            split_local_data[i], sycl_queue=queue, target_df=dataframe
+        )
+        if weighted:
+            local_dpt_weights = _convert_to_dataframe(
+                split_local_weights[i], sycl_queue=queue, target_df=dataframe
+            )
+            incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights)
+        else:
+            incbs_spmd.partial_fit(local_dpt_data)
```
icfaust, line 157:
```suggestion
        incbs_spmd.partial_fit(local_dpt_data, sample_weight=local_dpt_weights if weighted else None)
```

onedal/spmd/basic_statistics/incremental_basic_statistics.py
```diff
@@ -0,0 +1,78 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from daal4py.sklearn._utils import get_dtype
+
+from ..._device_offload import support_usm_ndarray
+from ...basic_statistics import (
+    IncrementalBasicStatistics as IncrementalBasicStatistics_nonSPMD,
```
icfaust, line 21:
```suggestion
    IncrementalBasicStatistics as IncBS_Standard,
```

Just as a note, this is setting a precedent with the naming convention for incremental SPMD inheritance and class definition (similar to how ```_Batch``` has been previously used, but is not correct in this circumstance). Are we sure defining something due to the absence of an attribute is a good thing? Wouldn't something like ```_Standard``` or ```_Orig``` be better?  I know this is a ridiculous semantic thing, but we are going to see this now in 4+ spots through the codebase. @ethanglaser what are your thoughts?  If we shorten it, it may be simpler to read since we can reduce multi-lining.

onedal/spmd/basic_statistics/incremental_basic_statistics.py
```diff
@@ -0,0 +1,78 @@
...
+            policy,
+            self._onedal_params,
+            self._partial_result,
+            X_table,
+            weights_table,
+        )
+
+    @support_usm_ndarray()
+    def finalize_fit(self, queue=None):
+        return super().finalize_fit(queue=queue)
```
icfaust, line 78: Doesn't inheritance handle this definition? Superfluous?

icfaust (commented):
Most of my comments are about the tests, as the inheritance is rather straightforward.  Most are semantic questions since we are setting precedence for the other incremental algos.  Please ping me when the changes related to the review are complete and I will re-review.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
olegkkruglov: what is your suggestion?

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
olegkkruglov: tbh i think it should be verified in separate _convert_to_dataframe test, not here. and this construction used a lot of times in our tests and i've never seen dtype is provided

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
olegkkruglov: https://github.com/intel/scikit-learn-intelex/pull/1961/files/ed1767bd506673d8f5636a9bd44df8767a1dfd94#r1732489632

olegkkruglov: https://github.com/intel/scikit-learn-intelex/pull/1961#discussion_r1732489632

onedal/spmd/basic_statistics/incremental_basic_statistics.py
olegkkruglov: it is done similar to all others algos in onedal/spmd and if you want to change that then it should be done in a separate pr I guess

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+            [0.0, 0.0, 0.0],
+            [0.0, 1.0, 2.0],
+            [0.0, 2.0, 4.0],
+            [0.0, 3.0, 8.0],
+            [0.0, 4.0, 16.0],
+            [0.0, 5.0, 32.0],
+            [0.0, 6.0, 64.0],
+            [0.0, 7.0, 128.0],
+        ]
+    ).astype(dtype=dtype)
```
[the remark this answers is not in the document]
olegkkruglov: I'm not sure it is necessary. The gold data for different algos could be different in future, so it's better to keep the style when gold data is defined directly in the test.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,303 @@
...
+            [0.0, 7.0, 128.0],
+        ]
+    )
+    data = data.astype(dtype=dtype)
+    dpt_data = _convert_to_dataframe(data, sycl_queue=queue, target_df=dataframe)
+    local_data = _get_local_tensor(data)
+    split_local_data = np.array_split(local_data, num_blocks)
+
+    if weighted:
+        weights = np.array([0, 1, 2, 3, 4, 5, 6, 7])
```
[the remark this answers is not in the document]
olegkkruglov: I prefer direct setting of weights to avoid let user think weights must be consecutive ints. dtype setting for weights is updated.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
olegkkruglov: https://github.com/intel/scikit-learn-intelex/pull/1961#discussion_r1732704687

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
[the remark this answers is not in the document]
icfaust: Now counting this and in other PRs of similar style, the count is 5+.  When this circumstance arises we can define a 'silver' data or a 'bronze' data. Please change it. If you are still against it we can get a 3rd person to weigh in.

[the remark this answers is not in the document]
olegkkruglov: I still see nothing bad in current implementation. as for me it is better to keep it as is so it would be possible easily and independently change it if necessary. but it does not really matter tbh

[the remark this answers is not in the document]
ethanglaser: Lets save it for a follow-up

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
ethanglaser: does incbs_spmd.option work? this would be preferred

onedal/spmd/basic_statistics/incremental_basic_statistics.py
ethanglaser: I believe its necessary because of the addition of support_usm_ndarray

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
olegkkruglov: it doesn't

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
icfaust: https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/tests/_utils.py#L307

This is a pattern used extensively in test_patching and test_run_to_run_stability.  This will pass the dtype setting to the dataframe constructor to insure proper dtyping.  I would like your conversions to be explicit in the dtype it requests, it isn't clear to me if it is guaranteed to convert to the same datatype.

onedal/spmd/basic_statistics/incremental_basic_statistics.py
icfaust: As per the documentation of (https://github.com/intel/scikit-learn-intelex/blob/main/onedal/_device_offload.py#L180): 
```    Handles USMArray input. Puts SYCLQueue from data to decorated function arguments.
    Converts output of decorated function to dpctl.tensor/dpnp.ndarray if input was of this type.

    Parameters
    ----------
    freefunc (bool) : Set to True if decorates free function.
    queue_param (bool) : Set to False if the decorated function has no `queue` parameter

    Notes
    -----
    Queue will not be changed if provided explicitly```

This method 1) only takes a queue as input 2) returns a copy of the estimator object.  Neither of which require the use of support_usm_ndarray. Therefore this code is not necessary.

File: sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py [hunk 16]
```diff
@@ -0,0 +1,291 @@
...
+            [0.0, 5.0, 32.0],
+            [0.0, 6.0, 64.0],
+            [0.0, 7.0, 128.0],
+        ],
+        dtype=dtype,
+    )
+    dpt_data = _convert_to_dataframe(data, sycl_queue=queue, target_df=dataframe)
+
+    local_dpt_data = _convert_to_dataframe(
+        _get_local_tensor(data), sycl_queue=queue, target_df=dataframe
```
icfaust, line 69: Please be explicit with dtype in _convert_to_dataframe as per the other comment

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,291 @@
...
+def test_incremental_basic_statistics_partial_fit_spmd_synthetic(
+    dataframe, queue, num_blocks, weighted, n_samples, n_features, dtype
+):
+    # Import spmd and batch algo
+    from sklearnex.basic_statistics import IncrementalBasicStatistics
+    from sklearnex.spmd.basic_statistics import (
+        IncrementalBasicStatistics as IncrementalBasicStatistics_SPMD,
+    )
+
+    tol = 2e-3 if dtype == np.float32 else 1e-7
```
icfaust, line 254: Woah, is difference in tolerance 4 orders of magnitude? If so then please make a ticket for investigation, this is really really bad.

icfaust (commented):
Follow up changes associated with tolerance and dtyping, please also see my note on support_usm_ndarray.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
olegkkruglov: but there are a lot of other places in the repo where it is not specified and it works well. I still think that verifying of _convert_to_dataframe work is not the goal of this pr and should be resolved separately

onedal/spmd/basic_statistics/incremental_basic_statistics.py
ethanglaser: Good catch

bdmoore1 (approved):

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
olegkkruglov: ticket 8350

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
icfaust: There is already a circumstance in this PR where variation in the datatype has wide impact on the accuracy of the resultant data.  That could be due to an issue in oneDAL, sklearnex interface, and/or within the tests themselves.  By explicitly typing your conversions, you are being clear to the user, developer, and reviewer the chain of a vital parameter in testing and saving effort for the future ticket.  There is good logical reasoning as to why I am asking for this, precedence is an insufficient argument in this case.

icfaust (changes requested):
I request that a 3rd party weigh in on the dtype conversion discussion @samir-nasibli @Alexsandruss .  I believe I have good logical grounds for this request, and until a sufficient agreement is met, I do not want this PR merged.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
samir-nasibli: Thank you guys. I understand your ideas, but as I see `option` here is a string value, so in this case valuable and we and leave it as is. If there are any other better ways to implement, please @ethanglaser @icfaust suggest

icfaust: I guess its fine, assert_allclose has an err_msg keyword, please make some sort of string using ```option``` such that if an assert fails, we know which attribute is causing it.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
samir-nasibli: > could you add `assert(dpt_data.dtype == dtype)` somewhere later in your tests, since dtype isn't being passed to _convert_to_dataframe, I would like to verify that the conversion hasn't corrupted the dtype. Either that or explicitly pass the dtype as a kwarg.

It is not recommended to explicitly add dtype, shape or other kwargs for the _convert_to_dataframe. It is expected that it will work as is, since all primitives inside uses input object metadata and attributes. Actually passing numpy dtype as a dtype param can break here for non numpy dataframes.
I think, this is not a scope for the current PR. We can separately cover with tests `_convert_to_dataframe` for this kind of checks (just needed a ticket now). @icfaust if there some bugs that you investigated related this please also report it as a bug for future investigation and work.

icfaust: > > could you add `assert(dpt_data.dtype == dtype)` somewhere later in your tests, since dtype isn't being passed to _convert_to_dataframe, I would like to verify that the conversion hasn't corrupted the dtype. Either that or explicitly pass the dtype as a kwarg.
> 
> It is not recommended to explicitly add dtype, shape or other kwargs for the _convert_to_dataframe. It is expected that it will work as is, since all primitives inside uses input object metadata and attributes. Actually passing numpy dtype as a dtype param can break here for non numpy dataframes. I think, this is not a scope for the current PR. We can separately cover with tests `_convert_to_dataframe` for this kind of checks (just needed a ticket now). @icfaust if there some bugs that you investigated related this please also report it as a bug for future investigation and work.

Please see the codebase: https://github.com/intel/scikit-learn-intelex/blob/main/onedal/tests/utils/_dataframes_support.py#L139 Issues related to dtype changing  have already been addressed, and are being used in testing all estimators and methods as per test_patching.py . There is a known issue that I have had @olegkkruglov make a ticket for here: https://github.com/intel/scikit-learn-intelex/pull/1961#discussion_r1734356127 which is correlated to the dtype of the input.  Since I am outvoted, you can remove my changes requested, but I only ask that someone else beyond @bdmoore1 provides an approval before merging.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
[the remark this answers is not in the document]
icfaust: Sounds good, please add a comment to the code that the values are intentional, and how the relate to the gold data. The values seemed arbitrary to me.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
olegkkruglov: done

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -260,6 +271,7 @@ def test_incremental_basic_statistics_partial_fit_spmd_synthetic(
     split_data = np.array_split(data, num_blocks)
 
     if weighted:
+        # Create weights array containing the weight for each sample in the data
```
icfaust, line 274: Why is 0-7 chosen? Its clear you are adding a weight for each sample in the data, but why are those values chosen? @olegkkruglov

olegkkruglov: do you have better option? it's just an example, why did this even catch your attention? does that mean that there are no more serious issues?

icfaust: If its this, then I think we may be misusing it: https://www.testenvironmentmanagement.com/what-are-test-data-gold-copies/

olegkkruglov: for me gold data is just small dataset making the final answer and intermediate results relatively easy to calculate and verify separately and independently in case if something is going wrong

ethanglaser: To me it seems fine for this PR. To be more "gold" maybe it would make sense to use an array of 0s and 1s for simplicity but this should also be fine. A broader refactor of spmd algos can be done after the incrementals are merged where we centralize some dataset generation logic but as a standalone PR I have no objections.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
samir-nasibli: I vote also for checking this.
I suggest:
1) I think it would be better just to add some asserts after `_convert_to_dataframe` call to check that returned object has the same dtype. But it also somewhere should be refactored and added for other algos.
2) Also good to have a ticket for covering the `_convert_to_dataframe` with unit tests.


Generally, I think we should avoid explicit shape, dtype, or other attr providing. I expect that all dataframes construct methods called for the input object handling all attributes and metadate, but make sense to check as proposed on 2).

sklearnex/tests/_utils_spmd.py
```diff
@@ -111,6 +111,13 @@ def _generate_clustering_data(
     return X_train, X_test
 
 
+def _generate_weights(n_samples, dtype=np.float64, random_state=42):
+    # Generates weights
+    gen = np.random.default_rng(random_state)
+    weights = gen.uniform(low=-0.3, high=+0.7, size=(n_samples)).astype(dtype)
+    return weights
```
samir-nasibli, line 118: Please reuse `_generate_statistic_data`, and avoid code duplication. Just update `_generate_statistic_data` primitive like this:


```python
def _generate_statistic_data(n_samples, n_features=None, dtype=np.float64, random_state=42):
    # Generates statistical data
    gen = np.random.default_rng(random_state)
    size = (n_samples, n_features) if n_features else (n_samples)
    data = gen.uniform(low=-0.3, high=+0.7, size=size).astype(dtype)
    return data
```

onedal/spmd/basic_statistics/incremental_basic_statistics.py
```diff
@@ -0,0 +1,72 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from daal4py.sklearn._utils import get_dtype
+
+from ...basic_statistics import (
+    IncrementalBasicStatistics as IncrementalBasicStatistics_nonSPMD,
```
samir-nasibli, line 20: There is another option with Base prefix or suffix. But definitely `nonSPMD` should be changed
```suggestion
    IncrementalBasicStatistics as IncrementalBasicStatistics_Base,
```

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
samir-nasibli: `_convert_to_dataframe` this will not work for array api for example, array api case should be updated in `_convert_to_dataframe`.
Even explicitly provided dtype doesn't guarantee that this will be the same dtpye. I think, asserts after `_convert_to_dataframe` more preferable.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
```diff
@@ -0,0 +1,307 @@
...
+
+@pytest.mark.skipif(
+    not _mpi_libs_and_gpu_available,
+    reason="GPU device and MPI libs required for test",
+)
+@pytest.mark.parametrize(
+    "dataframe,queue",
+    get_dataframes_and_queues(dataframe_filter_="dpnp,dpctl", device_filter_="gpu"),
+)
+@pytest.mark.parametrize("weighted", [True, False])
```
samir-nasibli, line 41: Just add indexes `weighted` and `non_weighted` for better logs.

samir-nasibli (commented):
Please provide latest green CI run after all comments addressed. Generally, looks to me good. Just minor comments should be addressed. I am ok to do some follow-up work after the merge

samir-nasibli:
Please rebase and run internal CI as well

icfaust (approved):
Approval contingent on:

@ethanglaser 's weigh-in on this comment: https://github.com/intel/scikit-learn-intelex/pull/1961/files#r1732104503

Implementing @samir-nasibli 's suggestions from his review (negotiate with him on which to implement/not implement)

@ethanglaser 's suggestion to have a follow-up on the gold data DRY issue https://github.com/intel/scikit-learn-intelex/pull/1961#discussion_r1736173958()

@ethanglaser 's suggestion to have a follow-up refactor of spmd (https://github.com/intel/scikit-learn-intelex/pull/1961#discussion_r1732987805)

@samir-nasibli 's suggestion to have testing implemented for ```_convert_to_dataframe``` and adding assert statements (https://github.com/intel/scikit-learn-intelex/pull/1961#discussion_r1736176039)

When we have these odds and ends complete then we can merge.  I can write the ticket for @samir-nasibli 's suggestion for ```_convert_to_supported``` testing, would @olegkkruglov or @ethanglaser write the tickets for the other two?  I think this should be doable today.

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
olegkkruglov: i still can't get why should it be only here and not in tons of other occurrences of this call in our repo. in my understanding it will become completely unnecessary after we test _convert_to_dataframe separately which is already planned. moreover, dtype check of the  _convert_to_dataframe result is not straightforward as far as i understand because the returning value may have different types depending on arguments of convert_to_dataframe. so it would lead to perceptible test size increase which i don't like and don't think it should be here

olegkkruglov:
> @ethanglaser 's suggestion to have a follow-up on the gold data DRY issue [#1961 (comment)](https://github.com/intel/scikit-learn-intelex/pull/1961#discussion_r1736173958)
> 
> @ethanglaser 's suggestion to have a follow-up refactor of spmd ([#1961 (comment)](https://github.com/intel/scikit-learn-intelex/pull/1961#discussion_r1732987805))

ticket 8352 covers both

sklearnex/spmd/basic_statistics/tests/test_incremental_basic_statistics_spmd.py
samir-nasibli: Could be done on follow up refactoring. Non critical
```python
@pytest.mark.parametrize("weighted", [True, False], ids=["weighted", "non_weighted"])
```

samir-nasibli (approved):
Expecting all follow up tickets addressed. Thank you for the work done!

PR #1974, opened 2024-07-30
[fix] solve various windows onedal build warnings

icfaust:
We have a number of these errors in the compilation stage in windows for onedal: ```warning C4101: 'e': unreferenced local variable```

Add static casts to double type where implicit conversion had been previously done.

Following the guide of this StackOverflow post: https://stackoverflow.com/questions/24428146/catching-exception-with-unreferenced-local-variable-warning by casting to void, it will tell the compiler it is unused, but will not impact the output machine code.

Fix belt-and-suspenders setting of variable to ```const std::int64_t``` to ```int``` as required by numpy interfaces.

Remove unused variables and type aliases (icx compiler complains on windows)


onedal/datatypes/data_conversion.cpp
```diff
@@ -237,7 +237,7 @@ template <int NpType, typename T = byte_t>
 static PyObject *convert_to_numpy_impl(const dal::array<T> &array,
                                        std::int64_t row_count,
                                        std::int64_t column_count = 0) {
-    const std::int64_t size_dims = column_count == 0 ? 1 : 2;
+    int size_dims = column_count == 0 ? 1 : 2;
```
icfaust, line 240: Required to be int type here: https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_SimpleNewFromData

ethanglaser: Can it still be `const` ?

ethanglaser (approved):

PR #1992, opened 2024-08-07
[enh] Conda recipe update and CI OS version pinning

Alexsandruss:
### Description

Scikit-learn-intelex conda recipe update.

Changes:
- Update content of recipe, building and testing scripts to sync with conda-forge feedstock
- Add recipe building on Linux and Windows to CI
- Pin CI images versions
- Add hint for CMake to get python executable from $PYTHON if set

Conda-forge checks: https://github.com/conda-forge/scikit-learn-intelex-feedstock/pull/38

---

**Testing**

**Performance**


scripts/build_backend.py
```diff
@@ -64,7 +64,7 @@ def custom_build_cmake_clib(
 
     if iface in ["dpc", "spmd_dpc"]:
         if IS_WIN:
-            cxx = "icx"
+            cxx = "icx-cl"
```
icfaust, line 67: how come icx-cl? so far my builds with icx seem to work...

conda-recipe/run_test.sh
```diff
@@ -15,54 +15,51 @@
...
     fi
-    daal4py_dir="$( dirname "${daal4py_dir}" )"
+    sklex_root="$( dirname "${sklex_root}" )"
     count=$(($count - 1))
 done
 
-echo "daal4py_dir=$daal4py_dir"
 if [[ count -eq 0 ]]; then
-    echo "run_test.sh must be in daal4py repository"
+    echo "run_test.sh did not find required testing directories"
```
icfaust, line 29:
```suggestion
    echo "run_test.sh did not find the required testing directories"
```

conda-recipe/meta.yaml
```diff
@@ -14,96 +14,104 @@
...
+  requires:
+    - pandas
+    - scikit-learn
+    - xgboost
+    - lightgbm
+    - catboost
+    - pytest
+    - pyyaml
+    - mpich  # [not win]
+    # - impi_rt  # [win]
```
icfaust, line 83: How is this list verified?

conda-recipe/meta.yaml
```diff
@@ -14,96 +14,104 @@
...
-    - DPCPPROOT
-    - DALROOT
-    ignore_run_exports:
-       - python
-       - mpich
+  skip: true  # [not (linux64 or win)]
+  number: {{ buildnumber }}
+  include_recipe: False
+  script_env:
+  - DPCPPROOT
```
icfaust, line 39: When I search sklearnex for ```DPCPPROOT``` I now only find it in the conda-recipe (no longer in build steps). Is this something we need, and if so, can we give a better description somewhere to a user?

conda-recipe/conda_build_config.yaml
```diff
@@ -15,4 +15,8 @@
 #===============================================================================
 
 numpy:
-  - 1.19
+  - 1.24
+c_compiler:    # [win]
+- vs2022       # [win]
```
icfaust, line 20: Private CI Visual Studio doesn't utilize this compiler, though public CI does. I think it worth having a discussion how we verify this support on private CI-side thoughts? @homksei @ethanglaser

.ci/pipeline/nightly.yml
```diff
@@ -34,7 +34,7 @@ variables:
 jobs:
 - job: Coverity
   pool:
-    vmImage: 'ubuntu-latest'
+    vmImage: 'ubuntu-22.04'
```
icfaust, line 37: This should be in a separate PR.  Forcing the VM OS for Coverity/Jupyter checks doesn't seem related and isn't mentioned in the description. Doesn't seem related to the conda build. (Will fast track review when split off).

.ci/pipeline/build-and-test-win.yml
```diff
@@ -51,7 +51,7 @@ steps:
   - script: |
       call activate CB
       cd ..
-      call s\conda-recipe\run_test.bat s
+      call s\conda-recipe\run_test.bat s\
```
icfaust, line 54: Interesting. For my edification, could you explain why this change is necessary (may impact a PR of mine)

icfaust (commented):
Overall pretty useful addition. I would beef up the description, separate out some of the other changes, and possibly drop sklearn 1.1 or 1.2 from the standard test matrix.  The reason being is that by adding these two recipes, we are starting to really push our luck on azure pipeline runners. I'll let others weigh in on this @samir-nasibli @ethanglaser .

.ci/pipeline/build-and-test-win.yml
Alexsandruss: It is to save path compatibility for custom path and non-set path. Our CI runs tests from `s` directory while conda recipe builder - from default path.

.ci/pipeline/nightly.yml
Alexsandruss: It is actually mentioned: `- Pin CI images versions`.
As CI has update in this PR anyway, I thought it's convenient to include pinning as no actual change in used images is happening.

scripts/build_backend.py
Alexsandruss: No actual reason or difference, it was added during debugging. I'm going to remove it.

conda-recipe/meta.yaml
Alexsandruss: It is needed to pass external DPC compiler for `conda-build`, I added docs section for this.

conda-recipe/conda_build_config.yaml
Alexsandruss: I don't see any need to re-specification of `vs` version until we are enable building of conda package in internal CI: azure CI images and most of users use vs2022 anyway

conda-recipe/meta.yaml
Alexsandruss: What do you mean as the verification in this case?

bdmoore1 (approved):

conda-recipe/meta.yaml
icfaust: I guess this looks to be another list similar to dependencies-dev, just worried about the maintenance of it. Not a big deal, just something I was thinking about, and not sure how it was getting maintained vs dependencies-dev.

conda-recipe/conda_build_config.yaml
icfaust: Seems reasonable.  I will still poke at private CI people as to why we are using an outdated vs version since most are using 2022 as you mentioned.

.ci/pipeline/nightly.yml
icfaust: I went on ahead and changed the title to make sure this was also there, I think splitting off to a separate PR is a waste of time, just as long as it doesn't get lost in the shuffle.

conda-recipe/run_test.bat
```diff
@@ -15,34 +15,18 @@ rem See the License for the specific language governing permissions and
...
-    call "%TBBROOT%\env\vars.bat" || set exitcode=1
-)
+%PYTHON% -m unittest discover -v -s %1%tests -p test*.py || set exitcode=1
 
-%PYTHON% -m unittest discover -v -s %1\tests -p test*.py || set exitcode=1
-
-pytest --verbose --pyargs %1\daal4py\sklearn || set exitcode=1
+pytest --verbose --pyargs daal4py || set exitcode=1
 pytest --verbose --pyargs sklearnex || set exitcode=1
-pytest --verbose --pyargs %1\onedal --deselect="onedal/common/tests/test_policy.py" || set exitcode=1
```
icfaust, line 46: Does private CI use run_test.bat? I know that the public windows doesn't build the dpc portion and test against dpctl/dpnp so some of these tests aren't verified against public CI.  May be worth running an /intelci: run and checking the testing. It looks like it was deselected in your #1115 but there isn't a comment as to why.

icfaust (commented):
Looks to be some sort of windows python conda issue. Your previous responses were sufficient for me. I found one issue with a re-enablement that may require running private CI, maybe some clarity could explain it away as an alternate solution. If we solve these small issues, I think it would be ready.

samir-nasibli:
@Alexsandruss please rebase your branch.

conda-recipe/run_test.bat
Alexsandruss: Private CI doesn't use run_test.bat, and I don't remember reason why deselection was introduced. Anyway, it works now.

conda-recipe/meta.yaml
Alexsandruss: I'm afraid there will be at least two files with dependencies: one for conda-recipe and others for non-conda build. Currently, all of them are tested in CIs.

INSTALL.md
```diff
@@ -212,6 +212,27 @@ back to the project source-code directory. That way, you can edit the source cod
 without reinstalling the package after a small change.
 * `--single-version-externally-managed` is an option for Python packages instructing the setup tools module to create a package the host's package manager can easily manage.
 
+## Build from Sources with `conda-build`
+
+Intel(R) Extension for Scikit-learn* is easily built from the sources using only one command and `conda-build` utility. 
+
+### Prerequisites for `conda-build`
+
+* any `conda` distribution (`miniforge` is preferrable)
```
icfaust, line 221:
```suggestion
* any `conda` distribution (`miniforge` is recommended)
```
spelling error in ```preferable```

conda-recipe/meta.yaml
```diff
@@ -14,96 +14,105 @@
...
+    - numpy {{ numpy }}
+    - mpich  # [not win]
+    # - impi-devel  # [win]
+    # dal-devel pinning depends on the recipe location (repo or feedstock)
+    - dal-devel
+    # - dal-devel =={{ version }}
+  run:
+    - python
+    - {{ pin_compatible('numpy') }}
+    - dpcpp-cpp-rt  # [linux64]
```
icfaust, line 69: Only dpc support on Linux (if I am understanding this correct) for the conda meta?

icfaust (commented):
So far its okay to have Ubuntu-22, but we should have a ticket in the backlog to check for the Ubuntu 24 drop for azure pipelines.  Or do we have renovatebot checking for this?

conda-recipe/meta.yaml
Alexsandruss: Yes, only linux, there are unresolved issues with dpc building on Windows using conda-forge dpcpp compiler package.

INSTALL.md
Alexsandruss: Corrected.

conda-recipe/meta.yaml
```diff
@@ -14,96 +14,105 @@
...
-        - tests
-        - daal4py/sklearn
-    commands:
-        - python -c "import daal4py"
-        - mpirun -n 4 python -m unittest discover -v -s tests -p test*spmd*.py # [not win]
-        - mpiexec -localonly -n 4 python -m unittest discover -v -s tests -p test*spmd*.py # [win]
-        - python -m unittest discover -v -s tests -p test*.py
-        - pytest --pyargs daal4py/sklearn/
-        - python tests/run_examples.py
+  requires:
```
icfaust, line 75: Actually now that I think of it, is ```tests:``` evaluated at all in ci?  I see that the azure pipelines job will build it, but do we try to evaluate the testing side of meta.yml?  If no, then I wouldn't do the full testing suite. To go off my previous review comment, I think model builders would best test the ```requres:``` list.

INSTALL.md
```diff
@@ -212,6 +215,27 @@ back to the project source-code directory. That way, you can edit the source cod
...
+Intel(R) Extension for Scikit-learn* is easily built from the sources using only one command and `conda-build` utility. 
+
+### Prerequisites for `conda-build`
+
+* any `conda` distribution (`miniforge` is recommended)
+* `conda-build` and `conda-verify` installed in a conda environment
+* (Windows only) Microsoft Visual Studio* 2022
+* (optional) Intel(R) oneAPI DPC++/C++ Compiler
+
+In order to enable DPC++ interfaces support, you need to set `DPCPPROOT` environment variable pointing to DPC++/C++ Compiler distribution.
```
icfaust, line 229: Please add commentary here about the divergence of windows and linux for dpc support. Also maybe MPI depending on the discussion in the other review comment.

conda-recipe/meta.yaml
icfaust: Seems reasonable, I guess that is a bit the reason to add a specific CI job for it. thank you,

conda-recipe/meta.yaml
```diff
@@ -14,96 +14,105 @@
...
+    - python
+    - setuptools
+    - cmake
+    - clang-format
+    - cython
+    - jinja2
+    - pybind11
+    - numpy {{ numpy }}
+    - mpich  # [not win]
+    # - impi-devel  # [win]
```
icfaust, line 62: I don't know much about the mpi-side of things. You have proposed switching the linux support from ```impi``` to ```mpich```.  Could you check with @ethanglaser about these things (out of my knowledge).  Also mpi support for windows is dropped (and is in a comment that may need to be deleted). Is this something to be readded in the future? Also, not to be annoying: Given that dpcpp_cpp_rt was on the ```run:``` side of things, did we have a regression in windows support via this at some point (i.e. did we support dpc building via conda for windows before, or was it just not properly tested, and was erroneously added to ```run:```)?

icfaust (commented):
Bit of a deeper-dive review. I think with these comments addressed, you'll get my approval.

.ci/pipeline/ci.yml
```diff
@@ -57,6 +57,57 @@ jobs:
...
+      bash .ci/scripts/install_dpcpp.sh
+    displayName: "dpcpp installation"
+  - script: |
+      conda config --add channels conda-forge
+      conda config --set channel_priority strict
+      conda update -y -q --all
+    displayName: "Conda update"
+  - script: |
+      conda create -y -q -n build-env python=$(PYTHON_VERSION) conda-build conda-verify
+    displayName: "Conda install"
```
samir-nasibli, line 80: Please address message from the logs.
The build conda recipe pipeline shows warning 
```
Free memory is lower than 5%; Currently used: 95.03%
```

.ci/pipeline/ci.yml
```diff
@@ -57,6 +57,57 @@ jobs:
...
+  - script: |
+      export DPCPPROOT=/opt/intel/oneapi/compiler/latest
+      . /usr/share/miniconda/etc/profile.d/conda.sh
+      conda activate build-env
+      conda build .
```
samir-nasibli, line 85: It make sense build with no tests (or just run package import test) and run functional tests separately in other step. What do you think?

samir-nasibli (commented):
initial review done.

.ci/pipeline/ci.yml
Alexsandruss: Doesn't make sense: more steps and script lines in CI, reading logs doesn't become easier anyway.

conda-recipe/meta.yaml
napetrov: there should be split between GPU mpi support and CPU mpi support.  As well as Intel build of scikit vs public one. 

Overall mpich and mpi are compatibe

ethanglaser: It's a good point - is there a reason we have mpich here instead of impi_rt/impi-devel?

conda-recipe/run_test.sh
ethanglaser, line 1: This looks much better, thanks for doing this

conda-recipe/meta.yaml
```diff
@@ -14,96 +14,100 @@
...
-        - python tests/run_examples.py
+  requires:
+    - pandas
+    - scikit-learn
+    - xgboost
+    - lightgbm
+    - catboost
+    - pytest
+    - pyyaml
+    - impi_rt  # [not win]
```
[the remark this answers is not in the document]
Alexsandruss: I updated conda recipe to sync with requirements-test.txt.
dpctl/dpnp are not going to be included since they are absent from conda-forge channel.

conda-recipe/meta.yaml
```diff
@@ -14,96 +14,105 @@
...
+    - pandas
+    - scikit-learn
+    - xgboost
+    - lightgbm
+    - catboost
+    - pytest
+    - pyyaml
+    - mpich  # [not win]
+    # - impi_rt  # [win]
+  source_files:
```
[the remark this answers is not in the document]
Alexsandruss: I removed `daal4py/sklearn`, it's not necessary at this point. Model builders tests are contained in `tests` and included.

conda-recipe/meta.yaml
Alexsandruss: `conda-build` seems to prefer `run_test.bat / run_test.sh` over test commands in recipe, so full testing is included in conda-recipe CI by default.

conda-recipe/meta.yaml
Alexsandruss: I reverted building and runtime deps to use impi.
Feedstock builds of sklearnex never included dpc part, so there is no regression in support on Windows.

INSTALL.md
Alexsandruss: I added OS specific details for DPC compiler in INSTALL.md.

conda-recipe/meta.yaml
icfaust: @Alexsandruss thank you for the note, this is what impacts #2107 and #2104 @david-cortes-intel

.ci/pipeline/ci.yml
```diff
@@ -57,6 +57,65 @@ jobs:
...
+  dependsOn: Lint
+  timeoutInMinutes: 120
+  strategy:
+    matrix:
+      Python3.12_Numpy1.26:
+        PYTHON_VERSION: '3.12'
+        NUMPY_VERSION: '1.26'
+      Python3.9_Numpy1.22:
+        PYTHON_VERSION: '3.9'
+        NUMPY_VERSION: '1.22'
```
icfaust, line 101: So hardcoding the numpy versions is going to increase maintenance burden, is there a reason why they are pinned, and why pinned to < 2.0?

.ci/pipeline/ci.yml
```diff
@@ -57,6 +57,65 @@ jobs:
...
+  - script: |
+      conda create -y -q -n build-env conda-build conda-verify
+    displayName: "Conda create"
+  - script: |
+      export DPCPPROOT=/opt/intel/oneapi/compiler/latest
+      . /usr/share/miniconda/etc/profile.d/conda.sh
+      conda activate build-env
+      conda build . --python $(PYTHON_VERSION) --numpy $(NUMPY_VERSION)
+    displayName: "Conda build"
+- job: WindowsCondaRecipe
```
icfaust, line 91: Where is the run_test.sh occur here? (sorry if its a dumb question, I don't know much about the conda building testing standard procedures) (given https://github.com/intel/scikit-learn-intelex/pull/1992/files#r1801556134)

icfaust (commented):
Just some last questions from me. Thanks for the updates based off the previous comments.

david-cortes-intel:
Question: are the commands in `meta.yaml` for executing tests actually needed if it will end up calling the `run_test.sh/bat` scripts in the end?

.ci/pipeline/ci.yml
samir-nasibli: @icfaust https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#test-section
If this section exists or if there is a `run_test.[py,pl,sh,bat,r]` file in the recipe, the package is installed into a test environment after the build is finished and the tests are run there.

About dpnp/dpctl test skipping. I think it should be noted somewhere and tracked when it could be enabled.

.ci/pipeline/ci.yml
```diff
@@ -57,6 +57,65 @@ jobs:
...
+  dependsOn: Lint
+  timeoutInMinutes: 120
+  strategy:
+    matrix:
+      Python3.12_Numpy1.26:
+        PYTHON_VERSION: '3.12'
+        NUMPY_VERSION: '1.26'
+      Python3.9_Numpy1.22:
+        PYTHON_VERSION: '3.9'
+        NUMPY_VERSION: '1.22'
```
samir-nasibli, line 70: Any justification for pinning such numpy versions?

.ci/pipeline/ci.yml
```diff
@@ -57,6 +57,65 @@ jobs:
...
+    displayName: "Conda update"
+  - script: |
+      conda create -y -q -n build-env conda-build conda-verify
+    displayName: "Conda create"
+  - script: |
+      export DPCPPROOT=/opt/intel/oneapi/compiler/latest
+      . /usr/share/miniconda/etc/profile.d/conda.sh
+      conda activate build-env
+      conda build . --python $(PYTHON_VERSION) --numpy $(NUMPY_VERSION)
+    displayName: "Conda build"
```
samir-nasibli, line 90: Since there an option only build without testing I suggest add `test` in the name of the job for better description of the flow.
```suggestion
    displayName: "Conda build and test"
```

.ci/scripts/gen_release_jobs.py
```diff
@@ -24,7 +24,7 @@
 
 CHANNELS = args.channels
 PYTHON_VERSIONS = ["3.9", "3.10", "3.11"]
-SYSTEMS = ["ubuntu-latest", "windows-latest"]
+SYSTEMS = ["ubuntu-22.04", "windows-2022"]
```
samir-nasibli, line 27: I think it worth to add some comment for better understanding and justification why not the latest versions is used.

conda-recipe/meta.yaml
```diff
@@ -14,96 +14,103 @@
 # limitations under the License.
 #===============================================================================
 
-{% set version = '2024.1' %}
+# NB: this recipe should be synced between sklearnex repo and feedstocks
+
+{% set name = "scikit-learn-intelex" %}
+{% set version = "2025.1" %}
```
samir-nasibli, line 20: I suggest to use git env variables here
https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#git-environment-variables

conda-recipe/meta.yaml
```diff
@@ -14,96 +14,103 @@
...
+    - cmake
+    - clang-format
+    - cython
+    - jinja2
+    - pybind11
+    - numpy {{ numpy }}
+    - impi-devel  # [not win]
+    # dal-devel pinning depends on the recipe location (repo or feedstock)
+    - dal-devel
+    # - dal-devel =={{ version }}
```
samir-nasibli, line 63: Maybe for the current iteration it is fine, but for a such project like `scikit-learn-intelex` I expect that it will be enhanced by using `pyproject.toml`, where requirements defined in one place. It will require to update `pyproject.toml` file, `[build-system]` table is recommended https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#writing-your-pyproject-toml

conda-recipe/conda_build_config.yaml
```diff
@@ -15,4 +15,8 @@
 #===============================================================================
 
 numpy:
-  - 1.19
+  - 1.24
```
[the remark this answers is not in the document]
david-cortes-intel: And actually on second thought: since these are **build-time** dependencies rather than runtime, perhaps would be better to switch them to numpy>=2, since a module built with numpy>=2 is usable with a runtime numpy<2, but not the other way around.

.ci/pipeline/ci.yml
Alexsandruss: Current numpy handling is similar to `dependencies-dev` and `requirements-test.txt`: lowest possible numpy version is set for building and `{{ pin_compatible(numpy)) }}` installs highest compatible version for testing. The only difference is `conda-build` ignores `numpy>=2.0`

.ci/pipeline/ci.yml
Alexsandruss: I added note about dpnp/dpctl.

conda-recipe/conda_build_config.yaml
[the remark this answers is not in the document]
Alexsandruss: I would like to leave numpy installation behavior similar to usual testing: https://github.com/intel/scikit-learn-intelex/pull/1992#discussion_r1804754817

.ci/pipeline/ci.yml
Alexsandruss: https://github.com/intel/scikit-learn-intelex/pull/1992#discussion_r1804754817

conda-recipe/conda_build_config.yaml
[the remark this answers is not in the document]
david-cortes-intel: @Alexsandruss Wouldn't that leave parts of the conda package broken if a user first install scikit-learn-intelex and then later on upgrades numpy to a version >= 2?

.ci/scripts/gen_release_jobs.py
Alexsandruss: Added.

conda-recipe/run_test.sh
```diff
@@ -15,55 +15,51 @@
...
     exit 1
 fi
 
-echo "Start testing ..."
 return_code=0
 
-python -c "import daal4py"
-return_code=$(($return_code + $?))
+if [ -z "${PYTHON}" ]; then
+    export PYTHON=python
```
david-cortes-intel, line 36: Is this one used anywhere when running on linux? I understood it was meant for the MPI invocation on windows only.

conda-recipe/run_test.bat
```diff
@@ -15,34 +15,18 @@ rem See the License for the specific language governing permissions and
 rem limitations under the License.
 rem ============================================================================
 
-rem %1 - scikit-learn-intelex repo root
+rem %1% - scikit-learn-intelex repo root (leave empty if it's PWD)
```
david-cortes-intel, line 18: Shouldn't this say "%cd%" since it's windows?

conda-recipe/run_test.bat
```diff
@@ -15,34 +15,18 @@ rem See the License for the specific language governing permissions and
...
 set exitcode=0
 
-IF DEFINED DPCPPROOT (
-    echo "Sourcing DPCPPROOT"
-    call "%DPCPPROOT%\env\vars.bat" || set exitcode=1
-    set "CC=dpcpp"
-    set "CXX=dpcpp"
-    dpcpp --version
-)
+IF NOT DEFINED PYTHON (set PYTHON="python")
```
david-cortes-intel, line 22:
```suggestion
IF NOT DEFINED PYTHON (set "PYTHON=python")
```

conda-recipe/run_test.bat
```diff
@@ -15,34 +15,18 @@ rem See the License for the specific language governing permissions and
...
 
-%PYTHON% -m pytest --verbose -s %1\tests || set exitcode=1
-
-pytest --verbose --pyargs %1\daal4py\sklearn || set exitcode=1
+pytest --verbose --pyargs daal4py || set exitcode=1
 pytest --verbose --pyargs sklearnex || set exitcode=1
-pytest --verbose --pyargs %1\onedal --deselect="onedal/common/tests/test_policy.py" || set exitcode=1
-pytest --verbose %1\.ci\scripts\test_global_patch.py || set exitcode=1
+pytest --verbose --pyargs onedal || set exitcode=1
+pytest --verbose %1%.ci\scripts\test_global_patch.py || set exitcode=1
```
david-cortes-intel, line 31: Wouldn't `%1%` be interpreted as an environment variable instead of CLI argument?

conda-recipe/run_test.sh
Alexsandruss: For `${PYTHON} -c "from sklearnex import patch_sklearn; patch_sklearn()"` at least. It might be required for some additional tests later, so no need to remove it.

conda-recipe/run_test.bat
Alexsandruss: `%1%` and `%1` works the same way here as I saw.

conda-recipe/run_test.bat
Alexsandruss: Done.

conda-recipe/run_test.bat
Alexsandruss: I added mention of both.

INSTALL.md
```diff
@@ -212,6 +215,28 @@ back to the project source-code directory. That way, you can edit the source cod
...
 
+## Build from Sources with `conda-build`
+
+Intel(R) Extension for Scikit-learn* is easily built from the sources using only one command and `conda-build` utility. 
+
+### Prerequisites for `conda-build`
+
+* any `conda` distribution (`miniforge` is recommended)
+* `conda-build` and `conda-verify` installed in a conda environment
+* (Windows only) Microsoft Visual Studio* 2022
```
napetrov, line 226: can we point to general VS/compiler requirements so we would have single place to track/update?

Alexsandruss: Not really in case of conda recipe, vs2022 is default for this repo and included in CI Windows image, vs2019 - for conda-forge feedstock, and build config change is required to set another version and there is no guarantee it will work as expected (but it is unlikely someone will use version other than 2019 or 2022).

I will add mention how to change compiler for conda-build on Win.

conda-recipe/run_test.bat
david-cortes-intel: Isn't that just a coincidence due to the way you placed the variable?

For example, you can try a small script like this:
```batch
echo %1% %2%
```

and another like this:
```batch
echo %1 %2
```

then call each with arguments like these:
```batch
file.bat arg1 arg2
```

and you'd see that once there is more than one argument or they are somewhere where an extra `%` wouldn't be ignored, it's not the same to use `%1%` instead of `%1`.

conda-recipe/run_test.sh
david-cortes-intel: Is `${PYTHON}` ever set to something other than "python" when executing the .sh script? If not, would be nice to add it as a comment in the file.

conda-recipe/meta.yaml
Alexsandruss: Added

conda-recipe/run_test.sh
Alexsandruss: It's used by `conda-build`: https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#environment-variables-set-during-the-build-process

conda-recipe/run_test.bat
Alexsandruss: Yes, it works as you described, I'll fix it

conda-recipe/run_test.sh
```diff
@@ -15,55 +15,51 @@
...
-echo "Pytest run of legacy unittest ..."
-echo ${daal4py_dir}
-pytest --verbose -s ${daal4py_dir}/tests
+${PYTHON} -c "from sklearnex import patch_sklearn; patch_sklearn()"
 return_code=$(($return_code + $?))
 
 echo "NO_DIST=$NO_DIST"
 if [[ ! $NO_DIST ]]; then
-    echo "MPI pytest run of legacy unittest ..."
+    echo "MPI pytest run"
```
david-cortes-intel, line 44: Why would this one have an `echo` entry while the `echo` entries of others were removed?

conda-recipe/run_test.sh
```diff
@@ -15,55 +15,51 @@
...
     mpirun --version
-    mpirun -n 4 pytest --verbose -s ${daal4py_dir}/tests/test*spmd*.py
+    mpirun -n 4 pytest --verbose -s ${sklex_root}/tests/test*spmd*.py
     return_code=$(($return_code + $?))
 fi
 
-echo "Pytest of daal4py running ..."
-pytest --verbose --pyargs ${daal4py_dir}/daal4py/sklearn
+pytest --verbose -s ${sklex_root}/tests
```
david-cortes-intel, line 50: I think it'd be quite helpful when inspecting failures to have this one execute before the MPI one, like it was before.

That way, if something fails and execution stops during the MPI part, one would then know whether the problem was with MPI specifically or with the tests in general.

david-cortes-intel (approved):

icfaust:
@Alexsandruss This was merged with an error in the WindowsCondaEnv Python3.12_Sklearn1.5 step

david-cortes-intel:
> @Alexsandruss This was merged with an error in the WindowsCondaEnv Python3.12_Sklearn1.5 step

The error appears to be sporadic though.

PR #1996, opened 2024-08-09
DEP: deprecate fallback to sklearn after onedal in LogReg

samir-nasibli:
### Description
deprecate fallback to sklearn after onedal in LogReg. There is no need for this in LogRegression estimator.

---

**Testing**

**Performance**


samir-nasibli:
@icfaust @ethanglaser @md-shafiul-alam @avolkov-intel @Alexsandruss Dear reviewers this PR pretty ready for the review and merge. Please take a look

ethanglaser:
Could you provide some additional context? I am not familiar with the original conditional

avolkov-intel:
What is the need for the removal? We have some places in oneDAL Log Reg implementation where we throw exceptions. We haven't observed such exceptions before but they potentially might happen. I think it is better to keep such fallback to ensure user will get the result computed.

icfaust:
> What is the need for the removal? We have some places in oneDAL Log Reg implementation where we throw exceptions. We haven't observed such exceptions before but they potentially might happen. I think it is better to keep such fallback to ensure user will get the result computed.

So far I see that the logistic Regression itself throws errors on malformed inputs (not sufficient in my opinion for the fallback). However, it also depends on the lapack and logloss/blas implementations (which use GEMV) which may also throw errors. If we don't expect errors to be thrown from logloss then the PR can move forward, otherwise I agree with @avolkov-intel that the PR should be closed.

icfaust:
@avolkov-intel @samir-nasibli status on this discussion?

samir-nasibli:
> @avolkov-intel @samir-nasibli status on this discussion?

I don't see any reason why we should not go forward, we are not expecting specific failures for LogRegression.


Even more if we are expecting some failures this kind of flows should be controlled by config params. So probably for LinReg (here https://github.com/intel/scikit-learn-intelex/pull/1907) make sense add to into config_contex params `sklearn_after_onedal`, smth like this.

samir-nasibli:
> What is the need for the removal? We have some places in oneDAL Log Reg implementation where we throw exceptions. We haven't observed such exceptions before but they potentially might happen. I think it is better to keep such fallback to ensure user will get the result computed.

If we have some exceptions, so it is better  to catch them and investigate. Exact cases should be properly dispatched, not after onedal backend. So this reason it make sense to remove it here. At least for Log Reg it is safe, both of CIs are green. So I strongly prefer to go with changes.

Alexsandruss (changes requested):
LogisticRegression is less faulty than Linear but common approach for all algorithms with runtime errors is required before making any changes. I guess PCA is applicable here too.

samir-nasibli:
> LogisticRegression is less faulty than Linear but common approach for all algorithms with runtime errors is required before making any changes. I guess PCA is applicable here too.

I will combine these changes with #1907 and bring common solution

samir-nasibli:
Closing this PR. Please follow proposals on #1907

PR #1998, opened 2024-08-11
FIX: incremental  estimators tests

samir-nasibli:
### Description

fixes incremental estimators tests.
Output array should be converted to numpy array before setting tol dependent on the dtype, otherwise the tol settings will not be set. Found on #1861 

---

**Testing**

**Performance**


icfaust (commented):
I tried to suggest this when it was first implemented, but was overridden.  Rather than cast it to numpy for checking dtype, couldn't we just use the dtype test parameter? i.e.
``` tol = 2e-6 if dtype == np.float32 else 1e-7```?

**Comment by samir-nasibli** [comment]
> Rather than cast it to numpy for checking dtype, couldn't we just use the dtype test parameter? i.e. ` tol = 2e-6 if dtype == np.float32 else 1e-7`?

All asserts require numpy inputs, so make sense convert them to numpy ndarrays.

File: sklearnex/preview/decomposition/tests/test_incremental_pca.py [hunk 1]
```diff
@@ -200,7 +200,7 @@ def test_sklearnex_partial_fit_on_gold_data(dataframe, queue, whiten, num_blocks
 
     X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
     transformed_data = incpca.transform(X_df)
-    check_pca_on_gold_data(incpca, dtype, whiten, transformed_data)
+    check_pca_on_gold_data(incpca, dtype, whiten, _as_numpy(transformed_data))
```
olegkkruglov, line 203: similar comment here, it's better to use dtype arg from check_pca_on_gold_data in the function for tolerance setting, it allows to avoid additional _as_numpy call

olegkkruglov: _as_numpy call already exists inside the method, only tolerance setting is necessary to be updated using dtype arg

olegkkruglov:
> Rather than cast it to numpy for checking dtype, couldn't we just use the dtype test parameter? i.e. ` tol = 2e-6 if dtype == np.float32 else 1e-7`?

agree, this is exactly what I asked in my comments

olegkkruglov (approved):
LGTM, assuming green CI

samir-nasibli:
Last CI run: http://intel-ci.intel.com/ef691edb-824e-f15b-993d-a4bf010d0e2e

samir-nasibli:
@mergify backport rls/2024.7.0-rls

PR #2027, opened 2024-09-04
Increase tolerance in sparse logistic regression test

avolkov-intel:
### Description

Lighten the relative tolerance requirement in sparse logistic regression test.
Follow up for https://github.com/intel/scikit-learn-intelex/pull/2013

---

**Testing**

**Performance**

### DIFF
--- a/onedal/linear_model/tests/test_logistic_regression.py
+++ b/onedal/linear_model/tests/test_logistic_regression.py
@@ -89,7 +89,7 @@ if daal_check_version((2024, "P", 700)):
         model_sp.fit(X_sp, y, queue=queue)
         pred_sp = model_sp.predict(X_sp, queue=queue)
 
-        rtol = 2e-4 if dtype == np.float32 else 1e-5
+        rtol = 2e-4
         assert_allclose(pred, pred_sp, rtol=rtol)
         assert_allclose(model.coef_, model_sp.coef_, rtol=rtol)
         assert_allclose(model.intercept_, model_sp.intercept_, rtol=rtol)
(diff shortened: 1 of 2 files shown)


olegkkruglov (approved):
LGTM, but I'm a little bit aware about big tolerance for float64, please take a look if possible

onedal/linear_model/tests/test_logistic_regression.py
```diff
@@ -89,6 +89,7 @@ def test_csr(queue, dtype, dims):
         model_sp.fit(X_sp, y, queue=queue)
         pred_sp = model_sp.predict(X_sp, queue=queue)
 
-        assert_allclose(pred, pred_sp)
-        assert_allclose(model.coef_, model_sp.coef_)
-        assert_allclose(model.intercept_, model_sp.intercept_)
+        rtol = 2e-4 if dtype == np.float32 else 1e-5
```
ethanglaser, line 92: Are there any fails at float64 level? Because if 1e-7 works without fail for that, then this value would be ideal

olegkkruglov: Anatoly said there are fails if the tolerance is less when I asked him. probably worth separate investigating

avolkov-intel: There were some sporadic failures with difference 2e-6

ethanglaser (approved):

onedal/linear_model/tests/test_logistic_regression.py
avolkov-intel: Seems it is not enough, the tolerance should be increased to 2e-4 for both floating point types. Definitely would need to further investigate it

Issue #2016, opened 2024-08-27
Different results from xgboost when using daal4py converted model 


kuza55:
**Describe the bug**
Different results from xgboost when using daal4py converted model 

**To Reproduce**

Use the attached model: 
[model.zip](https://github.com/user-attachments/files/16769910/model.zip)

```
import numpy as np
import xgboost as xgb
import daal4py as d4p

model = xgb.Booster()
model.load_model(model_path)
daal_model = d4p.mb.convert_model(model)

inputs = np.array([[0.0000e+00, 0.0000e+00, 0.0000e+00, 0.0000e+00, 0.0000e+00, 0.0000e+00,
2.1400e+02, 1.0752e+04, 1.0100e+03, 1.0000e+00, 0.0000e+00, 0.0000e+00,
0.0000e+00, 1.0000e+00, 0.0000e+00, 0.0000e+00, 0.0000e+00, 0.0000e+00,
1.0000e+00, 0.0000e+00, 0.0000e+00]], dtype=np.float64)

prediction_probs = daal_model.predict_proba(
    inputs
)

prediction = daal_model.predict(
    inputs
)

old_prediction = self.model.inplace_predict(
    inputs
)

print(prediction_probs, prediction, old_prediction)
```

**Expected behavior**

I expect to see the same results

**Output/Screenshots**
I get different results between xbg/d4p, and the d4p probs/predictions also seem inconsistent with themselves
```
[[0.53925217 0.46074783]] [0] [0.95738953]
```

**Environment:**
 - OS: Ubuntu 22.04.2 LTS via WSL
 - Version: daal/daal4py version 2024.6.0

razdoburdin:
Hi @kuza55,
Thanks for your report!

I have reproduced the bug on my end. We will work on a permanent fix for the issue you found, but it will take some time. As a quick workaround, you can save your model in `json` format and manually replace all values in the `feature_types` array from `int` to `float`.

Please let me know if you still encounter the problem after trying this fix.

PR #2053, opened 2024-09-16
fix int overflow in FTI model convertor

razdoburdin:
### Description

FTI model convertor for xgboost suffers from integer overflow error, since xgboost doesn't control numeric limits for integer valued features. In this PR we manually replace the feature type from `int` to `float `to avoid this problem.

Should fix https://github.com/intel/scikit-learn-intelex/issues/2016

---

**Testing**

**Performance**

### DIFF
--- a/src/gbt_convertors.pyx
+++ b/src/gbt_convertors.pyx
@@ -253,6 +253,15 @@ class TreeList(list):
         Note: We cannot type-hint the xgb.Booster without loading xgb as dependency in pyx code,
               therefore not type hint is added.
         """
+        # dump_format="json" is affected by the integer-overflow error,
+        # since XGBoost doesn't control numeric limits of integer features.
+        # For correct conversion we manulally replace feature types from 'int'->'float;
+        feature_types = booster.feature_types
+        for i in range(len(feature_types)):
+            if feature_types[i] == "int":
+                feature_types[i] = "float"
+        booster.feature_types = feature_types
+
         tl = TreeList()
         dump = booster.get_dump(dump_format="json", with_stats=True)
         for tree_id, raw_tree in enumerate(dump):
--- a/src/gbt_convertors.pyx
+++ b/src/gbt_convertors.pyx
@@ -256,12 +256,13 @@ class TreeList(list):
         # dump_format="json" is affected by the integer-overflow error,
         # since XGBoost doesn't control numeric limits of integer features.
         # For correct conversion we manulally replace feature types from 'int'->'float;
-        feature_types = booster.feature_types
-        if feature_types:
-            for i in range(len(feature_types)):
-                if feature_types[i] == "int":
-                    feature_types[i] = "float"
-            booster.feature_types = feature_types
+        if hasattr(booster, "feature_types"):
+            feature_types = booster.feature_types
+            if feature_types:
+                for i in range(len(feature_types)):
+                    if feature_types[i] == "int":
+                        feature_types[i] = "float"
+                booster.feature_types = feature_types
 
         tl = TreeList()
         dump = booster.get_dump(dump_format="json", with_stats=True)
--- a/src/gbt_convertors.pyx
+++ b/src/gbt_convertors.pyx
@@ -257,10 +257,11 @@ class TreeList(list):
         # since XGBoost doesn't control numeric limits of integer features.
         # For correct conversion we manulally replace feature types from 'int'->'float;
         feature_types = booster.feature_types
-        for i in range(len(feature_types)):
-            if feature_types[i] == "int":
-                feature_types[i] = "float"
-        booster.feature_types = feature_types
+        if feature_types:
+            for i in range(len(feature_types)):
+                if feature_types[i] == "int":
+                    feature_types[i] = "float"
+            booster.feature_types = feature_types
 
         tl = TreeList()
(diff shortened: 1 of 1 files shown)


src/gbt_convertors.pyx
```diff
@@ -253,6 +253,15 @@ class TreeList(list):
         Note: We cannot type-hint the xgb.Booster without loading xgb as dependency in pyx code,
               therefore not type hint is added.
         """
+        # dump_format="json" is affected by the integer-overflow error,
+        # since XGBoost doesn't control numeric limits of integer features.
+        # For correct conversion we manulally replace feature types from 'int'->'float;
+        feature_types = booster.feature_types
+        for i in range(len(feature_types)):
+            if feature_types[i] == "int":
```
david-cortes-intel, line 261: Can also be 'i' for boolean: https://xgboost.readthedocs.io/en/stable/python/python_api.html#xgboost.DMatrix

src/gbt_convertors.pyx
```diff
@@ -253,6 +253,15 @@ class TreeList(list):
         Note: We cannot type-hint the xgb.Booster without loading xgb as dependency in pyx code,
               therefore not type hint is added.
         """
+        # dump_format="json" is affected by the integer-overflow error,
```
david-cortes-intel, line 256: Same thing would happen with `ubjson`.

razdoburdin: yes, but we use `json` for conversion.

src/gbt_convertors.pyx
razdoburdin: Would Boolean features also be affected by overflow? I mean, if a column is marked as 'bool,' could the split condition for it be something other than '0' or '1'?

david-cortes-intel: I'm not  sure how booleans would be handled in `scikit-learn-intelex`, but in XGBoost, I think (though not 100% sure) they would be treated as numeric, meaning that the splits are of the form:
```
value <= threshold
```

For booleans, anything other than zero would be `true`, so:
```
(bool)(-0.5) -> true
(float)((bool)-0.5) <= 0 -> false
```

But without casting:
```
-0.5 <= 0 -> true
```

razdoburdin: From overflow point of view, model dump is free from this problem. It just don't convert split condition value for Boolean features.

david-cortes-intel: > From overflow point of view, model dump is free from this problem. It just don't convert split condition value for Boolean features.

Is it correct that it wouldn't cast to `bool` neither the split point, nor the values being passed? Just wanted to point out that one can still easily generate models like the one attached here.
[bad_model.json](https://github.com/user-attachments/files/17011614/bad_model.json)

razdoburdin: For the values we don't use `json `(or any other) conversion, so it shouldn't be any problems from this side.
As for splitting points, here is the conversion of the `bad_model`, you have attached. As you can see, `split_conditions `are not converted at all. 

```
model.load_model("bad_model.json")
print(model.get_dump(dump_format="json", with_stats=True))
```

`['  { "nodeid": 0, "depth": 0, "split": "cyl", "yes": 1, "no": 2 , "gain": 673.269775, "cover": 32, "children": [\n    { "nodeid": 1, "depth": 1, "split": "wt", "yes": 3, "no": 4 , "gain": 102.000031, "cover": 11, "children": [\n      { "nodeid": 3, "leaf": 2.56526804 , "cover": 6 }, \n      { "nodeid": 4, "leaf": 0.622343481 , "cover": 5 }\n    ]}, \n    { "nodeid": 2, "depth": 1, "split": "cyl", "yes": 5, "no": 6 , "gain": 88.5589447, "cover": 21, "children": [\n      { "nodeid": 5, "leaf": -0.0912892371 , "cover": 7 }, \n      { "nodeid": 6, "leaf": -1.39737535 , "cover": 14 }\n    ]}\n  ]}', '  { "nodeid": 0, "depth": 0, "split": "cyl", "yes": 1, "no": 2 , "gain": 364.500366, "cover": 32, "children": [\n    { "nodeid": 1, "depth": 1, "split": "disp", "yes": 3, "no": 4 , "gain": 63.5811615, "cover": 11, "children": [\n      { "nodeid": 3, "leaf": 2.05602694 , "cover": 5 }, \n      { "nodeid": 4, "leaf": 0.54339689 , "cover": 6 }\n    ]}, \n    { "nodeid": 2, "depth": 1, "split": "hp", "yes": 5, "no": 6 , "gain": 62.5995941, "cover": 21, "children": [\n      { "nodeid": 5, "leaf": -0.302962184 , "cover": 14 }, \n      { "nodeid": 6, "leaf": -1.38572836 , "cover": 7 }\n    ]}\n  ]}']`

samir-nasibli:
@razdoburdin could you please point out tests where this fix affected?

razdoburdin:
> @razdoburdin could you please point out tests where this fix affected?

Only the tests related to xgboost model_builders were affected.

icfaust:
please merge main into your branch @razdoburdin

icfaust:
Failures in pre-commit unreleated to the changes in this PR.

david-cortes-intel (approved):

PR #2055, opened 2024-09-17
[fix] automatically cancel older github action PR CI runs on new commit to branch

icfaust:
### Description

New commits on PRs with currently running CI jobs will not cancel those jobs/runs.  This will lead to a lock up of Github Actions runners if there are too many commits at the same time. This update will enable the same style of cancellation as is observed in Azure Pipelines.

Note this will not stop the github Actions CI checks on merges to main, only on PRs.

---

**Testing**

**Performance**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,10 @@ on:
       - main
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref_name }}-${{ github.event.number || github.sha }}
+  cancel-in-progress: true
+
 env:
   TBB_VERSION: 2021.13
   DPCPP_VERSION: 2024.2
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-# build artefacts test
+# build artefacts
 *.so
 *.egg
 *.pyc
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-# build artefacts
+# build artefacts test
 *.so
 *.egg
 *.pyc


icfaust:
Operational success can be found for the two test commits in the actions CI list https://github.com/intel/scikit-learn-intelex/actions

.github/workflows/ci.yml
```diff
@@ -23,6 +23,10 @@ on:
       - main
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref_name }}-${{ github.event.number || github.sha }}
```
napetrov, line 27: wouldn't this mean that we are grouping it to the same sha groups vs just PR?

icfaust: good question! the sha group is only used as a fallback in the case that a PR number isn't defined (```github.event.number```) which will be the case when it isn't a PR (push to main and in workflow_dispatch). https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-using-a-fallback-value

.github/workflows/ci.yml
```diff
@@ -23,6 +23,10 @@ on:
       - main
   workflow_dispatch:
 
+concurrency:
```
Alexsandruss, line 26: Should not it be on job level rather than on global?

icfaust: I guess you could do it on jobs, but I'm trying to understand why that would be wanted.  If you push a new commit, there isn't a granular control of the jobs in a workflow, you'd like to see all rerun.  Let me know, I can change it but it seems unnecessary.

Alexsandruss: I'm worried if concurrency limits number of simultaneous jobs in same workflow, Github Workflows docs are ambiguous about that: is one type of job limited or all of them.

icfaust: Ah, so what this will do is group all the jobs run under this workflow together, so there is concurrency for all the jobs in the workflow, what is not allowed is another workflow from this PR named CI to be run.  This concurrency check essentially makes it equivalent of what occurs on Azure Pipelines. Instead of checking job to job, this will check workflow to workflow, so all the jobs in this workflow should run concurrently.

Alexsandruss (approved):

PR #2072, opened 2024-09-27
FIX: Add covariance module to sklearnex init

olegkkruglov:
### Description

* Add covariance module to sklearnex `__init__.py` which allows to use `covariance` submodule after `import *`

The following code will work only after this change

```python
from sklearnex import *
c = covariance.IncrementalEmpiricalCovariance()
```

---

**Testing**

**Performance**

### DIFF
--- a/sklearnex/__init__.py
+++ b/sklearnex/__init__.py
@@ -32,6 +32,7 @@ __all__ = [
     "basic_statistics",
     "cluster",
     "config_context",
+    "covariance",
     "decomposition",
     "ensemble",
     "get_config",


samir-nasibli:
@olegkkruglov Why this wasn't before catch? I think make sense add test suit for the import

olegkkruglov:
> @olegkkruglov Why this wasn't before catch?

I guess because we don't use import * in our tests

samir-nasibli:
@olegkkruglov we have this import in the tests https://github.com/intel/scikit-learn-intelex/blob/f094bc40fd8194cbe652dcb17619baf4dc0a87f4/sklearnex/covariance/tests/test_incremental_covariance.py#L222
How this work then?

olegkkruglov:
> @olegkkruglov I think make sense add test suit for the import



> @olegkkruglov we have this import in the tests
> 
> https://github.com/intel/scikit-learn-intelex/blob/f094bc40fd8194cbe652dcb17619baf4dc0a87f4/sklearnex/covariance/tests/test_incremental_covariance.py#L222
> 
> 
> How this work then?

it is not `import *`

for example, the following code will work only after my change
`
from sklearnex import *
c = covariance.EmpiricalCovariance()
`

samir-nasibli (approved):
CI fails are not related with the changes. But I would suggest to run internal CI once again

sklearnex/__init__.py
```diff
@@ -32,6 +32,7 @@
     "basic_statistics",
     "cluster",
     "config_context",
+    "covariance",
```
icfaust, line 35: @olegkkruglov I believe your example in the description is incorrect. It should be ```IncrementalEmpiricalCovariance```. Importing ```EmpericalCovariance``` can only be done from the sklearnex.preview package. @samir-nasibli This would have been caught by https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/tests/test_patching.py#L289 but it only looks for matching estimators, which is not yet the case because EmpiricalCovariance is still in preview and IncrementalEmpericalCovariance is a sklearnex-unique estimator.

samir-nasibli: > @olegkkruglov I believe your example in the description is incorrect. It should be `IncrementalEmpiricalCovariance`. Importing `EmpericalCovariance` can only be done from the sklearnex.preview package. @samir-nasibli This would have been caught by https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/tests/test_patching.py#L289 but it only looks for matching estimators, which is not yet the case because EmpiricalCovariance is still in preview and IncrementalEmpericalCovariance is a sklearnex-unique estimator.

@icfaust thank you! I was confused by the description. Let's cover this with a test

olegkkruglov: i think the test is not really necessary here. after moving EmpiricalCovariance out of preview this will be covered

samir-nasibli: Discussed with @icfaust 
We acan adapt this test 
https://github.com/icfaust/scikit-learn-intelex/blob/dev/centralize_finite_assert/sklearnex/tests/test_common.py#L101
if he were to load all_estimators and then cross-reference to sklearnex.__all__
But this suggestion the scope for another PR.

Please @olegkkruglov make sure to create a ticket for this.

samir-nasibli (approved):
Thank you @olegkkruglov for the finding.

PR #2086, opened 2024-10-07
[fix] correct skips in design rule checks (```test_common.py```) caused by fragile whitelist_to_blacklist

icfaust:
## Description

fix errors in private CI and Azure Pipelines caused by differences in ```sys.path```.  This PR will enable those tests properly in those CI systems.  It deals with possible issues with user paths and symlinks, as well as erroneous inclusions of subdirectories (if they occur).

No benching necessary as it is purely testing code.

---

**Testing**

**Performance**


sklearnex/tests/test_common.py
```diff
@@ -49,6 +49,54 @@
     "IncrementalEmpiricalCovariance-score-call_validate_data",  #  must call clone of itself
     "SVC(probability=True)-fit-call_validate_data",  #  SVC fit can use sklearn estimator
     "NuSVC(probability=True)-fit-call_validate_data",  #  NuSVC fit can use sklearn estimator
+    "LogisticRegression-score-n_jobs_check",  # uses daal4py for cpu in sklearnex
```
samir-nasibli, line 52: We only have one list for all design violations vs all design rules?
Should be renamed for the specific rule I think and better to have for the specific rule specific maybe list of design violation.

icfaust: The design rule is in the string, I'm not sure I understand

samir-nasibli: I mean we have only one skip list for the all design rules?

samir-nasibli: I mean test should be parameterized like this
`def test_estimator(estimator, method, design_pattern, design_violation_list, estimator_trace):
`
Probably some design patterns should have some part of common design violation list, that should be defined.
But if we expect that all rules will not have the specific violations except some common list then make sense to leave as is

samir-nasibli: But probably the comment for the further update :)

samir-nasibli: Thank you @icfaust !
We have discussed it somehow, assuming we make the comment resolved.
We have one skip list for each test cases, where the design_rule is specified. Im good with this one.

Just for the further optimization we can add the string comment into pytest,xfail. to see this in the logs not only in code base. But this refactoring not the scope of the current PR.

samir-nasibli (approved):

sklearnex/tests/test_common.py
icfaust: Done, it will now give an informative string on xfail.

david-cortes-intel (approved):
Cannot comment on whether the list covers everything that it should, but can confirm that it solves test failures for me, at least in my local environment.

icfaust:
I'm going to merge this and will be responsible for private CI failures.

PR #2090, opened 2024-10-07
TEST: Run all tests through `pytest`

david-cortes-intel:
## Description

Currently, unit tests execute through a mixture of built-in `unittest` and `pytest`.

From the commit history, it looks like initial tests were added using python's built-in `unittest` framework, and then subsequent tests were introduced using `pytest`.

Compared to `unittest`, `pytest` has a much nicer output format, offers many options for how to execute tests and how to see their outputs, and has plugins that allow it do extra things which the built-in cannot when desired (e.g. JSON outputs, test coverage reports, etc.).

This PR switches the automated scripts to execute all of the tests using `pytest`. This doesn't change what tests were executed - just uses `pytest` as the runner.

One challenge here is how each of these interprets the path from where modules are imported - here I'm trying to control these with the `PYTHONPATH` variable when `pytest` is called within shell scripts, but don't know if this will work out for all the variants under which tests are executed.

---

**Testing**

**Performance**

### DIFF
--- a/.ci/scripts/test_global_patch.py
+++ b/.ci/scripts/test_global_patch.py
@@ -21,6 +21,10 @@ import sys
 
 import pytest
 
+# This is a workaround for older versions of Python on Windows
+# which didn't have it as part of the built-in 'os' module.
+EX_OK = os.EX_OK if hasattr(os, "EX_OK") else 0
+
 # Note: from the structure of this file, one might thing of adding a test
 # along the lines of 'test_patching_all_from_command_line'. There is however
 # an issue in that, after the first time a scikit-learn module is imported,
@@ -36,14 +40,14 @@ def patch_svc_from_command_line():
     err_code = subprocess.call(
         [sys.executable, "-m", "sklearnex.glob", "patch_sklearn", "-a", "svc"]
     )
-    assert err_code == os.EX_OK
+    assert err_code == EX_OK
 
     yield
 
     err_code = subprocess.call(
         [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn", "-a", "svc"]
     )
-    assert err_code == os.EX_OK
+    assert err_code == EX_OK
 
 
 def test_patching_svc_from_command_line(patch_svc_from_command_line):
@@ -59,7 +63,7 @@ def test_unpatching_svc_from_command_line(patch_svc_from_command_line):
     err_code = subprocess.call(
         [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"]
     )
-    assert err_code == os.EX_OK
+    assert err_code == EX_OK
     from sklearnex import unpatch_sklearn
 
     unpatch_sklearn()
--- a/.ci/scripts/test_global_patch.py
+++ b/.ci/scripts/test_global_patch.py
@@ -22,7 +22,7 @@ import pytest
 
 
 @pytest.fixture
-def patch_from_command_line(request):
+def patch_svc_from_command_line(request):
     err_code = subprocess.call(
         [sys.executable, "-m", "sklearnex.glob", "patch_sklearn", "-a", "svc"]
     )
@@ -38,7 +38,7 @@ def patch_from_command_line(request):
     return
 
 
-def test_patching_from_command_line(patch_from_command_line):
+def test_patching_svc_from_command_line(patch_from_command_line):
     from sklearn.svm import SVC, SVR
 
     assert SVC.__module__.startswith("daal4py") or SVC.__module__.startswith("sklearnex")
@@ -47,7 +47,7 @@ def test_patching_from_command_line(patch_from_command_line):
     )
 
 
-def test_unpatching_from_command_line(patch_from_command_line):
+def test_unpatching_svc_from_command_line(patch_from_command_line):
     err_code = subprocess.call(
         [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"]
     )
@@ -66,7 +66,7 @@ def test_unpatching_from_command_line(patch_from_command_line):
 
 
 @pytest.fixture
-def patch_from_function(request):
+def patch_svc_from_function(request):
     from sklearnex import patch_sklearn, unpatch_sklearn
 
     patch_sklearn(name=["svc"], global_patch=True)
@@ -78,7 +78,7 @@ def patch_from_function(request):
     return
 
 
-def test_patching_from_function(patch_from_function):
+def test_patching_svc_from_function(patch_from_function):
     from sklearn.svm import SVC, SVR
 
     assert SVC.__module__.startswith("daal4py") or SVC.__module__.startswith("sklearnex")
@@ -87,7 +87,7 @@ def test_patching_from_function(patch_from_function):
     )
 
 
-def test_unpatching_from_function(patch_from_function):
+def test_unpatching_svc_from_function(patch_from_function):
     from sklearnex import unpatch_sklearn
 
     unpatch_sklearn(global_unpatch=True)
--- a/.ci/scripts/test_global_patch.py
+++ b/.ci/scripts/test_global_patch.py
@@ -21,6 +21,15 @@ import sys
 
 import pytest
 
+# Note: from the structure of this file, one might thing of adding a test
+# along the lines of 'test_patching_all_from_command_line'. There is however
+# an issue in that, after the first time a scikit-learn module is imported,
+# further calls to 'patch_sklearn' with different arguments will have no effect
+# since sklearn is already imported. Reloading it through 'importlib.reload'
+# or deleting it from 'sys.modules' doesn't appear to have the intended effect
+# either. This also makes this first command-line fixture and tests that use
+# it not entirely idempotent, given that they need to import sklearn modules.
+
 
 @pytest.fixture
 def patch_svc_from_command_line():
@@ -62,42 +71,6 @@ def test_unpatching_svc_from_command_line(patch_svc_from_command_line):
     assert not SVR.__module__.startswith("daal4py") and not SVR.__module__.startswith(
         "sklearnex"
     )
-
-
-@pytest.fixture
-def patch_all_from_command_line():
-    err_code = subprocess.call([sys.executable, "-m", "sklearnex.glob", "patch_sklearn"])
-    assert err_code == os.EX_OK
-
-    yield
-
-    err_code = subprocess.call(
-        [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"]
-    )
-    assert err_code == os.EX_OK
-
-
-def test_patching_all_from_command_line(patch_all_from_command_line):
-    from sklearn.svm import SVC, SVR
-
-    assert SVC.__module__.startswith("daal4py") or SVC.__module__.startswith("sklearnex")
-    assert SVR.__module__.startswith("daal4py") or SVR.__module__.startswith("sklearnex")
-
-
-def test_unpatching_all_from_command_line(patch_all_from_command_line):
-    err_code = subprocess.call(
-        [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"]
-    )
-    assert err_code == os.EX_OK
-    from sklearnex import unpatch_sklearn
-
-    unpatch_sklearn()
-    from sklearn.svm import SVC, SVR
-
-    assert not SVC.__module__.startswith("daal4py")
-    assert not SVC.__module__.startswith("sklearnex")
-    assert not SVR.__module__.startswith("daal4py")
-    assert not SVR.__module__.startswith("sklearnex")
 
 
 @pytest.fixture
--- a/.ci/scripts/test_global_patch.py
+++ b/.ci/scripts/test_global_patch.py
@@ -18,52 +18,84 @@
 import subprocess
 import sys
 
-# test patching from command line
-err_code = subprocess.call(
-    [sys.executable, "-m", "sklearnex.glob", "patch_sklearn", "-a", "svc"]
-)
-assert not err_code
-from sklearn.svm import SVC, SVR
-
-assert SVC.__module__.startswith("daal4py") or SVC.__module__.startswith("sklearnex")
-assert not SVR.__module__.startswith("daal4py") and not SVR.__module__.startswith(
-    "sklearnex"
-)
+import pytest
 
 
-from sklearnex import patch_sklearn, unpatch_sklearn
+@pytest.fixture
+def patch_from_command_line(request):
+    err_code = subprocess.call(
+        [sys.executable, "-m", "sklearnex.glob", "patch_sklearn", "-a", "svc"]
+    )
+    assert not err_code
 
-# test unpatching from command line
-err_code = subprocess.call([sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"])
-assert not err_code
-unpatch_sklearn()
-from sklearn.svm import SVC, SVR
+    def unpatch_from_cmd():
+        err_code = subprocess.call(
+            [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"]
+        )
+        assert not err_code
 
-assert not SVC.__module__.startswith("daal4py") and not SVC.__module__.startswith(
-    "sklearnex"
-)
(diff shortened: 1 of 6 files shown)


david-cortes-intel:
@Alexsandruss Would you perhaps have any pointers about why the failing CI job here:
https://github.com/intel/scikit-learn-intelex/actions/runs/11218859163/job/31183555329?pr=2090

.. might be failing to import `daal4py._daal4py` in the new `pytest` call that I added, while it succeeds in the other calls?

icfaust:
I've had to fight against pytest's package discovery recently and had to use ```--rootdir``` to get things to work properly (https://github.com/intel/scikit-learn-intelex/blob/main/.ci/scripts/run_sklearn_tests.py#L47). Maybe that could help you?  I believe your ```daal4py._daal4py``` is coming from the use of the ```$PYTHONPATH``` which is using the uninstalled repo daal4py folder to find daal4py, but can't find the compiled _daal4py.so file there.

david-cortes-intel:
> I've had to fight against pytest's package discovery recently and had to use `--rootdir` to get things to work properly (https://github.com/intel/scikit-learn-intelex/blob/main/.ci/scripts/run_sklearn_tests.py#L47). Maybe that could help you? I believe your `daal4py._daal4py` error is coming from the `$PYTHONPATH` which is using the repo daal4py folder to find daal4py instead of the installed site-packages daal4py, because it can't find the compiled _daal4py.so file there.

Thanks for the tip. But that was actually for a different reason: some test files import other tests files - for example `test_daal4py_spmd_examples.py` imports `test_daal4py_examples` right here, as an absolute import instead of relative, so the folder with test files needs to be under `sys.path` for it to work:
https://github.com/intel/scikit-learn-intelex/blob/2f73b9de4a79205b67522f502ae140e6470871f3/tests/test_daal4py_spmd_examples.py#L22

But since now these tests should only be getting executed through `pytest`, I guess a better solution should be to use relative imports. I'm now trying that approach so let's see how the CI will end up working with it.

david-cortes-intel:
Looks like it was just a matter of deleting the `__init__.py` file in the tests folder. All CI jobs are executing them correctly now.

.ci/scripts/test_global_patch.py
```diff
@@ -18,52 +18,84 @@
...
+    assert not err_code
+
+    def unpatch_from_cmd():
+        err_code = subprocess.call(
+            [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"]
+        )
+        assert not err_code
+
+    request.addfinalizer(unpatch_from_cmd)
+    return
```
icfaust, line 38:
```suggestion
    err_code = subprocess.call(
        [sys.executable, "-m", "sklearnex.glob", "patch_sklearn", "-a", "svc"]
    )
    assert not err_code
    yield
    err_code = subprocess.call(
        [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"]
    )
    assert not err_code
```

Thoughts on using yield instead?

.ci/scripts/test_global_patch.py
```diff
@@ -18,52 +18,84 @@
...
+    assert not SVR.__module__.startswith("daal4py") and not SVR.__module__.startswith(
+        "sklearnex"
+    )
+
+
+@pytest.fixture
+def patch_from_function(request):
+    from sklearnex import patch_sklearn, unpatch_sklearn
+
+    patch_sklearn(name=["svc"], global_patch=True)
```
icfaust, line 72: Again possibly yield?

.ci/scripts/test_global_patch.py
```diff
@@ -18,52 +18,84 @@
...
-    "sklearnex"
-)
-assert not SVR.__module__.startswith("daal4py") and not SVR.__module__.startswith(
-    "sklearnex"
-)
+import pytest
+
+
+@pytest.fixture
+def patch_from_command_line(request):
```
icfaust, line 25: since this is all derived with the ```SVC``` estimator, could you add that into the fixture name, wasn't obvious to me at first

.ci/scripts/test_global_patch.py
```diff
@@ -18,52 +18,84 @@
...
+def test_patching_from_command_line(patch_from_command_line):
+    from sklearn.svm import SVC, SVR
+
+    assert SVC.__module__.startswith("daal4py") or SVC.__module__.startswith("sklearnex")
+    assert not SVR.__module__.startswith("daal4py") and not SVR.__module__.startswith(
+        "sklearnex"
+    )
+
+
+def test_unpatching_from_command_line(patch_from_command_line):
```
icfaust, line 50: why is the fixture used here again? Is this trying to expand functionality or match the previous one (because this is slightly different from what was before).  What was previous assumed that the global setting was returned to normal after the unpatch, so by reapplying the ```"sklearnex.glob", "patch_sklearn", "-a", "svc"``` the state of this test is different than what occurred in the file before.

.ci/scripts/test_global_patch.py
```diff
@@ -18,52 +18,84 @@
...
+    assert not SVC.__module__.startswith("daal4py") and not SVC.__module__.startswith(
+        "sklearnex"
+    )
+    assert not SVR.__module__.startswith("daal4py") and not SVR.__module__.startswith(
+        "sklearnex"
+    )
+
+
+@pytest.fixture
+def patch_from_function(request):
```
icfaust, line 69:
```suggestion
def patch_svc_from_function(request):

```

.ci/scripts/test_global_patch.py
```diff
@@ -18,52 +18,84 @@
...
+
+
+def test_unpatching_from_command_line(patch_from_command_line):
+    err_code = subprocess.call(
+        [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"]
+    )
+    assert not err_code
+    from sklearnex import unpatch_sklearn
+
+    unpatch_sklearn()
```
icfaust, line 57: https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/tests/test_monkeypatch.py#L65  Just be careful with these functions since they impact global state and aren't local to the test (anything with a patch_sklearn call should be with a 'try-finally' with an unpatch_sklearn)

.ci/scripts/test_global_patch.py
```diff
@@ -18,52 +18,84 @@
...
+def test_patching_from_function(patch_from_function):
+    from sklearn.svm import SVC, SVR
+
+    assert SVC.__module__.startswith("daal4py") or SVC.__module__.startswith("sklearnex")
+    assert not SVR.__module__.startswith("daal4py") and not SVR.__module__.startswith(
+        "sklearnex"
+    )
+
+
+def test_unpatching_from_function(patch_from_function):
```
icfaust, line 90: Now that I look at what the code had previously been, I am glad you have done this. Much better isolation of tests.

conda-recipe/meta.yaml
```diff
@@ -76,9 +76,9 @@ test:
         - daal4py/sklearn
     commands:
         - python -c "import daal4py"
-        - mpirun -n 4 python -m unittest discover -v -s tests -p test*spmd*.py # [not win]
-        - mpiexec -localonly -n 4 python -m unittest discover -v -s tests -p test*spmd*.py # [win]
-        - python -m unittest discover -v -s tests -p test*.py
+        - mpirun -n 4 pytest --pyargs --verbose -s tests/test*spmd*.py # [not win]
```
icfaust, line 79: is ```--pyargs``` necessary since you are using a path and not a package name? (same goes for the others in ```run_test.bat``` / ```run_test.sh```)

tests/test_examples_sklearnex.py
```diff
@@ -36,40 +38,31 @@
...
+            f
+            for f in os.listdir(examples_path)
+            if f.endswith(".py")
+            and "spmd" not in f
+            and "dpnp" not in f
+            and "dpctl" not in f
+        ],
+    )
+    def test_generator(file):
+        def testit(self):
```
icfaust, line 53: Dumb question, why is the ```testit``` function necessary now since we are no longer setting attributes to the class ```TestsklearnexExamples``` (is the class necessary anymore?)

icfaust (commented):
thank you for doing this.  Here are my initial thoughts.

.ci/scripts/test_global_patch.py
david-cortes-intel: What would be the advantage of using yield here?

.ci/scripts/test_global_patch.py
david-cortes-intel: The fixture has a finalizer. So on each test, it gets initialized (global patching is applied), and after the test finishes, gets finalized (global unpatching is applied). Then the same repeats for the next test that uses the fixture, ensuring that state doesn't propagate across tests. This is of course assuming that patching and unpatching are working correctly, which is what the tests are looking at, but I thought it better than the idea of merging patching and unpatching in a single test.

.ci/scripts/test_global_patch.py
david-cortes-intel: Yes, that's the idea behind the fixtures with finalizers.

conda-recipe/meta.yaml
david-cortes-intel: I don't know. But all the other `pytest` calls have it.

tests/test_examples_sklearnex.py
david-cortes-intel: Thanks for pointing this out. I just realized that `pytest.parametrize` doesn't actually work when put into methods that inherit from `unittest.TestCase`, so the parametrizations where not actually being executed. I changed it now to regular `pytest`.

.ci/scripts/test_global_patch.py
david-cortes-intel: Added.

.ci/scripts/test_global_patch.py
icfaust: Its not a must, but its the pytest recommended procedure over adding a finalizer https://docs.pytest.org/en/stable/how-to/fixtures.html#yield-fixtures-recommended

conda-recipe/meta.yaml
icfaust: I'd say cut them out and see what happens. If they fail you can readd them (and ping me)

.ci/scripts/test_global_patch.py
david-cortes-intel: Changed to `yield`.

.ci/scripts/test_global_patch.py
```diff
@@ -18,52 +18,80 @@
...
-    "sklearnex"
-)
-
-
-from sklearnex import patch_sklearn, unpatch_sklearn
-
-# test unpatching from command line
-err_code = subprocess.call([sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"])
-assert not err_code
-unpatch_sklearn()
```
icfaust, line 39: The state of the unpatch_sklearn() test was previously after a ```subprocess.call([sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"])``` not a 
```subprocess.call(
        [sys.executable, "-m", "sklearnex.glob", "patch_sklearn", "-a", "svc"]``` call, meaning the state of the test is different in the new implementation.  This is okay if this was what you desired


https://github.com/intel/scikit-learn-intelex/pull/2090#discussion_r1794733125

Comment (david-cortes-intel) [reply .ci/scripts/test_global_patch.py:39]: Thanks, I had missed that detail entirely.

So I tried adding a new test for global state separately from just `SVC`, but the test actually ends up failing:
```python
@pytest.fixture
def patch_all_from_command_line():
    err_code = subprocess.call(
        [sys.executable, "-m", "sklearnex.glob", "patch_sklearn"]
    )
    assert err_code == os.EX_OK

    yield

    err_code = subprocess.call(
        [sys.executable, "-m", "sklearnex.glob", "unpatch_sklearn"]
    )
    assert err_code == os.EX_OK

def test_patching_all_from_command_line(patch_all_from_command_line):
    from sklearn.svm import SVC, SVR

    assert SVC.__module__.startswith("daal4py")
    assert SVC.__module__.startswith("sklearnex")
    assert SVR.__module__.startswith("daal4py")
    assert SVR.__module__.startswith("sklearnex")
```

should that be expected to succeed?

icfaust: I assume that its failing on
```assert SVC.__module__.startswith("daal4py")```. You can remove
```assert SVC.__module__.startswith("daal4py")``` and ```assert SVR.__module__.startswith("daal4py")```

Comment (david-cortes-intel) [reply .ci/scripts/test_global_patch.py:39]: It also fails for `sklearnex`.

Comment (david-cortes-intel) [reply .ci/scripts/test_global_patch.py:39]: I added it this new test as an extra commit here, let's see if it also fails in the CI.

Comment (icfaust) [reply .ci/scripts/test_global_patch.py:39]: Could be a legit failure then, lets see

Comment (david-cortes-intel) [reply .ci/scripts/test_global_patch.py:39]: Looks like it also fails in the CI jobs:
```
    def test_patching_all_from_command_line(patch_all_from_command_line):
        from sklearn.svm import SVC, SVR
    
>       assert SVC.__module__.startswith("daal4py") or SVC.__module__.startswith("sklearnex")
E       AssertionError: assert (False or False)
E        +  where False = <built-in method startswith of str object at 0x7f8679a04300>('daal4py')
E        +    where <built-in method startswith of str object at 0x7f8679a04300> = 'sklearn.svm._classes'.startswith
E        +      where 'sklearn.svm._classes' = <class 'sklearn.svm._classes.SVC'>.__module__
E        +  and   False = <built-in method startswith of str object at 0x7f8679a04300>('sklearnex')
E        +    where <built-in method startswith of str object at 0x7f8679a04300> = 'sklearn.svm._classes'.startswith
E        +      where 'sklearn.svm._classes' = <class 'sklearn.svm._classes.SVC'>.__module__

scripts/test_global_patch.py:83: AssertionError
```

File: conda-recipe/meta.yaml (hunk 8 above)
Comment (david-cortes-intel) [reply conda-recipe/meta.yaml]: Looks like they weren't necessary.

File: .ci/scripts/test_global_patch.py (hunk 10 above)
Comment (icfaust) [reply .ci/scripts/test_global_patch.py:39]: Yeah this is bad, i'll admit I don't know much about the ```glob``` use.  I wonder if sklearn has been pre-imported via sklearnex imports, and the change in glob not affecting it.  There is definitely something here.

Comment (david-cortes-intel) [reply .ci/scripts/test_global_patch.py:39]: Yes, looks like that does indeed make a difference. If leaving it as the only test then it actually succeeds. Will try to find a workaround.

Comment (david-cortes-intel) [reply .ci/scripts/test_global_patch.py:39]: I didn't manage to find any solution so in the end I just removed the test and left a comment in the file.

File: .ci/scripts/test_global_patch.py (hunk 1 above)
Comment (david-cortes-intel) [reply .ci/scripts/test_global_patch.py]: Looks like the windows job doesn't like usage of 'yield':
```
INTERNALERROR>     ^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "C:\Miniconda\envs\CB\Lib\site-packages\_pytest\logging.py", line 790, in pytest_collection
INTERNALERROR>     return (yield)
INTERNALERROR>             ^^^^^

...

joblib.externals.loky.process_executor.BrokenProcessPool: A task has failed to un-serialize. Please ensure that the arguments of the function are all picklable.
```

Will try to see where that multi-processing invocation is coming from.

File: .ci/scripts/test_global_patch.py [hunk 11]
```diff
@@ -15,55 +15,135 @@
...
 
-assert SVC.__module__.startswith("daal4py") or SVC.__module__.startswith("sklearnex")
-assert not SVR.__module__.startswith("daal4py") and not SVR.__module__.startswith(
-    "sklearnex"
-)
+# This is a workaround for older versions of Python on Windows
+# which didn't have it as part of the built-in 'os' module.
+EX_OK = os.EX_OK if hasattr(os, "EX_OK") else 0
 
+# Note: from the structure of this file, one might thing of adding a test
```
icfaust, line 28:
```suggestion
# Note: from the structure of this file, one might think of adding a test

```

icfaust (approved):
Assuming small English correction ready to merge

PR #2099, opened 2024-10-09
TEST: using XFAIL instead of SKIP where it is required

samir-nasibli:
## Description

[in progress]
Tests will be checked.
Assuming that skip needed to be used for env or sigsevs 

---

**Testing**

**Performance**


icfaust:
just be careful, from the pytest docs: "Note that no other code is executed after the [pytest.xfail()](https://docs.pytest.org/en/stable/reference/reference.html#pytest.xfail) call, differently from the marker. That’s because it is implemented internally by raising a known exception"

samir-nasibli:
> just be careful, from the pytest docs: "Note that no other code is executed after the [pytest.xfail()](https://docs.pytest.org/en/stable/reference/reference.html#pytest.xfail) call, differently from the marker. That’s because it is implemented internally by raising a known exception"

Thank you @icfaust ! better to use [pytest.mark.xfail](https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-xfail-ref) instead. Will update

syakov-intel:
@samir-nasibli what needs to be done to get out of Draft state?

samir-nasibli:
[pytest.xfail()](https://docs.pytest.org/en/stable/reference/reference.html#pytest.xfail) This isn't working as I expected. After calling it, the code after doesn't work, so the proposed changes lose all value and meaning. I suggest closing the PR and in the future, if something needs to be marked as `xfail`, the type of error expected and the reason for it should be specified. For example:

```python
# Don't know the exact error type or error message, and how to provide exact message here to check.
# RuntimeError("the error message ") doesn't work.
@pytest.mark.xfail(raises=RuntimeError, "Expected to fail due to bug in histogram merges fixed in 2025.2.")
@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
def test_sklearnex_import_et_classifier(dataframe, queue):
...
```

@icfaust make sense to close and rethink about the design.

PR #2106, opened 2024-10-13
ENH: `PCA` via Array API based on #2096

samir-nasibli:
## Description

This enables array API support for the PCA and IncrementalPCA sklearnex estimators. This required a refactor of the related sklearnex and onedal estimators in order to match design guidelines and simplify integration.  This re-enables 50 sklearn tests.

This required the following changes:
 - Fix missing IncrementalPCA sklearn conformance tests (had been untested before)
 - cast device when the array supports the array API standard
 - Changed inheritance of the incrementalPCA onedal estimator to simplify code
 - Removed check_array use in onedal's IncrementalPCA as that is a sklearn conformance requirement
 - Remove use check_raw_input as its no longer relevant
 - Removed sklearn conformance in n_components in the oneDAL estimator
 - removed the double setting of self._queue
 - integrated use of the ```like``` keyword for proper return types
 - set ```_onedal_model``` as to None and to always exist as an attribute, following in line the new design rule
 - moved many sklearn conformance functions from the onedal estimator to sklearnex
 - removed the aliasing of ```predict``` to ```transform``` in the onedal estimator (and its use in testing)
 - Changed structure of ```fit``` and ```fit_transform``` for the PCA and IncrementalPCA sklearnex estimators to minimize maintenance (no longer duplicating sklearn code nor requiring version checking). This was done by  patching ```fit``` and required ```_onedal_fit``` to return the validated X data for use later.
 - Simplified the calculations made in ```_onedal_cpu_supported``` and ```_onedal_gpu_supported``` 
 - fixed problems in the IncrementalPCA patching which had a bug with respect to sparse inputs
 - sklearn conformance functions added (```_validate_n_components```, ```_postprocess_n_components``` and ```_compute_noise_variance```).  Added comments to clarify why things are done for simple future analysis
 - add ```get_namespace``` and modified corresponding ```validate_data``` dtypes throughout the code
 - set onedal estimator factory function  as a PCA class method in order to easily create an SPMD interface 
 - Added the necessary PCA SPMD sklearnex interface 
 - Moved modification of the output results from onedal to sklearn for PCA sklearn conformance.  This included adding copying behavior, generalized to array API.
 - created ```n_features_``` property for sklearn conformance, rather than duplicating the same data twice
 - Just like linear model algorithms like ```LinearRegression``` and ```Ridge```, attributes which are required to generate models for prediction/inference are set in a way where modifications in the sklearnex estimator (i.e. by a later fit which falls back to sklearn) are propagated properly to the onedal estimator.  This will mean that the output of transform will always reflect the last fit.
 - ```check_is_fitted``` is properly added to IncrementalPCA
 - missing sklearn conformance is added in the ```n_components``` check (which matches fixes for an underlying bug in sklearn IncrementalPCA)
 - A new parameter is added to the sklearnex IncrementalPCA interface (svd_solver) which exposes the ```onedal_svd``` approach which is available in oneDAL.  This matches the PCA keyword, but is not available in the sklearn IncrementalPCA implementation.  This will require a deselection in ```test_patching``` when IncrementalPCA is brought out of preview
 - The finalize_fit sklearn conformance changes are moved from onedal to sklearnex in IncrementalPCA
 - re-enable deselected sklearnex tests with respect to design and end-to-end array api tests (test_patching)
 - Added testing for cases where attributes are modified, or a fallback sklearn fit occurs after a onedal fit.
 - Removed unnecessary shape restriction on oneDAL support.
 - Deselected tests related to limitations in the array_api_strict numpy backend __dlpack__ support

No changes in performance are expected.

---

**Testing**

**Performance**


sklearnex/decomposition/pca.py
```diff
@@ -110,62 +116,319 @@ def __init__(
...
 
-                return U, S, Vt, X, x_is_centered, xp
+            # This is necessary as get_namespace may return a bare numpy,
+            # which ``asarray`` may not contain the copy keyword argument.
+            if xp is np:
+                copy = np.copy
             else:
-                return U, S, Vt
+                copy = lambda x: xp.asarray(x, copy=True)
+
```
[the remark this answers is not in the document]
icfaust: mehhhh nah.

onedal/datatypes/_data_conversion.py
```diff
@@ -111,7 +111,9 @@ def return_type_constructor(array):
                 else xp.asarray(backend.from_table(x), device=device)
             )
     elif hasattr(array, "__array_namespace__"):
-        func = array.__array_namespace__().from_dlpack
+        xp = array.__array_namespace__()
+        device = array.device
+        func = lambda inp: xp.from_dlpack(inp, device=device)
```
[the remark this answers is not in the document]
icfaust: no thanks.

onedal/decomposition/incremental_pca.py
```diff
@@ -146,45 +146,34 @@ def partial_fit(self, X, queue=None):
...
-        if not use_raw_input:
-            X = _check_array(X, dtype=[np.float64, np.float32], ensure_2d=True)
-
         n_samples, n_features = X.shape
         first_pass = not hasattr(self, "components_")
         if first_pass:
             self.components_ = None
             self.n_samples_seen_ = n_samples
             self.n_features_in_ = n_features
+            self.n_components_ = self.n_components if self.n_components else min(X.shape)
```
david-cortes-intel: What happens if the first batch has less rows than components, but then the second batch leaves the total amount of rows being larger than the number of components?

onedal/decomposition/incremental_pca.py
```diff
@@ -146,45 +146,34 @@ def partial_fit(self, X, queue=None):
...
+            self._outtype = return_type_constructor(X)
         self._queue = queue
 
         X_table = to_table(X, queue=queue)
 
         if not hasattr(self, "_dtype"):
             self._dtype = X_table.dtype
             self._params = self._get_onedal_params(X_table)
+        else:
+            assert X.dtype == self._dtype, f"data of different datatype given"
```
david-cortes-intel: Shouldn't it convert the data in this case? Scikit-learn wouldn't error out if passed data of different types across different batches.

sklearnex/decomposition/pca.py
```diff
@@ -110,62 +116,319 @@ def __init__(
...
+                if (
+                    sklearn_check_version("1.5")
+                    and n_features <= 1_000
+                    and n_samples >= 10 * n_features
+                ):
+                    return "covariance_eigh"
+                elif max(n_samples, n_features) <= 500 or n_components == "mle":
+                    return "full"
+                elif 1 <= n_components < 0.8 * n_sf_min:
+                    return "randomized"
```
david-cortes-intel, line 277: Does this make it not end up call oneDAL? Once a procedure with SYEVR is implemented, these will need to be revisited.

sklearnex/decomposition/pca.py
```diff
@@ -110,62 +116,319 @@ def __init__(
...
+                else:
+                    return "full"
+
+        else:
```
david-cortes-intel, line 281: Is it really necessary to have an 'else' here? I guess it's because the 'covariance_eigh' solver was introduced in a later sklearn version, but oneDAL will support it regardless of the sklearn version.

sklearnex/decomposition/pca.py
```diff
@@ -110,62 +116,319 @@ def __init__(
...
+            if xp is np:
+                copy = np.copy
             else:
-                return U, S, Vt
+                copy = lambda x: xp.asarray(x, copy=True)
+
+            self.components_ = copy(
+                self._onedal_estimator.components_[:n_components, ...]
+            )
+            # oneDAL PCA eigenvalues are not guaranteed to be >=0 and must be clipped.
```
david-cortes-intel, line 393: This should definitely be done on the oneDAL side. CC @Alexandr-Solovev .

sklearnex/preview/decomposition/incremental_pca.py
```diff
@@ -26,83 +28,147 @@
...
-    def __init__(self, n_components=None, *, whiten=False, copy=True, batch_size=None):
+    __doc__ = _sklearn_IncrementalPCA.__doc__
+
+    if sklearn_check_version("1.2"):
+        _parameter_constraints: dict = {
+            **_sklearn_IncrementalPCA._parameter_constraints,
+            "svd_solver": [StrOptions({"auto", "covariance_eigh", "onedal_svd"})],
+        }
+
+    def __init__(
```
david-cortes-intel, line 51: Shouldn't this have the same signature as sklearn?
https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.IncrementalPCA.html

sklearnex/preview/decomposition/incremental_pca.py
```diff
@@ -26,83 +28,147 @@
...
+        self.svd_solver = svd_solver
         self._need_to_finalize = False
+        # Note: use of the onedal_svd solver will cause partial result to grow proportionally
+        # to the input data and for that reason is not the default, which is contrary
+        # to the scikit-learn implementation.
 
     _onedal_incremental_pca = staticmethod(onedal_IncrementalPCA)
 
+    def _set_n_components(self, n_components, n_samples, n_features, first_pass):
+        # extracted from sklearn's ``IncrementalPCA.partial_fit``. This is a
```
david-cortes-intel, line 72: Could it just call the sklearn internal method, given that it inherits from the sklearn estimator?

sklearnex/decomposition/pca.py
```diff
@@ -110,62 +116,319 @@ def __init__(
...
+                # 3. solver is set to "full" and sklearn version < 1.5
+                # 4. solver is set to "auto" and dispatched to "full"
+                force_solver = self._fit_svd_solver == "full" and (
+                    not sklearn_check_version("1.5") or self.svd_solver == "auto"
+                )
+
+                patching_status.and_conditions(
+                    [
+                        (
+                            n_features < 2 * n_samples,
```
david-cortes-intel: Recap here: this would only apply when the underlying solver uses singular value decomposition instead of covariance+eigendecomposition, and it's also unclear whether it will fill with zeros or error out from the oneDAL side if passed such inputs.

Need to check whether the condition is still applicable. Also need to figure out which solver string will end up calling SVD from oneDAL side.

sklearnex/decomposition/pca.py
```diff
@@ -110,62 +116,319 @@ def __init__(
...
+                n_components = (
+                    n_sf_min if self.n_components is None else self.n_components
+                )
+                # This is matching aspects of sklearn.decomposition.PCA's ``_fit`` method
+                # Must be done this way as the logic hidden behind a ``validate_data`` call
+                # in sklearn cannot be reused without performance loss. This is likely to be
+                # high maintenance, but is written to be as simple and straightforward as
+                # possible.
+                if (
+                    sklearn_check_version("1.5")
```
david-cortes-intel, line 269: Recap here: when this condition is not met, it might end up still getting overridden later on.

sklearnex/decomposition/pca.py
```diff
@@ -110,62 +116,319 @@ def __init__(
...
-                return U, S, Vt, X, x_is_centered, xp
+            # This is necessary as get_namespace may return a bare numpy,
+            # which ``asarray`` may not contain the copy keyword argument.
+            if xp is np:
+                copy = np.copy
             else:
-                return U, S, Vt
+                copy = lambda x: xp.asarray(x, copy=True)
+
+            self.components_ = copy(
```
david-cortes-intel: Recap here: oneDAL internally does the conversion to have this in row-major order. This step might not be needed.

sklearnex/decomposition/pca.py
```diff
@@ -241,173 +500,70 @@ def inverse_transform(self, X):
...
-                )
+        @components_.deleter
+        def components_(self):
+            del self._components_
+
+        @property
+        def mean_(self):
+            return self._mean_
+
+        @mean_.setter
```
david-cortes-intel, line 529: Recap here: need to add tests checking that things work after modifying the attributes.

sklearnex/decomposition/pca.py
```diff
@@ -181,38 +444,34 @@ def transform(self, X):
...
+                    X,
+                    dtype=[xp.float64, xp.float32],
+                    reset=False,
+                )
 
             return self._onedal_estimator.predict(X, queue=queue)
 
-        def fit_transform(self, X, y=None):
-            if sklearn_check_version("1.5"):
-                U, S, Vt, X_fit, x_is_centered, xp = self._fit(X)
```
david-cortes-intel, line 201: Recap here: stock sklearn reuses the 'U' matrix during `fit_transform`, whereas oneDAL doesn't have this optimization. Would nevertheless be worth keeping this optimization from the sklearn side when oneDAL is not used.

icfaust: Sounds good, with the new separation of fit_transform the optimization will exist still by using the pure sklearn fit_transform: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/decomposition/_pca.py#L466

sklearnex/decomposition/pca.py
icfaust: Successful with removal, will address the one test case failure which is expected based on how we support 'covariance_eigh'

onedal/decomposition/incremental_pca.py
icfaust: Type promotion could have unintended consequences with respect to hysteresis and resultant fitted values, especially with the order of which the data is fed into the online estimator etc. I think we would be opening a can of worms. I think we are trading convenience for clarity.

sklearnex/decomposition/pca.py
icfaust: Thanks for the note, the value is explicitly discussed in the code comments in ```onedal_cpu_supported``` and ```onedal_gpu_supported``` that the value is forced for sklearnex at a later point. This is a pure replication of code for clarity and maintainability.

sklearnex/decomposition/pca.py
icfaust: Sounds good!  This is the stock logic for sklearn as when it switches to it's randomized approach.

sklearnex/decomposition/pca.py
icfaust: I assume this is about
```python
else:
    return "full"
```

This is to make sure that sklearn's logic is replicated exactly for ease of maintenance, but will  possibly emit a warning and swap in ```_onedal_fit``` as it is coerced to ```covariance_eigh```

sklearnex/preview/decomposition/incremental_pca.py
icfaust: Will change, but in general the signature match is only something we enforce on estimators which are out of preview.  The ```svd_solver``` is added from ```PCA``` in order to enable the oneDAL online SVD PCA solver (which is available on CPU via DAAL).

sklearnex/preview/decomposition/incremental_pca.py
icfaust: Is coupled with other calls which are computational expensive (i.e. the partial fit itself): https://github.com/scikit-learn/scikit-learn/blob/1fe659545c70d9f805c1c4097dd2fce9a6285a12/sklearn/decomposition/_incremental_pca.py#L296

sklearnex/decomposition/pca.py
Alexandr-Solovev: its done

onedal/decomposition/incremental_pca.py
Vika-F: Agree. It looks like there should be X.shape[0] instead of min(X.shape)

sklearnex/decomposition/pca.py
```diff
@@ -110,62 +116,319 @@ def __init__(
...
+            # post-process the number of components
+            if self.n_components is not None and not isinstance(
+                self.n_components, Integral
+            ):
+                n_components = self._postprocess_n_components()
+
+            # set attributes necessary for calls to transform, will modify
+            # self._onedal_estimator, and clear any previous fit models
+            # Follow guidance from sklearn PCA._fit_full and copy the data
+            self.n_components_ = n_components
```
david-cortes-intel, line 387: @icfaust What happens if the number of non-zero singular values is less than all of (a) user-supplied n_components, (b) number of rows in the data, (c) number of columns in the data ?

Could you please add a test where there's lots of duplicated or linearly dependent columns up to the point where such situation is encountered.

sklearnex/decomposition/pca.py
icfaust: @Alexandr-Solovev can you please link to the PR that changed it? I needed to add this here otherwise failures occur.  This is occurring using the nightly.

sklearnex/preview/decomposition/incremental_pca.py
icfaust: On review, the deviation from the sklearn signature is a feature not a bug and will not change, I had thought you had meant the signature beyond ```svd_solver``` was incorrect.

david-cortes-intel: In other cases where signatures differ, the extra parts are added at the end. Here, you are adding an additional argument before others. It could potentially cause issues if the initializer is called programmatically through low-level mechanisms.

sklearnex/decomposition/pca.py
david-cortes-intel: @Alexandr-Solovev Any comments about the current status of oneDAL's handling of small/negative eigenvalues?

sklearnex/decomposition/pca.py
david-cortes-intel: Dropping an example here in case it isn't very clear what kind of inputs would trigger this:
```python
import numpy as np
m = 4
n = 6
mu = rng.standard_normal(size=n)
S = rng.standard_normal(size=(n+1,n))
S = S.T@S
X = rng.multivariate_normal(mu, S, size=m)
X[:, 3:] = X[:, :3]
```

This data has 4 rows, 6 columns, and 3 non-zero singular values:
image

In this case, sklearn doesn't drop the last singular value even though it should, but at least it doesn't end up crashing or anything like that.

sklearnex/preview/decomposition/incremental_pca.py
icfaust: I'll push it to the end as requested.

onedal/decomposition/incremental_pca.py
icfaust: Changed.

onedal/decomposition/incremental_pca.py
icfaust: I have removed the logic entirely in contrast to my previous comment, if we want to address different datatype input affects on incremental algos, its worth making a ticket for them all and doing it in a structured way.

sklearnex/decomposition/pca.py
icfaust: Note, will be done as a follow up on oneDAL side.

sklearnex/preview/decomposition/incremental_pca.py
icfaust: Done.

sklearnex/decomposition/pca.py
icfaust: Added ```test_changed_estimated_attributes``` and ```test_create_model_behavior``` to test_incremental_pca.py and test_pca.py to test what was requested, and to test the ability to create_models.  A follow up would be to add this level of testing to other similar model-creating estimators like the linear regression algos.

sklearnex/decomposition/tests/test_pca.py
```diff
@@ -107,4 +110,55 @@ def test_non_batched_covariance(hyperparameters, dataframe, queue):
...
+    X = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+
+    est0 = PCA(n_components=4).fit(X_0)
+    est = clone(est0)
+    assert not hasattr(est, "_onedal_estimator")
+    est.fit(X)
+    assert not np.array_equal(_as_numpy(est.transform(X)), _as_numpy(est0.transform(X)))
+
+    # copy over parameters necessary for transform
+    est.mean_ = est0.mean_
```
david-cortes-intel, line 132: Not sure if supposed to work, but this crashes when replacing them with arrays of another type. For example:
```python
import dpnp
est.mean_ = dpnp.array(est0.mean_, device="gpu", dtype=np.float32)
est.components_ = dpnp.array(est0.components_, device="gpu", dtype=np.float32)
est.explained_variance_ = dpnp.array(est0.explained_variance_, device="gpu", dtype=np.float32)
```

david-cortes-intel:
The failing test here has been popping up in other PRs too, but I guess this would be the right PR to address it:
```
=================================== FAILURES ===================================
______________________ test_pca_dtype_preservation[full] _______________________

svd_solver = 'full'

    @pytest.mark.parametrize("svd_solver", PCA_SOLVERS)
    def test_pca_dtype_preservation(svd_solver):
>       check_pca_float_dtype_preservation(svd_solver)

decomposition/tests/test_pca.py:649: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
decomposition/tests/test_pca.py:669: in check_pca_float_dtype_preservation
    assert_allclose(pca_64.components_, pca_32.components_, rtol=2e-4)
utils/_testing.py:292: in assert_allclose
    np_assert_allclose(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<function assert_allclose.<locals>.compare at 0x7f62bf73beb0>, array([[-0.62022375, -0.15983497,  0.38316965,  0.6655...4071014,  0.9080595 , -0.21966814],
       [ 0.12497696,  0.8811203 , -0.16712739,  0.42435375]],
      dtype=float32))
kwds = {'equal_nan': True, 'err_msg': '', 'header': 'Not equal to tolerance rtol=0.0002, atol=0', 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Not equal to tolerance rtol=0.0002, atol=0
E           
E           Mismatched elements: 4 / 12 (33.3%)
E           Max absolute difference: 0.00014655
E           Max relative difference: 0.00087689
E            x: array([[-0.620224, -0.159835,  0.38317 ,  0.66555 ],
E                  [ 0.263179,  0.240851,  0.908007, -0.21966 ],
E                  [ 0.124977,  0.881087, -0.167274,  0.424366]])
E            y: array([[-0.620258, -0.159886,  0.38311 ,  0.66554 ],
E                  [ 0.263121,  0.24071 ,  0.908059, -0.219668],
E                  [ 0.124977,  0.88112 , -0.167127,  0.424354]], dtype=float32)

../../contextlib.py:79: AssertionError
```

david-cortes-intel (approved):
LGTM pending these two:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/2106#issuecomment-3187141823
https://github.com/uxlfoundation/scikit-learn-intelex/pull/2106#discussion_r2275615235

icfaust:
Issues related to pca dtype comparisons is underway in #2666 , the issues related to ```create_model``` are more extensive, as there is no prevention of trying to create a model from data on a non SYCL or CPU device, which would lead to odd failures in ```to_table``` and should be refactored for linear_regression, Ridge, and PCA algorithms simultaneously with centralized fixtures and testing.

PR #2108, opened 2024-10-14
TEST: increase threshold for memory usage on tests

david-cortes-intel:
## Description

ref https://github.com/intel/scikit-learn-intelex/pull/2107

This PR increases the threshold for the tests that check for memory usage increases to 25% of the input data size, as those tests were previously producing failures on some CI runs in that other PR.

Regardless of whether there are or aren't any memory leaks, it is not guaranteed that this condition would be met, since a memory allocator is not required to return all memory to the OS after freeing it, but this increase should allow the CI jobs to pass until a better solution is found.

Note that the PR contains an unrelated change - this is coming from the precommit hook, which required that change to go into the PR.

---

**Testing**

**Performance**


samir-nasibli (commented):
I think better to use this specific value for specific cases

david-cortes-intel:
> I think better to use this specific value for specific cases

Changed it use 25% for pandas inputs, while leaving it at 15% for other types of inputs.

Alexsandruss:
`Check PR Checklist / Close all checkboxes before moving from draft`

samir-nasibli (approved):
Looks good! Assuming green internal CI as well.
For me make sense also cover with ticket further investigation the reason of such number increase

sklearnex/tests/test_memory_usage.py
```diff
@@ -226,15 +227,20 @@ def _kfold_function_template(estimator, dataframe, data_shape, queue=None, func=
     else:
         name = estimator.__name__
 
+    threshold = (
+        EXTRA_MEMORY_THRESHOLD_PANDAS
+        if isinstance(dataframe, pd.DataFrame)
```
samir-nasibli, line 232: could it be `pd.Series` somehow?

david-cortes-intel: Thanks for pointing this out. It made me realize that I was actually checking the wrong parameter ("X" here would be the dataframe, whereas this "dataframe" is a string) - fixed it now.

PR #2109, opened 2024-10-15
]fox

icfaust:
## Description

---

**Testing**

**Performance**

### DIFF
--- a/sklearnex/linear_model/logistic_regression.py
+++ b/sklearnex/linear_model/logistic_regression.py
@@ -231,6 +231,7 @@ if daal_check_version((2024, "P", 1)):
                 "score",
             ]
 
+            check_is_fitted(self) #  Necessary for result sparsity checks
             class_name = self.__class__.__name__
             patching_status = PatchingConditionsChain(
                 f"sklearn.linear_model.{class_name}.{method_name}"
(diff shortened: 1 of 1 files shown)


ahuber21 (approved):
It may be a preferred alternative to use `issparse(getattr(self, "coeff_"))` and `issparse(getattr(self, "intercept_"))` instead.
I realize that right now there's no difference because there are no normal scenarios in which those attributes will be sparse. But we may want to come back to it later.

A second thought is that the order of operations
* check if GPU is supported
* check if model is fitted

could be swapped all together. That should also avoid the `AttributeError`.

I'm approving, asking you to merge if all CIs are green. Let's keep the comment above in mind for future refactorings.

icfaust:
> It may be a preferred alternative to use `issparse(getattr(self, "coeff_"))` and `issparse(getattr(self, "intercept_"))` instead. I realize that right now there's no difference because there are no normal scenarios in which those attributes will be sparse. But we may want to come back to it later.
> 
> A second thought is that the order of operations
> 
> * check if GPU is supported
> * check if model is fitted
> 
> could be swapped all together. That should also avoid the `AttributeError`.
> 
> I'm approving, asking you to merge if all CIs are green. Let's keep the comment above in mind for future refactorings.

Definitely agree. Checking model is fitted before other steps should be a rule.

PR #2110, opened 2024-10-15
[enhancement] enforce use of sklearn's ```check_is_fitted``` before checking oneDAL support

icfaust:
## Description

As a follow-up suggestion from #2109 , lets be consistent on where we call check_is_fitted.

This is stricter than sklearn's ```check_estimators_unfitted``` as it applies to anything that uses sklearnex's ```dispatch```.

---

**Testing**

**Performance**


sklearnex/tests/test_common.py
```diff
@@ -365,7 +365,26 @@ def runtime_property_check(text, estimator, method):
...
+            onedal_support = "_onedal_cpu_supported"
+        elif "_onedal_gpu_supported" in text["funcs"]:
+            onedal_support = "_onedal_gpu_supported"
+        else:
+            pytest.skip(f"onedal dispatching not used in {estimator}.{method}")
+        # get location of _onedal_*_supported
+        idx = len(text["funcs"]) - 1 - text["funcs"][::-1].index(onedal_support)
+        validfuncs = text["funcs"][:idx]
+        assert (
+            "check_is_fitted" in validfuncs
```
david-cortes-intel, line 380: Wouldn't it be better to use `mock` to ensure that the function is called in the test? Would also make it easier to understand what's going on during the tests.

david-cortes-intel: I mean the built-in `mock` module, by adding something like `@mock.patch("sklearn.utils.validation.check_is_fitted")` on top of the test (+ an extra argument for it), and then checking the property `called` on the mocked function.

icfaust: Very cool reference, it is definitely cleaner.

So your recommendation is as follows: refactor the general design tests procedure to instead of using trace to using a build-up fixture which collects the various spies, applies them, runs the estimator and method, and then stores the spies afterwards.

or run each estimator and method for this check_is_fitted test using a spy separate from the other procedures?

I guess I am trying to figure out how to get the ordering of various spies.

icfaust: I guess I am worried that as we add more design rules its going to become a slippery slope back to just wanting a full trace...

david-cortes-intel: I'm not sure about what would be the best way of arranging the tests.

As I see it, there's two places that use the "check that function was called" logic, which are for n_jobs and check_is_fitted. There's different ways that those could be added to the test, for example:
* Adding them as part of the test's body, by calling the "spy" initializer on the estimator fixture, which would need to move the `.fit` call to inside the body.
* Adding them as fixtures that would patch the function on initialize and unpatch at finalize, checking the function calls as part of the fixture finalization
* Adding them as a decorator to the test that would patch, unpatch, and assert.
* Turning the test into a class with an initializer that would fit the model and then the subsequent methods would test.

But that would leave the other design rules still as a list that gets checked inside the body, which might look inconsistent.

Not sure what's the best way here. My personal preference would be to have these two as fixtures and perhaps later on refactor the rest if needed.

samir-nasibli: Very good discussion. Thank you @david-cortes-intel for the suggestions. These could refactoring could be done as a follow -up.
Currently I am good with this impl, since it's consistent with other this kind of the design rules check.

sklearnex/svm/svr.py
```diff
@@ -98,6 +98,7 @@ def fit(self, X, y, sample_weight=None):
 
     @wrap_output_data
     def predict(self, X):
+        check_is_fitted(self)
```
samir-nasibli, line 101: many thanks for these alignments!

samir-nasibli (approved):

PR #2126, opened 2024-10-23
[enhancement] add oneDAL finiteness_checker implementation to onedal

icfaust:
## Description

A GPU side finiteness checker was added to oneDAL in 2024.7 (oneapi-src/oneDAL#2781), this becomes useful with the merge of #2045.  This will expose the first step, which is to include it in the onedal folder.  This may replace daal4py's assert_all_finite, but will occur in a later PR.  This adds limited testing on onedal side. This doesn't require performance benchmarking at this stage, but will be necessary in the next PR.

Aspects have been integrated in the following PRs:
#2152 
#2151

---

**Testing**

**Performance**


icfaust:
I believe the failures in private CI are unreleated.

icfaust:
Private CI fail unrelated to this implementation.

onedal/utils/tests/test_validation.py
```diff
@@ -0,0 +1,117 @@
+# ==============================================================================
```
Vika-F, line 1: A possibly dummy question:
Why this test needs to be added?
There is already https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/utils/tests/test_finite.py that tests about the same functionality.

icfaust: good question! that will be refactored in the next PR, this only creates the python oneDAL interface in the onedal folder and doesn't yet integrate it into sklearnex, so in short they test two different versions of ```_assert_all_finite```, the sklearnex tests are testing daal4py at the moment

onedal/datatypes/_data_conversion.py
samir-nasibli, line 1: I like refactoring for `to_table/from_table` primitives, but since this api used in several other PRs, I think it worth to move this common changes into separate PR, to be used on #2096, #2106, #2100

onedal/datatypes/_data_conversion.py
```diff
@@ -18,28 +18,43 @@
...
 
 def _apply_and_pass(func, *args, **kwargs):
     if len(args) == 1:
         return func(args[0], **kwargs)
     return tuple(map(lambda arg: func(arg, **kwargs), args))
 
 
-if _is_dpc_backend:
+def convert_one_to_table(arg):
+    return _backend.to_table(np.atleast_2d(arg) if np.isscalar(arg) else arg)
```
samir-nasibli, line 31: It worth to add a comment. Will this properly work with dpnp ndarrays and dpctl usm_ndarrays?

onedal/datatypes/_data_conversion.py
```diff
@@ -18,28 +18,43 @@
...
+    try:
         import dpnp
 
+        def _onedal_gpu_table_to_array(table, xp=None):
+            # By default DPNP ndarray created with a copy.
+            # TODO:
+            # investigate why dpnp.array(table, copy=False) doesn't work.
+            # Work around with using dpctl.tensor.asarray.
+            if xp == dpnp:
+                return dpnp.array(dpnp.dpctl.tensor.asarray(table), copy=False)
```
samir-nasibli, line 49: I don't know is this dpctl module in dpnp is an official part of API? This may work now, but I am not sure for further versions of DPNP. It worth to add some comments here

onedal/datatypes/_data_conversion.py
```diff
@@ -18,28 +18,43 @@
...
+def to_table(*args):
+    return _apply_and_pass(convert_one_to_table, *args)
 
-    if dpctl_available:
-        import dpctl.tensor as dpt
 
-    if dpnp_available:
+if _is_dpc_backend:
+
+    try:
```
samir-nasibli, line 40: please use dpnp_available from _dpep_helpers instead.

onedal/utils/validation.py
```diff
@@ -430,3 +435,31 @@ def _is_csr(x):
...
+    params = {
+        "fptype": "float" if X.dtype.name == "float32" else "double",
+        "method": "dense",
+        "allow_nan": allow_nan,
+    }
+    X_t = to_table(_convert_to_supported(policy, X))
+    if not _backend.finiteness_checker.compute.compute(policy, params, X_t).finite:
+        type_err = "infinity" if allow_nan else "NaN, infinity"
+        padded_input_name = input_name + " " if input_name else ""
+        msg_err = f"Input {padded_input_name}contains {type_err}."
```
samir-nasibli, line 451: Please cover with tests this message catch

onedal/common/_policy.py
```diff
@@ -48,12 +48,8 @@ def __init__(self):
 
 
 if _is_dpc_backend:
-    from onedal._device_offload import DummySyclQueue
```
samir-nasibli, line 51: This is the only place where `DummySyclQueue` is used. I am ok with this refactoring, but it worth also remove it definition as well. I didn't find the description for these changes, make sense to move into separate branch or add into commit messages on the merge this work as well, to be clear other dev/users why and when this were removed.

samir-nasibli (commented):
@icfaust Thank you for the excellent work! I have a few questions and clarifications, and I’ve left some comments

samir-nasibli:
@icfaust I believe it would be beneficial to include links to other dependent PRs in the description.

onedal/utils/validation.py
```diff
@@ -438,3 +443,31 @@ def _is_csr(x):
     return isinstance(x, sp.csr_matrix) or (
         hasattr(sp, "csr_array") and isinstance(x, sp.csr_array)
     )
+
+
+def _assert_all_finite(X, allow_nan=False, input_name=""):
+    policy = _get_policy(None, X)
+    params = {
+        "fptype": "float" if X.dtype.name == "float32" else "double",
```
samir-nasibli, line 451: What if `_convert_to_supported`  will convert it into float32 on the line 455? Does it make sense to move this params definition after _convert_to_support call?

onedal/utils/validation.py
```diff
@@ -438,3 +443,31 @@ def _is_csr(x):
...
+    X,
+    *,
+    allow_nan=False,
+    input_name="",
+):
+    _assert_all_finite(
+        X.data if sp.issparse(X) else X,
+        allow_nan=allow_nan,
+        input_name=input_name,
+    )
```
samir-nasibli, line 473: Is this main API for the feature? Is it used in the tests?

icfaust: Added test which uses the main function name, also testing sparse arrays as well.

onedal/utils/validation.py
icfaust: Changed. Thanks for pointing that out.

icfaust:
latest run: http://intel-ci.intel.com/efa730ba-581c-f119-ab8c-a4bf010d0e2d

onedal/datatypes/table.cpp
```diff
@@ -72,6 +72,10 @@ ONEDAL_PY_INIT_MODULE(table) {
         const auto column_count = t.get_column_count();
         return py::make_tuple(row_count, column_count);
     });
+    table_obj.def_property_readonly("dtype", [](const table& t){
+        // returns a numpy dtype, even if source was not from numpy
```
samir-nasibli, line 76: Why is this required? Will this be consistent for different array inputs?

onedal/utils/tests/test_validation.py
```diff
@@ -0,0 +1,149 @@
...
+from onedal.utils.validation import _assert_all_finite, assert_all_finite
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+@pytest.mark.parametrize(
+    "shape",
+    [
+        [16, 2048],
+        [
+            2**16 + 3,
```
samir-nasibli, line 37:
```suggestion
            65539, # 2**16 + 3,
```

samir-nasibli (commented):
Thank you @icfaust ! Looks good to me.
Left last minor comments

onedal/datatypes/table.cpp
icfaust: @samir-nasibli good question! This simplifies dtype checking for the backend handoff, what oneDAL cares about most is what the dtype of the tables to it are (i.e. the output of to_table). It is not guaranteed to match the dtype of the input into to_table, but it is guaranteed in matching the oneDAL table dtype.  It is by luck that dpnp/dpctl allow for dtype comparisons to numpy dtypes, which is not the case for array_api_strict inputs. By having a consistent numpy dtype to oneDAL tables in python, we will not have to worry about supporting all the different dtypes of all the various array inputs we will eventually support.

samir-nasibli: @icfaust Thank you for the clarification! I don't like this extension of table API which is wo for numpy and dpnp/dpctl.

I would suggest to solve the problem at python level https://github.com/intel/scikit-learn-intelex/blob/5bb54a520f5f3d1cf719e149df6526de943e393b/onedal/utils/validation.py#L452 without needing to extend table API

icfaust: @samir-nasibli I disagree, I think we should find someone to arbitrate on this.

samir-nasibli: Also you can name it as `numpy_dtype` instead of `dtype`, but again I don't see a lot of profit having this attr for the numeric table. Let's have a broader discussion.

ethanglaser: Could you clarify the situation that this would be helping? Like if an alternative array backend is used then the dtype of the input to to_table may not match the output of from_table?

icfaust: The large problem coming to us soon is handling zero-copy of arrays to the backend via the array_api standard.  Comparison to numpy-dtypes when an object has different dtypes (for example an array_api_strict float64 dtype is not equivalent to a numpy float64 dtype, and will yield a false). We will have to do one of two things:

adapt the often:
"float" if dtype==np.float64 else "double" which occurs when creating param dicts
or
always use the tables dtype, which has already been translated into a oneDAL, appropriate datatype.

By using numpy as the datatype, we are not having to re-invent the wheel with creating our own dtype standard, as pybind11 supports it natively in C++, its already all over the codebase in sklearnex/onedal.

Secondarily, I have talked about how actually our backend can ingest complex float and complex double types, but will treat them as float types or double types. See here:

https://github.com/intel/scikit-learn-intelex/pull/2172#issuecomment-2495364

onedal/utils/finiteness_checker.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+ONEDAL_PY_INIT_MODULE(finiteness_checker) {
+    using namespace dal::detail;
+    using namespace finiteness_checker;
+    using namespace dal::finiteness_checker;
+
+    using task_list = types<task::compute>;
+    auto sub = m.def_submodule("finiteness_checker");
+
+    #ifndef ONEDAL_DATA_PARALLEL_SPMD
```
ethanglaser, line 97: only instantiate if not data parallel spmd?

onedal/utils/tests/test_validation.py
```diff
@@ -0,0 +1,144 @@
...
+        [65539],  # 2**16 + 3,
+        [1000, 1000],
+        [
+            3,
+        ],
+    ],
+)
+@pytest.mark.parametrize("allow_nan", [False, True])
+@pytest.mark.parametrize(
+    "dataframe, queue", get_dataframes_and_queues("numpy,dpnp,dpctl")
```
ethanglaser, line 45: are there issues with pandas?

onedal/utils/tests/test_validation.py
```diff
@@ -0,0 +1,144 @@
...
+import numpy as np
+import numpy.random as rand
+import pytest
+import scipy.sparse as sp
+
+from onedal.tests.utils._dataframes_support import (
+    _convert_to_dataframe,
+    get_dataframes_and_queues,
+)
+from onedal.utils.validation import _assert_all_finite, assert_all_finite
```
ethanglaser, line 28: why both and not just assert_all_finite?

onedal/utils/tests/test_validation.py
```diff
@@ -0,0 +1,144 @@
...
+    X = rand.uniform(high=np.finfo(dtype).max, size=shape).astype(dtype)
+
+    if check:
+        loc = rand.randint(0, X.size - 1)
+        X.reshape((-1,))[loc] = float(check)
+
+    X = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+
+    if check is None or (allow_nan and check == "NaN"):
+        _assert_all_finite(X, allow_nan=allow_nan)
```
ethanglaser, line 85: are we checking that result of call is expected?

onedal/utils/tests/test_validation.py
icfaust: pandas inputs are not to be encountered by this function, check_array should always convert them to numpy inputs, we also do not support heterogeneous tables yet, which will be done at a later point.

onedal/utils/tests/test_validation.py
icfaust: _assert_all_finite will raise a failure if a non-finite occurs, meaning we are using the native functionality of the test in the test (i.e. it will raise an error if there is a problem, in this case, no error should be expected).

onedal/utils/finiteness_checker.cpp
icfaust: As of now there isn't an spmd implimentation in oneDAL.  If you forsee this as a problem, we will have to go back to oneDAL and add it (we can make that available for next release if important). @ethanglaser let me know.

onedal/utils/tests/test_validation.py
icfaust: This issue is handled in the follow up PR: #2177

Alexsandruss (approved):

PR #2130, opened 2024-10-24
DOC: Fix docs related to `n_jobs`

david-cortes-intel:
## Description

This PR fixes an issue in which the docs for classes that are not part of original scikit-learn were not setting the docs for `n_jobs` correctly and showing warnings about them, due to to three reasons:
* Using internal "glossary" link, which is from original scikit-learn but not present in sklearnex.
* Referencing objects from joblib without linking to them the way scikit-learn does.
* The decorator adding the parameter more than once to the docstrings if the keyword "Attributes" appeared more than once.

Before:
image
(note that glossary and backend are not clickable)

After:
image

---

**Testing**

**Performance**

Not applicable.

### DIFF
--- a/doc/sources/conf.py
+++ b/doc/sources/conf.py
@@ -68,7 +68,22 @@ extensions = [
     "sphinx_design",
     "sphinx_copybutton",
     "sphinx.ext.napoleon",
+    "sphinx.ext.intersphinx",
 ]
+
+intersphinx_mapping = {
+    "sklearn": ("https://scikit-learn.org/stable/", None),
+    # from scikit-learn, in case some object in sklearnex points to them:
+    # https://github.com/scikit-learn/scikit-learn/blob/main/doc/conf.py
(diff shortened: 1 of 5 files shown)


doc/sources/conf.py
```diff
@@ -68,8 +68,23 @@
     "sphinx_design",
     "sphinx_copybutton",
     "sphinx.ext.napoleon",
+    "sphinx.ext.intersphinx",
 ]
 
+intersphinx_mapping = {
```
icfaust, line 74: Wishful thinking: is there any way to get this from sklearn itself? I imagine this was a maintenance burden we just didn't track until your good work.

icfaust (commented):
otherwise looks good to me, just wondered something small.

doc/sources/conf.py
david-cortes-intel: I don't think so, because this is only part of the docs `conf.py` file, which is not part of the python package.

icfaust (approved):

PR #2131, opened 2024-10-25
ENH: Fix doc link in HTML representation of estimators

david-cortes-intel:
## Description

This PR fixes the interactive HTML representation of sklearnex estimators that don't have a scikit-learn analog to offer a clickable link to their docs, the same way scikit-learn estimators do, which is currently not working.

Before:
image

After:
image

Note: here I'm using the "latest" tag, as I wasn't sure how to get the link to the actual version of sklearnex that is used at runtime. Currently, this "latest" tag points to a 2023 release, so it won't work as-is until that issue is fixed.

---

**Testing**

**Performance**

Not applicable

### DIFF
--- a/sklearnex/_utils.py
+++ b/sklearnex/_utils.py
@@ -18,6 +18,7 @@ import logging
 import os
 import sys
 import warnings
+from abc import ABC
 
 from daal4py.sklearn._utils import (
     PatchingConditionsChain as daal4py_PatchingConditionsChain,
@@ -113,3 +114,19 @@ def register_hyperparameters(hyperparameters_map):
         return estimator_class
 
     return wrap_class
+
+
+# This abstract class is meant to generate a clickable doc link for classses
+# in sklearnex that are not present in scikit-learn. It should be inherited
+# before inheriting from a scikit-learn estimator, otherwise will get overriden
+# by the estimator's original.
+class NonSKLearnAlgorithm(ABC):
+    @property
+    def _doc_link_module(self) -> str:
+        return "sklearnex"
+
+    @property
+    def _doc_link_template(self) -> str:
+        module_path = ".".join(self.__class__.__module__.split(".")[:-1])
+        class_name = self.__class__.__name__
+        return f"https://intel.github.io/scikit-learn-intelex/latest/non-scikit-algorithms.html#{module_path}.{class_name}"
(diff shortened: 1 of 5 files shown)


sklearnex/_utils.py
```diff
@@ -113,3 +114,19 @@ def get_hyperparameters(self, op):
...
+# before inheriting from a scikit-learn estimator, otherwise will get overriden
+# by the estimator's original.
+class NonSKLearnAlgorithm(ABC):
+    @property
+    def _doc_link_module(self) -> str:
+        return "sklearnex"
+
+    @property
+    def _doc_link_template(self) -> str:
+        module_path = ".".join(self.__class__.__module__.split(".")[:-1])
```
icfaust, line 130:
```suggestion
        module_path, _ = ".".join(self.__class__.__module__.rsplit(".", 1))

```

sklearnex/_utils.py
```diff
@@ -113,3 +114,19 @@ def get_hyperparameters(self, op):
         return estimator_class
 
     return wrap_class
+
+
+# This abstract class is meant to generate a clickable doc link for classses
+# in sklearnex that are not present in scikit-learn. It should be inherited
+# before inheriting from a scikit-learn estimator, otherwise will get overriden
+# by the estimator's original.
+class NonSKLearnAlgorithm(ABC):
```
icfaust, line 123:
```suggestion
class IntelEstimator(ABC):

```

icfaust (commented):
Good PR, just some subtle stuff:
1) I think describing what something has rather than what it doesn't is better, also switch ```Algorithm``` to ```Estimator``` to follow sklearn convention. I'd prefer naming it after Intel in this case since the source is from Intel, though we could swap to oneDAL or something.
2) Just some small thing on using rsplit instead of split.

We can merge this today no problem.

sklearnex/_utils.py
david-cortes-intel: Renamed.

icfaust (approved):

sklearnex/_utils.py
```diff
@@ -113,3 +114,20 @@ def get_hyperparameters(self, op):
...
+# before inheriting from a scikit-learn estimator, otherwise will get overriden
+# by the estimator's original.
+class IntelEstimator(ABC):
+    @property
+    def _doc_link_module(self) -> str:
+        return "sklearnex"
+
+    @property
+    def _doc_link_template(self) -> str:
+        module_path, _ = ".".join(self.__class__.__module__.rsplit(".", 1))
```
icfaust, line 130:
```suggestion
        module_path, _ = self.__class__.__module__.rsplit(".", 1)

```
Sorry, made a mistake in my original suggestion. Totally my fault.

PR #2135, opened 2024-10-25
TEST: Fix json report for MPI

david-cortes-intel:
## Description

This PR is an attempt at fixing the JSON report generated by the test script that is ran with MPI.

Currently, the MPI call works in such a way that every process in the MPI call gets passed the argument to produce the JSON report, and each of them tries to write simultaneously to the same file, which results in clashes.

This PR should hopefully fix it by generating the JSON report only in the MPI process that has rank=0.

---

**Testing**

**Performance**

Not applicable

### DIFF
--- a/conda-recipe/run_test.bat
+++ b/conda-recipe/run_test.bat
@@ -74,7 +74,7 @@ if "%with_json_report%"=="1" (
     pytest --verbose "%1.ci\scripts\test_global_patch.py" || set exitcode=1
     if NOT "%NO_DIST%"=="1" (
         %PYTHON% -m pytest -k spmd --with-mpi --verbose --pyargs sklearnex || set exitcode=1
-        %PYTHON% -m pytest --with-mpi --verbose -s "%1tests\test_daal4py_spmd_examples.py" || set exitcode=1
+        %PYTHON% -m pytest --verbose -s "%1tests\test_daal4py_spmd_examples.py" || set exitcode=1
     )
 )
 
--- a/conda-recipe/run_test.bat
+++ b/conda-recipe/run_test.bat
@@ -19,7 +19,13 @@ rem %1 - scikit-learn-intelex repo root (should end with '\', leave empty if it'
 
 set exitcode=0
 
-IF NOT DEFINED PYTHON (set "PYTHON=python")
+IF NOT DEFINED PYTHON (
+    set "PYTHON=python"
+    set NO_DIST=1
+)
+if %PYTHON% EQU "python" (
+    set NO_DIST=1
+)
 
 %PYTHON% -c "from sklearnex import patch_sklearn; patch_sklearn()" || set exitcode=1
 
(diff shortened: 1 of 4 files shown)


conda-recipe/run_test.bat
```diff
@@ -1,4 +1,4 @@
-@echo off
+@echo on
```
Vika-F, line 1: Can you please attach the log examples "before" and "after" to this PR.

david-cortes-intel: It doesn't add any extra prints to the logs produced here. I _think_ it requires the caller script to not have supressed echo either.

When it works, this `@echo on` should  print the commands that it will execute before it executes them, but after variable substitutions, which can be helpful when testing things outside of the CI pipelines.

Vika-F: The comment is related not to this line, but to the whole PR. Sorry for the misunderstanding.
The description says that it fixes the clashes in the JSON repost in case of distributed execution.
So, please attach the reports "before" and "after" the changes.

tests/helper_mpi_tests.py
```diff
@@ -0,0 +1,25 @@
+# The purpose of this helper script is to be able to execute
```
Vika-F, line 1: Secure development lifecycle requirements do not allow to start the execution of some commands passed to the script without the additional safety checks to prevent the malware execution.

That's why the codefactor check is failing below.

From my understanding, it would be better not to have this helper script, but to modify `test_daal4py_spmd_examples.py` and other similar scripts. I.e. let them use JSON related arguments only at rank==0.

david-cortes-intel: Problem is, (a) the pytest plugin works in such a way that it requires being passed arguments in the same call to `pytest`, while the tests require running `pytest` under MPI, so it's not possible to pass different arguments to different MPI ranks within the same CLI call; (b) the file `test_daal4py_spmd_examples.py` is also used by the custom runner here: https://github.com/intel/scikit-learn-intelex/blob/main/tests/run_examples.py

Would like to hear some suggestions about how to better achieve the goal here (make only the first rank save a file, which is achieved by passing arguments to `pytest`) if you could think of others.

conda-recipe/run_test.bat
david-cortes-intel: I unfortunately don't have report files to attach.

But the issue is quite simple: multiple processes were writing to the same output file at the same time, which results in malformed json - as a simple example, if two processes both write "out" at the same time to the same file, then the output might be one of "outout", "oouutt", "ououtt", "ooutut", etc., which in the case of JSON text results in a file which doesn't even follow json format as `{}`, `:`, `[]` and so on are not respected.

tests/helper_mpi_tests.py
david-cortes-intel: > Secure development lifecycle requirements do not allow to start the execution of some commands passed to the script without the additional safety checks to prevent the malware execution.

Not sure what's the idea behind it - we have .sh files which run arbitrary commands already.

david-cortes-intel: Changed the `os.system` invocation to a `pytest.main` call.

tests/helper_mpi_tests.py
```diff
@@ -0,0 +1,34 @@
...
```
Vika-F, line 1: Please add a copyright notice.

david-cortes-intel: Added.

Vika-F (approved):
LGTM

PR #2147, opened 2024-10-31
[CI] make GitHub actions more robust by filtering events and using oneDAL build's lnx env

icfaust:
## Description

Two things are added:

1) Linux will download the ```oneDAL_env``` artifact which describes which external dependencies to download from apt (like MKL, TBB and compiler versions)

2) Filter the list of Nightly-runs for those triggered by 'workflow_dispatch' or 'schedule' which then opens up the possibility on oneDAL side to run the Nightly-build on PRs which are modifying capabilities in CI.  This means changes in oneDAL can be done in a less-blind fashion for their impact on the nightly (and will have a follow-up PR after this is merged). This means the Nightly-build can be used for checking when CI is changed in oneDAL.

---

**Testing**

**Performance**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,11 +51,6 @@ jobs:
     steps:
       - name: Checkout Scikit-learn-intelex
         uses: actions/checkout@v4
-      - name: Checkout oneDAL
-        uses: actions/checkout@v4
-        with:
-          repository: oneapi-src/oneDAL
-          path: oneDAL
       - name: Install Python
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
samir-nasibli, line 1: It make sense add also py311 into the matrix, since dpctl/dpnp already support py311.

.github/workflows/ci.yml
```diff
@@ -92,9 +96,11 @@ jobs:
       - name: dpcpp installation
         run: |
           # This CI system yields oneAPI dependencies from the oneDAL repository
-          bash ./oneDAL/.ci/env/apt.sh dpcpp
+          bash .ci/env/apt.sh dpcpp
```
samir-nasibli, line 99: Doesn't it worth update comment for this command then?

.ci/scripts/describe_system.sh
```diff
@@ -57,3 +57,8 @@ if [ -x "$(command -v icpx)" ]; then
     echo "ICPX:"
     icpx --version
 fi
+echo
+# SYCL devices
+if [ -x "$(command -v sycl-ls)" ]; then
+    sycl-ls
+fi
```
samir-nasibli, line 64: 👍

.github/workflows/ci.yml
icfaust: Good question @samir-nasibli, its still using the oneAPI dependencies from oneDAL, but its now using an artifact, as the main repo could be updated but the nightly build hasn't seen it yet, so the original comment is still valid.

.github/workflows/ci.yml
samir-nasibli: This enhancement could be done as follow up.

icfaust: I'd like one of the builds not to use dpctl/dpnp so as to test other pathways of queues for code-coverage purposes.

PR #2083, opened 2024-10-03
ENH: Enable accelerated linear regression for overdetermined systems

david-cortes-intel:
## Description

As a follow-up from oneDAL PR: https://github.com/oneapi-src/oneDAL/pull/2930

The PR above enables fitting linear regressions to overdetermined systems, both in regular regression and multi-output regression. Currently, the `sklearnex` estimators fall back to `scikit-learn` for overdetermined systems:
https://github.com/intel/scikit-learn-intelex/blob/2fccf444f800fb8ae1cce844b35727efc623e6d3/sklearnex/linear_model/linear.py#L180

But after the PR above, those should be supported in oneDAL too. Hence, this PR enables using oneDAL for those situations and adds tests for them.

One thing to note though: the mechanism in oneDAL is implemented as a fallback from Cholesky - that is, it will first perform a Cholesky factorization, and if that fails, will then use the fallback solver which can handle overdetermined systems. But if e.g. there are more columns that rows, we'll already know beforehand that Cholesky will invariably fail, and the procedure could be made faster if such kind of solver could be requested straight away from oneDAL. Implementation of it as a selectable solver is left for future PRs.

Note that since the oneDAL PR hasn't been merged at the time of writing, this PR will of course end up failing tests, but will keep it here in the meantime.

---

Checklist to comply with **before moving PR from draft**:

**PR completeness and readability**

- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes or created a separate PR with update and provided its number in the description, if necessary.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/intel/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have added a respective label(s) to PR if I have a permission for that.
- [x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [x] I have run it locally and tested the changes extensively.
- [x] All CI jobs are green or I have provided justification why they aren't.
- [x] I have extended testing suite if new functionality was introduced in this PR.

**Performance**

- [x] I have measured performance for affected algorithms using [scikit-learn_bench](https://github.com/IntelPython/scikit-learn_bench) and provided at least summary table with measured data, if performance change is expected.
- [x] I have provided justification why performance has changed or why changes are not expected.
- [x] I have provided justification why quality metrics have changed or why changes are not expected.
[...]

PR #2154, opened 2024-11-06
ENH: Moving Ridge out of preview

DDJHB:
## Description

This PR moves Ridge out of preview namespace to main sklearnex. It drives sklearnex Ridge away from daal4py by switching underlying implementation to onedal4py, which enables GPU support in turn. Minor usage of daal4py remains in case of individual (non-scalar) penalties since oneDAL still does not support such scenario. Performance comparison has been recorded on CPU:
image
Performance comparison on GPU is irrelevant since there was not real GPU implementation before. 
Separate ticket to update documentation has been created.

---

**Testing**

**Performance**


DDJHB:
Private CI errors seem unrelated

sklearnex/linear_model/tests/test_ridge.py
```diff
@@ -37,7 +41,7 @@ def test_sklearnex_import_ridge(dataframe, queue):
     ridge_reg = Ridge(alpha=0.5).fit(X_c, y_c)
 
     if daal_check_version((2024, "P", 600)):
-        assert "preview" in ridge_reg.__module__
+        assert "sklearnex" in ridge_reg.__module__
```
ethanglaser, line 44: Would it make sense to also assert that preview is not in the module string?

ethanglaser:
This should also be updated - https://github.com/intel/scikit-learn-intelex/blob/main/doc/sources/preview.rst?plain=1#L49-L76, remove ridge from bottom part and replace it with EmpiricalCovariance or IncrementalPCA in snippet

sklearnex/tests/test_n_jobs_support.py
```diff
@@ -49,7 +49,7 @@ def _check_n_jobs_entry_in_logs(records, function_name, n_jobs):
         if f"{function_name}: setting {expected_n_jobs} threads" in rec:
             return True
     # False if n_jobs is set and not found in logs
-    return n_jobs is None
+    return n_jobs is None or n_jobs == -1
```
ethanglaser, line 52: What is the reason for this?

sklearnex/linear_model/ridge.py
```diff
@@ -14,11 +14,416 @@
...
+            self.n_features_in_ = self._onedal_estimator.n_features_in_
+            self._coef = self._onedal_estimator.coef_
+            self._intercept = self._onedal_estimator.intercept_
+
+        fit.__doc__ = _sklearn_Ridge.fit.__doc__
+        predict.__doc__ = _sklearn_Ridge.predict.__doc__
+        score.__doc__ = _sklearn_Ridge.score.__doc__
+
+else:
+    from daal4py.sklearn.linear_model._ridge import Ridge
```
ethanglaser, line 420:
```suggestion
    from daal4py.sklearn.linear_model import Ridge

```

ethanglaser (commented):
Looking great!

onedal/linear_model/linear_model.py
```diff
@@ -293,15 +293,15 @@ def fit(self, X, y, queue=None):
         """
         module = self._get_backend("linear_model", "regression")
 
-        X = _check_array(
```
ethanglaser, line 296: Is this what was causing the slowdown between sklearnex and daal4py?

sklearnex/tests/test_n_jobs_support.py
DDJHB: this is a small bug i discovered with the n_jobs_support test - when you run it in batch with other tests (what happens in CI) it works perfectly, yet, run standalone - it fails on n_jobs=-1. upon little investigation found following:
n_jobs_wrapper in daal4py does not log anything since for -1 case it matches with old thread count which means nothing is to be set so nothing is rly logged, which fails the test since it relies on logs to pass. agreed with @icfaust to add this clause so whenever we have n_jobs=-1 and "return True" on line 50 did not trigger, meaning that no update to n_threads has been done, success is still returned for the check in the test to pass.

icfaust:
With the CI fixes merged, please rebase @DDJHB

doc/sources/preview.rst
```diff
@@ -49,11 +49,11 @@ For example, you can set the environment variable in the following way:
 Then, you can import Scikit-learn estimator patched with a preview one from `sklearnex.preview` module::
 
      from sklearnex import patch_sklearn
-     patch_sklearn()
-     from sklearn.linear_model import Ridge
-     print(Ridge.__module__)
+     patch_sklearn(preview=True)
```
ethanglaser, line 52:
```suggestion
     patch_sklearn()
```
arg should not needed due to env variable setting above

sklearnex/linear_model/tests/test_ridge.py
ethanglaser, line 1: Requires linting

doc/sources/preview.rst
DDJHB: yeah, did not notice env var setting in the file. quick question, why show customer a way to enable preview by setting env variable instead of explicit indication in code?

onedal/linear_model/linear_model.py
```diff
@@ -293,15 +293,15 @@ def fit(self, X, y, queue=None):
...
-            X,
-            dtype=[np.float64, np.float32],
-            force_all_finite=False,
-            ensure_2d=False,
-            copy=self.copy_X,
-        )
+        if not isinstance(X, np.ndarray):
```
icfaust, line 296: @samir-nasibli  Is there a ticket for Ridge Array_api support? I know you had mentioned you wanted this to be more in the direction of array_api, but I think maybe it should be split into 2 PRs - out of preview and array_api support.  If thats okay, could you make sure to send @DDJHB and me the ticket, or we can have @DDJHB make it.  You can see this code is rather numpy-oriented.

sklearnex/linear_model/ridge.py
```diff
@@ -14,11 +14,416 @@
...
+
+    from .._device_offload import dispatch, wrap_output_data
+    from .._utils import PatchingConditionsChain
+
+    if sklearn_check_version("1.6"):
+        from sklearn.utils.validation import validate_data
+    else:
+        validate_data = _sklearn_Ridge._validate_data
+
+    def _is_numeric_scalar(value):
```
icfaust, line 49: Still not sure why this was/is a separate function...

sklearnex/linear_model/ridge.py
```diff
@@ -14,11 +14,416 @@
...
+                        self.solver == "auto",
+                        f"'{self.solver}' solver is not supported. "
+                        "Only 'auto' solver is supported.",
+                    ),
+                    (
+                        not issparse(X) and not issparse(y),
+                        "Sparse input is not supported.",
+                    ),
+                    (
+                        not is_underdetermined,
```
icfaust, line 222: This is worth checking against https://github.com/intel/scikit-learn-intelex/pull/2083 where underdetermined checks are now only necessary for GPU.

sklearnex/linear_model/ridge.py
```diff
@@ -14,11 +14,416 @@
...
+else:
+    from daal4py.sklearn.linear_model import Ridge
+    from onedal._device_offload import support_input_format
+
+    Ridge.fit = support_input_format(queue_param=False)(Ridge.fit)
+    Ridge.predict = support_input_format(queue_param=False)(Ridge.predict)
+    Ridge.score = support_input_format(queue_param=False)(Ridge.score)
+
+    logging.warning(
+        "Preview Ridge requires oneDAL version >= 2024.6 but it was not found"
```
icfaust, line 428:
```suggestion
        "Ridge requires oneDAL version >= 2024.6 but it was not found"

```

sklearnex/linear_model/tests/test_ridge.py
```diff
@@ -100,3 +106,15 @@ def test_ridge_predict_before_fit(dataframe, queue):
 
         with pytest.raises(NotFittedError):
             model.predict(X_c)
+
```
icfaust, line 109: Maybe add some tests like from #2083?

icfaust (commented):
Some questions about overdetermined systems, a little bit about a warning.  Finally, we should make Docs green. Close to merge.

sklearnex/linear_model/ridge.py
DDJHB: single point of change for expression applied in different parts of the class

DDJHB: resolving as per our conversation @icfaust

doc/sources/preview.rst
ethanglaser: Good point and I tend to agree on setting using code instead of env variable. Any objections to this @Alexsandruss ?

david-cortes-intel:
> @david-cortes-intel does this PR look properly aligned with #2083 (specifically see changes from commit: [4e18d8a#diff-ed94dee2dafa5407c12f73420aaee87ee7748a2fedeeea8d4bd7b1d50fbf4a53](https://github.com/intel/scikit-learn-intelex/commit/4e18d8a8c5039d911d2c8a4f13d59bb8a13a4c60#diff-ed94dee2dafa5407c12f73420aaee87ee7748a2fedeeea8d4bd7b1d50fbf4a53))

Yes, changes look aligned - I had made a mistake in the condition namings which this PR is correcting.

onedal/linear_model/linear_model.py
icfaust: @samir-nasibli ping

samir-nasibli: Currently only numpy.ndarray array inputs are expected. No need for branching. I think now this branching is for avoiding sparse inputs.

onedal/linear_model/linear_model.py
```diff
@@ -293,15 +293,15 @@ def fit(self, X, y, queue=None):
...
+            X = np.asarray(X)
+
+        dtype = get_dtype(X)
+        if dtype not in [np.float32, np.float64]:
+            dtype = np.float64
+            X = X.astype(dtype, copy=self.copy_X)
 
-        y = np.asarray(y).astype(dtype=get_dtype(X))
+        y = np.asarray(y).astype(dtype=dtype)
```
samir-nasibli, line 304: This is still unclear to me why we have this perf drop here. How much this impl differ from daal4py CPU branch?

sklearnex/linear_model/ridge.py
```diff
@@ -14,11 +14,433 @@
...
+                super().__init__(
+                    alpha=alpha,
+                    fit_intercept=fit_intercept,
+                    normalize=normalize,
+                    copy_X=copy_X,
+                    max_iter=max_iter,
+                    tol=tol,
+                    solver=solver,
+                    random_state=random_state,
+                )
```
samir-nasibli, line 138: For the new estimators or functionality we are not supporting skl < 1.0

sklearnex/linear_model/tests/test_ridge.py
```diff
@@ -100,3 +106,48 @@ def test_ridge_predict_before_fit(dataframe, queue):
...
+
+@pytest.mark.skip(reason="Forced to use original sklearn for now.")
+def test_sklearnex_multivariate_ridge_alpha_shape():
+    from sklearnex.linear_model import Ridge
+
+    _test_multivariate_ridge_alpha_shape(Ridge, random_state=0)
+
+
+if daal_check_version((2025, "P", 100)):
```
david-cortes-intel, line 125: This one could be changed to `pytest.mark.skipif`.

sklearnex/linear_model/tests/test_ridge.py
```diff
@@ -100,3 +106,48 @@ def test_ridge_predict_before_fit(dataframe, queue):
...
+
+    @pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+    @pytest.mark.parametrize("overdetermined", [True, False])
+    @pytest.mark.parametrize("alpha", [0.1, 0.5, 1.0])
+    def test_ridge_overdetermined_system(dataframe, queue, overdetermined, alpha):
+        if not overdetermined and queue and queue.sycl_device.is_gpu:
+            pytest.skip(reason="Underdetermined system is not yet supported on GPU")
```
david-cortes-intel, line 132: I'm not 100% sure how ridge is implemented, but if it uses the same code as the non-regularized linear regression, you could just enable this in the fit conditions and it should work (provided the regularization is sufficiently higher than zero for the dtype), because the regularization would make the linear systems positive.

DDJHB: this is a very nice catch, thanks

doc/sources/preview.rst
Alexsandruss: I think it's better to replace env. variable setting with patching call but add note below for variable.

sklearnex/linear_model/tests/test_ridge.py
```diff
@@ -100,3 +106,48 @@ def test_ridge_predict_before_fit(dataframe, queue):
...
+            y = numpy.random.rand(10)
+
+        X_c = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+        y_c = _convert_to_dataframe(y, sycl_queue=queue, target_df=dataframe)
+        ridge_reg = Ridge(alpha=alpha, fit_intercept=False).fit(X_c, y_c)
+
+        # computing the coefficients manually
+        lambda_identity = alpha * numpy.eye(X.shape[1])
+        inverse_term = numpy.linalg.inv(numpy.dot(X.T, X) + lambda_identity)
+        xt_y = numpy.dot(X.T, y)
```
ethanglaser, line 150: Let's import numpy as np

sklearnex/linear_model/ridge.py
```diff
@@ -236,18 +231,21 @@ def _onedal_gpu_supported(self, method_name, *data):
...
-                patching_status.and_conditions(
-                    [
-                        (
-                            not is_underdetermined,
-                            "The shape of X (fitting) does not satisfy oneDAL requirements:"
-                            "Number of features + 1 >= number of samples.",
-                        ),
-                    ]
-                )
+                if isinstance(self.alpha, numbers.Real) and np.isclose(self.alpha, 0):
```
david-cortes-intel, line 234: I don't think `np.isclose` would suffice here. This regularizaation number is meant to be added to a large number which is the result of sums of squares. Therefore, at the very least, it should be larger than machine epsilon for the desired dtype, and likely higher in order to ensure that $ssq + \alpha > ssq$.

Perhaps 1e-5 could be a more reasonable minimum threshold.

sklearnex/linear_model/ridge.py
```diff
@@ -236,18 +231,21 @@ def _onedal_gpu_supported(self, method_name, *data):
...
+                    n_features = _num_features(data[0], fallback_1d=True)
+                    is_underdetermined = n_samples < (
+                        n_features + int(self.fit_intercept)
+                    )
+                    patching_status.and_conditions(
+                        [
+                            (
+                                not is_underdetermined,
+                                "The shape of X (fitting) does not satisfy oneDAL requirements:"
+                                "Number of features + 1 >= number of samples.",
```
david-cortes-intel, line 245: Here it could also mention the regularization being non-zero as another requirement.

sklearnex/linear_model/ridge.py
```diff
@@ -14,11 +14,374 @@
...
+
+            return patching_status
+
+        def _onedal_gpu_supported(self, method_name, *data):
+            patching_status = PatchingConditionsChain(
+                f"sklearn.linear_model.{self.__class__.__name__}.{method_name}"
+            )
+
+            if method_name == "fit":
+                if isinstance(self.alpha, numbers.Real) and np.isclose(self.alpha, 0):
```
david-cortes-intel, line 234: Also since the PR that fixes this for GPU was already merged (https://github.com/oneapi-src/oneDAL/pull/2976), perhaps you could also skip this check for version >= 2025.2.

icfaust (commented):
I hate to be the barer of bad news, please remove the following skips and rerun private CI:
https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/tests/test_patching.py#L147
https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/tests/test_patching.py#L138
https://github.com/intel/scikit-learn-intelex/blob/main/sklearnex/tests/test_memory_usage.py#L75

icfaust (commented):
@DDJHB Please merge main into your PR and rerun, some conventions in the codebase have changes, and some CI failures have been remedied.

david-cortes-intel (approved):
LGTM

icfaust (approved):

Alexsandruss (approved):

PR #2160, opened 2024-11-11
Update of DPC dependencies for CI and conda-recipe

Alexsandruss:
## Description

Changes:
- Update DPC compiler, DPC runtime libraries and oneDAL development files to 2025 version in CI
- Update DPCTL and DPNP versions and their installation channel to PyPI for environment consistency
- Pin linux kernel headers in conda recipe to reflect GLIBC>=2.28 dependency of dal-devel dpc part
- Move DPC runtime libraries to optional dependencies in conda recipe

---

**Testing**

**Performance**

N/A, CI update.

### DIFF
--- a/.ci/pipeline/build-and-test-lnx.yml
+++ b/.ci/pipeline/build-and-test-lnx.yml
@@ -20,7 +20,7 @@ steps:
       conda config --add channels conda-forge
       conda config --set channel_priority strict
       conda update -y -q conda
-      conda create -q -y -n CB -c conda-forge python=$(PYTHON_VERSION) dal-devel=2025 mpich pyyaml dpcpp_linux-64=2025.0 dpcpp-cpp-rt=2025.0
+      conda create -q -y -n CB -c conda-forge python=$(PYTHON_VERSION) dal-devel=2025.0 impi-devel=2021.14 pyyaml dpcpp_linux-64=2025.0 dpcpp-cpp-rt=2025.0
     displayName: "Conda create"
   - script: |
       . /usr/share/miniconda/etc/profile.d/conda.sh
--- a/.ci/pipeline/build-and-test-lnx.yml
+++ b/.ci/pipeline/build-and-test-lnx.yml
@@ -43,6 +43,7 @@ steps:
       . /usr/share/miniconda/etc/profile.d/conda.sh
       conda activate CB
       if [ $(SKLEARN_VERSION) != "1.0" ]; then conda install -q -y -c https://software.repos.intel.com/python/conda/ dpctl=0.18 dpnp=0.16; fi
+      conda uninstall --force numpy numpy-base
       bash .ci/scripts/setup_sklearn.sh $(SKLEARN_VERSION)
       pip install --upgrade -r requirements-test.txt
       pip install $(python .ci/scripts/get_compatible_scipy_version.py)
--- a/.ci/pipeline/build-and-test-lnx.yml
+++ b/.ci/pipeline/build-and-test-lnx.yml
@@ -27,7 +27,7 @@ steps:
       conda config --add channels conda-forge
       conda config --set channel_priority strict
       conda update -y -q conda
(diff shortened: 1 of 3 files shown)


.ci/pipeline/build-and-test-lnx.yml
```diff
@@ -48,7 +45,7 @@ steps:
       bash .ci/scripts/setup_sklearn.sh $(SKLEARN_VERSION)
       pip install --upgrade -r requirements-test.txt
       pip install $(python .ci/scripts/get_compatible_scipy_version.py)
-      if [ $(echo $(PYTHON_VERSION) | grep '3.9\|3.11') ] && [ $(SKLEARN_VERSION) != "1.0" ]; then conda install -q -y -c https://software.repos.intel.com/python/conda/ dpctl=0.17.0 dpnp=0.15.0; fi
+      if [ $(SKLEARN_VERSION) != "1.0" ]; then pip install dpctl==0.18.* dpnp==0.16.*; fi
```
samir-nasibli, line 48: Are we installing dpctl/dpnp for all py versions?

Alexsandruss: Yes, it has same supported python versions as sklearnex.

icfaust: that does, however, cause certain functionality within sklearnex to become untested (see DummySyclQueue)

ethanglaser: Is DummySyclQueue needed if dpctl supports all sklearnex versions?

ethanglaser: Are there circumstances at this point where dpctl would not be available?

samir-nasibli: I can give at least two examples:
2. Installation Issues: problems on the user’s side during the installation of dpctl.
3. Environment Compatibility: for example, dpctl does not currently support Python 3.13. While scikit-learn already supports Python 3.13 and we may soon add it to our matrices, dpctl will not be available for this Python version at this time.

conda-recipe/meta.yaml
```diff
@@ -58,29 +61,30 @@ requirements:
     - cython
     - jinja2
     - pybind11
-    - numpy {{ numpy }}
-    - impi-devel  # [not win]
+    - numpy
+    - mpich  # [not win]
```
ethanglaser, line 65: Why mpich instead of impi-devel?

ethanglaser (approved):
Thank you for resolving CI. preference would be to keep impi in meta.yaml if it makes sense.

conda-recipe/meta.yaml
Alexsandruss: It was for testing purposes, mpi version doesn't make difference in public build of conda-recipe.

conda-recipe/meta.yaml
```diff
@@ -58,15 +61,15 @@ requirements:
...
     - impi-devel  # [not win]
     # dal-devel pinning depends on the recipe location (repo or feedstock)
     - dal-devel
     # - dal-devel =={{ version }}
   run:
     - python
-    - {{ pin_compatible('numpy') }}
-    - dpcpp-cpp-rt  # [linux64]
+    - numpy
+    - scikit-learn
```
ethanglaser, line 72: Also, is scikit-learn needed in build deps or should it remain as test dep?

samir-nasibli: `scikit-learn` is only run time dependency.

icfaust (approved):
Before merge, @Alexsandruss could you leave a comment in the PR on the switch from conda to pip for dpctl/dpnp? This way when we face an upgrade issue with dpctl/dpnp in the future there is a bit of a note for someone to understand (doesn't need to be in the codebase, but something like research via git-blame to this PR would be nice)

icfaust:
Also thank you for doing this, very much appreciated @Alexsandruss

PR #2170, opened 2024-11-18
MAINT: Implement option to oneDAL's lib path to sklearnex's RPATH

david-cortes-intel:
## Description

This PR adds the path to the .so files of oneDAL to the RPATH of the generated .so files for `daal4py` and `onedal` in sklearnex, so that env. variables pointing to oneDAL would not need to be defined before the library can be loaded.

Note: this is an experimental PR, I'm not yet sure of the full implications.

---

**Testing**

**Performance**

Not applicable.


scripts/CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ elseif(UNIX)
         set(CMAKE_INSTALL_RPATH "@loader_path/../../../")
     else()
         set(ONEDAL_PY_LINK_OPTIONS "-Wl,-z,noexecstack,-z,relro,-z,now,-fstack-protector-strong,-fno-strict-overflow,-fno-delete-null-pointer-checks,-fwrapv")
-        set(CMAKE_INSTALL_RPATH "$ORIGIN/../../../")
+        set(CMAKE_INSTALL_RPATH "$ORIGIN/../../../:${oneDAL_LIBRARY_DIR}")
```
Alexsandruss, line 43: Usage of absolute paths has two issues if package is built for public distribution:
- It is non-portable
- It is absolute path on building machine which can be a security issue

david-cortes-intel: Thanks for looking into it.

What would you think of enabling it for non-release builds through an optional argument?

I think it shouldn't be much of an issue that it isn't portable, because it would still be able to load the library if it finds it elsewhere (e.g. in the first RPATH entry which is relative, or in `LD_LIBRARY_PATH`).

Alexsandruss: Optional argument is fine.

napetrov: More over we would need to ensure that we are not adding those accidentally to release builds

david-cortes-intel: Made it into an optional behavior that can be triggered by passing an argument `--dev` to the `setup.py` script.

I think this should be a sufficiently high barrier to avoid accidents, because it's not a recognized built-in argument for setuptools and thus calls to setup.py with that argument would error out before this PR.

setup.py
```diff
@@ -41,6 +41,13 @@
 from scripts.package_helpers import get_packages_with_tests
 from scripts.version import get_onedal_shared_libs, get_onedal_version
 
+USE_ABS_RPATH = False
+abs_rpath_args = ["abs-rpath", "-abs-rpath", "--abs-rpath"]
```
Alexsandruss, line 45: Why three types of prefix? Double is closest to a standard.

david-cortes-intel: Modified to accept only a double-dashed argument.

Alexsandruss (approved):

icfaust:
@david-cortes-intel does this add to the daal4py so file? im having trouble seeing it here and it says it does in the description

PR #2172, opened 2024-11-19
[Fix] add dtype attribute to python oneDAL table objects

icfaust:
## Description

Taken from #2126 convert_to_supported will set the data to float, but the fptype param may still be double.  This is most simply solved by querying the data for its dtype at the parameter setting stage. Ideally fptype should be instead managed entirely in the backend. Unfortunately, that change is very core to the backend offloading and should be done more carefully.

---

**Testing**

**Performance**


david-cortes-intel:
> convert_to_supported will set the data to float, but the fptype param may still be double

I'm not understanding the issue. What would be fp64 when the data is fp32?

ahuber21 (approved):

icfaust:
> > convert_to_supported will set the data to float, but the fptype param may still be double
> 
> I'm not understanding the issue. What would be fp64 when the data is fp32?

It isn't the main problem, the main problem is as follows: array_api dtypes are not supposed to match numpy dtypes. When we set the dtype of the oneDAL calculation, we are 

> > convert_to_supported will set the data to float, but the fptype param may still be double
> 
> I'm not understanding the issue. What would be fp64 when the data is fp32?

our tables can ingest complex float32 types (https://github.com/intel/scikit-learn-intelex/blob/main/onedal/datatypes/utils/numpy_helpers.hpp#L71) and will return float32 types: https://github.com/intel/scikit-learn-intelex/blob/main/onedal/datatypes/utils/numpy_helpers.cpp#L34, our checks for dtype are weak. We assume that dtype can only be of float32 or float64 which must be enforced at the very beginning of the code on sklearnex side using validate_data (just as an example from Linear Regression: https://github.com/intel/scikit-learn-intelex/blob/main/onedal/linear_model/linear_model.py#L46). We should use the datatype of the oneDAL table to describe how oneDAL is to operate.

david-cortes-intel:
> > > convert_to_supported will set the data to float, but the fptype param may still be double
> > 
> > 
> > I'm not understanding the issue. What would be fp64 when the data is fp32?
> 
> our tables can ingest complex float32 types (https://github.com/intel/scikit-learn-intelex/blob/main/onedal/datatypes/utils/numpy_helpers.hpp#L71) and will return float32 types: https://github.com/intel/scikit-learn-intelex/blob/main/onedal/datatypes/utils/numpy_helpers.cpp#L34, our checks for dtype are weak. We assume that dtype can only be of float32 or float64 which must be enforced at the very beginning of the code on sklearnex side using validate_data (just as an example from Linear Regression: https://github.com/intel/scikit-learn-intelex/blob/main/onedal/linear_model/linear_model.py#L46). We should use the datatype of the oneDAL table to describe how oneDAL is to operate.

Got it. But if complex numbers are the only issue, wouldn't it be better to create a helper function along the lines of "is_fp32_dtype" or so, and convert the data in python before it reaches oneDAL? It'd avoid double copies or conversions when the inputs are not in array formats (e.g. arrow[complex64] -> numpy[complex64] -> array[float32]).

Perhaps it could just use `np.info(data.dtype).dtype`.

onedal/datatypes/table.cpp
```diff
@@ -72,6 +72,10 @@ ONEDAL_PY_INIT_MODULE(table) {
         const auto column_count = t.get_column_count();
         return py::make_tuple(row_count, column_count);
     });
+    table_obj.def_property_readonly("dtype", [](const table& t){
```
samir-nasibli, line 75:
```suggestion
    table_obj.def_property_readonly("get_numpy_dtype", [](const table& t){
```
I suggest to use this or just `numpy_dtype` naming, just not to be confused

samir-nasibli: And again, from my point of view for tables API, this seems unnecessary and looks like a workaround or a hack.

PR #2185, opened 2024-11-23
[enhancement] remove contiguous check from ```_check_array```

icfaust:
## Description

This PR makes ```to_table``` handle non-contiguous arrays natively.  The checks in ```_check_array``` are hardcoded for numpy inputs, and are applied regardless of the circumstance. This makes the check seamless, will remove a blocker for other non-numpy data types, and will help in the rollout of the new finite checker, where a future refactor will remove ```_check_array``` entirely and enforce the use of ```validate_data``` and ```_check_sample_weight``` in sklearnex.  This also removes three TODOs listed in the codebase.

---

**Testing**

**Performance**

### DIFF
--- a/onedal/datatypes/data_conversion.cpp
+++ b/onedal/datatypes/data_conversion.cpp
@@ -155,9 +155,13 @@ dal::table convert_to_table(PyObject *obj) {
     }
     if (is_array(obj)) {
         PyArrayObject *ary = reinterpret_cast<PyArrayObject *>(obj);
-        if (!array_is_behaved_C(ary) && !array_is_behaved_F(ary)) {
+        if (!PyArray_ISCARRAY_RO(ary) && !PyArray_ISFARRAY_RO(ary)) {
             // NOTE: this will make a C-contiguous deep copy of the data
+            // this is expected to be a special case
             ary = PyArray_GETCONTIGUOUS(ary);
+            res = convert_to_table(ary);
+            Py_DECREF(ary);
+            return res;
         }
 #define MAKE_HOMOGEN_TABLE(CType) res = convert_to_homogen_impl<CType>(ary);
         SET_NPY_FEATURE(array_type(ary),
--- a/onedal/datatypes/data_conversion.cpp
+++ b/onedal/datatypes/data_conversion.cpp
@@ -159,9 +159,15 @@ dal::table convert_to_table(PyObject *obj) {
             // NOTE: this will make a C-contiguous deep copy of the data
             // this is expected to be a special case
             ary = PyArray_GETCONTIGUOUS(ary);
-            res = convert_to_table(reinterpret_cast<PyObject *>(ary));
-            Py_DECREF(ary);
-            return res;
+            if (ary) {
+                res = convert_to_table(reinterpret_cast<PyObject *>(ary));
+                Py_DECREF(ary);
+                return res;
+            } 
+            else {
+                throw std::invalid_argument(
+                "[convert_to_table] Numpy input could not be converted into onedal table.");
+            }
         }
 #define MAKE_HOMOGEN_TABLE(CType) res = convert_to_homogen_impl<CType>(ary);
         SET_NPY_FEATURE(array_type(ary),
--- a/onedal/datatypes/data_conversion.cpp
+++ b/onedal/datatypes/data_conversion.cpp
@@ -159,7 +159,7 @@ dal::table convert_to_table(PyObject *obj) {
             // NOTE: this will make a C-contiguous deep copy of the data
             // this is expected to be a special case
             ary = PyArray_GETCONTIGUOUS(ary);
-            res = convert_to_table(ary);
+            res = convert_to_table(reinterpret_cast<PyObject *>(ary));
             Py_DECREF(ary);
             return res;
         }
--- a/onedal/datatypes/data_conversion.cpp
+++ b/onedal/datatypes/data_conversion.cpp
@@ -155,18 +155,16 @@ dal::table convert_to_table(PyObject *obj) {
     }
     if (is_array(obj)) {
         PyArrayObject *ary = reinterpret_cast<PyArrayObject *>(obj);
-        if (array_is_behaved_C(ary) || array_is_behaved_F(ary)) {
+        if (!array_is_behaved_C(ary) && !array_is_behaved_F(ary)) {
+            // NOTE: this will make a C-contiguous deep copy of the data
+            ary = PyArray_GETCONTIGUOUS(ary);
+        }
 #define MAKE_HOMOGEN_TABLE(CType) res = convert_to_homogen_impl<CType>(ary);
-            SET_NPY_FEATURE(array_type(ary),
-                            array_type_sizeof(ary),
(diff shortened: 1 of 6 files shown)


onedal/utils/validation.py
```diff
@@ -153,15 +153,6 @@ def _check_array(
 
     if sp.issparse(array):
         return array
-
-    # TODO: Convert this kind of arrays to a table like in daal4py
-    if not array.flags.aligned and not array.flags.writeable:
```
icfaust, line 158: From numpy documentation: https://numpy.org/devdocs/dev/alignment.html only structured arrays will cause this, meaning that this is a non issue based on the dtypes of use for oneDAL (float32, float64, int32, and int64), and would failt to convert in table.cpp anyway.

david-cortes-intel: I think numpy also allows creating non-aligned arrays out of custom non-owned pointers, for example through [PyArray_SimpleNewFromData](https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_SimpleNewFromData), so in theory there could be a non-aligned float32 array or similar. But it'd be a very unlikely input, since default allocators in most platforms are aligned.

Also an easier conversion could be through [np.require](https://numpy.org/doc/stable/reference/generated/numpy.require.html).

icfaust: Good point, I'll probably because of this too I will make a stricter check on C++ side https://github.com/intel/scikit-learn-intelex/blob/main/onedal/decomposition/pca.py#L150 probably using https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_FromArray where I can specify the requirements to be aligned and owned (rather than use PyArray_GETCONTIGUOUS)

icfaust: With respect to non-ownership, I don't want to step on anyone's toes, as changes to PCA are occuring here which may inpact the ownership check there: https://github.com/intel/scikit-learn-intelex/pull/2106

onedal/datatypes/data_conversion_sua_iface.cpp
```diff
@@ -64,6 +67,23 @@ dal::table convert_to_homogen_impl(py::object obj) {
...
+            const auto space = obj.attr("__array_namespace__")();
+            copy = space.attr("asarray")(obj, "copy"_a = true);
+        }
+        else {
+            throw std::runtime_error("Wrong strides");
+        }
+        res = convert_to_homogen_impl<Type>(copy);
+        copy.dec_ref();
+        return res;
+    }
```
samir-nasibli, line 85: Please add comments here for other reviewers

onedal/datatypes/data_conversion.cpp
```diff
@@ -155,18 +155,20 @@ dal::table convert_to_table(PyObject *obj) {
     }
     if (is_array(obj)) {
         PyArrayObject *ary = reinterpret_cast<PyArrayObject *>(obj);
-        if (array_is_behaved_C(ary) || array_is_behaved_F(ary)) {
+        if (!PyArray_ISCARRAY_RO(ary) && !PyArray_ISFARRAY_RO(ary)) {
+            // NOTE: this will make a C-contiguous deep copy of the data
+            // this is expected to be a special case
+            ary = PyArray_GETCONTIGUOUS(ary);
```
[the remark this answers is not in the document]
icfaust: The Decref is necessary to prevent a memory leak (otherwise leaks in BasicStatitistics occur in all CIs), the memory is effectively owned by the table object, who will call its destructor with the pycapsule.  This work makes to_table support non-contiguous arrays, which will simplify the python coding occurring before to_table.  This work will enable ```_check_array``` to be moved from ```onedal``` to ```sklearnex``` simplifying the rollout of the new finite checker to SVM and neighbors algorithms. There is no error handling available (that I can find) in the numpy C-api as also numpy generally doesn't raise errors like that.  The checks that it is 1) a numpy array and 2) of certain aligned and type characteristics will prevent errors, and the use of the numpy c-api is very standardized.

onedal/datatypes/data_conversion_sua_iface.cpp
icfaust: done!

ahuber21 (approved):

samir-nasibli:
@icfaust please add proper description to the PR and share benchmark validation results

onedal/datatypes/data_conversion.cpp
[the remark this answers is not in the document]
david-cortes-intel: It looks like it can fail to allocate, in which case it will return `NULL`:
https://github.com/numpy/numpy/blob/cf9598572528318a54489b3c9ed5f65ef042e8c8/numpy/_core/src/multiarray/convert.c#L495

Alexsandruss (approved):
Please, provide detailed explanation of changes in next time.

icfaust:
Sorry about that, updated the description.

onedal/datatypes/data_conversion.cpp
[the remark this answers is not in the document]
icfaust: Ahhh thank you for doing that research @david-cortes-intel , that would mean in the case of a failure in allocation, it will fail to decref. I will add a check there.

onedal/datatypes/data_conversion_sua_iface.cpp
```diff
@@ -64,6 +67,25 @@ dal::table convert_to_homogen_impl(py::object obj) {
     // Get oneDAL Homogen DataLayout enumeration from input object shape and strides.
     const auto layout = get_sua_iface_layout(sua_iface_dict, r_count, c_count);
 
+    if (layout == dal::data_layout::unknown){
+        // NOTE: this will make a C-contiguous deep copy of the data
+        // if possible, this is expected to be a special case
+        py::object copy;
+        if (py::hasattr(obj, "copy")){
```
david-cortes-intel, line 74: Is it guaranteed that the copy will be C-contiguous? It will always be the case with numpy, but what about other packages?

ahuber21 (approved):
LGTM, please wait for green CI

onedal/datatypes/data_conversion_sua_iface.cpp
icfaust: as of now only dpctl and dpnp use this standard: 
https://github.com/IntelPython/dpnp/blob/master/dpnp/dpnp_container.py#L137 will specify "K" to dpctl.tensor's copy: https://github.com/IntelPython/dpctl/blob/master/dpctl/tensor/_copy_utils.py#L574 which if its not F aligned (which is the case) it will default to C alignment. asarray in dpctl https://github.com/IntelPython/dpctl/blob/master/dpctl/tensor/_ctors.py#L483 also has "K" as default.  We test this circumstance in the test that is modified in this PR for dpnp and dpctl.  If a new sycl_usm_namespace array type comes out, we can come back to this.

david-cortes-intel: What about other libraries that implement the array protocol? XArray also has "copy" for example:
https://docs.xarray.dev/en/latest/generated/xarray.DataArray.copy.html

PR #2201, opened 2024-12-02
[enhancement] Enable Array API in ensemble algos

icfaust:
## Description

This PR refactors the Ensemble algorithms (RandomForestRegressor, RandomForestClassifier, ExtraTreesRegressor and ExtraTreesClassifier) to follow repository standards and add array API support. This reduced the code by 500+ lines and required the following changes:

 - Remove `BaseEstimator` inheritance from onedal ensemble estimators
 - Change estimator `__init__` signatures to remove sklearn conformant kwargs in onedal ensemble estimators
 - Inline code comments added for function of various aspects for future maintenance
 - Remove `random_state` use from onedal estimators
 - Add `class_count` kwarg to `fit` as calculating it in python is scikit-learn conformance (oneDAL expects it a priori)
 - Remove input parameter checks from the onedal estimators
 - generalize return of out of bag values from `oneDAL` for use by Classifiers and Regressors
 - Remove unused `_create_model` function
 - Centralized `predict` method
 - Create `ForestRegressor` and `ForestClasssifier` objects to minimize maintenance
 - swap away from `max_samples` to `observations_per_tree_fraction` to follow oneDAL values
 - Modify tests for onedal to use numpy arrays (which can be consumed, where lists cannot)
 - Reorder warnings and errors based on type (e.g. parameter checks vs input checks etc.)
 - Refactor `_save_attributes` method to be specific to Classifiers vs Regressors
 - Refactor `_onedal_fit_ready`, `_onedal_cpu_supported` and `_onedal_gpu_supported` to reduce code duplication via inheritance and make array API enabled
 - Add `enable_array_api` decorators to public-facing estimators
 - Place `_check_parameters` function behind `sklearn_check_version` for future removal
 - Remove check for `min_impurity_split` which was removed in sklearn 0.25
 - Add array API-enabled `_validate_y_class_weight` method designed specifically for sklearnex estimators (missing some functionality which is irrelevant to the sklearnex estimator)
 - Remove `check_n_features` from `sklearnex.utils.validation` as it is no longer necessary
 - Enable weighted fitting support for gpu
 - Removed `sample_weight` checks for sparsity (blocked by `_check_sample_weight`)
 - Added documentation to the nature of set attributes and array API support

---

**Testing**

### DIFF
--- a/doc/sources/conf.py
+++ b/doc/sources/conf.py
@@ -272,6 +272,7 @@ notfound_urls_prefix = "/scikit-learn-intelex/"
 linkcheck_ignore = [
     re.compile(r"https://github\.com/.+"),  # Avoid rate error
     re.compile(r"https://.*intel\.com/.+"),  # Avoid permission error
+    re.compile(r"https://medium\.com/.*"),  # Avoid Medium paywall/rate limit
 ]
 
 # Speed up link-check and fail faster


sklearnex/tests/utils/base.py
```diff
@@ -396,8 +396,6 @@ def _get_processor_info():
 
 
 class DummyEstimator(BaseEstimator):
-
-    def fit(self, X, y=None):
```
david-cortes-intel, line 402: This looks like an accidental change.

onedal/ensemble/forest.py
```diff
@@ -200,81 +176,6 @@ def _get_onedal_params(self, data):
             onedal_params["splitter_mode"] = self.splitter_mode
         return onedal_params
 
-    def _check_parameters(self):
-        if isinstance(self.min_samples_leaf, numbers.Integral):
-            if not 1 <= self.min_samples_leaf:
-                raise ValueError(
-                    "min_samples_leaf must be at least 1 "
```
david-cortes-intel, line 207: I see there would be no such message issued now. Was this originally an error message thrown by sklearn? If so, should it retain it for some older version of sklearn?

sklearnex/ensemble/_forest.py
```diff
@@ -87,7 +87,7 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
                 y,
                 multi_output=True,
                 accept_sparse=False,
-                dtype=[np.float64, np.float32],
+                dtype=[xp.float64, xp.float32],
```
david-cortes-intel, line 103: Shouldn't the call to get this 'xp' pass all of the data arrays? (X+y+sample_weight)

sklearnex/ensemble/_forest.py
```diff
@@ -337,6 +375,10 @@ def estimators_(self, estimators):
         self._cached_estimators_ = estimators
 
     def _estimators_(self):
+        """This attribute provides lazy creation of scikit-learn conformant
+        Decision Trees used for analysis in such as 'apply'. This will stay
```
david-cortes-intel:
```suggestion
        Decision Trees used for analysis in methods such as 'apply'. This will stay
```

sklearnex/ensemble/_forest.py
```diff
@@ -581,26 +619,17 @@ def _onedal_fit_ready(self, patching_status, X, y, sample_weight):
...
+                        f"Number of outputs is not 1.",
                     ),
                     (
-                        y.dtype in [np.float32, np.float64, np.int32, np.int64],
+                        y.dtype in [xp.float32, xp.float64, xp.int32, xp.int64],
                         f"Datatype ({y.dtype}) for y is not supported.",
                     ),
                 ]
             )
             # TODO: Fix to support integers as input
```
david-cortes-intel: Would this be implemented with the changes in this PR?

icfaust: Not sure yet. we will see.

onedal/ensemble/forest.py
icfaust: This is all sklearn conformance that needs to be transferred to the sklearnex estimator for easier maintenance.  I will go about making sure CIs pass, but as of now I am starting with stripping down the onedal estimator.

sklearnex/ensemble/_forest.py
```diff
@@ -239,86 +426,95 @@ def _to_absolute_max_features(self, max_features, n_features):
...
+                        "greater than 1 or a float in (0.0, 1.0]; "
+                        "got the float %s" % self.min_samples_split
+                    )
+            if not 0 <= self.min_weight_fraction_leaf <= 0.5:
+                raise ValueError("min_weight_fraction_leaf must in [0, 0.5]")
+            if hasattr(self, "min_impurity_split"):
+                warnings.warn(
+                    "The min_impurity_split parameter is deprecated. "
+                    "Its default value has changed from 1e-7 to 0 in "
+                    "version 0.23, and it will be removed in 0.25. "
```
david-cortes-intel: Since these versions are not supported, could this validation code be simplified?

icfaust: removed.

sklearnex/tests/utils/base.py
icfaust: removed

sklearnex/ensemble/_forest.py
icfaust: Will not be done.

david-cortes-intel:
I'm not sure if this is a bug in DPNP or an issue with this PR where not all inputs are converted as necessary, but I see this error if I try to make predictions on numpy arrays from a model that was fitted to dpnp arrays:
```
File /export/users/dcortes/repos/sklex-dpnp/sklearnex/ensemble/_forest.py:867, in ForestClassifier._onedal_predict(self, X, queue)
    864 res = self._onedal_estimator.predict(X, queue=queue)
    866 if is_array_api_compliant:
--> 867     return xp.take(self.classes_, xp.astype(xp.reshape(res, (-1,)), xp.int64))
    868 else:
    869     return xp.take(self.classes_, res.ravel().astype(xp.int64, casting="unsafe"))
````

Reproducer:
```python
import os, sys
import numpy as np
import dpnp
from sklearnex import config_context, set_config
from sklearnex.ensemble import RandomForestClassifier
from sklearn.datasets import make_classification
X, y = make_classification(random_state=123)
Xd = dpnp.array(X, dtype=np.float32, device="cpu")
yd = dpnp.array(y, dtype=np.float32, device="cpu")

set_config(array_api_dispatch=True)
model = RandomForestClassifier(n_estimators=1, max_depth=5).fit(Xd, yd)
model.predict(X[:5])
```

david-cortes-intel:
@icfaust Since the equivalent classes in sklearn do not have array API support: how is this meant to work for attributes? I see that the arrays in the `Tree` object class for example are always numpy regardless of what was passed as inputs, while other internal attributes appear to have the array API class that was used as input, with data being on the corresponding device.

I guess this might be intended, but if that is the case, please add this kind of behavior to the documentation.

sklearnex/ensemble/_forest.py
```diff
@@ -144,69 +196,214 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
...
+            patching_status.and_conditions(
+                [
+                    (
+                        daal_check_version((2023, "P", 200))
+                        or self.estimator.__class__ == DecisionTreeClassifier,
+                        "ExtraTrees only supported starting from oneDAL version 2023.2",
+                    ),
+                    (
+                        not sp.issparse(data[2]),
+                        "sample_weight is sparse. " "Sparse input is not supported.",
```
david-cortes-intel: Was this perhaps meant to have an additional check on 'X'?

icfaust:
> I'm not sure if this is a bug in DPNP or an issue with this PR where not all inputs are converted as necessary, but I see this error if I try to make predictions on numpy arrays from a model that was fitted to dpnp arrays:
> 
> ```
> File /export/users/dcortes/repos/sklex-dpnp/sklearnex/ensemble/_forest.py:867, in ForestClassifier._onedal_predict(self, X, queue)
>     864 res = self._onedal_estimator.predict(X, queue=queue)
>     866 if is_array_api_compliant:
> --> 867     return xp.take(self.classes_, xp.astype(xp.reshape(res, (-1,)), xp.int64))
>     868 else:
>     869     return xp.take(self.classes_, res.ravel().astype(xp.int64, casting="unsafe"))
> ```
> 
> Reproducer:
> 
> ```python
> import os, sys
> os.environ["SCIPY_ARRAY_API"] = "1"
> import numpy as np
> import dpnp
> from sklearnex import config_context, set_config
> from sklearnex.ensemble import RandomForestClassifier
> from sklearn.datasets import make_classification
> X, y = make_classification(random_state=123)
> Xd = dpnp.array(X, dtype=np.float32, device="cpu")
> yd = dpnp.array(y, dtype=np.float32, device="cpu")
> 
> set_config(array_api_dispatch=True)
> model = RandomForestClassifier(n_estimators=1, max_depth=5).fit(Xd, yd)
> model.predict(X[:5])
> ```

@david-cortes-intel  #2747  This is an issue beyond just this estimator, and should be addressed as a follow up with additional testing added in a central way.

sklearnex/ensemble/_forest.py
```diff
@@ -58,38 +56,58 @@
...
+else:
+    _check_array = partial(
+        check_array,
+        force_all_finite=False,
+        dtype=None,
+        ensure_2d=False,
+        ensure_min_samples=0,
+        ensure_min_features=0,
+        accept_sparse=True,
+    )
```
ahuber21:
```suggestion
__check_kwargs = {
  "dtype": None,
  "ensure_2d": False,
  "ensure_min_samples": 0,
  "ensure_min_features": 0,
  "accepts_sparse": True,
}
if sklearn_check_version("1.6"):
  __check_kwargs["ensure_all_finite"] = False
else:
  __check_kwargs["force_all_finite"] = False
_check_array = partial(check_array, **__check_kwargs)
```

Or any other pattern that avoids the duplication...

sklearnex/ensemble/_forest.py
```diff
@@ -119,22 +144,49 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
...
+            warnings.warn(
+                "Setting 'random_state' value is not supported. "
+                "State set by oneDAL to default value (777).",
+                RuntimeWarning,
+            )
+
+        rs = check_random_state(self.random_state)
+        # use numpy here due to lack of array API support in sklearn random state
+        # seed is a python integer
+        seed = rs.randint(0, np.iinfo("i").max)
```
ahuber21, line 181: This means the seed is generated from `random_state` and, given deterministic `random_state`, will be deterministic, too?
Everything else would be confusing, especially in tests.

sklearnex/ensemble/_forest.py
```diff
@@ -1226,11 +1043,32 @@ def score(self, X, y, sample_weight=None):
             sample_weight=sample_weight,
         )
 
+    def _onedal_predict(self, X, queue=None):
+        check_is_fitted(self, "_onedal_estimator")
+        xp, _ = get_namespace(X)
+        use_raw_input = get_config().get("use_raw_input", False) is True
+
+        if not use_raw_input:
```
ahuber21, line 1045: Wouldn't now be a good time to remove raw input support again?

ahuber21 (commented):
Only minor questions.

ethanglaser:
Please address real private CI fails:
test_estimators[ExtraTreesClassifier()-check_estimators_nan_inf]
test_estimators[ExtraTreesRegressor()-check_estimators_nan_inf]
test_estimators[RandomForestClassifier()-check_estimators_nan_inf]
test_estimators[RandomForestRegressor()-check_estimators_nan_inf]

before merging

onedal/ensemble/tests/test_random_forest.py
```diff
@@ -74,9 +79,9 @@ def test_rf_classifier_random_splitter(queue):
         shuffle=False,
     )
     rf = RandomForestClassifier(max_depth=2, random_state=0, splitter_mode="random").fit(
-        X, y, queue=queue
+        X, y, class_count=2, queue=queue
```
ethanglaser, line 82: Just curious - why does class_count need to be specified?

onedal/ensemble/tests/test_random_forest.py
```diff
@@ -44,20 +49,20 @@ def test_rf_regression(queue):
     X, y = make_regression(
         n_samples=100, n_features=4, n_informative=2, random_state=0, shuffle=False
     )
-    rf = RandomForestRegressor(max_depth=2, random_state=0).fit(X, y, queue=queue)
+    rf = RandomForestRegressor(max_depth=2, random_state=209652396).fit(X, y, queue=queue)
```
ethanglaser, line 52: Can this be restored to 0? Nearly all of our other tests use 0

sklearnex/ensemble/_forest.py
```diff
@@ -144,69 +196,214 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
...
-            "oob_score": self.oob_score,
-            "n_jobs": self.n_jobs,
-            "random_state": self.random_state,
-            "verbose": self.verbose,
-            "warm_start": self.warm_start,
-            "error_metric_mode": self._err if self.oob_score else "none",
-            "variable_importance_mode": "mdi",
-            "class_weight": self.class_weight,
+            "random_state": seed,
+            "observations_per_tree_fraction": (
```
ethanglaser, line 199: I am not super familiar with this parameter but it is not present in sklearn's estimator so not sure if we want to add it here. Anyone familiar with it?

sklearnex/ensemble/_forest.py
```diff
@@ -119,22 +144,49 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
...
             # try catch needed for raw_inputs + array_api data where unlike
             # numpy the way to yield unique values is via `unique_values`
             # This should be removed when refactored for gpu zero-copy
-            try:
-                self.classes_ = xp.unique(y)
-            except AttributeError:
-                self.classes_ = xp.unique_values(y)
-            self.n_classes_ = len(self.classes_)
-        self.n_features_in_ = X.shape[1]
+            if hasattr(self, "class_weight"):
```
ethanglaser: Can we leverage sklearn's is_classifier functionality here? Assuming the intention for this conditional to identify whether classifier

sklearnex/ensemble/_forest.py
```diff
@@ -58,38 +56,58 @@
...
                 X,
                 y,
                 multi_output=True,
                 accept_sparse=False,
-                dtype=[np.float64, np.float32],
-                ensure_all_finite=False,
-                ensure_2d=True,
+                dtype=[xp.float64, xp.float32],
+                ensure_all_finite=False,  # completed in offload check
+                y_numeric=not hasattr(self, "predict_proba"),  # trigger for Regressors
```
ethanglaser: bumping:
Can we leverage sklearn's is_classifier functionality here? Assuming the intention for this conditional to identify whether classifier

sklearnex/ensemble/_forest.py
```diff
@@ -104,12 +122,19 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
                 stacklevel=2,
             )
 
+        if not sklearn_check_version("1.2") and self.criterion == "mse":
+            warnings.warn(
+                "Criterion 'mse' was deprecated in v1.0 and will be "
+                "removed in version 1.2. Use `criterion='squared_error'` "
+                "which is equivalent.",
+                FutureWarning,
+            )
```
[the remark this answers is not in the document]
david-cortes-intel: AI comment is wrong.

sklearnex/ensemble/_forest.py
david-cortes-intel: Wouldn't it just be a different way of specifying the numbers that is calculable from sklearn's parameters?

sklearnex/ensemble/_forest.py
```diff
@@ -144,69 +196,214 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
...
+
+    def _save_attributes(self, xp):
+        if self.oob_score:
+            self.oob_score_ = self._onedal_estimator.oob_error_
+
+        self._validate_estimator()
+
+    def _onedal_fit_ready(self, patching_status, X, y, sample_weight):
+        # Normally this would be handled in _onedal_fit, but this condition is required
+        # here for scikit-learn conformance as following checks require non-sparse data
```
david-cortes-intel: Wouldn't it still be conformant if this is added to the patching conditions and it lets sklearn throw the error instead?

sklearnex/ensemble/_forest.py
```diff
@@ -222,14 +419,14 @@ def _to_absolute_max_features(self, max_features, n_features):
                             FutureWarning,
                         )
                     return (
-                        max(1, int(np.sqrt(n_features)))
+                        max(1, int(math.sqrt(n_features)))
```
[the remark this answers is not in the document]
david-cortes-intel: AI comment can be ignored.

onedal/ensemble/forest.py
```diff
@@ -14,41 +14,24 @@
...
-    _check_array,
-    _check_n_features,
-    _check_X_y,
-    _column_or_1d,
-    _validate_targets,
-)
 
 
-class BaseForest(BaseEnsemble, metaclass=ABCMeta):
+class BaseForest(metaclass=ABCMeta):
```
david-cortes-intel:
```suggestion
class BaseForest(ABC):
```
I think this is more idiomatic. Result is the same though.

onedal/ensemble/forest.py
```diff
@@ -186,259 +128,102 @@ def _get_onedal_params(self, data):
...
+            result = self.infer(params, hparams.backend, self._onedal_model, X)
         else:
-            result = self.infer(params, model, X)
+            result = self.infer(params, self._onedal_model, X)
 
         # TODO: fix probabilities out of [0, 1] interval on oneDAL side
         pred = from_table(result.probabilities)
-        return pred.clip(0.0, 1.0)
+        pred[pred > 1.0] = 1.0
+        pred[pred < 0.0] = 0.0
```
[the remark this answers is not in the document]
david-cortes-intel: Array API does offer clip, so suggestion is valid:
https://data-apis.org/array-api/latest/API_specification/generated/array_api.clip.html

sklearnex/ensemble/_forest.py
```diff
@@ -144,69 +196,214 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
...
+            sample_weight,
+            getattr(self, "n_classes_", 0),
+            queue=queue,
+        )
+
+        self._save_attributes(xp)
+
+        return self
+
+    def _save_attributes(self, xp):
```
david-cortes-intel, line 231: Is argument `xp` used for something here?

sklearnex/ensemble/_forest.py
```diff
@@ -466,166 +664,137 @@ def _estimators_(self):
...
 
-            self.n_outputs_ = y.shape[1]
+        patching_status = super()._onedal_fit_ready(patching_status, X, y, sample_weight)
+
+        if patching_status.get_status():
+            xp, is_array_api_compliant = get_namespace(X, y, sample_weight)
+
+            try:
+                # properly verifies all non array API inputs without conversion
+                correct_target = type_of_target(y) in ["binary", "multiclass"]
```
david-cortes-intel, line 730: Looks like this sklearn function is already array-api-compliant and also has a try-catch for IndexError:
https://github.com/scikit-learn/scikit-learn/blob/6a9ce612bcc0a75bfdeca5a4f6fc626e88b89d9b/sklearn/utils/multiclass.py#L381

david-cortes-intel:
The CI error for sklearn==1.0.2 on `ExtraTreesClassifier`:
```
>               raise AssertionError(err_msg)
E               AssertionError: Estimator doesn't check for NaN and inf in fit.
```

Looks like something here is not working correctly.

sklearnex/ensemble/_forest.py
icfaust: @david-cortes-intel is right on this, and this question is highlight a big point of something, namely that we are not bound in any way to match sklearn inputs in onedal module estimators. We should be using oneDAL-native parameters in our calls to onedal objects, as if we do the translation of sklearn parameters in the onedal module, then it is subject to influence from sklearn and is therefore a maintenance burden.

onedal/ensemble/tests/test_random_forest.py
icfaust: The seed state calculation was being improperly used from sklearn in the onedal side of things. I can swap it, but then I would need to change the results of the test. I can do that, just let me know.

onedal/ensemble/tests/test_random_forest.py
icfaust: We've been doing sneaky stuff in our classifier estimator implementations in sklearnex. oneDAL expects that we know *a priori* the number of classes (https://uxlfoundation.github.io/oneDAL/api/algorithms/ensembles/decision-forest.html#_CPPv4N6oneapi3dal15decision_forest10descriptor11class_countE) which we've been relying on sklearn to calculate. That functionality should either be natively implemented in oneDAL, or its sklearn conformance (which means it should not exist in the oneDAL estimator).

sklearnex/ensemble/_forest.py
icfaust: good catch. will do

onedal/ensemble/forest.py
```diff
@@ -186,259 +128,102 @@ def _get_onedal_params(self, data):
...
         params = self._get_onedal_params(X_table)
         if hparams is not None and not hparams.is_default:
-            result = self.infer(params, hparams.backend, model, X_table)
+            result = self.infer(params, hparams.backend, self._onedal_model, X_table)
         else:
-            result = self.infer(params, model, X_table)
+            result = self.infer(params, self._onedal_model, X_table)
 
-        y = from_table(result.responses, like=X)
+        y = from_table(result.responses, like=X)[:, 0]
```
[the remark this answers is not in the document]
icfaust: They will need to read the documentation of `from_table` which makes the 2d assumption explicit.

onedal/ensemble/forest.py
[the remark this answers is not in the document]
icfaust: We would need to yield the namespace, while this may be standardized for array API, yielding the namespace is not trivial and is avoided in the onedal module objects (done with sklearn's get_namespace which builds on array_api_compat which shoe-horns in some data frameworks into being quasi-array API compliant, see pytorch).  If you want I can move this logic to the sklearnex estimator where we can there use `clip`.

sklearnex/ensemble/_forest.py
icfaust: Changed.

sklearnex/ensemble/_forest.py
icfaust: That's right, it converts a seed number into a numpy random state, which we (rather than using the random_state object) convert back to an integer for use in oneDAL. https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html So in the end its just a random number spitting out a different random number for use in oneDAL

sklearnex/ensemble/_forest.py
icfaust: included for use in `ForestRegressor`'s version.

sklearnex/ensemble/_forest.py
icfaust: Moved it into the patching conditions.

sklearnex/ensemble/_forest.py
icfaust: Don't understand. Sparse checks occur in `_onedal_fit_ready` on `X` and `y`.

sklearnex/ensemble/_forest.py
icfaust: Will do it in a coordinated fashion by someone before 2026.0 (dependent on LogReg, kMeans, Neighbors algos).

icfaust:
> @icfaust Since the equivalent classes in sklearn do not have array API support: how is this meant to work for attributes? I see that the arrays in the `Tree` object class for example are always numpy regardless of what was passed as inputs, while other internal attributes appear to have the array API class that was used as input, with data being on the corresponding device.
> 
> I guess this might be intended, but if that is the case, please add this kind of behavior to the documentation.

The `Tree` object is part of the `estimators_` attribute which is defined as a list of `DecisionTree` estimators which do not support array API. The ensemble algorithms are special in the way that they include other estimator objects in their estimator object. It would seem unnecessarily burdensome to have to recursively define all objects as being array API support even if they themselves cannot be. Ideally we try to best represent the datatypes described in the documentation of the estimator to follow array API best practices (https://scikit-learn.org/stable/modules/array_api.html#input-and-output-array-type-handling), which I believe this is properly doing.

onedal/ensemble/forest.py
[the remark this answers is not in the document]
david-cortes-intel: Guess it doesn't have much of a performance penalty, maybe could be left for later.

sklearnex/ensemble/_forest.py
david-cortes-intel: In that case, shouldn't this check also be moved there, since all of the fit variants support this argument?

david-cortes-intel:
> > @icfaust Since the equivalent classes in sklearn do not have array API support: how is this meant to work for attributes? I see that the arrays in the `Tree` object class for example are always numpy regardless of what was passed as inputs, while other internal attributes appear to have the array API class that was used as input, with data being on the corresponding device.
> > I guess this might be intended, but if that is the case, please add this kind of behavior to the documentation.
> 
> The `Tree` object is part of the `estimators_` attribute which is defined as a list of `DecisionTree` estimators which do not support array API. The ensemble algorithms are special in the way that they include other estimator objects in their estimator object. It would seem unnecessarily burdensome to have to recursively define all objects as being array API support even if they themselves cannot be. Ideally we try to best represent the datatypes described in the documentation of the estimator to follow array API best practices (https://scikit-learn.org/stable/modules/array_api.html#input-and-output-array-type-handling), which I believe this is properly doing.

I do not see anything in that documentation page that would imply that classed subobjects within array-api-compatible estimators can be non-array-api-compatible. Please add a note about it to the documentation to make it explicit.

sklearnex/ensemble/_forest.py
icfaust: Difference between CPU and GPU support (see `onedal_gpu_supported`).

sklearnex/ensemble/_forest.py
icfaust: Done! Thanks for the recommendation.

sklearnex/ensemble/_forest.py
icfaust: Changed to reuse some logic from `sklearnex.utils.validation` to minimize maintenance. Thanks for the suggestion.

onedal/ensemble/forest.py
icfaust: Done.

icfaust:
> The CI error for sklearn==1.0.2 on `ExtraTreesClassifier`:
> 
> ```
> >               raise AssertionError(err_msg)
> E               AssertionError: Estimator doesn't check for NaN and inf in fit.
> ```
> 
> Looks like something here is not working correctly.

Thanks to you and @ethanglaser for the headsup. Because I have removed duplicated finite checks, I forgot to trigger the finite check in `_onedal_fit` to be based on the sklearn version.

sklearnex/ensemble/_forest.py
david-cortes-intel: I cannot find any hint that the GPU version would support sparse weights. I see it just declares the weights as a dense array:
https://github.com/uxlfoundation/oneDAL/blob/b21771586e1d5fd0877ab128e41685f4c7b6f320/cpp/oneapi/dal/algo/decision_forest/backend/gpu/train_splitter_impl_dpc.cpp#L342

ethanglaser:
Can you rebase and rerun CI? Looks like the conformance issue is resolved, but the current treelite version is leading to yellow CI

sklearnex/ensemble/_forest.py
```diff
@@ -239,86 +432,83 @@ def _to_absolute_max_features(self, max_features, n_features):
                 "Invalid value for max_features. Allowed string "
                 f"values are {allowed_string_values}."
             )
+
         if isinstance(max_features, (numbers.Integral, np.integer)):
-            return max_features
+            return int(max_features)
```
ethanglaser, line 432: Why is this necessary if its already determined to be an int?

sklearnex/ensemble/_forest.py
icfaust: The GPU version doesn't support weights at all.

EDIT: @Alexandr-Solovev is weight support fully tested and available on oneDAL GPU RF now? This support criteria was from the original implementation I did 2 years ago, so maybe things have changed.  If we do have weight support, this logic can be refactored, but any test case failures from enablement I will then add to the GPU deselections.

sklearnex/ensemble/_forest.py
icfaust: The input into the oneDAL estimator and pybind11 should be a python int type, as numbers.Integral and np.integer are either more general or other object types.

sklearnex/ensemble/_forest.py
```diff
@@ -684,185 +846,34 @@ def score(self, X, y, sample_weight=None):
...
-        try:
-            return xp.take(
-                xp.asarray(self.classes_, device=res.sycl_queue),
-                xp.astype(xp.reshape(res, (-1,)), xp.int64),
-            )
-        except AttributeError:
-            return np.take(self.classes_, res.ravel().astype(np.int64, casting="unsafe"))
+
+        if is_array_api_compliant:
+            return xp.take(self.classes_, xp.astype(xp.reshape(res, (-1,)), xp.int64))
```
ethanglaser:
```suggestion
            return xp.take(xp.asarray(self.classes_, device=res.sycl_queue), xp.astype(xp.reshape(res, (-1,)), xp.int64))
```

icfaust:
> > > @icfaust Since the equivalent classes in sklearn do not have array API support: how is this meant to work for attributes? I see that the arrays in the `Tree` object class for example are always numpy regardless of what was passed as inputs, while other internal attributes appear to have the array API class that was used as input, with data being on the corresponding device.
> > > I guess this might be intended, but if that is the case, please add this kind of behavior to the documentation.
> > 
> > 
> > The `Tree` object is part of the `estimators_` attribute which is defined as a list of `DecisionTree` estimators which do not support array API. The ensemble algorithms are special in the way that they include other estimator objects in their estimator object. It would seem unnecessarily burdensome to have to recursively define all objects as being array API support even if they themselves cannot be. Ideally we try to best represent the datatypes described in the documentation of the estimator to follow array API best practices (https://scikit-learn.org/stable/modules/array_api.html#input-and-output-array-type-handling), which I believe this is properly doing.
> 
> I do not see anything in that documentation page that would imply that classed subobjects within array-api-compatible estimators can be non-array-api-compatible. Please add a note about it to the documentation to make it explicit.

I'll add that as a last step.

sklearnex/ensemble/_forest.py
Alexandr-Solovev: @icfaust Yes, oneDAL GPU RF has support for weights, and this functionality at least covered with bazel tests.
https://github.com/uxlfoundation/oneDAL/blob/main/cpp/oneapi/dal/algo/decision_forest/test/batch.cpp#L317
Inside we call https://github.com/uxlfoundation/oneDAL/blob/main/cpp/oneapi/dal/algo/decision_forest/test/fixture.hpp#L227

and provide pre-last column as weights https://github.com/uxlfoundation/oneDAL/blob/main/cpp/oneapi/dal/algo/decision_forest/test/fixture.hpp#L234
And train is called with weights https://github.com/uxlfoundation/oneDAL/blob/main/cpp/oneapi/dal/algo/decision_forest/test/fixture.hpp#L237

From the history of adding weights support:
CPU: https://github.com/uxlfoundation/oneDAL/commit/a760cb24c27c1664abba9363a8d79f0e8baf89ec#diff-a006a7279fd1dde5210dd5ea4fd501fa147e64c4835c544915dbe3c0958ff870
GPU: https://github.com/uxlfoundation/oneDAL/commit/14ad1ababe5601a62983c49b0ac2e046c990c445

david-cortes-intel: @Alexandr-Solovev Does it support **sparse** weights?

Alexandr-Solovev: No, it does not

sklearnex/ensemble/_forest.py
icfaust: Can you clarify why `asarray` is required here? Is this to conform to `use_raw_inputs`? If so then I need to add a note. I also had to adapt it, as `device=res.sycl_queue` is not array API compliant (needed to use `device=res.device`).

ethanglaser: Error I was getting in sklbench was that self.classes_ and res were not offloaded to the same sycl queue

icfaust: Gotcha! thanks for doing such indepth checking!

sklearnex/ensemble/_forest.py
ethanglaser: Previously this value was determined based on max_samples - is this still the case? I am having a hard time following the changes

ethanglaser: /if not I guess we don't need to document this anywhere since observations_per_tree_fraction isn't documented to begin this

sklearnex/ensemble/_forest.py
icfaust: @david-cortes-intel  here is something to note: https://github.com/scikit-learn/scikit-learn/blob/47621fec9c88727fcf06e1acfb9493480454edf9/sklearn/utils/validation.py#L2157
https://github.com/scikit-learn/scikit-learn/blob/47621fec9c88727fcf06e1acfb9493480454edf9/sklearn/utils/validation.py#L2154

Sparse support in sample_weights is not supported in sklearn, and is guarded against using `_check_sample_weight`. The old check was then not necessary and has been removed. As a consequence of this discussion, I have removed the None check in `onedal_gpu_supported` for `sample_weight`. I will rerun CI for this change.

onedal/ensemble/forest.py
```diff
@@ -549,25 +287,19 @@ def __init__(
...
-        ccp_alpha=0.0,
-        max_samples=None,
+        observations_per_tree_fraction=1.0,
         max_bins=256,
         min_bin_size=1,
-        infer_mode="class_responses",
+        infer_mode="class_responses",  # not used (see forest.cpp)
         splitter_mode="best",
-        voting_mode="weighted",
+        voting_mode="weighted",  # not used (see forest.cpp)
```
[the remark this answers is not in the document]
icfaust: That can be done as a follow-up, the comments are step in that direction.

sklearnex/ensemble/_forest.py
```diff
@@ -222,14 +410,14 @@ def _to_absolute_max_features(self, max_features, n_features):
                             FutureWarning,
                         )
                     return (
-                        max(1, int(np.sqrt(n_features)))
+                        max(1, int(math.sqrt(n_features)))
```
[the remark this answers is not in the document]
icfaust: The value must be an int, which is pretty obvious.

sklearnex/ensemble/_forest.py
```diff
@@ -58,38 +56,55 @@
...
+__check_kwargs = {
+    "dtype": None,
+    "ensure_2d": False,
+    "ensure_min_samples": 0,
+    "ensure_min_features": 0,
+    "accept_sparse": True,
+    _finite_keyword: False,
+}
+
+_check_array = partial(check_array, **__check_kwargs)
```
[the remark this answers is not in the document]
icfaust: Followed suggestion, change can be a follow-up.

onedal/ensemble/forest.py
```diff
@@ -186,259 +128,102 @@ def _get_onedal_params(self, data):
...
 
-        model = self._onedal_model
         if hparams is not None and not hparams.is_default:
-            result = self.infer(params, hparams.backend, model, X)
+            result = self.infer(params, hparams.backend, self._onedal_model, X)
         else:
-            result = self.infer(params, model, X)
+            result = self.infer(params, self._onedal_model, X)
 
         # TODO: fix probabilities out of [0, 1] interval on oneDAL side
```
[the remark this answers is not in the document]
icfaust: Already done. Hopefully someone addresses this soon.

onedal/ensemble/tests/test_random_forest.py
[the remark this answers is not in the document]
icfaust: Its definitely a magic number, but derived from mechanations. The reason we don't refactor the value is because of legacy testing support.

icfaust:
latest CI run: http://intel-ci.intel.com/f0d3b0ab-f859-f1ef-91c9-a4bf010d0e2d

deselected_tests.yaml
```diff
@@ -525,12 +525,8 @@ gpu:
   - tests/test_common.py::test_estimators[ExtraTreesClassifier()-check_class_weight_classifiers]
   - tests/test_common.py::test_estimators[ExtraTreesRegressor()-check_sample_weights_invariance(kind=zeros)]
   - tests/test_common.py::test_estimators[RandomForestRegressor()-check_regressor_data_not_an_array]
-
-  # GPU implementation of Extra Trees doesn't support sample_weights
-  # comparisons to GPU with sample weights will use different algorithms
-  - tests/test_common.py::test_estimators[ExtraTreesClassifier()-check_sample_weights_invariance(kind=ones)]
   - tests/test_common.py::test_estimators[ExtraTreesClassifier()-check_sample_weights_invariance(kind=zeros)]
-  - tests/test_common.py::test_estimators[ExtraTreesRegressor()-check_sample_weights_invariance(kind=ones)]
+  - ensemble/tests/test_forest.py::test_min_weight_fraction_leaf
```
david-cortes-intel, line 529: CC @Alexandr-Solovev - [this test](https://github.com/scikit-learn/scikit-learn/blob/47621fec9c88727fcf06e1acfb9493480454edf9/sklearn/ensemble/tests/test_forest.py#L1048) in particular is very straighforward and not expected to fail, yet it does here.

PR #2205, opened 2024-12-03
DOC: Clear some warnings for daal4py doc builds

david-cortes-intel:
## Description

This PR resolves many warnings that pop out when building the `daal4py` docs and which prevent the CI jobs from building those docs correctly.

Note that it doesn't solve all of the warnings - after this PR, one would still see many of these cases:
```
scikit-learn-intelex/doc/daal4py/docstring of daal4py._daal4py.classifier_prediction_result:1: WARNING: duplicate object description of daal4py.classifier_prediction_result, other instance in algorithms, use :noindex: for one of them
```

(I'm not sure what to do about those)

---

**Testing**

**Performance**

Not applicable.


david-cortes-intel:
This worked for a local build, but it seems it was not enough for the CI job named "Docs":
https://dev.azure.com/daal/daal4py/_build/results?buildId=40827&view=logs&j=7e620c85-24a8-5ffa-8b1f-642bc9b1fc36&t=dadee17e-671e-5fd8-4599-c783ec4d74d0

It still fails to import `daal4py` during the doc build step. Wonder if the issue might be that it is not building it.

doc/daal4py/conf.py
```diff
@@ -29,9 +29,10 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath("../.."))
```
icfaust, line 35: Since I don't know this side, but the PR looks simple enough, small explanation here as a comment and I'll approve (given it makes sense).

david-cortes-intel: The docs are generated from the python docstrings through `autodoc`. That means it needs to be able to import the package in order to build the docs.

This line adds the folder where `daal4py` would be to the paths where python searches, assuming you built the extension beforehand. Looks like it's not working in the CI job "docs" though, but the same logic is used in sklearnex, and works if you try `make html` locally:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/95bd1eabcca2e4d0565423be1dba3b58deaf8901/doc/sources/conf.py#L35

icfaust (approved):

PR #2207, opened 2024-12-03
[Enhancement] enable array API support in ```EmpiricialCovariance``` and ```IncrementalEmpiricalCovariance```

icfaust:
## Description

Enables array API zero copy dispatching for ```EmpiricalCovariance``` and ```IncrementalEmpiricalCovariance```, this required the following changes:

 - Added array API-enabled  ```log_likelihood``` and ```pinvh``` functions (as they are unavailable in sklearn) to ```sklearnex.utils._array_api```
 - fix ```sklearnex.preview.covariance.EmpiricalCovariance.mahalanobis.__doc__``` which was a bug
 - Implement array API enabled ```mahalanobis```, ```score```, and ```error_norm``` methods. Likely to change due to the nature of how we support dpnp and dpctl.
 - Added array API enabling from #2566
 - Added ```check_is_fitted``` (which is even missing from sklearn)
 - Added ```get_namespace``` and swapped namespace support away from numpy
 - Create a ```get_precision``` function to use internal ```pinvh```.  This is important for array API conformance, where attributes will no longer only be numpy arrays
 - moved sklearn conformance of return values to sklearnex estimators instead of onedal
 - general fixes for array_api_strict inputs (ellipsis use)
 - Move ```validate_params``` before fit's dispatch (will be set as a design rule in a follow-up PR
 - Deal with an issue with ```pairwise_distances``` kwargs and support_input_format, which do not interact well. A follow up development ticket for fixing this issue will be made
 - set proper array API conformance for return types of oneDAL tables using ```return_type_constructor``` in IncrementalEmpiricalCovariance

---

**Testing**

**Performance**


icfaust:
This PR is now dependent on the developments in #2096

sklearnex/covariance/incremental_covariance.py
```diff
@@ -156,13 +154,14 @@ def _onedal_finalize_fit(self):
         assert hasattr(self, "_onedal_estimator")
         self._onedal_estimator.finalize_fit()
         self._need_to_finalize = False
+        lp, _ = get_namespace(self._onedal_estimator.location_)
```
david-cortes-intel: How about naming it `xp_location` or similar?

sklearnex/covariance/incremental_covariance.py
```diff
@@ -351,40 +310,112 @@ def _onedal_fit(self, X, queue=None):
...
 
-    # expose sklearnex pairwise_distances if mahalanobis distance eventually supported
-    def mahalanobis(self, X):
-        check_feature_names(self, X, reset=False)
+    @wrap_output_data
+    def score(self, X_test, y=None):
 
-        xp, _ = get_namespace(X)
+        check_is_fitted(self)
+        xp, _ = get_namespace(self.covariance_)
```
david-cortes-intel: Shouldn't this consider both the `X` and the `covariance_` arrays when getting the namespace? E.g. like SciPy does:
https://docs.scipy.org/doc/scipy/dev/api-dev/array_api.html#adding-array-api-standard-support-to-a-scipy-function

sklearnex/covariance/incremental_covariance.py
```diff
@@ -351,40 +310,112 @@ def _onedal_fit(self, X, queue=None):
...
+        location = self.location_
         precision = self.get_precision()
-        # compute mahalanobis distances
-        # pairwise_distances will check n_features (via n_feature matching with
-        # self.location_) , and will check for finiteness via check array
-        # check_feature_names will match _validate_data functionally
-        location = self.location_[np.newaxis, :]
+
+        est = clone(self)
+        est.set_params(**{"assume_centered": True})
```
david-cortes-intel: Why not pass this directly as argument?

sklearnex/covariance/incremental_covariance.py
```diff
@@ -351,40 +310,112 @@ def _onedal_fit(self, X, queue=None):
...
+            # must be done this way is it does not inherit from sklearn
+            return _sklearn_EmpiricalCovariance.error_norm(
+                self, c_cov, norm=norm, scaling=scaling, squared=squared
+            )
+
+        # compute the error
+        error = c_cov - self.covariance_
+        # compute the error norm
+        if norm == "frobenius":
+            squared_norm = xp.sum(error**2)
```
david-cortes-intel: Wouldn't it be better to flatten `error` and calculate the dot product with itself?

sklearnex/utils/_array_api.py
```diff
@@ -151,3 +154,67 @@ class Ridge():
...
+    atol = 0.0 if atol is None else atol
+    rtol = max(a.shape) * xp.finfo(u.dtype).eps if (rtol is None) else rtol
+
+    if (atol < 0.0) or (rtol < 0.0):
+        raise ValueError("atol and rtol values must be positive.")
+
+    val = atol + maxS * rtol
+    above_cutoff = xp.nonzero(abs(s) > val)[0]
+
+    psigma_diag = 1.0 / xp.take(s, above_cutoff)
```
david-cortes-intel, line 189: These always come in sorted order. Could take them by index position instead of boolean mask.

sklearnex/utils/_array_api.py
```diff
@@ -151,3 +154,67 @@ class Ridge():
...
+def log_likelihood(emp_cov, precision):
+    # this is to compensate for a lack of array API support in sklearn
+    # even though it exists for ``fast_logdet``
+    xp, _ = get_namespace(emp_cov, precision)
+    p = precision.shape[0]
+    # extract sklearn.utils.extmath.fast_logdet for dpnp/dpctl support
+    sign, ld = xp.linalg.slogdet(precision)
+    if not sign > 0:
+        ld = -xp.inf
+    log_likelihood_ = -xp.sum(emp_cov * precision) + ld
```
david-cortes-intel, line 211: Shouldn't this be equal to the negative of Mahalanobis distance plus 'ld'?

sklearnex/utils/_array_api.py
```diff
@@ -151,3 +154,67 @@ class Ridge():
...
+
+    B = (u * psigma_diag) @ uconj.T
+
+    if return_rank:
+        return B, len(psigma_diag)
+    else:
+        return B
+
+
+pinvh.__doc__ = linalg.pinvh.__doc__
```
david-cortes-intel: Is this an exported function? Otherwise, why add the docstring from scipy? I guess it will become outdated later on.

onedal/covariance/incremental_covariance.py
```diff
@@ -143,13 +137,14 @@ def finalize_fit(self):
...
 
-            if daal_check_version((2024, "P", 1)) or (not self.bias):
-                self.covariance_ = from_table(result.cov_matrix)
-            else:
+            self.covariance_ = from_table(result.cov_matrix, like=self._outtype)
+
+            if self.bias and not daal_check_version((2024, "P", 1)):
                 n_rows = self._partial_result.partial_n_rows
-                self.covariance_ = from_table(result.cov_matrix) * (n_rows - 1) / n_rows
+                self.covariance_ = self.covariance_ * (n_rows - 1) / n_rows
```
david-cortes-intel:
```suggestion
                self.covariance_ = self.covariance_ * ((n_rows - 1) / n_rows)
```

sklearnex/utils/_array_api.py
david-cortes-intel: On second thought: the calculation of covariance from oneDAL side currently is imprecise enough that it might produce negative eigenvalues that are far off from zero, for example if it ends up having negative entries in the diagonal (which a covariance cannot have in theory). In such case, the result won't end up making sense, regardless of whether it discards small singular values by absolute value or by real value.

I think the better option here would be to error out if it finds that there are large negative eigenvalues, or to fall back to scikit-learn for the full covariance calculation (since theirs is more precise and won't end up with negative eigenvalues beyond small roundoff error). Thoughts @Alexsandruss @icfaust ?

icfaust: I would refer to this first https://github.com/scipy/scipy/blob/fa4f00d896306fc5dd60ea55a857b16712652c84/scipy/linalg/_basic.py#L1950

sklearnex/utils/_array_api.py
icfaust: https://github.com/scikit-learn/scikit-learn/blob/adb1ae76d798e822984d4e312c90c6ca8f01a3df/sklearn/covariance/_empirical_covariance.py#L52

icfaust: https://github.com/scikit-learn/scikit-learn/blob/adb1ae76d798e822984d4e312c90c6ca8f01a3df/sklearn/covariance/_empirical_covariance.py#L52

sklearnex/utils/_array_api.py
david-cortes-intel: @icfaust If oneDAL computes a very imprecise covariance that ends up with negative eigenvalues, the result will still be finite. Also, the function from scipy is meant for general symmetric matrices which is why they do things that way, but the input here is a covariance matrix, which by definition cannot have negative eigenvalues if computed to infinite precision.

sklearnex/utils/_array_api.py
david-cortes-intel: Did some quick testing and looks like it matches with SciPy's results.

sklearnex/utils/_array_api.py
icfaust: This implementation was intentional. Purely replicating the scipy functionality in the array API standard was done minimize the development cycle/ reviewing burden while guaranteeing no regressions/ performance degradations from sklearn.  Optimizations were specifically avoided as I had taken in to account possible future development that you suggested: https://uxlfoundation.github.io/oneDAL/contribution/ideas.html#cholesky-based-precision-calculation-medium The code comments in this section of the code are to that effect.

I think your concern is valid.  If there is likely a mathematical error associated with the eigenvalues associated with our covariance implementation, then this is a bug that should be seen in the current code. In that case it requires an example which then can be put into a ticket/issue and a separate investigation where we can also address the implications of oneDAL support or scikit-learn fallback w.r.t this algorithm as you suggest via the link or comment above respectively.

With multiple possible directions for future development and the context of the PR, I kindly ask that we consider this as a follow-up. @david-cortes-intel

onedal/covariance/incremental_covariance.py
icfaust: Changed it to a ```*=```

sklearnex/utils/_array_api.py
icfaust: I could copy out the docs if you'd like, just thought it would highlight the similarity.

sklearnex/covariance/incremental_covariance.py
icfaust: Depends on memory, but I see what you are getting at. If you'd like I can swap it, but I would like to leave a note about its variation from the sklearn version.

sklearnex/covariance/incremental_covariance.py
icfaust: Changed.

EDIT: I added a note since I had forgotten why I had done it this way in the first place (as it had been a while).  Because we support dpnp/dpctl out of the box (e.g. with array_api_dispatch=False) it only checks covariance

sklearnex/utils/_array_api.py
david-cortes-intel: I think it'd be better to leave out the docs, and maybe add a comment that it is not an exported function, or name it differently, like with a leading underscore.

sklearnex/utils/_array_api.py
david-cortes-intel: Tickets have already been created. Sharing an earlier example:
image

sklearnex/covariance/incremental_covariance.py
david-cortes-intel: It'd be better to swap it as that way it'll end up calling BLAS.

sklearnex/covariance/incremental_covariance.py
david-cortes-intel: What would happen if the covariance is on GPU but 'X' is on CPU and it's not a numpy array?

icfaust: @david-cortes-intel please provide concrete changes you would like.

david-cortes-intel: @icfaust A test checking that things work when the covariance is on GPU and 'X' is on CPU and it's array-api-compliant but not NumPy (e.g. a DPNP array on CPU, or torch tensor if it has the necessary properties) would be good to have here.

sklearnex/covariance/incremental_covariance.py
icfaust: Changed logic to use the standard ```xp``` and put it in the proper location, as it was necessary for supporting an older oneDAL version.

david-cortes-intel: Did you perhaps forget to push the commit?

sklearnex/covariance/incremental_covariance.py
```diff
@@ -353,40 +310,123 @@ def _onedal_fit(self, X, queue=None):
...
+                self, c_cov, norm=norm, scaling=scaling, squared=squared
+            )
+
+        # compute the error
+        error = c_cov - self.covariance_
+        # compute the error norm
+        if norm == "frobenius":
+            # variance from sklearn version to leverage BLAS GEMM
+            # squared_norm = xp.sum(error**2)
+            squared_norm = xp.matmul(
```
david-cortes-intel: Wouldn't this produce a 1x1 matrix that would then need to be subsetted? Also, what was wrong with 'dot'?

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -285,3 +285,25 @@ def test_IncrementalEmpiricalCovariance_against_sklearn(monkeypatch, sklearn_tes
...
+
+    rng = np.random.seed(42)
+    X = np.random.rand(5, 10)
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+
+    # calculate and store data on SYCL device (CPU or GPU)
+    with config_context(array_api_dispatch=True):
+        est.fit(X_df)
+
+    with pytest.raises(TypeError, match="Multiple namespaces for array inputs: .*"):
```
david-cortes-intel: Would this work correctly if put under a config context with `array_api_dispatch=True`?

sklearnex/covariance/incremental_covariance.py
icfaust: Unfortunately ```dot``` isn't defined in the array API standard, and ```vecdot``` isn't available in numpy <2.0.  I have swapped it to not defining 2d arrays, but make them 1d arrays where a 0d return is expected.

sklearnex/covariance/tests/test_incremental_covariance.py
icfaust: I used sklearn's Ridge + numpy and torch as an example of what to expect: (https://scikit-learn.org/stable/modules/array_api.html#input-and-output-array-type-handling)
image
When attempting to use any non-numpy input after fitting with ```array_api_dispatch=True``` will lead to some sort of error associated with the fitted framework, as ```get_namespace``` and ```validate_data``` will default force data to numpy (https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_array_api.py#L394) and then comparing numpy to the array api framework will fail.

If we were to use ```array_api_dispatch=True``` throughout, it will error at this point in an external package if get_namespace is used: https://github.com/data-apis/array-api-compat/blob/main/array_api_compat/common/_helpers.py#L665

sklearnex/covariance/incremental_covariance.py
icfaust: Test was added in last commits. Please review.

sklearnex/covariance/tests/test_incremental_covariance.py
david-cortes-intel: Isn't the idea with array API support in stock sklearn to make it work under such types of situations? (e.g. fitting on a torch array, then predicting on a different kind of array).

david-cortes-intel: Actually no, I see that sklearn throws the same error.

icfaust:
Private CI failure comes from an infrastructure timeout.

PR #2209, opened 2024-12-04
[enhancement] WIP new finite checking in SVM algorithms

icfaust:
## Description

---

**Testing**

**Performance**

### DIFF
--- a/onedal/svm/svm.py
+++ b/onedal/svm/svm.py
@@ -22,6 +22,7 @@ from scipy import sparse as sp
 
 from onedal import _backend
 
+from ..common._base import BaseEstimator
 from ..common._estimator_checks import _check_is_fitted
 from ..common._mixin import ClassifierMixin, RegressorMixin
 from ..common._policy import _get_policy
@@ -35,14 +36,7 @@ from ..utils import (
 )
 
 
-class SVMtype(Enum):
-    c_svc = 0
-    epsilon_svr = 1
-    nu_svc = 2
-    nu_svr = 3
-
-
-class BaseSVM(metaclass=ABCMeta):
+class BaseSVM(BaseEstimator, metaclass=ABCMeta):
     @abstractmethod
     def __init__(
         self,
@@ -63,8 +57,6 @@ class BaseSVM(metaclass=ABCMeta):
         decision_function_shape,
         break_ties,
         algorithm,
-        svm_type=None,
-        **kwargs,
     ):
         self.C = C
         self.nu = nu
@@ -82,21 +74,20 @@ class BaseSVM(metaclass=ABCMeta):
         self.decision_function_shape = decision_function_shape
         self.break_ties = break_ties
         self.algorithm = algorithm
-        self.svm_type = svm_type
 
     def _validate_targets(self, y, dtype):
         self.class_weight_ = None
         self.classes_ = None
         return _column_or_1d(y, warn=True).astype(dtype, copy=False)
 
-    def _get_onedal_params(self, data):
+    def _get_onedal_params(self, dtype):
         max_iter = 10000 if self.max_iter == -1 else self.max_iter
         # TODO: remove this workaround
         # when oneDAL SVM starts support of 'n_iterations' result
         self.n_iter_ = 1 if max_iter < 1 else max_iter
         class_count = 0 if self.classes_ is None else len(self.classes_)
         return {
-            "fptype": data.dtype,
+            "fptype": dtype,
             "method": self.algorithm,
             "kernel": self.kernel,
             "c": self.C,
@@ -174,9 +165,9 @@ class BaseSVM(metaclass=ABCMeta):
             self._scale_, self._sigma_ = _gamma, np.sqrt(0.5 / _gamma)
 
         policy = _get_policy(queue, *data)
-        X = _convert_to_supported(policy, X)
-        params = self._get_onedal_params(X)
-        result = module.train(policy, params, *to_table(*data))
+        data_t = to_table(*_convert_to_supported(policy, *data))
+        params = self._get_onedal_params(data_t[0].dtype)
+        result = module.train(policy, params, *data_t)
 
         if self._sparse:
             self.dual_coef_ = sp.csr_matrix(from_table(result.coeffs).T)
@@ -206,9 +197,6 @@ class BaseSVM(metaclass=ABCMeta):
         m.support_vectors = to_table(self.support_vectors_)
         m.coeffs = to_table(self.dual_coef_.T)
         m.biases = to_table(self.intercept_)
-
-        if self.svm_type is SVMtype.c_svc or self.svm_type is SVMtype.nu_svc:
-            m.first_class_response, m.second_class_response = 0, 1
         return m
 
     def _predict(self, X, module, queue):
@@ -370,7 +358,6 @@ class SVR(RegressorMixin, BaseSVM):
             break_ties=False,
             algorithm=algorithm,
         )
-        self.svm_type = SVMtype.epsilon_svr
 
     def fit(self, X, y, sample_weight=None, queue=None):
         return super()._fit(X, y, sample_weight, _backend.svm.regression, queue)
@@ -422,7 +409,11 @@ class SVC(ClassifierMixin, BaseSVM):
             break_ties=break_ties,
             algorithm=algorithm,
         )
-        self.svm_type = SVMtype.c_svc
+
+    def _create_model(self, module):
+        m = super()._create_model(module)
+        m.first_class_response, m.second_class_response = 0, 1
+        return m
 
     def _validate_targets(self, y, dtype):
         y, self.class_weight_, self.classes_ = _validate_targets(
@@ -483,7 +474,6 @@ class NuSVR(RegressorMixin, BaseSVM):
             break_ties=False,
             algorithm=algorithm,
         )
-        self.svm_type = SVMtype.nu_svr
 
     def fit(self, X, y, sample_weight=None, queue=None):
         return super()._fit(X, y, sample_weight, _backend.svm.nu_regression, queue)
@@ -535,7 +525,11 @@ class NuSVC(ClassifierMixin, BaseSVM):
             break_ties=break_ties,
             algorithm=algorithm,
         )
-        self.svm_type = SVMtype.nu_svc
+
+    def _create_model(self, module):
+        m = super()._create_model(module)
+        m.first_class_response, m.second_class_response = 0, 1
+        return m
 
     def _validate_targets(self, y, dtype):
         y, self.class_weight_, self.classes_ = _validate_targets(


onedal/svm/svm.py
```diff
@@ -79,128 +58,65 @@ def __init__(
...
+            self.support_vectors_ = from_table(result.support_vectors, like=X)
 
-        if getattr(self, "classes_", None) is not None:
-            indices = y.take(self.support_, axis=0)
-            self._n_support = np.array(
-                [np.sum(indices == i) for i, _ in enumerate(self.classes_)]
-            )
-        self._gamma = self._scale_
+        self.intercept_ = from_table(result.biases, like=X)[0, ...]
+        self.support_ = from_table(result.support_indices, like=X)[:, 0]
```
david-cortes-intel: Shouldn't this be casted to an integer type like it was in the previous code?

onedal/svm/svm.py
```diff
@@ -211,150 +127,56 @@ def _create_model(self):
...
-            )
-
-        if isinstance(self, ClassifierMixin):
-            sv = self.support_vectors_
-            if not self._sparse and sv.size > 0 and self._n_support.sum() != sv.shape[0]:
-                raise ValueError(
-                    "The internal representation "
-                    f"of {self.__class__.__name__} was altered"
-                )
+            if not sp.isspmatrix(X):
```
david-cortes-intel: The sparse matrix classes from SciPy are deprecated in favor of sparse arrays. This would return `False` for the `csr_array` class, and would cast the input to `csr_matrix`. Would there be any problems in supporting CSR arrays here?

onedal/svm/svm.py
```diff
@@ -381,41 +199,42 @@ def infer(self, *args, **kwargs): ...
...
         cache_size=200.0,
         max_iter=-1,
         tau=1e-12,
-        class_weight=None,
-        decision_function_shape="ovr",
-        break_ties=False,
         algorithm="thunder",
-        **kwargs,
     ):
+        # This hard-codes nu=.5, which may be bad. Needs to be investigated
```
david-cortes-intel: Would this 'nu' parameter be applicable to `NuSVC` instead of `SVC`? Shouldn't SVC simply ignore it?

onedal/svm/tests/test_nusvc.py
```diff
@@ -178,12 +178,14 @@ def test_cancer_linear_compare_with_sklearn(queue, nu):
...
-
-    clf = SklearnNuSVC(kernel="poly", **params)
+    class_count = len(np.unique(cancer.target))
+    # gamma="scale"
+    _gamma = 1.0 / (cancer.data.shape[1] * cancer.data.var())
+    clf = NuSVC(kernel="poly", gamma=_gamma, **params)
+    clf.fit(cancer.data, cancer.target, class_count=class_count, queue=queue)
+    result = accuracy_score(cancer.target, clf.predict(cancer.data, queue=queue))
+
+    clf = SklearnNuSVC(kernel="poly", gamma=_gamma, **params)
```
david-cortes-intel, line 188: If the idea here is to verify that `gamma` is processed correctly, shouldn't it instead pass the string `"scale"` to the sklearnex class?

sklearnex/svm/__init__.py
```diff
@@ -17,10 +17,7 @@
 from .._utils import get_sklearnex_version
 
 if get_sklearnex_version((2021, "P", 300)):
```
david-cortes-intel: Looks like this is the only place in the codebase where this function is used. Could it use the one that checks for daal version and do away with this function instead?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,646 @@
+# ==============================================================================
+# Copyright 2021 Intel Corporation
```
david-cortes-intel: We are using a different copyright header now:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#copyright-guidelines-for-contributions

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,646 @@
...
+        self.classes_ = self.estimator.classes_
+
+        predictions, _ = _get_response_values(
+            estimator,
+            X,
+            response_method=["decision_function", "predict_proba"],
+        )
+        if predictions.ndim == 1:
+            # Reshape binary output from `(n_samples,)` to `(n_samples, 1)`
+            predictions = predictions.reshape(-1, 1)
```
david-cortes-intel: Should't this sort of thing also be using array API? E.g. `xp.reshape(...)`.

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,646 @@
...
+        return self._dualcoef_
+
+    @_dual_coef_.setter
+    def _dual_coef_(self, value):
+        self._dualcoef_ = value
+        if hasattr(self, "_onedal_estimator"):
+            self._onedal_estimator.dual_coef_ = value
+            self._onedal_estimator._onedal_model = None
+
+    @_dual_coef_.deleter
```
david-cortes-intel, line 119: Is there some use-case where it'd make sense to delete this without causing crashes further down the line?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,646 @@
...
+        # the onedal estimator as a float type. This replicates functionality
+        # directly out of scikit-learn to enable various variable gamma values.
+        if self.kernel == "linear":
+            return 1.0
+
+        if isinstance(self.gamma, str):
+            if self.gamma == "scale":
+                if sp.issparse(X):
+                    # var = E[X^2] - E[X]^2
+                    X_sc = (X.multiply(X)).mean() - (X.mean()) ** 2
```
david-cortes-intel, line 197: Maybe this could call `BasicStatistics` instead.

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,646 @@
...
+                "onedal": self.__class__._onedal_predict_proba,
+                "sklearn": sklearn_pred_proba,
+            },
+            X,
+        )
+
+    def _fit_proba(self, X, y, sample_weight=None, queue=None):
+        # TODO: rewrite this method when probabilities output is implemented in oneDAL
+
+        # LibSVM uses the random seed to control cross-validation for probability generation
```
david-cortes-intel, line 358: Wouldn't oneDAL get used in this case instead of libSVM from sklearn? And even if the latter, wouldn't it either way get passed the `random_state` as seed?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,646 @@
...
+        self._icept_ = self._onedal_estimator.intercept_
+        self._sparse = False
+        self.fit_status_ = 0
+        self.shape_fit_ = X.shape
+
+        self._gamma = self._onedal_estimator.gamma
+        length = (self.classes_.shape[0] ** 2 - self.classes_.shape[0]) // 2
+
+        if self.probability:
+            self._probA = np.zeros(length)
```
david-cortes-intel: Does this get used for anything? Could it be removed instead?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,646 @@
...
+
+        if sklearn_check_version("1.1"):
+            self.n_iter_ = xp.full((length,), self._onedal_estimator.n_iter_)
+
+    def _onedal_predict(self, X, queue=None):
+        sv = self.support_vectors_
+
+        xp, _ = get_namespace(X)
+
+        if not self._sparse and sv.size > 0 and xp.sum(self._n_support) != sv.shape[0]:
```
david-cortes-intel, line 443: Could it use `n_features_in_` instead?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,646 @@
...
+        patching_status.and_conditions([(False, "GPU offloading is not supported.")])
+        return patching_status
+
+    def _onedal_cpu_supported(self, method_name, *data):
+        class_name = self.__class__.__name__
+        patching_status = PatchingConditionsChain(
+            f"sklearn.svm.{class_name}.{method_name}"
+        )
+        if method_name == "fit":
+            patching_status.and_conditions(
```
david-cortes-intel, line 154: Please update the support docs too, since it says all parameters should be supported:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/754d32dde74fbfa3cd9a1af2e0612798a4d5d02c/doc/sources/algorithms.rst?plain=1#L43

sklearnex/svm/_classes.py
```diff
@@ -0,0 +1,359 @@
...
+            # else if added to correct issues with
+            # sklearn tests:
+            # svm/tests/test_sparse.py::test_error
+            # svm/tests/test_svm.py::test_bad_input
+            # for sklearn versions < 1.2 (i.e. without
+            # validate_params parameter checking)
+            # Without this, a segmentation fault with
+            # Windows fatal exception: access violation
+            # occurs
+            raise ValueError("nu <= 0 or nu > 1")
```
david-cortes-intel:
```suggestion
            raise ValueError("'nu' must be in the range (0, 1].")
```

sklearnex/svm/_classes.py
```diff
@@ -0,0 +1,359 @@
...
+            # else if added to correct issues with
+            # sklearn tests:
+            # svm/tests/test_sparse.py::test_error
+            # svm/tests/test_svm.py::test_bad_input
+            # for sklearn versions < 1.2 (i.e. without
+            # validate_params parameter checking)
+            # Without this, a segmentation fault with
+            # Windows fatal exception: access violation
+            # occurs
+            raise ValueError("C <= 0")
```
david-cortes-intel:
```suggestion
            raise ValueError("'C' must be strictly positive.")
```

sklearnex/svm/tests/test_svm.py
```diff
@@ -98,6 +106,80 @@ def test_sklearnex_import_nusvr(dataframe, queue):
...
+
+@pass_if_not_implemented_for_gpu(reason="csr svm is not implemented")
+@pytest.mark.parametrize("queue", get_queues())
+@pytest.mark.parametrize("kernel", ["linear", "rbf", "poly", "sigmoid"])
+def test_iris(queue, kernel):
+    from sklearnex import config_context
+    from sklearnex.svm import SVC
+
+    if kernel == "rbf":
+        pytest.skip("RBF CSR SVM test failing in 2025.0.")
```
david-cortes-intel: If this is the same issue as in this PR, maybe it could mention it as a comment:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/2123

sklearnex/svm/_base.py
icfaust: Still in draft on purpose.  Iterating now on the scikit-learn-intelex estimators now that I've finished adapting the onedal to a close-to-final form.

sklearnex/svm/_base.py
icfaust: Definitely. I hadn't looked deeply on this `_n_support` attribute, but I am definitely going to remove it.

onedal/svm/svm.py
icfaust: Thats sklearn conformance, taken care of in the sklearnex estimator. Technically oneDAL is capable of returning an int32 table, but it looks like it sets it as float data.

sklearnex/svm/_base.py
icfaust: The statement about LibSVM gives background on the existence of `random_seed` in the SVM estimator. We use a questionable method with `prefit` in `CalibratedClasisfierCV` which wouldn't use `random_seed` anyway.

david-cortes-intel:
@icfaust Please remember to add the classes that this PR covers to this list in the docs:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/91885a302e7516f762424f8af2cac27a50c25849/doc/sources/array_api.rst?plain=1#L84

onedal/svm/svm.py
icfaust: Added.

onedal/svm/svm.py
icfaust: Centralized and commented the onedal estimators to disambiguate the situation while maximizing code reuse.  This should help maintenance. I have verified this with @Vika-F by looking at the source code.

onedal/svm/tests/test_nusvc.py
icfaust: This is in `onedal` testing and using a `onedal` estimator this aspect must now be supplied in the test as it is scikit-learn conformance.  Testing is maintained in capability, but a refactor of the testing should be done in a follow-up.

sklearnex/svm/__init__.py
icfaust: Removed, good find.

sklearnex/svm/_base.py
icfaust: Changed copyright headers as necessary.

sklearnex/svm/_base.py
icfaust: modified as requested.

sklearnex/svm/_base.py
icfaust: I would say removal of the deleter can be discussed as follow up as we may need to add a rule for this. As of now it conforms to other array API standardized sklearnex estimators.

sklearnex/svm/_base.py
icfaust: Would like to add this as a follow up. Investigative work would be necessary to verify performance and in corner cases. This duplicates the old capability which means no performance loss.

sklearnex/svm/_base.py
icfaust: This is for scikit-learn conformance providing a plausible value which oneDAL does not calculate (as per the design rules).

sklearnex/svm/_base.py
icfaust: This duplicates the logic out of scikit-learn for ease of maintenance.

sklearnex/svm/_base.py
icfaust: Documentation has been updated.

sklearnex/svm/tests/test_svm.py
icfaust: Skip removed and passes tests. Thanks for the heads up.

ethanglaser:
Looks like great work here, and comprehensive description of changes! It is quite a large PR that might benefit from a dedicated review session?

Also looks like we are hitting a limit on docbuild job related to medium.com

doc/sources/array_api.rst
```diff
@@ -96,6 +96,10 @@ The following patched classes have support for array API inputs:
 - :obj:`sklearn.linear_model.Ridge`
 - :obj:`sklearnex.linear_model.IncrementalLinearRegression`
 - :obj:`sklearnex.linear_model.IncrementalRidge`
+- :obj:`sklearnex.svm.NuSVC`
+- :obj:`sklearnex.svm.NuSVR`
+- :obj:`sklearnex.svm.SVC`
+- :obj:`sklearnex.svm.SVR`
```
david-cortes-intel:
```suggestion
- :obj:`sklearn.svm.NuSVC`
- :obj:`sklearn.svm.NuSVR`
- :obj:`sklearn.svm.SVC`
- :obj:`sklearn.svm.SVR`
```

Otherwise they wouldn't link to anything, since sklearnex patched classes do not have separate docs.

sklearnex/svm/tests/test_svm.py
```diff
@@ -98,6 +107,70 @@ def test_sklearnex_import_nusvr(dataframe, queue):
...
 
 
+@pass_if_not_implemented_for_gpu(reason="csr svm is not implemented")
+@pytest.mark.parametrize(
+    "queue",
+    get_queues("cpu")
+    + [
+        pytest.param(
+            get_queues("gpu"),
+            marks=pytest.mark.xfail(
```
david-cortes-intel, line 117: Sounds like a skip would be more appropriate than an xfail.

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,678 @@
...
+            # support.
+            patching_status.and_conditions(
+                [(hasattr(self, "_onedal_estimator"), "oneDAL model was not trained.")]
+            )
+            return patching_status
+        raise RuntimeError(f"Unknown method {method_name} in {class_name}")
+
+    def _svm_sample_weight_check(self, sample_weight, y, xp):
+        # This is purely for sklearn conformance. SVM algos in SVM raise unique errors.
+        if sample_weight is not None and xp.all(sample_weight <= 0):
```
david-cortes-intel: @Vika-F @Alexandr-Solovev Do oneDAL's SVMs work correctly if some weight is negative?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,678 @@
...
+                raise ValueError(
+                    "When 'gamma' is a string, it should be either 'scale' or "
+                    "'auto'. Got '{}' instead.".format(self.gamma)
+                )
+        else:
+            if sklearn_check_version("1.1") and not sklearn_check_version("1.2"):
+                if isinstance(self.gamma, Real):
+                    if self.gamma <= 0:
+                        msg = (
+                            f"gamma value must be > 0; {self.gamma!r} is invalid. Use"
```
david-cortes-intel, line 214: Does adding `!r` change anything about the variable is printed?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,678 @@
...
+            # oneDAL only accepts sample_weights, apply class_weight directly
+
+        # due to the nature of how sklearn checks nu in NuSVC (by not checking
+        # class_weight-adjusted values, this check must be done here regardless)
+        # Logic can be found in libsvm/svm.cpp in scikit-learn.
+        self._svm_sample_weight_check(sample_weight, y, xp)
+
+        if self.class_weight is not None:
+            for i, v in enumerate(self.class_weight_):
+                sample_weight[y == i] *= v
```
david-cortes-intel, line 325: Shouldn't this be making a deep copy beforehand so as not to modify the input data?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,678 @@
...
+
+                self.clf_prob = CalibratedClassifierCV(
+                    clf_base,
+                    ensemble=False,
+                    cv="prefit",
+                    method="sigmoid",
+                ).fit(X, y, sample_weight=sample_weight)
+
+    def _save_attributes(self, X, y, xp=np):
+        # This function requires array API adaptation.
```
david-cortes-intel: What is it missing for array API? Would changes be needed on the BaseSVR class too?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,678 @@
...
+    def predict_proba(self, X):
+        check_is_fitted(self)
+        return self._predict_proba(X)
+
+    @available_if(_sklearn_BaseSVC._check_proba)
+    @wraps(_sklearn_BaseSVC.predict_log_proba, assigned=["__doc__"])
+    def predict_log_proba(self, X):
+        xp, _ = get_namespace(X)
+
+        return xp.log(self.predict_proba(X))
```
david-cortes-intel, line 600: This will produce infinites if some probability is exactly equal to zero. Better to clip it as done in logistic regression:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/b63146e398739f06071702abf69b5dfa3fad4eb2/onedal/linear_model/logistic_regression.py#L223

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,678 @@
...
+class BaseSVR(BaseSVM):
+
+    # overwrite _validate_targets for array API support
+    def _validate_targets(self, y):
+        xp, is_array_api_compliant = get_namespace(y)
+
+        if not is_array_api_compliant:
+            return super()._validate_targets(y)
+
+        return xp.astype(column_or_1d(y, warn=True), xp.float64, copy=False)
```
david-cortes-intel: Shouldn't 'y' need to end up with the same dtype as 'X'?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,678 @@
...
+        self._onedal_estimator = self._onedal_factory(**onedal_params)
+        self._onedal_estimator.fit(X, y, sample_weight, queue=queue)
+        self._save_attributes(X, xp=xp)
+
+    def _save_attributes(self, X, xp=None):
+        self.support_vectors_ = self._onedal_estimator.support_vectors_
+        self.fit_status_ = 0
+        self.dual_coef_ = self._onedal_estimator.dual_coef_
+        self.shape_fit_ = X.shape
+        self.support_ = xp.asarray(self._onedal_estimator.support_, dtype=xp.int32)
```
david-cortes-intel: Is the resulting number guaranteed to fit into int32?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,678 @@
...
+            {
+                "onedal": self.__class__._onedal_score,
+                "sklearn": RegressorMixin.score,
+            },
+            X,
+            y,
+            sample_weight=sample_weight,
+        )
+
+    predict.__doc__ = _sklearn_BaseLibSVM.predict.__doc__
```
david-cortes-intel, line 719: Shouldn't it inherit these from sklearn's SVR class?

david-cortes-intel:
@icfaust I'm not sure if this is specific to SVMs or whether this PR was meant to fix it, but `allow_fallback_to_host` still doesn't work with the dispatching logic changes in this PR:

```python
import os
os.environ["SKLEARNEX_VERBOSE"] = "INFO"
import numpy as np
from sklearnex import config_context
from sklearnex.svm import SVC
X = rng.standard_normal(size=(100,10)).astype(np.float32)
y = rng.integers(3, size=100).astype(np.float32)
model = SVC(random_state=123)
with config_context(target_offload="gpu", allow_fallback_to_host=True):
    model.fit(X, y) 
```
```
RuntimeError: SVM with multiclass support is not implemented for GPU
```

Logs show:
```
INFO:sklearnex: sklearn.svm.SVC.fit: running accelerated version on CPU
```

sklearnex/svm/tests/test_svm.py
icfaust: This is legacy testing, as stated in the description, a full refactoring of the testing would be good, but should be done at a later date. This is to guarantee no functional loss in testing capability.

sklearnex/svm/_base.py
icfaust: Under such logic, then all other classifiers with `predict_log_proba` should be modified to clip. (warnings for other implementations can be found in the pytest logs)  I recommend this as a separate theme for a separate PR.

david-cortes-intel: That is not actually the case, because some types of models can produce log-probabilities directly without first calculating probabilities, which is what this kind of method is meant to do.

icfaust: https://github.com/uxlfoundation/scikit-learn-intelex/blob/b63146e398739f06071702abf69b5dfa3fad4eb2/sklearnex/ensemble/_forest.py#L659

icfaust: https://github.com/scikit-learn/scikit-learn/blob/a3a3b5ad15c7566746288ed32d9875cd8542db95/sklearn/ensemble/_forest.py#L969

sklearnex/svm/_base.py
icfaust: SVR inherits from BaseLibSVM (there is no BaseSVR in sklearn). As it is already around, might as well use it. This is actually equivalent of what is done for the Classifiers.

sklearnex/svm/_base.py
icfaust: https://github.com/scikit-learn/scikit-learn/blob/c60dae2060/sklearn/svm/_base.py#L875

icfaust:
> @icfaust I'm not sure if this is specific to SVMs or whether this PR was meant to fix it, but `allow_fallback_to_host` still doesn't work with the dispatching logic changes in this PR:
> 
> ```python
> import os
> os.environ["SKLEARNEX_VERBOSE"] = "INFO"
> import numpy as np
> from sklearnex import config_context
> from sklearnex.svm import SVC
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(100,10)).astype(np.float32)
> y = rng.integers(3, size=100).astype(np.float32)
> model = SVC(random_state=123)
> with config_context(target_offload="gpu", allow_fallback_to_host=True):
>     model.fit(X, y) 
> ```
> 
> ```
> RuntimeError: SVM with multiclass support is not implemented for GPU
> ```
> 
> Logs show:
> 
> ```
> INFO:sklearnex: sklearn.svm.SVC.fit: running accelerated version on CPU
> ```

Error is unrelated to the PR.  It looks like there is an issue associated with the fallback queue logic in `onedal.utils._sycl_queue_manager.manage_global_queue` when done in a nested fashion (which is the case when offloading for `assert_all_finite` in `onedal.utils.validation`). The fallback will be set in `sklearnex._device_offload`, but the use of `manage_global_queue(None, X)` in `assert_all_finite` will reset it back to the target_offload value upon completion in the `finally` step.  Someone else has to fix this and add tests. I do not have time.

sklearnex/svm/_base.py
icfaust: This is a copy over movement from the previous code. It handles the string and numerical values of `gamma` gracefully using `repr`.

sklearnex/svm/_base.py
icfaust: removed.

sklearnex/svm/_base.py
david-cortes-intel: > This is a copy over movement from the previous code. It handles the string and numerical values of `gamma` gracefully using `repr`.

Isn't that how they would be handled by default if you remove the `!r` part? Do you observe any difference in how they are printed?

sklearnex/svm/_base.py
icfaust: Done.

sklearnex/svm/_base.py
icfaust: https://github.com/scikit-learn/scikit-learn/blob/1.1.X/sklearn/svm/_base.py#L233
If you have an absolutely compelling reason to vary this code from an in-between supported sklearn version, let me know.

sklearnex/svm/_base.py
icfaust: Then we shouldn't use sklearn's `_validate_targets` whatsoever (https://github.com/scikit-learn/scikit-learn/blob/f1261a9356fe4519ac25af3a2e2b7ec31de3be96/sklearn/svm/_base.py#L308) and would be a bug in scikit-learn-intelex main. (See interop for dataconversion: https://github.com/uxlfoundation/oneDAL/blob/main/cpp/oneapi/dal/algo/svm/backend/cpu/train_reg_kernel.cpp#L91)

david-cortes-intel: So looks like it'd end up converted if the types do not match. Why not better convert it to the target type here to avoid another conversion later on?

sklearnex/svm/_base.py
david-cortes-intel: Looks like the right methods are not being wrapped, since DAAL does offer direct log-probability calculations for some classifiers, like in here:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/b63146e398739f06071702abf69b5dfa3fad4eb2/daal4py/sklearn/linear_model/logistic_path.py#L537

sklearnex/svm/_base.py
Vika-F: @david-cortes-intel I think that the condition is correct here, I do not think that SVM supports negative weights.

david-cortes-intel: @Vika-F In that case, the condition here would be incorrect since it would allow negative weights as long as at least one is positive.

Vika-F: @david-cortes-intel you are right.
Should it be something like: ` if sample_weight is not None and (xp.any(sample_weight < 0) or xp.all(sample_weight <= 0)):` ?

david-cortes-intel: Yes, so @icfaust looks like this should get checked in the patching conditions chain, avoiding oneDAL if there are negative weights.

sklearnex/svm/_base.py
icfaust: Just for posterity: only Gradient Boosted Trees and Logistic Regression DAAL algorithms use `computeClassLogProbabilities`, neither of which have oneDAL interfaces.

doc/sources/array_api.rst
```diff
@@ -96,6 +96,10 @@ The following patched classes have support for array API inputs:
 - :obj:`sklearn.linear_model.Ridge`
 - :obj:`sklearnex.linear_model.IncrementalLinearRegression`
 - :obj:`sklearnex.linear_model.IncrementalRidge`
+- :obj:`sklearn.svm.NuSVC`
+- :obj:`sklearn.svm.NuSVR`
```
Vika-F, line 109: A stupid question, but why some classes are listed here as 'sklearn.*' and the others as 'sklearnex.*'?
Shouldn't they all be 'sklearnex.*'?

onedal/svm/tests/test_nusvr.py
```diff
@@ -196,7 +209,7 @@ def test_synth_poly_compare_with_sklearn(queue, params):
 def test_pickle(queue):
     diabetes = datasets.load_diabetes()
 
-    clf = NuSVR(kernel="rbf", C=10.0)
+    clf = NuSVR(kernel="linear", C=10.0)
```
Vika-F, line 212: Were there some problems with rbf?

doc/sources/array_api.rst
david-cortes-intel: These link to the docs of the corresponding class. There is no doc page for sklearnex classes that patch something from sklearn, so they link instead to the docs from sklearn. The sklearnex classes listed here have separate docs under this page:
https://uxlfoundation.github.io/scikit-learn-intelex/2025.8/non-scikit-algorithms.html

onedal/svm/svm.py
```diff
@@ -365,12 +191,8 @@ def __init__(
             cache_size=cache_size,
             max_iter=max_iter,
             tau=tau,
-            class_weight=None,
-            decision_function_shape=None,
-            break_ties=False,
             algorithm=algorithm,
```
Vika-F, line 194: Maybe I do not understand something, but why aren't some of the parameters passed to BaseSVM constructor?
For example, 'tol', 'schrinking', 'gamma', etc?

Vika-F: I mean, why some parameters like (cache_size) got propagated to BaseSVM while others do not?
Looks like an inconsistency to me.

doc/sources/array_api.rst
Vika-F: Thank you for the explanation.

onedal/svm/svm.py
icfaust: Sorry @Vika-F I don't understand, lines 187-190 pass `tol` `shinking` and `gamma, am I misunderstanding things?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,691 @@
...
+        predictions, _ = _get_response_values(
+            estimator,
+            X,
+            response_method=["decision_function", "predict_proba"],
+        )
+        if predictions.ndim == 1:
+            # Reshape binary output from `(n_samples,)` to `(n_samples, 1)`
+            predictions = xp.reshape(predictions, (-1, 1))
+
+        calibrated_classifier = _fit_calibrator(
```
ethanglaser: FYI array API was enabled for CalibratedClassifierCV (https://github.com/scikit-learn/scikit-learn/pull/32246/files) might slightly impact some of the code here - at least this function call now requires an xp arg

onedal/svm/tests/test_nusvr.py
icfaust: I think it was to simplify the logic of the test by not having do to rbf kernel specific values needing to be set.

sklearnex/svm/_base.py
icfaust: I have made a modification to the PR to manage the effects of scikit-learn/scikit-learn#32246, please see the change on line 82 (and good catch)

onedal/svm/svm.py
Vika-F: Yes, those are passed as arguments to the constructor, but they are not propagated further.
I expected to see something like `self.tol_ = tol` or passing those args to super().__init__.
I mean it doesn't look like those variables are propagated to oneDAL.

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,692 @@
...
+            accept_sparse="csr",
+        )
+
+        y = self._onedal_validate_targets(X, y)
+
+        if (sw_flag := sample_weight is not None) or self.class_weight is not None:
+            sample_weight = _check_sample_weight(sample_weight, X)
+            # oneDAL only accepts sample_weights, apply class_weight directly
+
+        # due to the nature of how sklearn checks nu in NuSVC (by not checking
```
david-cortes-intel, line 311: Does this perhaps have some misplaced parenthesis? Or is it perhaps missing some sentence?

david-cortes-intel:
The CI issues:
> E               AssertionError: SVC failed when fitted on one label after sample_weight trimming. Error message is not explicit, it should have 'class'.

Looks like it could be solved by adding an extra check for single-class data in the patching conditions.

For the other issue:
> WARNING: The candidate selected for download or install is a yanked version

It should be solvable by merging the latest master.

onedal/svm/svm.py
```diff
@@ -79,128 +58,72 @@ def __init__(
...
-            "max_iteration_count": int(max_iter),
-            "scale": self._scale_,
-            "sigma": self._sigma_,
-            "shift": self.coef0,
+            "kernel": self.kernel,
             "degree": self.degree,
-            "tau": self.tau,
+            "shift": self.coef0 if self.kernel != "linear" else 0.0,
+            "scale": gamma if self.kernel != "linear" else 1.0,
+            "sigma": np.sqrt(0.5 / gamma) if self.kernel != "linear" else 1.0,
```
david-cortes-intel: Is this meant to accept scalars from array API libraries? If so, shouldn't it use `math.sqrt` instead?

sklearnex/svm/_base.py
```diff
@@ -0,0 +1,692 @@
...
+        )
+        if method_name == "fit":
+            xp, _ = get_namespace(*data)
+            _, _, sample_weight = data
+            patching_status.and_conditions(
+                [
+                    (
+                        self.kernel in ["linear", "rbf", "poly", "sigmoid"],
+                        f'Kernel is "{self.kernel}" while '
+                        '"linear", "rbf", "poly" and "sigmoid" are only supported.',
```
david-cortes-intel:
```suggestion
                        'only "linear", "rbf", "poly" and "sigmoid" are supported.',
```

sklearnex/svm/_classes.py
```diff
@@ -0,0 +1,404 @@
...
+    def _onedal_gpu_supported(self, method_name, *data):
+        class_name = self.__class__.__name__
+        patching_status = PatchingConditionsChain(
+            f"sklearn.svm.{class_name}.{method_name}"
+        )
+        conditions = [
+            (
+                self.kernel in ["linear", "rbf"],
+                f'Kernel is "{self.kernel}" while '
+                '"linear" and "rbf" are only supported on GPU.',
```
david-cortes-intel:
```suggestion
                'only "linear" and "rbf" are supported on GPU.',
```

david-cortes-intel:
Looks like these changes will be required for sklearn1.8, as otherwise conformance tests throw errors about 'xp' argument in some methods.

sklearnex/svm/_base.py
Vika-F: I think nothing is missing.
Those are two different sentences that describe the code above (the first one) and below (the second one).

david-cortes-intel:
There will be some changes required for sklearn1.8 that generate merge conflicts with this PR:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/2801

Perhaps they could be all incorporated here instead if it makes the merging easier.

Vika-F:
> There will be some changes required for sklearn1.8 that generate merge conflicts with this PR: #2801
> 
> Perhaps they could be all incorporated here instead if it makes the merging easier.

@david-cortes-intel Ok, I will do that. Anyway I will be fixing pre-commit issues here.

sklearnex/svm/_classes.py
```diff
@@ -0,0 +1,424 @@
...
+    def _onedal_cpu_supported(self, method_name, *data):
+        class_name = self.__class__.__name__
+        patching_status = PatchingConditionsChain(
+            f"sklearn.svm.{class_name}.{method_name}"
+        )
+        conditions = []
+        if method_name == "fit":
+            xp, _ = get_namespace(*data)
+            _, y, sample_weight = data
+            y_nonzero = y[sample_weight > xp.full_like(sample_weight, 0)] if sample_weight is not None else y
```
david-cortes-intel: Why not compare against scalar 0 instead? It should be supported by array API.

icfaust: @david-cortes-intel @Vika-F I moved the logic to `BaseSVC` since this issue should apply for all SVM Classifiers. I also used native python to make sure that the base checks are included, and that it only adds the check if necessary. Please give it a look.

icfaust: I assume that these test failures exist for GPU too? will we need to do it there as well @Vika-F (haven't checked private CI for it)

onedal/svm/svm.py
david-cortes-intel: Fixed.

david-cortes-intel:
CI failure is not related to this PR:
```
libgomp: Thread creation failed: Resource temporarily unavailable
```

PR #1785, opened 2024-04-03
Automatic CI Doc Build

aepanchi:
# Description
Enable automatic CI doc build once the main is updated. Propose a better version control solution. 
Output is available at https://aepanchi.github.io/scikit-learn-intelex/2024.3/

Changes proposed in this pull request:
- Introduce:
    - ``.github/workflows/publish.yml`` to enable automatic CI doc build
    - ``doc/index.html`` to redirect user to the latest doc version
    - ``doc/set_version.sh`` to set the doc version
    - ``doc/sources/_static/version_switcher.js`` and ``doc/sources/_templates/versions.html`` to switch versions
    - ``doc/versions.json`` to specify the doc versions to build
- Rename ``doc/sources/_static/style.css`` to ``doc/sources/_static/custom.css`` to align with conf.py; update the file with the version dropdown button configurations
- Update conf.py to include new configurations 

Signed-off-by: Alexandra Epanchinzeva alexandra.epanchinzeva@intel.com

PR #2214, opened 2024-12-07
Version support and automatic CI build

intelkevinputnam:
## Description

replaces #1785 

Version of docs built using CI: https://intelkevinputnam.github.io/scikit-learn-intelex/

There are two major changes in this PR:

1. Version switching support
2. CI deployment to GitHub pages via GitHub action

### Version Switching Support

All published versions must be tracked in doc/versions.json as this will allow all subsequent versions of the docs to automatically be able to switch to the latest without needing to rebuild the docs.

The version is supplied by the **installed** version of scikit-learn-intelex (latest by default in the CI script). This way building the docs is not dependent on the oneDAL environment with the goal of making it easier to contribute to and edit the docs. It should also make builds faster.

Since the move to the UXL Foundation, version switching is broken in the latest version of the docs because of there is a hard coded path to the versions.json that no longer exists (it was intel.github.com/scikit-learn-intelex). It also is now settable in the conf.py (html_context)

### CI Deployment to GitHub Pages via GitHub Action

GitHub pages will be updated automatically.  A new directory will be **added** for each version of scikit-learn-intelex. When the same version is built twice, it is overwritten. See ``.github/workflows/publish.yml`` for the script. The branch will need to be changed to ``main`` before it will work on the main branch. I'm not sure it is necessary to have the docs built on every merge into ``main``. Adding ``workflow_dispatch:`` to the list of conditions for running the action can provide a handy "deploy whenever you feel like it" option. 

To build the docs locally, run ``./build-docs.sh``. Docs can be previewed locally by starting a web server (``python3 -m http.server``) in the ``_build`` directory and using ``localhost:8000/scikit-learn-intelex`` as the URL.

---

- [X] I have reviewed my changes thoroughly before submitting this pull request.
- [X] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/intel/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [X] I have resolved any merge conflicts that might occur with the base branch.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,47 @@
+name: publish
+'on':
+  push:
+    branches:
+      docs-ci
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: install dependencies
+      run: |
+        sudo apt update
+        sudo apt install -y pandoc git
+        pip3 install -r requirements-doc.txt
+        pip3 install scikit-learn-intelex
+        cd doc
+        source set_version.sh
+        echo Generating version $DOC_VERSION
+    - name: build docs
+      run: |
+        cd doc
+        chmod +x build-doc.sh
+        ./build-doc.sh
+        mv _build/scikit-learn-intelex ../../output
+    - name: deploy docs
+      run: |
+        cd doc
+        source set_version.sh
+        cd ..
+        git fetch
+        git checkout -- doc/build-doc.sh
+        git checkout gh-pages
+        rm -rf $DOC_VERSION
+        mv -f ../output/$DOC_VERSION .
+        mv ../output/versions.json .
+        mv ../output/index.html .
+        rm -rf doc
+        git config --global user.name "${GITHUB_ACTOR}"
+        git config --global user.email "${GITHUB_ACTOR}@github.com"
+        git add .
+        git commit -m "latest HTML output"
+        git push -f https://${GITHUB_ACTOR}:${{secrets.GITHUB_TOKEN}}@github.com/${GITHUB_REPOSITORY}.git HEAD:gh-pages
+
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,7 @@ name: publish
 'on':
   push:
     branches:
-      docs-ci
+      ktp-docs-ci
 
 jobs:
   build:
--- a/doc/build-doc.sh
+++ b/doc/build-doc.sh
@@ -27,6 +27,13 @@ mkdir $SAMPLES_DIR
 cd ..
 rsync -a --exclude='doc/$SAMPLES_DIR/daal4py_data_science.ipynb' examples/notebooks/*.ipynb doc/$SAMPLES_DIR
 
-# build the documentation
 cd doc
-make html
+
+source ./set_version.sh
+export SPHINXPROJ=scikit-learn-intelex
+export BUILDDIR=_build
+export SOURCEDIR=sources
+
+sphinx-build -b html $SOURCEDIR $BUILDDIR/$SPHINXPROJ/$DOC_VERSION
+cp versions.json $BUILDDIR/$SPHINXPROJ
+echo "<meta http-equiv=\"refresh\" content=\"0; URL='/$SPHINXPROJ/$DOC_VERSION/'\" / >" >> $BUILDDIR/$SPHINXPROJ/index.html
--- a/doc/get_doc_version.py
+++ b/doc/get_doc_version.py
@@ -1,0 +1,15 @@
+import subprocess
+import os
+
+def get_version():
+    major = ''
+    minor = ''
+    pip_output = subprocess.run(["pip","list"],capture_output=True,encoding='utf-8')
+    lines = pip_output.stdout.split("\n")
+    for line in lines:
+        if line.startswith("scikit-learn-intelex"):
+            version = line.split()[1].split(".")
+            major = version[0]
+            minor = version[1]
+
+    return major + "." + minor
--- a/doc/sources/_static/custom.css
+++ b/doc/sources/_static/custom.css
@@ -121,3 +121,11 @@ div.quotation {
 
 a#wap_dns {display: none;}
 
+
+button#version-switcher-button {
+   background-color:  #2980b9;
+}
+
+div#version-switcher-dropdown {
+   display: inline;
+}
--- a/doc/sources/conf.py
+++ b/doc/sources/conf.py
@@ -23,7 +23,11 @@
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/master/config
 
+bench_test = False
+
 # -- Path setup --------------------------------------------------------------
+
+import json
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
@@ -34,7 +38,6 @@ import sys
 
 sys.path.insert(0, os.path.abspath("../"))
 
-
 # -- Project information -----------------------------------------------------
 
 project = "Intel(R) Extension for Scikit-learn*"
@@ -42,9 +45,9 @@ copyright = "Intel"
 author = "Intel"
 
 # The short X.Y version
-version = "2025.0.0"
+version = os.environ.get('DOC_VERSION','')
 # The full version, including alpha/beta/rc tags
-release = "2025.0.0"
+release = version
 
 
 # -- General configuration ---------------------------------------------------
@@ -154,6 +157,13 @@ html_theme_options = {
     "titles_only": False,
 }
 
+switcher_url = "/scikit-learn-intelex/versions.json"
+if bench_test:
+    switcher_url = "/versions.json"
+
+html_context = {"current_version": version,
+                "project_name": "scikit-learn-intelex",
+                "switcher_url": switcher_url}
 
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
@@ -163,6 +173,7 @@ html_static_path = ["_static"]
 
 def setup(app):
     app.add_css_file("custom.css")
+    app.add_js_file("version_switcher.js")
 
 
 # -- Options for HTMLHelp output ---------------------------------------------
--- a/doc/sources/conf.py
+++ b/doc/sources/conf.py
@@ -23,8 +23,6 @@
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/master/config
 
-bench_test = False
-
 # -- Path setup --------------------------------------------------------------
 
 import json
@@ -158,8 +156,6 @@ html_theme_options = {
 }
 
 switcher_url = "/scikit-learn-intelex/versions.json"
-if bench_test:
-    switcher_url = "/versions.json"
 
 html_context = {"current_version": version,
                 "project_name": "scikit-learn-intelex",
--- a/doc/sources/conf.py
+++ b/doc/sources/conf.py
@@ -43,7 +43,7 @@ copyright = "Intel"
 author = "Intel"
 
 # The short X.Y version
-version = os.environ.get('DOC_VERSION','')
+version = os.environ.get("DOC_VERSION", "")
 # The full version, including alpha/beta/rc tags
 release = version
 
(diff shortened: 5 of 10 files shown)


doc/build-doc.sh
```diff
@@ -27,6 +27,13 @@ mkdir $SAMPLES_DIR
 cd ..
 rsync -a --exclude='doc/$SAMPLES_DIR/daal4py_data_science.ipynb' examples/notebooks/*.ipynb doc/$SAMPLES_DIR
 
-# build the documentation
 cd doc
-make html
+
+source ./set_version.sh
+export SPHINXPROJ=scikit-learn-intelex
```
david-cortes-intel, line 33: We also have `daal4py` with docs under `doc/daal4py`.

doc/get_doc_version.py
```diff
@@ -0,0 +1,15 @@
+import subprocess
+import os
+
+def get_version():
```
david-cortes-intel, line 4: @icfaust Could you please look into whether this captures the version number correctly.

doc/sources/_static/custom.css
```diff
@@ -121,3 +121,11 @@ div.quotation {
 
 a#wap_dns {display: none;}
 
+
+button#version-switcher-button {
+   background-color:  #2980b9;
+}
+
+div#version-switcher-dropdown {
```
david-cortes-intel, line 129: Question: if we have this line right here, do we still need to add a version selector in the `conf.py` file?
https://github.com/uxlfoundation/scikit-learn-intelex/blob/54aa18b56b853add58d8ea078efae36fb1209741/doc/sources/conf.py#L146

I understand that version selector plugin is meant for ReadTheDocs, so don't know how it plays with these.

doc/sources/conf.py
```diff
@@ -34,17 +36,16 @@
...
-
 # -- Project information -----------------------------------------------------
 
 project = "Intel(R) Extension for Scikit-learn*"
 copyright = "Intel"
 author = "Intel"
 
 # The short X.Y version
-version = "2025.0.0"
+version = os.environ.get('DOC_VERSION','')
```
david-cortes-intel, line 46: How about changing the version to "latest" here when it doesn't find the env. variable?

doc/versions.json
```diff
@@ -0,0 +1,19 @@
+[
+    {
+        "name": "2025.0 (latest)",
```
david-cortes-intel, line 3: Could we make this file auto-generated from the releases that we have as github tags instead? It's BTW missing some version numbers, such as 2024.7.0:
https://github.com/uxlfoundation/scikit-learn-intelex/releases

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+        rm -rf $DOC_VERSION
+        mv -f ../output/$DOC_VERSION .
+        mv ../output/versions.json .
+        mv ../output/index.html .
+        rm -rf doc
+        git config --global user.name "${GITHUB_ACTOR}"
+        git config --global user.email "${GITHUB_ACTOR}@github.com"
+        git add .
+        git commit -m "latest HTML output"
+        git push -f https://${GITHUB_ACTOR}:${{secrets.GITHUB_TOKEN}}@github.com/${GITHUB_REPOSITORY}.git HEAD:gh-pages
```
david-cortes-intel, line 46: Question: what happens if something goes wrong during the build? For example, we might run into errors such as failing to import the modules from which autodoc generates documentation, which would manifest as a "warning" being issued but _some_ docs still being built.

Could there be a check for build warnings that would make this script fail (avoiding pushing anything to GH) if something goes wrong?

doc/versions.json
```diff
@@ -0,0 +1,19 @@
+[
```
david-cortes-intel, line 1: What about "stable" version tag in addition to latest?

david-cortes-intel (changes requested):

david-cortes-intel:
Perhaps this PR could be a good opportunity to address one topic around the version numbers: in open source projects, version "latest" or "dev" is typically meant for docs from "latest commit of the master branch", while the "last released version" is typically "stable".

This is the case for the docs of the scikit-learn package which this repository builds upon, and for most other machine-learning-related libraries such as gradient boosting frameworks and similar.

Could we make the switch here so that the freshest release would be "stable" and have "latest" as something built periodically from the main branch?

intelkevinputnam:
> @intelkevinputnam I'm seeing several misrendered things in the docs from your branch.
> 
> For example: image
> 
> This is how it looks for me in a local build: image

Since the build script is dependent on the latest **released** version of the scikit-learn-intelex package, the changes must not have propagated to it as of the first build. The changes you mention appear to have propagated to the released package now, so the you'll see what you expect if you check again.

doc/sources/_static/custom.css
intelkevinputnam: The version selector isn't related to ReadTheDocs at all. This is a custom version switcher implemented on top of it.

doc/sources/conf.py
intelkevinputnam: If there isn't a version, the build should fail.

.github/workflows/publish.yml
intelkevinputnam: Any errors in the github workflow will stop the docs from publishing. We could turn on "error on warning" in the sphinx build, if we wanted to gate against any issue in the Sphinx build.

doc/sources/_static/custom.css
david-cortes-intel: So can we remove this `"version_selector": True` in our `conf.py` then?

doc/versions.json
david-cortes-intel: The list of releases can be obtained programmatically. This repository keeps branches with "rls" in the name for all released versions, which can be obtained with a command like this from the root of the repository:
```
git branch -r | grep rls
```

And since it follows date-based versioning, the most up-to-date one will always be the last one in alphabetical order, as returned by that command. So it should be possible to auto-generate this file with all the releases that exist, and keep it auto-updated to point to the last when a new branch comes in.

.github/workflows/publish.yml
david-cortes-intel: Yes, that'd be ideal, so that we don't accidentally end up with empty docs. But would be better if it could ignore _some_ warnings selectively - currently there are some warnings when building `daal4py` which I'm not sure how to solve and which do not impact the resulting docs that get built.

doc/versions.json
david-cortes-intel: Although the issue would be that some versions are missing docs.

.github/workflows/publish.yml
david-cortes-intel: Example benign warnings when building `daal4py`:
```
/home/dcortes/repos/scikit-learn-intelex/doc/daal4py/docstring of daal4py._daal4py.classifier_prediction_result:1: WARNING: duplicate object description of daal4py.classifier_prediction_result, other instance in algorithms, use :noindex: for one of them
```

doc/sources/conf.py
david-cortes-intel: I mean: would be ideal to have a version "latest" building the docs directly from the master branch, for unreleased versions. So in that case, the release versions would have the env. variable whereas the master branch wouldn't.

doc/sources/conf.py
```diff
@@ -25,6 +25,8 @@
 
 # -- Path setup --------------------------------------------------------------
 
+import json
```
david-cortes-intel, line 28: Looks like this import is not used.

doc/sources/_static/custom.css
david-cortes-intel: Removed the option.

david-cortes-intel:
Verified that we can indeed push changes to this PR (CC @yuejiaointel )

.github/workflows/publish.yml
icfaust: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#example-1-passing-the-github_token-as-an-input  let me know if i am missing something, sentence just before this link: "Commits pushed by a GitHub Actions workflow that uses the GITHUB_TOKEN do not trigger a GitHub Pages build."

doc/get_doc_version.py
icfaust: i would prefer if we got rid of this file entirely and replaced it with
```
DOC_VERSION=$(python -m pip freeze | grep -oP "(?<=scikit-learn-intelex==)\d*\.\d*")
```

doc/versions.json
icfaust: I have to agree with @david-cortes-intel on this one, we could technically have the ```(latest)``` tag be updated on a github ```create``` trigger when it has ```rls``` in it.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: install dependencies
+      run: |
+        sudo apt update
+        sudo apt install -y pandoc git
```
icfaust, line 17:
```suggestion
        sudo apt install -y pandoc

```
How come git is necessary?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+        pip3 install -r requirements-doc.txt
+        pip3 install scikit-learn-intelex
+        cd doc
+        source set_version.sh
+        echo Generating version $DOC_VERSION
```
icfaust, line 22:
```suggestion
        DOC_VERSION=$(python -m pip freeze | grep -oP "(?<=scikit-learn-intelex==)\d*\.\d*")
        echo Generating version $DOC_VERSION
        echo "DOC_VERSION=${DOC_VERSION}" >> "$GITHUB_ENV"
```
Add it to the github environment to save some work/lines going forward.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+      run: |
+        cd doc
+        chmod +x build-doc.sh
+        ./build-doc.sh
+        mv _build/scikit-learn-intelex ../../output
+    - name: deploy docs
+      run: |
+        cd doc
+        source set_version.sh
+        cd ..
```
icfaust, line 33:
```suggestion
```
With the previous suggestion these 3 lines are unnecessary as ```DOC_VERSION``` will be propagated in the environment between steps.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+        pip3 install scikit-learn-intelex
+        cd doc
+        source set_version.sh
+        echo Generating version $DOC_VERSION
+    - name: build docs
+      run: |
+        cd doc
+        chmod +x build-doc.sh
+        ./build-doc.sh
+        mv _build/scikit-learn-intelex ../../output
```
icfaust, line 28:
```suggestion
        doc/build-doc.sh
        mv doc/_build/scikit-learn-intelex ../output
```
Sorry since I am not so knowledgeable on the docs side. If build-doc.sh doesn't have the right permissions, we can set those in the repo to what is necessary (does it hurt to store build-doc.sh as 755 instead)?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+        source set_version.sh
+        cd ..
+        git fetch
+        git checkout -- doc/build-doc.sh
+        git checkout gh-pages
+        rm -rf $DOC_VERSION
+        mv -f ../output/$DOC_VERSION .
+        mv ../output/versions.json .
+        mv ../output/index.html .
+        rm -rf doc
```
icfaust, line 41: Does this ```rm``` wipe the ```git checkout``` of the build-doc.sh?

icfaust (commented):
Here is my initial review, we probably need someone to review the js, since I have no idea on that.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: install dependencies
+      run: |
+        sudo apt update
+        sudo apt install -y pandoc git
+        pip3 install -r requirements-doc.txt
+        pip3 install scikit-learn-intelex
```
david-cortes-intel, line 19: I had forgotten to comment here but: we'd probably want to change this logic towards either building from source, or downloading artifacts from github - otherwise this wouldn't work out when pushing to release branches (would build docs for the wrong version), nor when trying to build docs for unreleased versions.

icfaust: I totally agree.

PR #2229, opened 2024-12-17
feature: new tests added for tsne to expand test coverage

yuejiaointel:
## Description

Added additional tests in sklearnex/manifold/tests/test_tsne.py to expand the test coverage for t-SNE algorithm. The associated issue number is 8325, https://jira.devtools.intel.com/browse/DAALL-8325

**Testing**

### DIFF
--- a/sklearnex/manifold/tests/test_tsne.py
+++ b/sklearnex/manifold/tests/test_tsne.py
@@ -16,7 +16,12 @@
 
 import numpy as np
 from numpy.testing import assert_allclose
-
+import pytest
+from onedal.tests.utils._dataframes_support import (
+    _as_numpy,
+    _convert_to_dataframe,
+    get_dataframes_and_queues,
+)
 
 def test_sklearnex_import():
     from sklearnex.manifold import TSNE
@@ -24,3 +29,87 @@ def test_sklearnex_import():
     X = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
     tsne = TSNE(n_components=2, perplexity=2.0).fit(X)
     assert "daal4py" in tsne.__module__
+
+from sklearnex.manifold import TSNE
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+def test_sklearnex_tsne_import(dataframe, queue):
+    """Test TSNE compatibility with different backends and queues, and validate sklearnex module."""
+    X = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+    tsne = TSNE(n_components=2, perplexity=2.0).fit(X_df)
+    assert "daal4py" in tsne.__module__
+    assert hasattr(tsne, "n_components"), "TSNE missing 'n_components' attribute."
+    assert tsne.n_components == 2, "TSNE 'n_components' attribute is incorrect."
+
+def test_valid_tsne_functionality():
+    """Test TSNE with valid data: basic functionality, random data, reproducibility, and edge cases."""
+    # Test basic functionality
+    X_basic = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
+    tsne = TSNE(n_components=2, perplexity=2.0).fit(X_basic)
+    assert tsne.embedding_.shape == (4, 2)
+    
+    # Test with random data
+    np.random.seed(42)
+    X_random = np.random.rand(100, 10)
+    tsne_random = TSNE(n_components=2, perplexity=30.0).fit(X_random)
+    assert tsne_random.embedding_.shape == (100, 2)
+
+    # Test reproducibility
+    X_repro = np.random.rand(50, 10)
+    tsne_1 = TSNE(n_components=2, random_state=42).fit_transform(X_repro)
+    tsne_2 = TSNE(n_components=2, random_state=42).fit_transform(X_repro)
+    assert_allclose(tsne_1, tsne_2, rtol=1e-5)
+
+    # Test perplexity edge case (close to dataset size)
+    X_perplexity = np.random.rand(10, 5)
+    tsne_perplexity = TSNE(n_components=2, perplexity=9).fit(X_perplexity)
+    assert tsne_perplexity.embedding_.shape == (10, 2)
+
+    # Test large data
+    X_large = np.random.rand(1000, 50)
+    tsne_large = TSNE(n_components=2, perplexity=50.0).fit(X_large)
+    assert tsne_large.embedding_.shape == (1000, 2)
+
+    # Test valid minimal data
+    X_valid = np.array([[0, 0], [1, 1], [2, 2]])
+    tsne_valid = TSNE(n_components=2, perplexity=2).fit(X_valid)
+    assert tsne_valid.embedding_.shape == (3, 2)
+
+def test_tsne_edge_cases_and_errors():
+    """Test TSNE with invalid, constant, and edge-case data."""
+    # Edge case: constant data
+    X_constant = np.ones((10, 10))
+    with pytest.raises(ValueError) as excinfo:
+        TSNE(n_components=2, perplexity=20).fit(X_constant)
+    assert "perplexity must be less than n_samples" in str(excinfo.value)
+
+    # Edge case: empty data
+    X_empty = np.empty((0, 10))
+    with pytest.raises(ValueError):
+        TSNE(n_components=2).fit(X_empty)
+
+    # Edge case: data with NaN or infinite values
+    X_invalid = np.array([[0, 0], [1, np.nan], [2, np.inf]])
+    with pytest.raises(ValueError):
+        TSNE(n_components=2).fit(X_invalid)
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+@pytest.mark.parametrize("num_blocks", [1, 2, 4])
+def test_tsne_full_fit_with_blocks(dataframe, queue, dtype, num_blocks):
+    """Test TSNE fitted on the full dataset, after splitting into blocks."""
+    np.random.seed(42)
+    X = np.random.rand(100, 20).astype(dtype)  # 100 samples, 20 features
+    X_blocks = np.array_split(X, num_blocks)   # Split into `num_blocks`
+
+    # Combine blocks back into a single dataset
+    X_combined = np.vstack(X_blocks)
+    X_df = _convert_to_dataframe(X_combined, sycl_queue=queue, target_df=dataframe)
+
+    # Fit TSNE on the combined dataset
+    tsne = TSNE(n_components=2, perplexity=30.0, random_state=42).fit(X_df)
+
+    # Validate embedding shape
+    assert tsne.embedding_.shape == (100, 2)
+


sklearnex/manifold/tests/test_tsne.py
```diff
@@ -16,11 +16,175 @@
 
 import numpy as np
 from numpy.testing import assert_allclose
-
+import pytest
+#Note: n_componets must be 2 for now
```
icfaust, line 20:
```suggestion
#Note: n_components must be 2 for now

```

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,192 @@
...
+    tsne_1 = TSNE(n_components=2, random_state=42).fit_transform(X_repro)
+    tsne_2 = TSNE(n_components=2, random_state=42).fit_transform(X_repro)
+    assert_allclose(tsne_1, tsne_2, rtol=1e-5)
+
+    # Test perplexity close to dataset size
+    X_perplexity = np.random.rand(10, 5)
+    tsne_perplexity = TSNE(n_components=2, perplexity=9).fit(X_perplexity)
+    assert tsne_perplexity.embedding_.shape == (10, 2)
+
+    # Test large data
```
david-cortes-intel, line 75: It feels like this one is perhaps not needed, considering that there's already a similar test earlier on with shape (100,10).

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,192 @@
...
+    # Test valid minimal data
+    X_valid = np.array([[0, 0], [1, 1], [2, 2]])
+    tsne_valid = TSNE(n_components=2, perplexity=2).fit(X_valid)
+    assert tsne_valid.embedding_.shape == (3, 2)
+
+    # Edge case: constant data
+    X_constant = np.ones((10, 10))
+    tsne = TSNE(n_components=2, perplexity=5, random_state=42)
+    embedding = tsne.fit(X_constant).embedding_
+    assert embedding.shape == (10, 2), f"Unexpected embedding shape: {embedding.shape}"
```
david-cortes-intel, line 89: I believe `pytest` already does the substitution for you in the error message when the test fails.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,192 @@
...
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+    tsne = TSNE(n_components=2, perplexity=2.0).fit(X_df)
+    assert "daal4py" in tsne.__module__
+    assert hasattr(tsne, "n_components"), "TSNE missing 'n_components' attribute."
+    assert tsne.n_components == 2, "TSNE 'n_components' attribute is incorrect."
+
+
+def test_basic_tsne_functionality():
+    """Test TSNE with valid data: basic functionality, random data, reproducibility, and edge cases."""
+    # Test basic functionality
```
david-cortes-intel, line 53: Any chance that all of these could be parameterized instead? Otherwise, if one of them fails, then the rest wouldn't execute.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,192 @@
...
+        TSNE(n_components=2).fit(X_empty)
+
+    # Edge case: data with NaN or infinite values
+    X_invalid = np.array([[0, 0], [1, np.nan], [2, np.inf]])
+    with pytest.raises(ValueError):
+        TSNE(n_components=2).fit(X_invalid)
+
+    # Edge Case: Sparse-Like High-Dimensional Data
+    np.random.seed(42)
+    X_sparse_like = np.random.rand(50, 10000) * (np.random.rand(50, 10000) > 0.99)
```
david-cortes-intel, line 103: Perhaps 10k is a bit too large for _just_ a test case. Could it be shrunk to e.g. 500 or so? If needed, perhaps you could use the random generator from `scipy.sparse` and then cast it to dense.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,192 @@
...
+        pytest.fail(f"TSNE failed on the specific complex dataset: {e}")
+
+
+@pytest.mark.parametrize(
+    "dataframe,queue", get_dataframes_and_queues(device_filter_="gpu")
+)
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_tsne_gpu_validation(dataframe, queue, dtype):
+    """
+    GPU validation test for TSNE with a specific complex dataset.
```
david-cortes-intel, line 168: Could this one be merged into the one above? (`test_tsne_with_specific_complex_dataset`)

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,192 @@
 # ===============================================================================
 
 import numpy as np
+import pytest
 from numpy.testing import assert_allclose
 
+# Note: n_components must be 2 for now
```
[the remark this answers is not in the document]
yuejiaointel: Hi David, 
Thx for the comments, I made a new commit that addressed other comments but for this one, Ethan double checked with Victoriya that TSNE works only for n_components == 2, here is the specific message:
yes our TSNE works only for n_components == 2.
See in DAAL (though the error message is not very descriptive): https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/src/algorithms/tsne/tsne_gradient_descent_impl.i#L1000
 
The reason of this behavior is that TSNE is a visualization algorithm. Its results are plotted to see the hidden structures in originally multidimensional data (usually 2d plots are used, more rarely - 3d).
So, having TSNE patched only for n_components == 2 looks like a reasonable compromise.

sklearnex/manifold/tests/test_tsne.py
david-cortes-intel: Looks like the tests for non-missingness were lost after merging them. Could be good to have them just in case, considering that it tests GPU execution where more things can go wrong.

sklearnex/manifold/tests/test_tsne.py
david-cortes-intel: Thanks for looking into it. But what I meant was to parameterize this whole function by turning the inputs and expectations, so that one parameterization would be "Test basic functionality", another parameterization "Test with random data", and so on.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,192 @@
...
+
+def test_basic_tsne_functionality():
+    """Test TSNE with valid data: basic functionality, random data, reproducibility, and edge cases."""
+    # Test basic functionality
+    X_basic = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
+    tsne = TSNE(n_components=2, perplexity=2.0).fit(X_basic)
+    assert tsne.embedding_.shape == (4, 2)
+
+    # Test with random data
+    np.random.seed(42)
```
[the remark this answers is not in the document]
david-cortes-intel: What I meant here was this module: https://numpy.org/doc/stable/reference/random/generator.html

To use it, you need to generate a `Generator` object, for example like this:
```python
```

Then, you need to call methods from that generator object, like this:
```python
rng.random(size=(100,10))
```

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David, 
Thx for the comments, I brought back the deleted test and merged with the previous complex test, one of the check is only valid on gpu device so I moved get_dataframes_and_queues from decorator to code to accept filter device as a parameter.
Best, Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David,
Thx for the suggestion, I refactored this function again to have parametrize each sub test, hopefully this is the right idea  
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David, I removed this test. Best, Yue

sklearnex/manifold/tests/test_tsne.py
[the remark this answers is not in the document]
yuejiaointel: Hi David, 
Thx for the suggestion, I fixed the code to use rng generator for all sub tests in test_tsne_functionality_and_edge_cases for independent results.
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,242 @@
...
+        (
+            "Test with random data",
+            lambda rng: rng.random((100, 10)),
+            2,
+            30.0,
+            (100, 2),
+            False,
+        ),
+        (
+            "Test reproducibility",
```
david-cortes-intel, line 71: I think the reproducibility test here was lost.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,242 @@
...
+
+        try:
+            tsne = TSNE(n_components=n_components, perplexity=perplexity, random_state=42)
+            embedding = tsne.fit_transform(X_df)
+
+            # Validate results
+            assert (
+                embedding.shape == expected_shape
+            ), f"{description}: Incorrect embedding shape."
+            if device_filter == "gpu":
```
david-cortes-intel, line 248: I think this doesn't need to be specific to GPU.

sklearnex/manifold/tests/test_tsne.py
david-cortes-intel: Thanks, although if the "Test reproducibility" test is also checking the shape, it still feels like "Test large data" could be removed altogether since it's not testing anything different. Or does the algorithm have some size-dependent behavior that would change between `(50,10)` and (1000, 50)`?

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,242 @@
...
+            "Edge case: data with NaN or infinite values",
+            lambda rng: np.array([[0, 0], [1, np.nan], [2, np.inf]]),
+            2,
+            5.0,
+            None,
+            True,
+        ),
+        (
+            "Edge Case: Sparse-Like High-Dimensional Data",
+            lambda rng: rng.random((50, 500)) * (rng.random((50, 500)) > 0.99),
```
david-cortes-intel, line 120: For less-random reproducibility, perhaps it could add a test where one column is constant or an exact duplicate of another column.

david-cortes-intel:
It looks like we don't have any test here nor in daal4py that would be checking that the results from TSNE make sense beyond having the right shape and non-missingness.

Since there's a very particular dataset here for the last test, it'd be helpful to add other assertions there along the lines of checking that the embeddings end up making some points closer than others as would be expected given the input data.

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David, 
Thx for the suggestion! I removed test large data, it used to be much larger but I agree 1000 and 50 are not that different
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David!
Good catch I moved reproducibility test to a new test function test_tsne_reproducibility to avoid ad hoc code in test_tsne_functionality_and_edge_cases
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David,
You are right, I tried on my testing machine again and somehow it works this time, must be some env settings not correctly yesterday. I removed the device filter and put the parametrization decorator back for get_dataframes_and_queues
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,242 @@
...
+    assert "daal4py" in tsne.__module__
+    assert hasattr(tsne, "n_components"), "TSNE missing 'n_components' attribute."
+    assert tsne.n_components == 2, "TSNE 'n_components' attribute is incorrect."
+
+
+@pytest.mark.parametrize(
+    "description,X_generator,n_components,perplexity,expected_shape,should_raise",
+    [
+        (
+            "Test basic functionality",
```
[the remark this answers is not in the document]
yuejiaointel: Hi David,
Thx for the pointer! I read the doc and used pytest.param and put the names as id. Hope this is the right idea!
Best,
Yue

yuejiaointel:
Hi David, 
About the last comment, I think that is a good test to add! I spent some time thinking it through and have added a logic check in the final test to evaluate the overlap of close neighbors. Here’s a summary of the steps I implemented:
1. get a distance array where [i, j] is Euclidean distance of point i and j in original space, same for tsne embedding space
2. rank distances for each point wrt first column in original space, also for embedding space
3. get top 5 neighbors of  each point in original and embedding space see how many are same by dividing them
4. get a mean of all fractions it should represent how the original and embedding space are similar for the most 5 closest points
5. check if that mean is > 0.6
Let me know your thoughts on this approach or if you believe it could be improved further.
Thx a lot :D
Yue

david-cortes-intel:
> Hi David, About the last comment, I think that is a good test to add! I spent some time thinking it through and have added a logic check in the final test to evaluate the overlap of close neighbors. Here’s a summary of the steps I implemented:
> 
> 1. get a distance array where [i, j] is Euclidean distance of point i and j in original space, same for tsne embedding space
> 2. rank distances for each point wrt first column in original space, also for embedding space
> 3. get top 5 neighbors of  each point in original and embedding space see how many are same by dividing them
> 4. get a mean of all fractions it should represent how the original and embedding space are similar for the most 5 closest points
> 5. check if that mean is > 0.6
>    Let me know your thoughts on this approach or if you believe it could be improved further.
>    Thx a lot :D
>    Yue

I think given the characteristics of the data that you are passing, it could be done by selecting some hard-coded set of points by index from "Complex Dataset1" that should end up being similar, and some selected set of points that should end up being dissimilar to the earlier ones; with the test then checking that the euclidean distances in the embedding space among each point from the first set are smaller than the distances between each point in the first set and each point in the second set.

Also maybe "Complex Dataset2" is not needed.

yuejiaointel:
> > Hi David, About the last comment, I think that is a good test to add! I spent some time thinking it through and have added a logic check in the final test to evaluate the overlap of close neighbors. Here’s a summary of the steps I implemented:
> > 
> > 1. get a distance array where [i, j] is Euclidean distance of point i and j in original space, same for tsne embedding space
> > 2. rank distances for each point wrt first column in original space, also for embedding space
> > 3. get top 5 neighbors of  each point in original and embedding space see how many are same by dividing them
> > 4. get a mean of all fractions it should represent how the original and embedding space are similar for the most 5 closest points
> > 5. check if that mean is > 0.6
> >    Let me know your thoughts on this approach or if you believe it could be improved further.
> >    Thx a lot :D
> >    Yue
> 
> I think given the characteristics of the data that you are passing, it could be done by selecting some hard-coded set of points by index from "Complex Dataset1" that should end up being similar, and some selected set of points that should end up being dissimilar to the earlier ones; with the test then checking that the euclidean distances in the embedding space among each point from the first set are smaller than the distances between each point in the first set and each point in the second set.
> 
> Also maybe "Complex Dataset2" is not needed.

Hi David!
I fixed the logic based on your suggestion. Now it will try to get a group A with similar points and group B with different points from group A, then I check in embedding space distance bt any 2 points in group A should be less than that point to any point in group B. I hope I understand you correctly. I also removed complex test 2.
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,256 @@
...
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+    tsne_1 = TSNE(n_components=2, random_state=42).fit_transform(X_df)
+    tsne_2 = TSNE(n_components=2, random_state=42).fit_transform(X_df)
+    # in case of dpctl.tensor.usm_ndarray covert to numpy array
+    tsne_1 = _as_numpy(tsne_1)
+    tsne_2 = _as_numpy(tsne_2)
+    assert_allclose(tsne_1, tsne_2, rtol=1e-5)
+
+
+def compute_pairwise_distances(data):
```
david-cortes-intel, line 191: Better to use the implementation in this same package:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/metrics/pairwise.py

Or failing that, in the dependencies, like these:
https://scikit-learn.org/dev/modules/generated/sklearn.metrics.pairwise_distances.html
https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.euclidean.html

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,256 @@
...
+
+    # Validate results
+    assert embedding.shape == expected_shape, f"Incorrect embedding shape."
+    embedding = _as_numpy(embedding)
+    assert np.all(np.isfinite(embedding)), f"Embedding contains NaN or infinite values."
+    assert np.any(embedding != 0), f"Embedding contains only zeros."
+
+    # Ensure close points in original space remain close in embedding
+    group_a_indices = [0, 1, 2]  # Hardcoded index of similar points
+    group_b_indices = [3, 4, 5]  # Hardcoded index of dissimilar points from a
```
david-cortes-intel, line 259: How about using the points that were already available and had also differences in signatures - e.g.
```
[2e9, 2e-9, -2e9, -2e-9]
```

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,256 @@
...
+    """
+    data = _as_numpy(data)
+    distances = np.linalg.norm(data[:, np.newaxis, :] - data[np.newaxis, :, :], axis=-1)
+    return distances
+
+
+@pytest.mark.parametrize(
+    "X,n_components,perplexity,expected_shape",
+    [
+        pytest.param(
```
david-cortes-intel, line 204: Since there's only one parameterization, maybe these could be defined inside the function body.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,256 @@
...
+    n_components,
+    perplexity,
+    expected_shape,
+    should_raise,
+    dataframe,
+    queue,
+    dtype,
+):
+    """
+    TSNE test covering multiple functionality and edge cases using parameterization.
```
david-cortes-intel, line 156: I think this comment is redundant given the function name.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,256 @@
...
+    # Ensure close points in original space remain close in embedding
+    group_a_indices = [0, 1, 2]  # Hardcoded index of similar points
+    group_b_indices = [3, 4, 5]  # Hardcoded index of dissimilar points from a
+    embedding_distances = compute_pairwise_distances(
+        embedding
+    )  # Get an array of distance where [i, j] is distance b/t i and j
+    # Check for distance b/t two points in group A < distance of this point and any point in group B
+    for i in group_a_indices:
+        for j in group_a_indices:
+            if i != j:
```
david-cortes-intel, line 266: This condition should not be needed,

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,256 @@
...
+            False,
+            id="Sparse-like data with every tenth element zero",
+        ),
+        pytest.param(
+            lambda rng: rng.random((10, 5)),
+            2,
+            0.5,
+            (10, 2),
+            False,
+            id="Extremely low perplexity",
```
david-cortes-intel, line 139: Considering what this is testing, perhaps it could add a check for no infinites , no NaNs, and no all-zeros in the embeddings in this test too.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,256 @@
...
+
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+def test_sklearnex_tsne_import(dataframe, queue):
+    """Test TSNE compatibility with different backends and queues, and validate sklearnex module."""
+    X = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+    tsne = TSNE(n_components=2, perplexity=2.0).fit(X_df)
+    assert "daal4py" in tsne.__module__
+    assert hasattr(tsne, "n_components"), "TSNE missing 'n_components' attribute."
```
david-cortes-intel, line 47: This one should already get tested as part of the line that comes next.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -15,12 +15,256 @@
...
+    assert tsne.n_components == 2, "TSNE 'n_components' attribute is incorrect."
```
david-cortes-intel, line 48: Could also check for 'perplexity' as it was passed to the constructor too.

yuejiaointel: Hi David,
Thx for the suggestion! Added
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David,
Thx for the suggestion! removed
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David,
Thx for the suggestion! Added
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David,
Thx for the suggestion! Removed
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David,
Thx for the suggestion! Removed
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David,
Thx for the suggestion! Moved parameterized def into function body
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David,
That is good note! I used the sklearn one and it works
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi David,
Thx for the suggestion! Now I picked a few ones like his one in later rows, it works, I'll still run multiple times of ci just to make sure 
Best,
Yue

david-cortes-intel:
Odd that the CI fails for sklearn1.0 and 1.1. I see that there's many places throughout the code with conditions for sklearn<1.2 though:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/daal4py/sklearn/manifold/_t_sne.py
Perhaps something is going wrong there.

yuejiaointel:
> Odd that the CI fails for sklearn1.0 and 1.1. I see that there's many places throughout the code with conditions for sklearn<1.2 though: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/daal4py/sklearn/manifold/_t_sne.py Perhaps something is going wrong there.

Hi David!
The tests are failing when checking 0 embedding when data is constant. But actually after I read the implementation I think it make sense. About the version difference, **default initialization in TSNE is changed from random to pca from 1.2** and that is why we see different behavior for different versions. Here is my understanding why we are getting non 0 embedding when we are using self._init = "random":
1. First we are generating a random embedding at first
X_embedded = 1e-4 * random_state.standard_normal(size=(n_samples, n_components))
2. Next the joint probabilities are meaningless since the pairwise distance are 0 but there could still be a value so it will still push the embedding in the gradient decent process and in the end we see some degenerated embedding.
P = _joint_probabilities(distances, self.perplexity, self.verbose)

I actually tested using sklearn and see same behavior here is my simple script and I also see a non zero embedding in the end.
from sklearn.manifold import TSNE
import numpy as np
constant_data = np.full((10, 10), fill_value=1)
tsne = TSNE(n_components=2, init='random',random_state=42, perplexity=5)
embedding = tsne.fit_transform(constant_data)
print(embedding)

I removed the check for 0 embedding because we see zero and non zero embedding for different versions, hope this helps!
Yue

david-cortes-intel:
> > Odd that the CI fails for sklearn1.0 and 1.1. I see that there's many places throughout the code with conditions for sklearn<1.2 though: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/daal4py/sklearn/manifold/_t_sne.py Perhaps something is going wrong there.
> 
> Hi David! The tests are failing when checking 0 embedding when data is constant. But actually after I read the implementation I think it make sense. About the version difference, **default initialization in TSNE is changed from random to pca from 1.2** and that is why we see different behavior for different versions. Here is my understanding why we are getting non 0 embedding when we are using self._init = "random":
> 
> 1. First we are generating a random embedding at first
>    X_embedded = 1e-4 * random_state.standard_normal(size=(n_samples, n_components))
> 2. Next the joint probabilities are meaningless since the pairwise distance are 0 but there could still be a value so it will still push the embedding in the gradient decent process and in the end we see some degenerated embedding.
>    P = _joint_probabilities(distances, self.perplexity, self.verbose)
> 
> I actually tested using sklearn and see same behavior here is my simple script and I also see a non zero embedding in the end. from sklearn.manifold import TSNE import numpy as np constant_data = np.full((10, 10), fill_value=1) tsne = TSNE(n_components=2, init='random',random_state=42, perplexity=5) embedding = tsne.fit_transform(constant_data) print(embedding)
> 
> I removed the check for 0 embedding because we see zero and non zero embedding for different versions, hope this helps! Yue

Then please move it into a separate test, using `init="pca"` in the constructor.

Also I don't think the test should look for exact zeros everywhere. For constant data, should test for a constant first dimension, and a zero second dimension.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -24,3 +33,222 @@ def test_sklearnex_import():
...
+        assert np.all(
+            np.isfinite(embedding)
+        ), f"Embedding contains NaN or infinite values."
+
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_tsne_reproducibility(dataframe, queue, dtype):
+    """
+    Test reproducibility
```
david-cortes-intel, line 189: Let's better remove obvious or redundant comments. Same for example for error messages in assertions - you will see the line that failed in a log when that happens, and many of those cases will be pretty obvious from the code line, line "isfinite failed".

yuejiaointel: Hi David,
Thx for the advice! I removed all obvious comments
Best,
Yue

yuejiaointel:
> > > Odd that the CI fails for sklearn1.0 and 1.1. I see that there's many places throughout the code with conditions for sklearn<1.2 though: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/daal4py/sklearn/manifold/_t_sne.py Perhaps something is going wrong there.
> > 
> > 
> > Hi David! The tests are failing when checking 0 embedding when data is constant. But actually after I read the implementation I think it make sense. About the version difference, **default initialization in TSNE is changed from random to pca from 1.2** and that is why we see different behavior for different versions. Here is my understanding why we are getting non 0 embedding when we are using self._init = "random":
> > 
> > 1. First we are generating a random embedding at first
> >    X_embedded = 1e-4 * random_state.standard_normal(size=(n_samples, n_components))
> > 2. Next the joint probabilities are meaningless since the pairwise distance are 0 but there could still be a value so it will still push the embedding in the gradient decent process and in the end we see some degenerated embedding.
> >    P = _joint_probabilities(distances, self.perplexity, self.verbose)
> > 
> > I actually tested using sklearn and see same behavior here is my simple script and I also see a non zero embedding in the end. from sklearn.manifold import TSNE import numpy as np constant_data = np.full((10, 10), fill_value=1) tsne = TSNE(n_components=2, init='random',random_state=42, perplexity=5) embedding = tsne.fit_transform(constant_data) print(embedding)
> > I removed the check for 0 embedding because we see zero and non zero embedding for different versions, hope this helps! Yue
> 
> Then please move it into a separate test, using `init="pca"` in the constructor.
> 
> Also I don't think the test should look for exact zeros everywhere. For constant data, should test for a constant first dimension, and a zero second dimension.

Hi David,
Thx for the advice, I created a test for constant data and check both random and pca initialization
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -24,3 +33,213 @@ def test_sklearnex_import():
...
+        embedding = tsne.fit_transform(X_df)
+        embedding = _as_numpy(embedding)
+        assert embedding.shape == expected_shape
+        assert np.all(np.isfinite(embedding))
+        assert np.any(embedding != 0)
+
+
+@pytest.mark.parametrize("init", ["pca", "random"])
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_tsne_constant_data(init, dtype):
```
david-cortes-intel, line 170: This one could also test on GPU.

david-cortes-intel (approved):
LGTM, pending CI results.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -24,3 +33,216 @@ def test_sklearnex_import():
...
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+def test_sklearnex_tsne_import(dataframe, queue):
+    """Test TSNE compatibility with different backends and queues, and validate sklearnex module."""
+    X = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+    tsne = TSNE(n_components=2, perplexity=2.0).fit(X_df)
+    assert "daal4py" in tsne.__module__
+    assert tsne.n_components == 2
+    assert tsne.perplexity == 2.0
```
ethanglaser, line 49: In my opinion the only thing missing from test scope is an exact result check, which I agree would be unrealistic for all tests, but since the import test is small would it be possible to add a specific embedding value check here after transform? Something like we have for pca here: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/decomposition/tests/test_pca.py#L35-L41

ethanglaser (commented):
Great work on this so far! A few small comments from my side but looks like its almost ready for merge.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -24,3 +33,216 @@ def test_sklearnex_import():
     X = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
     tsne = TSNE(n_components=2, perplexity=2.0).fit(X)
     assert "daal4py" in tsne.__module__
+
+
+from sklearnex.manifold import TSNE
+
+
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+def test_sklearnex_tsne_import(dataframe, queue):
```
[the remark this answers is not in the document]
yuejiaointel: Hi Ethan,
Thx for reviewing this! I removed first test and renamed the second import test
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -24,3 +33,216 @@ def test_sklearnex_import():
     X = np.array([[0, 0, 0], [0, 1, 1], [1, 0, 1], [1, 1, 1]])
     tsne = TSNE(n_components=2, perplexity=2.0).fit(X)
     assert "daal4py" in tsne.__module__
+
+
+from sklearnex.manifold import TSNE
```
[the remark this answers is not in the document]
yuejiaointel: Hi Ethan,
Thx for the suggestion! I moved the import to inside every test 
Best,
Yue

sklearnex/manifold/tests/test_tsne.py
yuejiaointel: Hi Ethan,
That is a good idea, thx for suggestion, I added a test for exact value in import test. Now just waiting to see if CI can pass
Best,
Yue

david-cortes-intel: @ethanglaser TSNE doesn't have a unique global optimum, so adding a test for exact numbers now is likely going to lead to problems later when it gets executed in e.g. newer machines, newer GPUs, etc. But there is a test checking for general goodness of the results at the end, only the criteria it needs to satisfy is not strict.

PR #2230, opened 2024-12-19
TEST: Use deterministic seeds in tests

david-cortes-intel:
## Description

This PR modifies tests parameterized by non-deterministic seeds to use fixed seeds.

Having these randomly-determined seeds has the issue that it makes it hard to compare a specific test over several runs (e.g. for different configurations), in addition to making it harder to reproduce potential failures.

---

**Testing**

**Performance**

Not applicable.

### DIFF
--- a/onedal/utils/tests/test_validation.py
+++ b/onedal/utils/tests/test_validation.py
@@ -14,8 +14,6 @@
 # limitations under the License.
 # ==============================================================================
 
-import time
-
 import numpy as np
 import numpy.random as rand
 import pytest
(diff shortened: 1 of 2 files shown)


samir-nasibli (approved):
👍

icfaust:
@david-cortes-intel @samir-nasibli i wish i had been consulted on this change. 1) because I introduced this code and 2) the proper management of determism shouldn't rely on assumptions of the functionality of oneDAL (e.g. that default seeds exist for oneDAL) 3) if run to run determism is necessary for certain tests in multiple configurations we should be doing comparisons using only subsets of the tests. if this is with respect to another testing framework, we should define and publish requirements to those. Please note on the horizon it is planned to introduce fuzzy testing (as an initiative). as a note, i do not think reverting is necessary, what is done is done.

PR #2237, opened 2025-01-01
[testing, CI] fix coverage statistics issue caused by ```test_common.py``` tracer patching

icfaust:
## Description

Python code coverage has been introduced in #2222 and #2225. Some code, while properly covered with tests, was showing incorrect coverage.  It was discovered that the tracing used in ```test_common.py``` interferes with the tracing collected by pytest-cov.  Any tests alphabetically after ```test_common.py``` have been neglected in codecov and by coverage.py.  Correcting this can be done by tracing the estimators in a separate process and passing the data back to the pytest-cov parent process.

This has been implemented using python multiprocessing due to the time overhead of loading sklearnex for each estimator and method.  The process persists for the period of all ```test_common.py``` tests as a daemon process.  The test passing/xfail/etc rate has been verified to match to main.

No performance metrics are generated due to purely for testing and CI.

The effectiveness of this change can be observed by the ~1.2% improvement in code coverage (as indirect changes).

---

**Testing**

**Performance**


ethanglaser:
Is 1.2% increase what you would expect? In general, where does the other missing coverage come from?

icfaust:
> Is 1.2% increase what you would expect? In general, where does the other missing coverage come from?

@ethanglaser good question! Its mainly 2 aspects: 1) The utils/validation.py tests coverage was not being recorded, 2) centralized testing (especially test_patching.py) was not being recorded, which covers all other methods of estimators (especially ```score```) which otherwise might not be specifically tested in the individual estimator tests. With the fix these are now recorded, and that means these tests provide an additional 1.2% of the codebase coverage not seen in sklearn conformance or estimator/onedal tests.

sklearnex/tests/test_common.py
```diff
@@ -225,23 +228,120 @@ def _commonpath(inp):
 _TRACE_BLOCK_LIST = _whitelist_to_blacklist()
 
 
+def sklearnex_trace(estimator, method):
```
ahuber21, line 231:
```suggestion
def sklearnex_trace(estimator_name, method_name):
```

sklearnex/tests/test_common.py
```diff
@@ -225,23 +228,120 @@ def _commonpath(inp):
...
+    """Generate a trace of all function calls in calling estimator.method.
+
+    Parameters
+    ----------
+    estimator : str
```
ahuber21, line 236:
```suggestion
    estimator_name : str
```

sklearnex/tests/test_common.py
```diff
@@ -225,23 +228,120 @@ def _commonpath(inp):
...
+    text: str
+        Returns a string output (captured stdout of a python Trace call). It is a
+        modified version to be more informative, completed by a monkeypatching
+        of trace._modname.
+    """
+    # get estimator
+    try:
+        est = PATCHED_MODELS[estimator]()
+    except KeyError:
+        est = SPECIAL_INSTANCES[estimator]
```
ahuber21, line 253:
```suggestion
    estimator = (SPECIAL_INSTANCES | PATCHED_MODELS)[estimator_name]
```

sklearnex/tests/test_common.py
```diff
@@ -225,23 +228,120 @@ def _commonpath(inp):
...
 
+def sklearnex_trace(estimator, method):
+    """Generate a trace of all function calls in calling estimator.method.
+
+    Parameters
+    ----------
+    estimator : str
+        name of estimator which is a key from PATCHED_MODELS or SPECIAL_INSTANCES
+
+    method : str
```
ahuber21, line 239:
```suggestion
    method_name : str
```

sklearnex/tests/test_common.py
```diff
@@ -225,23 +228,120 @@ def _commonpath(inp):
...
+    """
+    # get estimator
+    try:
+        est = PATCHED_MODELS[estimator]()
+    except KeyError:
+        est = SPECIAL_INSTANCES[estimator]
+
+    # get dataset
+    X, y = gen_dataset(est)[0]
+    # fit dataset if method does not contain 'fit'
```
ahuber21, line 257:
```suggestion
    # methods other than `fit` require a fitted estimator
```

sklearnex/tests/test_common.py
```diff
@@ -225,23 +228,120 @@ def _commonpath(inp):
...
+        trace._modname = _fullpath
+        tracer = trace.Trace(
+            count=0,
+            trace=1,
+            ignoredirs=_TRACE_BLOCK_LIST,
+        )
+        # call trace on method with dataset
+        f = io.StringIO()
+        with redirect_stdout(f):
+            tracer.runfunc(call_method, est, method, X, y)
```
ahuber21, line 275: Just to be sure, we don't care that we potentially use the same dataset for `fit` and whatever is requested in `method_name`?

ahuber21 (changes requested):

sklearnex/tests/test_common.py
icfaust: I had to revert this, because SPECIAL_INSTANCES is a special dictionary of estimators which uses sklearn's clone (in order to guarantee that there is no hysteresis between uses of the instance). And the patched models are classes. To be honest, the difference is tech debt that I introduced at the beginning of 2024, as I was trying to unify the centralized testing. Hindsight I would structure things like SPECIAL_INSTANCES.

sklearnex/tests/test_common.py
ahuber21: I thought so. In this case it shouldn't make a difference.

sklearnex/tests/test_common.py
ahuber21: Then I'd go with `PATCHED_MODELS.get(estimator_name, None) or SPECIAL_INSTANCES[estimator_name]`. I don't want to waste 4 lines on something that doesn't contribute to the function logic.

ahuber21: Sorry for sending you down a rabbit hole. I would still prefer a different implementation because I think `try/except` is a bit of an overkill

```python
estimator = PATCHED_MODELS[estimator_name] if estimator_name in PATCHED_MODELS else SPECIAL_INSTANCES[estimator_name]
```

icfaust: no apologies necessary, it showed i hadnt handled a failure case properly and would lead to pytest hanging, which would have been a nightmare to debug. now it should error 'gracefully' in CI (testing it now)

icfaust:
test if new failure system works on last commit

ahuber21 (approved):

PR #2240, opened 2025-01-03
[enhancement] replace ```NTHREADS``` threading control with ```MAKEFLAGS```, remove daal4py extension build parallelization

icfaust:
## Description

Follow on from #1196 which solves a leftover todo.  This sets the cmake processes and cythonize multithreading based on the ```MAKEFLAGS``` ```-j``` flag.  Written in a way that if a command line argument ```-j 2``` is given, it will override ```MAKEFLAGS```, but still maintains a default of the number of cpus if flag is not set in ```MAKEFLAGS```.

Secondly, the parallelization of daal4py building was not possible with the previous code, and is not possible without significant rework. The custom ```build_ext``` setuptools Command object is replaced by the Cython version to extend capabilities in daal4py building.

Changes reduce building on github actions for DPC++ builds by 2-3 minutes, probably because the previous metric limited the number of processes.

No performance benchmarks necessary

---

**Testing**

**Performance**


scripts/build_backend.py
```diff
@@ -131,16 +131,34 @@ def custom_build_cmake_clib(
         cmake_args += ["-DADD_ONEDAL_RPATH=ON"]
 
     cpu_count = multiprocessing.cpu_count()
+    # convert to max supported pointer to a memory size in kB
```
david-cortes-intel, line 134: I'm not getting the idea here. `sys.maxsize` returns you the maximum possible addressable memory range by the operating system in bytes, which in a modern machine would be 2^63 (expressing it in KiB would require division by 2^10).

Dividing that by 128 would not have any effect, since the available memory might be in the range of 2^33 through 2^38 or so for a modern machine.

icfaust: Changed, was confused a little about Py_ssize_t definition and how it relates to size_t.  I have switched division operators to bitshifts for clarity and directness.

david-cortes-intel: Given the exact numbers that you are arriving at, is this trying to limit addressable space to accommodate a `float64` casted to `size_t`? What's the idea behind these?

icfaust: Yeah, I am assuming that the underlying processes have a memory bound (not just linux or windows). I wanted a general value I could easily get out of python to have a conservative upper bound of memory that can be used. I then use that as a very large initial value for splitting it in the multiprocessing. I am using Py_ssize_t like ```sizeof``` to yield the largest byte value possible for a platform. This at least makes all platforms to have some sort of bound without having to explicitly handle other platforms beyond windows/linux (e.g. mac).

david-cortes-intel: Am I understanding it correctly that the purpose of this code is to spawn multiple compilation processes with a memory limit for each such that the sum would not exceed system memory? Or does this code serve some other purpose?

icfaust: @Alexsandruss can probably better comment on this, but from the looks of #1196 that is the case. Maybe he can comment more.

Alexsandruss: The purpose is to limit number of processes so their projected memory usage doesn't exceed free memory.

scripts/build_backend.py
```diff
@@ -131,16 +131,32 @@ def custom_build_cmake_clib(
...
+                "Select",
+                "FreePhysicalMemory",
+            ],
+            stdout=subprocess.PIPE,
+            text=True,
+        )
+        memfree = int(txt.stdout.strip().split()[-1])  # total free physical memory in kB
+
+    mem_per_proc = 20  # 2**20 kB or 1GB
+    cpu_count = min(cpu_count, floor(max(1, memfree >> mem_per_proc)))
```
Alexsandruss, line 159: `psutil` will significantly simplify code but introduces new building dependency:
```suggestion
    mem_per_proc = 20  # 2**20 kB or 1GB
    cpu_count = min(psutil.cpu_count(), floor(max(1, psutil.virtual_memory().available >> mem_per_proc)))
```

scripts/build_backend.py
david-cortes-intel: Then I'm not sure the logic would work as intended.

It's not possible to predict beforehand how much memory a compilation process will use - the estimate here of 1GiB for example would not suffice for compiling something like DuckDB in parallel with flags like LTO (some parallel processes might take north of 4GiB with GCC). I haven't measured how much memory sklearnex takes though.

Additionally, if the processes run out of memory by a small margin, they might still succeed after switching to swap/page files.

scripts/build_backend.py
david-cortes-intel: Note that `psutil.cpu_count` can return `None`.

scripts/build_backend.py
david-cortes-intel: Maybe a reasonable compromise would be to just get total RAM in the system and divide by 4GB to get a target number of max parallel processes. I don't think the problem is approachable otherwise - new processes can be spawned and terminated by the user as the compilation is running either way, which also changes the calculation.

It's also quite an odd thing to have in a project's build script. I am not aware of any other ML library limiting the number of parallel jobs for the user based on available RAM - typically they just OOM on their own and the user then needs to adjust threads as needed.

Was there a CI failure on windows due to OOM?

Alexsandruss: 1GB constraint was chosen in #1196 based on CI VM parameters. Change to some other working weak constraint will be fine. Also, `MemFree` might be changed to `MemAvailable`.

david-cortes-intel: Yes, I think a user-configurable variable would be the best solution here.

scripts/build_backend.py
```diff
@@ -131,16 +131,8 @@ def custom_build_cmake_clib(
...
-    # TODO: add on all platforms
-    if IS_LIN:
-        with open("/proc/meminfo", "r") as meminfo_file_obj:
-            memfree = meminfo_file_obj.read().split("\n")[1].split(" ")
-            while "" in memfree:
-                memfree.remove("")
-            memfree = int(memfree[1])  # total memory in kB
-        cpu_count = min(cpu_count, floor(max(1, memfree / 2**20)))
-
+    max_jobs = os.getenv("MAX_JOBS")
```
david-cortes-intel, line 134: `getenv` supports a default value when the variable doesn't exist:
https://docs.python.org/3/library/os.html#os.getenv

setup.py
```diff
@@ -16,6 +16,29 @@
...
+#   NO_DPC (default: false)
+#      build oneDAL interfaces without SYCL support
+#
+#   NO_STREAM (default: false)
+#      disables building daal4py with streaming mode
+#
+#   OFF_ONEDAL_IFACE (default: false)
+#      do not build or use oneDAL interfaces (e.g. DAAL support only)
+#
+#   MAX_JOBS (default: number of processors)
```
david-cortes-intel, line 39: Please add it to the docs in the .md files too. Particularly this one:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/INSTALL.md

setup.py
```diff
@@ -16,6 +16,29 @@
 # limitations under the License.
 # ==============================================================================
 
+# Environment variables which impact build process:
+#
+#   DALROOT (required)
+#      folder location of oneDAL
```
david-cortes-intel, line 22: This is not quite clear. Should it point to the folder having the .so, or the `/lib`, or something else, etc.

david-cortes-intel (changes requested):

scripts/build_backend.py
icfaust: Improved logic, please check.

setup.py
icfaust: Clarified.

setup.py
icfaust: Done.

setup.py
```diff
@@ -16,6 +16,29 @@
 # limitations under the License.
 # ==============================================================================
 
+# Environment variables which impact the build process:
```
david-cortes-intel, line 19: This one is not quite clear on what exactly it impacts.

The build process involves building a cython module and some pybind11 modules, but this applies only to the pybind11 ones, and gets parallelized within each module sequentially but not across files of different modules.

icfaust: Thank you for asking this question. Turns out there is some dysfunction in the daal4py build_ext that we can clean up, and there is a hidden environment variable NTHREADS already existing. Will switch to using it for consistency, and will add it to the docs. There is also duplicated code from setuptools which has existed for 11 year in the setuptools codebase. Sigh.

setup.py
```diff
@@ -52,8 +75,7 @@
 IS_LIN = False
 
 dal_root = os.environ.get("DALROOT")
-n_threads = int(os.environ.get("NTHREADS", os.cpu_count() or 1))
-
+n_threads = min(os.cpu_count(), max(1, os.getenv("NTHREADS", os.cpu_count())))
```
david-cortes-intel, line 78: I get:
```
  File "/home/dcortes/repos/scikit-learn-intelex/setup.py", line 78, in <module>
    n_threads = min(os.cpu_count(), max(1, os.getenv("NTHREADS", os.cpu_count())))
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'str' and 'int'
```

setup.py
```diff
@@ -16,6 +16,29 @@
...
+#      build oneDAL interfaces without SYCL support
+#
+#   NO_STREAM (default: false)
+#      disables building daal4py with streaming mode
+#
+#   OFF_ONEDAL_IFACE (default: false)
+#      do not build or use oneDAL interfaces (e.g. DAAL support only)
+#
+#   NTHREADS (default: number of processors)
+#      number of processes used for compilation, can be overridden with -j
```
david-cortes-intel, line 40: This `-j` doesn't appear to have any effect on the cython part (the `daal4py` part), because it parallelizes across extension modules and there's only one module (`daal4py`).

setup.py
```diff
@@ -405,45 +427,40 @@ def get_onedal_py_libs():
...
-        with ThreadPoolExecutor(max_workers=n_threads) as executor:
-            result_list = [
-                executor.submit(self.build_extension, ext) for ext in self.extensions
-            ]
-        assert all(
-            f.exception() is None for f in result_list
-        ), "There were errors building the extensions"
+    def build_extension(self, ext):
+        # monkeypatch a mulitprocess pool to multithread daal4py compilation
+        try:
```
david-cortes-intel, line 442: This doesn't work either:
```
AttributeError: Can't pickle local object 'parallel_build_ext.build_extension.<locals>.parallel_compile.<locals>.<lambda>'
```
I don't think it's possible to parallelize a cython extension build.

icfaust: Yeah, certainly not worth the effort. Will probably delete all of the mentions of the parallel_build_ext and leave a note.

setup.py
```diff
@@ -577,7 +590,7 @@ def run(self):
     author_email="onedal.maintainers@intel.com",
     maintainer_email="onedal.maintainers@intel.com",
     project_urls=project_urls,
-    cmdclass={"develop": develop, "build": build, "build_ext": parallel_build_ext},
```
david-cortes-intel, line 580: This is very troublesome for local development, as it doesn't allow easily building `daal4py` with options like `--force` (cython is quite bad at determining when something needs recompile) .If you want a default builder, you should set `"build_ext": Cython.Distutils.build_ext`.

setup.py
```diff
@@ -469,13 +469,26 @@ def post_build(self):
 
 
 class develop(orig_develop.develop, custom_build):
+    def finalize_options(self):
```
david-cortes-intel, line 472: Why would these be needed now if the compilation was already parallelized before?

david-cortes-intel (changes requested):

david-cortes-intel:
@icfaust If the build is being handled by CMake, is it actually necessary to have a custom variable for settings threads? I think it should follow `MAKEFLAGS` by default.

setup.py
icfaust: Sounds like a good use case, and a small change. I will add it.

setup.py
```diff
@@ -16,6 +16,29 @@
 # limitations under the License.
 # ==============================================================================
 
+# Environment variables which impact the build process:
+#
+#   DALROOT (required)
```
david-cortes-intel, line 21: It also requires `${TBROOT}` now.

setup.py
```diff
@@ -16,6 +16,29 @@
...
+#      oneDAL root path (folder containing `bin`,`include`, and `lib` folders)
+#
+#   MPIROOT (required if NO_DIST not set)
```
david-cortes-intel, line 24: Important here to mention that the MPI library needs to be the same between sklearnex and mpi4py.

setup.py
```diff
@@ -52,8 +72,10 @@
 IS_LIN = False
 
 dal_root = os.environ.get("DALROOT")
-n_threads = int(os.environ.get("NTHREADS", os.cpu_count() or 1))
-
+makeflags = os.getenv("MAKEFLAGS", "")
```
david-cortes-intel, line 75: Thanks, but what I meant here is: if `${MAKEFLAGS}` is consumed by CMake, and it already contains a `-j` argument, is it actually necessary to manually create a variable `n_threads` here and pass it around?

Also, doesn't `setup.py build` also recognize a flag `-j` which it then passes to the `build` command?

setup.py
icfaust: removed.

setup.py
icfaust: removed.

setup.py
icfaust: made a function which respects the following hierarchy: command line flag, MAKEFLAGS, default.

david-cortes-intel: But is it actually necessary? What would happen if you don't set anything and instead let pybind11/cmake consume the user config for `setup.py` args and `MAKEFLAGS` directly?

icfaust: default values will then be single threaded, as the cmake is a custom solution, and MAKEFLAGS aren't set by setup.py (so much as I have researched). Without having to rewrite our CIs both public and private, this is has the least impact. If you have suggestions please let me know.

david-cortes-intel: Aren't environment variables inherited in subprocesses launched by `setup.py`?

david-cortes-intel: In case it helps, I just gave it a try removing the context function `set_nthreads` and its usages, and it did end up parallelizing the pybind11 part with the value from `MAKEFLAGS`.

icfaust: The goal is to introduce a default value of cpu_count to MAKEFLAGS if ```-j``` isn't in there. This is in line with how it is currently in main, so in a way were are talking around the same thing we both want to use ```MAKEFLAGS```.

icfaust:
> I get:
> 
> ```
>   File "/home/dcortes/repos/scikit-learn-intelex/setup.py", line 338, in getpyexts
>     exts.extend(cythonize(ext, nthreads=n_threads))
>                                         ^^^^^^^^^
> NameError: name 'n_threads' is not defined. Did you mean: 'set_nthreads'?
> ```

Fixed!

setup.py
icfaust: Mainly because of the impact this change would require throughout all of the related CI systems. If they don't currently set ```-j``` in their sklearnex build process now in ```MAKEFLAGS```, it will switch to single threaded overnight. We can introduce this, which makes everything respect ```MAKEFLAGS```, then we can introduce it in all of the CI systems one at a time/ in a smooth way. This change will allow for that.

david-cortes-intel: Ok, but then please add a note in `INSTALL.md` that the number of threads used for building the pybind11 module are controlled through either `MAKEFLAGS` or `-j`, and mentioning that it defaults to using all threads if not set otherwise.

setup.py
```diff
@@ -312,6 +312,41 @@ def get_build_options():
...
+def set_nthreads(n_threads):
+    """MAKEFLAGS is used by the onedal cmake and cythonize to control the number
+    of processes. If it is set via the setup.py commmand with the --parallel or
+    -j argument, it will supercede this value. When both are not set, it will
+    default to the number of cpus, n_threads should be a positive integer, None,
+    or True"""
+    makeflags = os.getenv("MAKEFLAGS", "")
+    # True is used by setuptools to indicate cpu_count for `parallel`
+    # None is default for setuptools for single threading
+    orig_n_threads = re.findall(r"(?<=(?<!-)-j)\d*|$", makeflags)[0]
```
david-cortes-intel, line 325: I think this regex is too eager to detect flags. A user could potential pass an include with a path having "-j" somewhere and it'd pick the wrong thing.

setup.py
```diff
@@ -312,6 +312,41 @@ def get_build_options():
...
+
+    if makeflags:
+        if orig_n_threads:
+            # sub the value out
+            os.environ["MAKEFLAGS"] = re.sub(
+                r"(?<=(?<!-)-j)\d*", str(n_threads), makeflags, 1
+            )
+        else:
+            # add the value to MAKEFLAGS since it is not set
+            os.environ["MAKEFLAGS"] += f" -j{n_threads}"
```
david-cortes-intel, line 339: Is this supported on windows?

setup.py
icfaust: changed to \S to make it more selective

setup.py
```diff
@@ -312,6 +312,41 @@ def get_build_options():
...
+def set_nthreads(n_threads):
+    """MAKEFLAGS is used by the onedal cmake and cythonize to control the number
+    of processes. If it is set via the setup.py commmand with the --parallel or
+    -j argument, it will supercede this value. When both are not set, it will
+    default to the number of cpus, n_threads should be a positive integer, None,
+    or True"""
+    makeflags = os.getenv("MAKEFLAGS", "")
+    # True is used by setuptools to indicate cpu_count for `parallel`
+    # None is default for setuptools for single threading
+    orig_n_threads = re.findall(r"(?<=(?<!\S)-j)\d*|$", makeflags)[0]
```
david-cortes-intel, line 325: * If a flag is specified multiple times, the last occurrence ends up being taken. This one would take the first occurrence. Is it perhaps possible to use some argparsing module or similar to handle this?
* Maybe it could define the regex in a separate variable as there's more than one occurrence. Someone might try to modify it in the future.

icfaust: Solution I came up with should only replace the last occurrence. I have set a separate variable and added an explanation.

setup.py
```diff
@@ -312,6 +312,45 @@ def get_build_options():
     return eca, ela, include_dir_plat
 
 
+@contextmanager
```
david-cortes-intel, line 315: Does it actually need to be a context manager? This is a `setup.py` file, whose execution will end after the call to `setup`. There should be no harm in setting these globally and leaving them like that.

setup.py
david-cortes-intel: Thanks for investigating. Then it's starting to sound like it might be easier to just set `MAKEFLAGS=${MAKEFLAGS} -j$(nproc)` on all the linux jobs and remove all subclassing and custom argument passing logic.

> python setup.py install doesn't have a flag for parallelization

Maybe it's time to update the instructions and CIs to discourage this method, since it's in any event deprecated by `setuptools`. The right way to install would be to create a wheel through `setup.py` and install the result with `pip`. But that's better left for a different PR.

icfaust: Very good points for some follow up work, it is due time we swap to a fully kitted pyproject.toml. 

We still have the parallelization in cythonize. Also there is the possibility of clashing with MAKEFLAGS in the case that a user specifies both a ```-j``` flag to a ```python setup.py build``` and a ```MAKEFLAGS``` value.  I think these can be deprecated/removed when we coordinate with internal teams on internal CI systems. I'd like to ship this close to this current form, and better plan out future development on this. Let me know what you think, or what concrete steps are necessary to complete the PR.

david-cortes-intel: > We still have the parallelization in cythonize.

No, that part is not parallelized, even if you pass it the right arguments, because it only parallelizes across extension modules to compile, of which we have only one (`daal4py`). You can check that it only compiles using a single thread with `MAKEFLAGS=-j4 python setup.py build_ext --inplace --force -j 4` for example.

> Also there is the possibility of clashing with MAKEFLAGS in the case that a user specifies both a -j flag to a python setup.py build and a MAKEFLAGS value.

Yes, but that's also the case when using the original classes from pybind11/cython/etc.

> what concrete steps are necessary to complete the PR.

I'd say documenting these behaviors and defaults in the `INSTALL.md` file is quite important.

setup.py
david-cortes-intel: But would something happen if they are not reverted? They wouldn't be used past the call to `setup()` if you define them globally _before_ the call to `setup()` instead of in the build commands.

conda-recipe/bld.bat
```diff
@@ -30,4 +30,5 @@ IF DEFINED DPCPPROOT (
     call "%DPCPPROOT%\env\vars.bat"
 )
 
+set MAKEFLAGS=-j1
```
david-cortes-intel, line 33: Shouldn't this one be removed? Or does it lead to crashes elsewhere?

icfaust: Good question, on this latest CI run, windows runs are now >10 minutes showing control in windows works via ```MAKEFLAGS```. I will remove it, and add the docs in the next commit.

setup.py
```diff
@@ -312,6 +311,28 @@ def get_build_options():
...
+    -j argument, it will supersede this value. When both are not set, it will
+    default to the number of cpus, n_threads should be a positive integer, None,
+    or True"""
+    makeflags = os.getenv("MAKEFLAGS", "")
+    # True is used by setuptools to indicate cpu_count for `parallel`
+    # None is default for setuptools for single threading
+    # take the last defined value in MAKEFLAGS, as it will be the one
+    # used by cmake/make
+    regex = r"(?<=(?<!\S)-j)\d*(?<!\S)|$"
+    orig_n_threads = re.findall(regex, makeflags)[-1]
```
david-cortes-intel, line 326: This doesn't work for me:
image

setup.py
```diff
@@ -405,51 +427,34 @@ def get_onedal_py_libs():
...
-            ]
-        assert all(
-            f.exception() is None for f in result_list
-        ), "There were errors building the extensions"
-
+class onedal_build:
 
-class custom_build:
     def run(self):
+        self.onedal_run()
```
david-cortes-intel, line 433: Throws:
```
TypeError: onedal_build.onedal_run() missing 1 required positional argument: 'n_threads'
```

setup.py
icfaust: Don't worry, I misunderstood the scope of python set environment variables.

setup.py
icfaust: fixed

david-cortes-intel:
@icfaust  I'm seeing one issue now which doesn't happen in the master branch: a call to `setup.py build` will re-compile the cython part (daal4py) even if it was already built before through `build_ext`, whereas before it would reuse the `build_ext` result. No idea what causes this behavior.

david-cortes-intel:
> @icfaust I'm seeing one issue now which doesn't happen in the master branch: a call to `setup.py build` will re-compile the cython part (daal4py) even if it was already built before through `build_ext`, whereas before it would reuse the `build_ext` result. No idea what causes this behavior.

On a second look: perhaps it could be because `Cython.Distutils.build_ext` involves both .pyx -> .cpp and .cpp -> .so in addition to a manual call to 'cythonize' elsewhere, whereas the `build_ext` command it had before was only .cpp -> .so. Maybe it could just pass `setuptools.command.build_ext` if it works.

setup.py
```diff
@@ -34,8 +34,8 @@
 import setuptools.command.build as orig_build
 import setuptools.command.develop as orig_develop
 from Cython.Build import cythonize
+from Cython.Distutils import build_ext
```
david-cortes-intel, line 37:
```suggestion
from setuptools.command import build_ext
```

Verified that this solves the issue with `build` not reusing `build_ext` results.

icfaust: Removed entirely so that it uses this as default (and verified that the rebuild issue is taken care of).

david-cortes-intel (approved):
LGTM, pending CI results.

PR #2242, opened 2025-01-06
fix: trigger doc build all condition

yuejiaointel:
## Description

in this PR I fixed several problems so now doc can build without autodoc warnings, also I added code so in future if autodoc gives a warning it will be marked as an error so pipeline will fail. It still has duplicate warnings because we are having multiple same classes def in daal4py.

_List associated issue number(s) if exist(s): #8786 

**Testing**

**Performance**


david-cortes-intel:
Would the problem that this PR addresses get fixed as part of this other PR?
https://github.com/uxlfoundation/scikit-learn-intelex/pull/2214

yuejiaointel:
> Would the problem that this PR addresses get fixed as part of this other PR? #2214

Hi David,
I think it is different, the ref pr seems about making sure the doc are generated with the right version, but this one is mainly about fix the warning about doc, making sure doc can be generated completely, and bring back doc in ci workflow, now it is skipped. 
Best,
Yue

.ci/pipeline/docs.yml
```diff
@@ -17,8 +17,7 @@
 trigger:
   branches:
     include:
-    - main
-    - rls/*
+    - "*"
```
david-cortes-intel, line 20: What would be the motivation for including other branches here?

.ci/pipeline/docs.yml
```diff
@@ -65,16 +64,37 @@ jobs:
...
       export PREFIX=$(dirname $(dirname $(which python)))
-      export DALROOT=$PREFIX
+      export e=$PREFIX
       ./conda-recipe/build.sh
+      echo "Python executable being used1: $(which python)"
     displayName: 'Build daal4py/sklearnex'
   - script: |
+      echo "Current directory3: $(pwd)"
+      ls -l daal4py/
+      ls -l build/
```
david-cortes-intel, line 74: This is a temporary folder which is created during package build. It should not import modules from it.  These get copied elsewhere where the importable package lives.

.ci/pipeline/docs.yml
```diff
@@ -65,16 +64,37 @@ jobs:
...
+      ls -l build/lib.linux-x86_64-cpython-311
+      echo "Checking for shared libraries in build directory:"
+      find build/ -name "*.so"
       export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
+      echo "LD_LIBRARY_PATH: $LD_LIBRARY_PATH"
```
david-cortes-intel, line 79: `LD_LIBRARY_PATH` should not be needed. If this is for the oneDAL C++ library, be aware that there's an option `--abs-rpath` in the `setup.py` file as described in the installation instructions:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/INSTALL.md#build-intelr-extension-for-scikit-learn

.ci/pipeline/docs.yml
```diff
@@ -65,16 +64,37 @@ jobs:
     displayName: 'Install requirements'
   - script: |
       export PREFIX=$(dirname $(dirname $(which python)))
-      export DALROOT=$PREFIX
+      export e=$PREFIX
```
david-cortes-intel, line 67: Is this an arbitrary name that you chose, or does some other step in the current code reference a variable 'e'?

david-cortes-intel (changes requested):

.ci/pipeline/docs.yml
yuejiaointel: Hi David,
I am testing pipeline behavior on my own branch, some of these code are temporary and will be removed. I am trying to see why make html works locally but gives warnings on pipeline. If you have any suggestions or have encountered similar behaviors, your insights would be greatly appreciated!
Best,
Yue

.ci/pipeline/docs.yml
yuejiaointel: I can see the daal4py build *.so files in build folder on pipeline, locally I see them in daal4py folder, I'll find out where they get copied to. I did not see them in the sys.path folders on pipeline
Best,
Yue

.ci/pipeline/docs.yml
yuejiaointel: Hi David,
thx for the suggestion! I am not really sure what does variable do yet. I am trying to find out why the pipeline are giving warnings about some modules not being found, and my suspect is we are not settings the ci correctly and the build files are not being referred correctly
Best,
Yue

.ci/pipeline/docs.yml
yuejiaointel: Hi David,
This is just temp code I am having to print values, will be removed
Best,
Yue

.ci/pipeline/docs.yml
```diff
@@ -48,13 +47,14 @@ jobs:
     vmImage: 'ubuntu-22.04'
   steps:
   - script: |
+      source /opt/intel/oneapi/compiler/latest/env/vars.sh
```
david-cortes-intel, line 50: Shouldn't this source the whole oneAPI suite rather than just the compiler? I think it also needs MPI, TBB, etc.

doc/Makefile
```diff
@@ -33,4 +33,5 @@ help:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
-	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
\ No newline at end of file
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O) 2>&1 | tee build.log
+	@grep -i "autodoc" build.log && { echo "Autodoc Warnings, build failed!"; cat build.log; exit 1; } || { cat build.log; }
```
david-cortes-intel, line 37: The Makefile is auto-generated by Sphinx as I understand it. This line could be in `build-doc.sh` (for sklearnex) or in the `.yaml` files (for daal4py) instead. The Makefile can also be used locally outside of the CI, in which case one might want to get its original behaviors, like reporting the warnings in red and to stderr.

Also, currently the docs for sklearnex should build without any warnings, unlike those of daal4py. It would be ideal to set _all_ warnings as errors for sklearnex's docs. It can also throw other important warnings from outside autodoc that we need to look at, like misrendered things, intersphinx links not working, etc.

david-cortes-intel (changes requested):

david-cortes-intel:
@yuejiaointel @icfaust Is there some way that the generated HTMLs could be linked as a clickable and rendable link from the "Docs" CI that we see here at the bottom of PRs without having to manually download them from AzureDevOps?

.ci/pipeline/docs.yml
```diff
@@ -66,16 +66,23 @@ jobs:
   - script: |
       export PREFIX=$(dirname $(dirname $(which python)))
       export DALROOT=$PREFIX
+      export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
       ./conda-recipe/build.sh
+      cp ${DALROOT}/lib/python3.11/site-packages/daal4py/_daal4py.cpython*.so daal4py/
```
icfaust, line 71:
```suggestion
      NO_DPC=1
      python setup.py build_ext --inplace
```
If I understand correctly, I think you want the _daal4py.so just so you can import the modules of the repo in place.  You can just build the extension, which will suffice. Do you also need to import ```onedal``` for the docs?

icfaust (commented):
Some initial thoughts, I've never looked at this pipeline before, and seeing a ```daal-devel``` install makes me worried how outdated these scripts are.

david-cortes-intel:
@yuejiaointel I'm seeing some missing things in the docs generated here. Particularly, autoclass is missing the constructors of `daal4py` algorithm classes. This is for example how Cholesky looks like in the current online docs:
image

And this is how it renders in the generated artifact:
image

I have the same issue when building the docs locally so not sure if it's related to this PR.

.ci/pipeline/docs.yml
david-cortes-intel: I think `NO_DPC` should not be necessary if the compiler doesn't support `NO_DPC`, but `NO_DIST=1` might be required.

.ci/pipeline/docs.yml
icfaust: I believe those are handled by pip install later on, I think the intel compiler doesn't need to be installed at all.

.ci/pipeline/docs.yml
david-cortes-intel: Note that `build_ext` will only build `daal4py`. For `sklearnex` you need `build`.

.ci/pipeline/docs.yml
```diff
@@ -66,16 +66,23 @@ jobs:
   - script: |
       export PREFIX=$(dirname $(dirname $(which python)))
       export DALROOT=$PREFIX
+      export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
```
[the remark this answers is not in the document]
yuejiaointel: Hi David,
I think by setting DALROOT, and since this pipeline is not running in conda, in build.sh will set DALRPATH automatically by entering the elif statement since "${DALROOT}" != "${CONDA_PREFIX}"
Best,
Yue

.ci/pipeline/docs.yml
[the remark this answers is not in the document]
yuejiaointel: Hi David,
If I don't have this line I would get errors like ModuleNotFoundError: No module named 'daal4py._daal4py'. when we try to import daal4py. For example python -c "import daal4py; print('daal4py imported successfully')".This would happen locally as well if you delete the *.so file in daal4py folder. 

I think this happens because in sys path first one is '' which means we are always prioritizing local dir. Unless we delete sklearn/daal4py or putting the the dir contains daal4py build *.so file in front of ''. But I think cp to daal4py has less impact.
Best,
Yue

.ci/pipeline/docs.yml
yuejiaointel: That's good catch. Yes I agree this one should be deleted, I was using it to get tbb.so file but the daal-devel already has it. Best, Yue

.ci/pipeline/docs.yml
```diff
@@ -48,13 +47,14 @@ jobs:
...
   - script: |
+      source /opt/intel/oneapi/compiler/latest/env/vars.sh
       bash .ci/scripts/describe_system.sh
     displayName: 'System info'
   - task: UsePythonVersion@0
     inputs:
       versionSpec: '3.11'
       addToPath: true
-  - script: sudo apt-get update && sudo apt-get install -y clang-format
+  - script: sudo apt-get update && sudo apt-get install -y clang-format && sudo apt-get install -y pandoc
```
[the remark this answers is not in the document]
yuejiaointel: Hi David,
Thx for reviewing this PR, merged the apt command
Best,
Yue

doc/Makefile
yuejiaointel: Hi David,
That is a good suggestion! I moved the pipeline checking from make file to build-doc.sh for sklearnex doc and docs.yml for daal4py doc. I also now mark any warning as errors in sklearnex doc building.
Best,
Yue

.ci/pipeline/docs.yml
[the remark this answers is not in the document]
yuejiaointel: Hi Ethan,
Good catch! I merged the apt get command
Best,
Yue

.ci/pipeline/docs.yml
yuejiaointel: Hi Ethan! Thx for this suggestion, this works and I can remove setting LD_LIBRARY_PATH and cp the build file back. I do need the NO_DIST=1 otherwise it will complain MPIROOT not being set. But I still cannot remove setting LD_LIBRARY_PATH in build doc stage it will gives the module not found warning if not set.
Best,
Yue

yuejiaointel:
> @yuejiaointel I'm seeing some missing things in the docs generated here. Particularly, autoclass is missing the constructors of `daal4py` algorithm classes. This is for example how Cholesky looks like in the current online docs: image
> 
> And this is how it renders in the generated artifact: image
> 
> I have the same issue when building the docs locally so not sure if it's related to this PR.

Hi David, I will investigate and open another PR for this problem, I saw there is a ticket related to this problem and I assigned myself to it https://jira.devtools.intel.com/browse/DAALL-8892
Best,Yue

.ci/pipeline/docs.yml
```diff
@@ -64,18 +63,26 @@ jobs:
...
+      export DALROOT=$(dirname $(dirname $(which python)))
+      export NO_DIST=1
+      python setup.py build_ext --inplace
       ./conda-recipe/build.sh
     displayName: 'Build daal4py/sklearnex'
   - script: |
       export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
       cd doc/daal4py
-      make html
+      make html 2>&1 | tee build.log
```
david-cortes-intel, line 74: Please add this file to gitignore.

doc/build-doc.sh
```diff
@@ -25,8 +25,16 @@ mkdir $SAMPLES_DIR
 
 # copy jupyter notebooks
 cd ..
-rsync -a --exclude='doc/$SAMPLES_DIR/daal4py_data_science.ipynb' examples/notebooks/*.ipynb doc/$SAMPLES_DIR
```
david-cortes-intel, line 28: Why would this change be needed?

doc/daal4py/Makefile
```diff
@@ -33,4 +33,4 @@ help:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
-	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
david-cortes-intel, line 36: Better to leave the empty line at the end here.

.ci/pipeline/docs.yml
```diff
@@ -64,18 +63,26 @@ jobs:
       pip list
     displayName: 'Install requirements'
   - script: |
-      export PREFIX=$(dirname $(dirname $(which python)))
-      export DALROOT=$PREFIX
+      export DALROOT=$(dirname $(dirname $(which python)))
+      export NO_DIST=1
+      python setup.py build_ext --inplace
       ./conda-recipe/build.sh
```
david-cortes-intel, line 69: There's a call to both `build_ext` and conda recipe's build. What's the idea for having both? That'd leave you with one importable module (`daal4py`) being in two places, with one of them having oneDAL in the rpath and the other without.

If you want to build both extensions in-place, you should replace the call to `build.sh` with this:
```
python setup.py build --abs-rpath
```

If that doesn't work, perhaps would require these two:
```
python setup.py build_ext --inplace --abs-rpath
python setup.py build --abs-rpath
```

doc/build-doc.sh
```diff
@@ -25,8 +25,16 @@ mkdir $SAMPLES_DIR
...
 
 # build the documentation
 cd doc
-make html
+SPHINXOPTS="-W" make html 2>&1 | tee build.log
+
+cat build.log
+
+# check for autodoc warnings
+if grep "WARNING" build.log; then
```
david-cortes-intel, line 37: On a second look, maybe it'd be easier to follow if having both of these checks (daal4py/sklearnex) in the .yaml files.

doc/build-doc.sh
```diff
@@ -25,8 +25,16 @@ mkdir $SAMPLES_DIR
...
+    echo "Warnings detected, build failed!"
+    exit 1
```
david-cortes-intel, line 39: This block should not be needed if you are passing option `-W` to sphinx. Also please leave the empty lines at the end of files.

.ci/pipeline/docs.yml
david-cortes-intel: @yuejiaointel A call to `python setup.py build` also builds the module in-place. One is for `daal4py` and the other for `onedal`+`sklearnex`. They are built using different frameworks which follow different setup commands.

.ci/pipeline/docs.yml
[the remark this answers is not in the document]
david-cortes-intel: It will set it in the call to `build.sh`, but not in the call to `setup.py build_ext`.

doc/build-doc.sh
yuejiaointel: Hi David,
The exclude command does not recognize a path it just read as a string literal, so I changed to just file name. If we don't exclude this file when cp from examples, sklearnex doc will give a warning that this file is not included in any of the rst files
Best,
Yue

david-cortes-intel: That's curious - it does work for me when I try it locally. It also appears to show up like that in stackexchange posts:
https://askubuntu.com/questions/1420321/how-to-exclude-the-multiple-directory-in-rsync

.. but it looks like just pattern matching should also work. Would it is still work if you prefix it with `.*`?

david-cortes-intel:
@yuejiaointel Looks like something went wrong, despite the CI status and the warning failure logic:
image

doc/build-doc.sh
```diff
@@ -25,8 +25,8 @@ mkdir $SAMPLES_DIR
 
 # copy jupyter notebooks
 cd ..
-rsync -a --exclude='doc/$SAMPLES_DIR/daal4py_data_science.ipynb' examples/notebooks/*.ipynb doc/$SAMPLES_DIR
+rsync -a --exclude='daal4py_data_science.ipynb' examples/notebooks/*.ipynb doc/$SAMPLES_DIR
 
 # build the documentation
 cd doc
-make html
+SPHINXOPTS="-W" make html
```
david-cortes-intel, line 32: Looks like this doesn't work, because it gets overridden by the makefile. You need to define this option "-W" in an environment variable "O" instead.

doc/daal4py/Makefile
david-cortes-intel: My bad, I now realized that the PR wasn't removing the empty line at the end - instead, github shows a blank line at the end as a red negative symbol regardless of what the PR is doing with it.

yuejiaointel: Hi David,
Got it, thx for the info! I removed the empty lines again so the make files do not show up in the file changed section
Best,
Yue

.ci/pipeline/docs.yml
david-cortes-intel: Amendment: looks like `LD_LIBRARY_PATH` is necessary for TBB regardless of `PYTHONPATH`. (experiments here: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2269)

yuejiaointel:
> @yuejiaointel Looks like something went wrong, despite the CI status and the warning failure logic: image

Hi David,
I did lots of experiments and I think using 

 python setup.py build_ext --inplace --abs-rpath
 python setup.py build --abs-rpath

does not work.  python setup.py build_ext --inplace --abs-rpath
 will still copy the _daal4py.cpython*.so back into ./daal4py, and then in later sklearnex documentation build it will interpret the daal4py folder in doc as the folder to import daal4py. As a result, the sklearnex doc is not build because it could not import the modules.

The reason is:
python -c "import daal4py; print('Using daal4py from:', daal4py.__file__)"
Before entering doc daal4py.__file__ is referring to daal4py in root
after entering doc daal4py.__file__ is None


But using:

./conda-recipe/build.sh
cp ${DALROOT}/lib/python3.11/site-packages/daal4py/_daal4py.cpython*.so daal4py/

daal4py.__file__ is referring site-packages daal4py which I think is more ideal
What I am still not understanding is why we need to cp the *.so file back to daal4py folder in root project.

doc/build-doc.sh
yuejiaointel: Hi David,
Thx for the investigation! Changed
Best,
Yue

david-cortes-intel:
> > @yuejiaointel Looks like something went wrong, despite the CI status and the warning failure logic: image
> 
> Hi David, I did some experiments on this problem and I think using
> 
> python setup.py build_ext --inplace --abs-rpath python setup.py build --abs-rpath
> 
> does not work. First of all, python setup.py build_ext --inplace --abs-rpath will still copy the _daal4py.cpython*.so back into ./daal4py, and then in later sklearnex documentation build it will interpret the daal4py folder in doc as the folder to import daal4py. As a result, the sklearnex doc is not build because it could not import the modules. I think the reason is using the --inplace pritorize local dir more than the site package and we have a daal4py in root project and another in doc and it is prioritizing local daal4py folder depending where we are that is why we see different behavior in different folders.
> 
> Experiment result I did: python -c "import daal4py; print('Using daal4py from:', daal4py.**file**)" Before entering doc daal4py.**file** is referring to daal4py in root after entering doc daal4py.**file** is None
> 
> But using:
> 
> ./conda-recipe/build.sh cp ${DALROOT}/lib/python3.11/site-packages/daal4py/_daal4py.cpython*.so daal4py/
> 
> daal4py.**file** is referring site-packages daal4py which I think is more ideal What I am still not understanding is why we need to cp the *.so file back to daal4py folder in root project even when we can referring daal4py in the site packages.

It is not required to manually copy the file.

If you use the build script, it will install the modules into `site-packages` (they are directly importable), whereas if you use `setup.py build`, it will build it in-place in the root folder, so attempting to import them directly requires either manually adding them to `${PYTHONPATH}` (or to `sys.path`), or launching the python interpreter from the folder where they are (the root of the repository, not `/doc`).

I guess the issue could be that the root of the docs folder might be in the python's path and thus `import daal4py` makes it try to import the module from the root folder instead of site packages. There's also these lines here which could be having an effect:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/324ab24bf9463f9327e79fcfd593b5a5511964b8/doc/daal4py/conf.py#L35
https://github.com/uxlfoundation/scikit-learn-intelex/blob/324ab24bf9463f9327e79fcfd593b5a5511964b8/doc/sources/conf.py#L35

See this example which is working with in-place builds:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/2269/files#diff-3f4a99ddd4f584a56e293b58a9a1a9c0d514c381b336d3f1ceaa7b2fcc55a902L78

yuejiaointel:
> > > @yuejiaointel Looks like something went wrong, despite the CI status and the warning failure logic: image
> > 
> > 
> > Hi David, I did some experiments on this problem and I think using
> > python setup.py build_ext --inplace --abs-rpath python setup.py build --abs-rpath
> > does not work. First of all, python setup.py build_ext --inplace --abs-rpath will still copy the _daal4py.cpython*.so back into ./daal4py, and then in later sklearnex documentation build it will interpret the daal4py folder in doc as the folder to import daal4py. As a result, the sklearnex doc is not build because it could not import the modules. I think the reason is using the --inplace pritorize local dir more than the site package and we have a daal4py in root project and another in doc and it is prioritizing local daal4py folder depending where we are that is why we see different behavior in different folders.
> > Experiment result I did: python -c "import daal4py; print('Using daal4py from:', daal4py.**file**)" Before entering doc daal4py.**file** is referring to daal4py in root after entering doc daal4py.**file** is None
> > But using:
> > ./conda-recipe/build.sh cp ${DALROOT}/lib/python3.11/site-packages/daal4py/_daal4py.cpython*.so daal4py/
> > daal4py.**file** is referring site-packages daal4py which I think is more ideal What I am still not understanding is why we need to cp the *.so file back to daal4py folder in root project even when we can referring daal4py in the site packages.
> 
> It is not required to manually copy the file.
> 
> If you use the build script, it will install the modules into `site-packages` (they are directly importable), whereas if you use `setup.py build`, it will build it in-place in the root folder, so attempting to import them directly requires either manually adding them to `${PYTHONPATH}` (or to `sys.path`), or launching the python interpreter from the folder where they are (the root of the repository, not `/doc`).
> 
> I guess the issue could be that the root of the docs folder might be in the python's path and thus `import daal4py` makes it try to import the module from the root folder instead of site packages. There's also these lines here which could be having an effect:
> 
> https://github.com/uxlfoundation/scikit-learn-intelex/blob/324ab24bf9463f9327e79fcfd593b5a5511964b8/doc/daal4py/conf.py#L35
> 
> 
> https://github.com/uxlfoundation/scikit-learn-intelex/blob/324ab24bf9463f9327e79fcfd593b5a5511964b8/doc/sources/conf.py#L35
> 
> See this example which is working with in-place builds: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2269/files#diff-3f4a99ddd4f584a56e293b58a9a1a9c0d514c381b336d3f1ceaa7b2fcc55a902L78

Hi David!
Thx! This make senses now. I commented out the lines in both conf and removed the cp command. Now I just need one line of running build.sh and it works
Best,
Yue

doc/daal4py/Makefile
david-cortes-intel, line 36: Looks like this file is not being modified. Please reset it to how it has before the PR so as not to mess the history.

.ci/pipeline/docs.yml
david-cortes-intel, line 20: Please remember to change this before merging.

david-cortes-intel (approved):
LGTM, some small comments.

david-cortes-intel:
@yuejiaointel Before merging, could you do an experiment here checking that the CI job would show in red if it fails to import daal4py?

yuejiaointel:
> @yuejiaointel Before merging, could you do an experiment here checking that the CI job would show in red if it fails to import daal4py?

Hi David,
I can confirm that if failed to import daal4py ci for doc will fail, can see it from commit 71ed216
Best,
Yue

doc/daal4py/Makefile
yuejiaointel: Fixed thx for the comment
Best,
Yue

.ci/pipeline/docs.yml
yuejiaointel: Hi David,
Reverted
Best,
Yue

PR #2249, opened 2025-01-09
[CI, enhancement] Add ability to build with ```gcov```,  adding C++ code coverage for the ```onedal``` folder in codecov

icfaust:
## Description

Coverage statistics have been recently added to sklearnex via codecov.  This was focused on the Python code, but C++ is the other important aspect for integrating oneDAL (representing a fifth of the code in the repo). This PR does the following:

1) Adds ability to build with gcov for GNU C++ and Intel DPC++ compilers in cmake set by the SKLEARNEX_GCOV environment variable
2) Integrates C++ code coverage statistics in the Windows and Linux DPC++ Github Actions (currently Python 3.9 and Python3.11)
3) Switches coverage.py data format from json format to lcov (with branching enabled) to match C++ addition

It does this by using the ```gcovr``` package, and by a bash script (```generate_coverage_files.sh```). Note, to use gcov, the build Numpy must be installed, otherwise it will fail (usually changed in the requirements-test.txt installation).  Therefore, the bash script uses a ```NUMPY_BUILD``` environment variable in order to store this information. This does not impact the daal4py build, as it is considered legacy code.

Consequences:
1) Adds C++ coverage metrics
2) Greatly reduces coverage percentages (by using branch rather than line coverage which is stricter)
3) Adds ~3 minute total via the longer builds and report generation, increasing the CI runtime by 10%
4) lcov files are now to be considered the standard code coverage format for sklearnex

Note: windows visual studio does not easily maintain a code coverage tool, and is therefore neglected.  The majority of code is operated with the DPC build, with only a small segment being missed.

---

**Testing**

**Performance**


scripts/CMakeLists.txt
```diff
@@ -122,6 +125,18 @@ if(IFACE STREQUAL "host")
         endif()
     endif()
 
+    if(SKLEARNEX_GCOV)
+        if(CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM" AND WIN32)
+            set(CMAKE_CXX_FLAGS "/clang:--coverage ${CMAKE_CXX_FLAGS}")
+            list(APPEND ONEDAL_LIBRARIES "clang_rt.profile-x86_64.lib")
+        elseif(CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
david-cortes-intel, line 132: * What about clang?
* What happens if you invoke the compiler through `ccache`?

david-cortes-intel: If we know that it will work out with clang, I think there's no harm in adding it here. Someone might try to generate these reports locally too.

icfaust: Done! Generalized to work with the various compile id's which are clang-based.

david-cortes-intel:
@icfaust What would be the right way to generate a report locally? I haven't been able to get an interactive html report from this PR.

icfaust:
> @icfaust What would be the right way to generate a report locally? I haven't been able to get an interactive html report from this PR.

You should be able to use ```generate_coverage_reports.sh``` or most of the commands in there. The big thing is the gcov is touchy with cmake, and the build directory is the root for gcovr.  If you are separating the build install and the test requirements installs, make sure to be careful with the numpy version.

david-cortes-intel:
> > @icfaust What would be the right way to generate a report locally? I haven't been able to get an interactive html report from this PR.
> 
> You should be able to use `generate_coverage_reports.sh` or most of the commands in there. The big thing is the gcov is touchy with cmake, and the build directory is the root for gcovr. If you are separating the build install and the test requirements installs, make sure to be careful with the numpy version. Are you using g++ or icpx?

Meaning: it's not generated as part of the `run.sh` files for coverage? If that's the case, would be ideal to start documenting these things in the .md files.

david-cortes-intel:
@icfaust Can you attach here an .html report of the C++ code coverage that would be generated?

icfaust:
> @icfaust Can you attach here an .html report of the C++ code coverage that would be generated?

I recommend looking at codecov: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2249#issuecomment-2579588945

.github/scripts/generate_coverage_reports.sh
```diff
@@ -0,0 +1,46 @@
...
+    FILTER=$(realpath ./onedal).*
+    echo $FILTER
+    
+    NUMPY_TEST=$(python -m pip freeze | grep numpy)
+    # install dependencies
+    # proper operation of gcov with sklearnex requires the header files from
+    # the build numpy, this must be previously set as NUMPY_BUILD
+    python -m pip install gcovr $NUMPY_BUILD
+    
+    gcovr --gcov-executable "${GCOV_EXE}" -r . -v --lcov --filter "${FITLER}" -o coverage_cpp_"${1}".info
```
david-cortes-intel, line 42:
```suggestion
    gcovr --gcov-executable "${GCOV_EXE}" -r . -v --lcov --filter "${FILTER}" -o coverage_cpp_"${1}".info
```

icfaust: Oh wow, not sure how I didn't pick that up.

david-cortes-intel (approved):

PR #2253, opened 2025-01-13
Add sklearnex.BasicStatistics API for CSR inputs on GPU and a test for it

Vika-F:
* sklearnex.BasicStatistics API was extended to accept CSR inputs on GPU
* The test for CSR inputs on GPU were added

---

**Testing**


sklearnex/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -173,6 +174,53 @@ def test_multiple_options_on_random_data(
     assert_allclose(gtr_sum, res_sum, atol=tol)
 
 
+# @pytest.mark.skipif(not hasattr(sp, "random_array"), reason="requires scipy>=1.12.0")
```
david-cortes-intel, line 177: For earlier versions, could use the earlier `random` which returns sparse matrices instead of sparse arrays:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.random.html

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -167,9 +172,30 @@ def __getattr__(self, attr):
...
 
     def _onedal_supported(self, method_name, *data):
+
         patching_status = PatchingConditionsChain(
             f"sklearnex.basic_statistics.{self.__class__.__name__}.{method_name}"
         )
+
+        X, sample_weight = data
+        is_data_supported = (
+            _is_csr(X) and daal_check_version((2025, "P", 200))
```
david-cortes-intel, line 182: Wouldn't this disable support with CSC on CPU?

Vika-F: We do not support CSC format on CPU.
If this is a typo and you meant CSR, then the users with the previous versions of sklearnex do not have this check in the code.

Vika-F: You're right, didn't understand that for the first time. I've reverted the changes for CPU.

onedal/basic_statistics/tests/utils.py
```diff
@@ -31,3 +31,21 @@
...
+    "variation": (lambda X: np.std(X, axis=0) / np.mean(X, axis=0), (0.1, 0.1)),
+    #    TODO: There is a bug in oneDAL's max computations on GPU
+    #    "sum_squares": (lambda X: np.sum(np.square(X), axis=0), (2e-4, 1e-7)),
+    "sum_squares_centered": (
+        lambda X: np.sum(np.square(X - np.mean(X, axis=0)), axis=0),
+        (0.1, 1e-6),
+    ),
+    "standard_deviation": (lambda X: np.std(X, axis=0), (1e-2, 2e-3)),
+    "second_order_raw_moment": (lambda X: np.mean(np.square(X), axis=0), (1e-6, 1e-7)),
+}
```
icfaust, line 51:
```suggestion
options_and_tests_sparse = options_and_tests.copy()
# oneDAL's `max` computatios on GPU are incorrect 
del options_and_tests_sparse["max"]
# oneDAL's `sum_squares` computation on GPU are incorrect 
del options_and_tests_sparse["sum_squares"]
options_and_tests_sparse["variation"] = (lambda X: np.std(X, axis=0) / np.mean(X, axis=0), (5e-2, 5e-2))
options_and_tests_sparse["sum_squares_centered"] = (
        lambda X: np.sum(np.square(X - np.mean(X, axis=0)), axis=0),
        (0.1, 1e-6)
    )
options_and_tests_sparse["standard_deviation"] = (lambda X: np.std(X, axis=0), (1e-2, 2e-3))

```

I think it would be more useful to see the difference from the standard algorithm parameters.

sklearnex/basic_statistics/basic_statistics.py
```diff
@@ -166,21 +168,50 @@ def __getattr__(self, attr):
...
+        patching_status = PatchingConditionsChain(
+            f"sklearnex.basic_statistics.{self.__class__.__name__}.{method_name}"
+        )
+        X, sample_weight = data
+
+        is_data_supported = (
+            _is_csr(X) and daal_check_version((2025, "P", 200))
+        ) or not issparse(X)
+
+        is_sample_weight_supported = sample_weight is None or not issparse(X)
```
icfaust, line 187:
```suggestion
        is_sample_weight_supported = not (sample_weight and issparse(X))
```

sklearnex/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -17,16 +17,29 @@
...
+)
 from onedal.tests.utils._dataframes_support import (
     _convert_to_dataframe,
     get_dataframes_and_queues,
+    get_queues,
 )
 from sklearnex.basic_statistics import BasicStatistics
 
 
+# Generate random sparse data using scipy.sparse.random or scipy.sparse.random_array
```
icfaust, line 35: Since this isn't a function which is specific to basic_statistics, I would recommend adding it to ```sklearnex/tests/utils/_base.py``` and importing it from there

icfaust (commented):
Generally looks good, just some small stylistic things.  It looks to function properly from my understanding.

sklearnex/basic_statistics/basic_statistics.py
Vika-F: Tried this suggestion, but it resulted in `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`.
To make it work it should look like  `is_sample_weight_supported = not (sample_weight is not None and issparse(X))`.
So, I left it as it was.

icfaust: Oh crap. Good call, forgot about iterables.

sklearnex/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -212,6 +309,43 @@ def test_all_option_on_random_data(
...
+    )
+    X_dense = X_sparse.toarray()
+
+    result = compute_sparse_result(X_sparse, "all", queue)
+
+    for result_option in options_and_tests:
+        if result_option in ["max", "sum_squares"] and (
+            queue is None or queue.sycl_device.is_cpu
+        ):
+            # TODO: here is a bug in oneDAL's max and sum_squares computations on CPU
```
icfaust, line 337: Please share a ticket. Please also add a more descriptive string to help guide where to look rather than labeling it as a bug alone.

sklearnex/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -128,6 +144,45 @@ def test_single_option_on_random_data(
...
+@pytest.mark.parametrize("row_count", [1000, 10000])
+@pytest.mark.parametrize("column_count", [10, 100])
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_single_option_on_random_sparse_data(
+    queue, result_option, row_count, column_count, dtype
+):
+    if result_option in ["max", "sum_squares"] and (
+        queue is None or queue.sycl_device.is_cpu
+    ):
+        pytest.skip("max and sum_squares computations are buggy on CPU")
```
icfaust, line 158: The english in the skip is too informal and not informative enough.

sklearnex/basic_statistics/basic_statistics.py
icfaust: This is a nitpick, but ```not issparse(X)``` is much more likely to be True than ```_is_csr(X)``` by normal usage of the estimator. Swapping the statements will be probably on average faster by the nature of ```or```.

onedal/basic_statistics/tests/utils.py
```diff
@@ -22,11 +22,23 @@
     "max": (lambda X: np.max(X, axis=0), (1e-7, 1e-7)),
     "mean": (lambda X: np.mean(X, axis=0), (5e-7, 1e-7)),
     "variance": (lambda X: np.var(X, axis=0), (2e-3, 2e-3)),
-    "variation": (lambda X: np.std(X, axis=0) / np.mean(X, axis=0), (5e-2, 5e-2)),
+    "variation": (
+        lambda X: (
```
icfaust, line 26: Rather than have this in a lambda function, this has enough logic in it to warrant a separate function and some commentary as to it use. I assume there is a case where we have an explicit zero mean, or was this just for safety reasons?

onedal/basic_statistics/tests/utils.py
```diff
@@ -22,11 +22,23 @@
...
+                ]
+            )
+        ),
+        (0.1, 0.1),
+    ),
     "sum_squares": (lambda X: np.sum(np.square(X), axis=0), (2e-4, 1e-7)),
     "sum_squares_centered": (
         lambda X: np.sum(np.square(X - np.mean(X, axis=0)), axis=0),
-        (2e-4, 1e-7),
+        (1e-2, 1e-7),
```
icfaust, line 41: This impacts the bound range on all of the tests, when its just for GPU sparse anaylsis (changing it by 2 orders of magnitude). Are we sure this a good idea?

At that point doing it in the specific GPU tests may improve locality and explainability for those who end up looking at this in the future (and since the number of new 3 test function definitions). We are already doing something similar with the pytest.skips in those tests.

icfaust (commented):
The nitpick doesn't need to be done, but the change in tolerances in general should be changed.

onedal/basic_statistics/tests/utils.py
Vika-F: Yes, there are the cases of CSR data with the columns that contain only zeros. That's why the check is added.
Will move to a separate function.

onedal/basic_statistics/tests/utils.py
Vika-F: I am not sure that this is due to GPU and not due to the bug in computations.
In this PR I added new tests for both CPU and GPU.
Will try to merge this fix first https://github.com/uxlfoundation/oneDAL/pull/3047 and re-evaluate the thresholds.

onedal/basic_statistics/basic_statistics.py
```diff
@@ -82,7 +82,13 @@ def fit(self, data, sample_weight=None, queue=None):
             sample_weight = _check_array(sample_weight, ensure_2d=False)
 
         is_single_dim = data.ndim == 1
-        data_table, weights_table = to_table(data, sample_weight, queue=queue)
+
+        data_table = to_table(data, queue=queue)
+        weights_table = (
```
icfaust, line 87: Not sure I understand why this change is necessary? Does to_table fail if sample_weight is None and queue is not None?

sklearnex/basic_statistics/basic_statistics.py
icfaust: @Vika-F let me know what you think

onedal/basic_statistics/tests/utils.py
```diff
@@ -16,18 +16,35 @@
...
+    "variance": (lambda X: np.var(X, axis=0, ddof=1), (2e-4, 1e-7)),
+    "variation": (lambda X: variation(X), (1e-3, 1e-6)),
     "sum_squares": (lambda X: np.sum(np.square(X), axis=0), (2e-4, 1e-7)),
     "sum_squares_centered": (
         lambda X: np.sum(np.square(X - np.mean(X, axis=0)), axis=0),
-        (2e-4, 1e-7),
+        (1e-3, 1e-7),
     ),
-    "standard_deviation": (lambda X: np.std(X, axis=0), (2e-3, 2e-3)),
+    "standard_deviation": (lambda X: np.std(X, axis=0, ddof=1), (2e-3, 1e-7)),
```
icfaust, line 48: I assume that the oneDAL documentation for low_order_moments has a description that it is using Bessel's correction?  The documentation for sklearnex doesn't exist for this difference, which means we should make a tiny ticket to make sure someone goes through and verifies the docs for this here: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/basic_statistics/basic_statistics.py#L54 since variance and standard_devation should have a note for this

sklearnex/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -128,6 +144,49 @@ def test_single_option_on_random_data(
...
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_single_option_on_random_sparse_data(
+    queue, result_option, row_count, column_count, dtype
+):
+    if not daal_check_version((2025, "P", 200)) and result_option in [
+        "max",
+        "sum_squares",
+    ]:
+        pytest.skip(
+            "There is a bug in 'max' and 'sum_squares' computations in oneDAL version < 2025.2"
```
icfaust, line 160:
```suggestion
            "'max' and 'sum_squares' calculate using a subset of the data in oneDAL version < 2025.2"
```

sklearnex/basic_statistics/tests/test_basic_statistics.py
```diff
@@ -212,6 +313,44 @@ def test_all_option_on_random_data(
...
+    X_dense = X_sparse.toarray()
+
+    result = compute_sparse_result(X_sparse, "all", queue)
+
+    for result_option in options_and_tests:
+        if not daal_check_version((2025, "P", 200)) and result_option in [
+            "max",
+            "sum_squares",
+        ]:
+            # TODO: There is a bug in 'max' and 'sum_squares' computations in oneDAL version < 2025.2
```
icfaust, line 342:
```suggestion
```
Completed with uxlfoundation/oneDAL#3047 right?

icfaust (commented):
Looks pretty close to me, just some subtle stuff and some things we should make tickets for (as future work).

onedal/basic_statistics/tests/utils.py
Vika-F: Yes, oneDAL uses Bessel's correction. See 'variance' here: https://uxlfoundation.github.io/oneDAL/daal/algorithms/moments/moments-of-low-order.html
I'll add the note.

onedal/basic_statistics/basic_statistics.py
Vika-F: Yes, it fails without the change for some reason with CSR data. I provide an exact error.

onedal/basic_statistics/tests/utils.py
icfaust: Btw, good catch on this.

onedal/basic_statistics/basic_statistics.py
icfaust: Yeah! It means there is probably a bug in to_table that needs to be addressed.

sklearnex/basic_statistics/tests/test_basic_statistics.py
Vika-F: Yes, the fix was merged: https://github.com/uxlfoundation/oneDAL/pull/3047.
This check is only needed to make public CI green.

sklearnex/basic_statistics/basic_statistics.py
Vika-F: Ah, just missed this. I'll place `issparse(X)` first.

sklearnex/basic_statistics/tests/test_basic_statistics.py
icfaust: Great! So then I would remove the todo, since you've got it under control.

icfaust (approved):
Merge when CI's show green 👍

PR #2195, opened 2024-11-28
[enhancement] Refactor ```onedal/datatypes``` in preparation for dlpack support

icfaust:
## Description

PR #1568 is too large and too out of date to use for refactoring the onedal ```table``` object. Subsequent PRs which introduced sycl_usm_array support mixed in additional C++ code in an unsustainable/ poorly extensible manner.  This PR tries to find an intermediate solution which does not rewrite ```onedal/datatypes``` but only shifts the code in a way which is clearer and better separated following #1568 in spirit.  This way adding dlpack support to ```table``` will be separated cleanly from ```numpy``` and ```sycl_usm_array```.  It defines additional namespaces so that it is clear in the code which conversion mechanism is being used where.  It begins standardization of the naming schemes of functions and files for simple extensibility in the future (say supporting other array or possibly dataframe types).

Therefore, because it is only namespace definitions and file movements, performance benchmarks aren't necessary.

NOTE: The ```SET_CTYPE_FROM_DAL_TYPE``` macro is moved out of the sycl_usm folder because of its generality and re-use within dlpack.

NOTE: A function ```convert_from_sua_iface``` is renamed ```convert_to_table``` in the ```sycl_usm``` namespace.

NOTE: changes to pickling for incremental algorithms must also be done. It is clear with the changes what types of inputs/outputs are expected for ```__getstate__``` and ```__setstate__```, which will default to numpy (not explicitly discussed in those PRs, but may be important in future array_api discussions).

NOTE: there are definitely inconsistencies between the sycl_usm and numpy implementations, and changes to those implementations in any functional manner should be done in a follow-up PR/ticket.  They are likely to be time-intensive and require benchmarking.

---

PR should start as a draft, then move to ready for review state after CI is passed and all applicable checkboxes are closed.
This approach ensures that reviewers don't spend extra time asking for regular requirements.

You can remove a checkbox as not applicable only if it doesn't relate to this PR in any way.
For example, PR with docs update doesn't require checkboxes for performance while PR with any change in actual code should have checkboxes and justify how this code change is expected to affect performance (or justification should be self-evident).

Checklist to comply with **before moving PR from draft**:

**PR completeness and readability**

- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes or created a separate PR with update and provided its number in the description, if necessary.
[...]

PR #2275, opened 2025-01-27
[enhancement] add dlpack support to ```to_table```

icfaust:
## Description

This PR introduces ```__dlpack__``` tensor (https://github.com/dmlc/dlpack) consumption by ```to_table``` allowing for zero-copy use of data in oneDAL.  This is important for enabling array_api support and is a pre-requisite for #2096 (array api dispatching).  That PR is then a pre-requisite for #2100 #2106 #2189 #2206 #2207 and #2209.  Sklearn provides array_api support for some algorithms. If we wish to fully support zero copy of sycl_usm inputs, we need to be able to consume array_api inputs due to underlying sklearn dependencies (```validate_data```, ```check_array```, etc.). While we support Sycl usm ndarrays (```dpctl```, ```dpnp```) via the ```__sycl_usm_array_interface__``` method in the onedal folder estimators, to properly interface estimators in the sklearnex folder, we need to support the ```__dlpack__``` method of arrays/tensors.  This PR does that and greatly simplifies the necessary logic in #2096 and the follow-up PRs. This PR also provides the added benefit of working with other frameworks which support SYCL gpu data which have ```__dlpack__``` interfaces (i.e. PyTorch).

NOTES:

- This PR continues from #2195 to integrate dlpack support, and it takes code from #1568.  Please use #1568 as reference, though it contained some mistakes.

- This new functionality is not yet exposed publicly and therefore must be added to the documentation with #2096.

- This aspect is not yet benchmarkable/ has nothing to benchmark against.

- Memory leak checking using the infrastructure from #2045 (only CPU at the moment, will be modified if/when PyTorch support is brought online)

- It does numeric testing via onedal's ```assert_all_finite``` testing using ```array_api_strict``` arrays, as that has a simple interface to the backend with no checking of array aspects on the python-side before ```to_table```.

- A special testing class is created to verify for SYCL device support and is used in ```test_data.py```.

- This does not have any work with returning ```__dlpack__``` supported arrays as results. And must be done as a follow-up PR (probably also a good idea in order to ease reviewing). Therefore, this is only about array/tensor consumption.
 
TODO: add a onedal function which checks a dlpack tensor for C-contiguity or F-contiguity similar to the ```flags``` attribute of numpy/dpctl/dpnp. This is out of the scope of this PR, but is necessary for assert_all_finite support for the next step in array_api work.

---

**Testing**

**Performance**


icfaust:
Segfault with fortran orders, need to investigate.

icfaust:
Something weird is happening to codecov with respect to seeing the new files. Will need to be investigated at some point

onedal/datatypes/dlpack/dlpack.h
```diff
@@ -0,0 +1,332 @@
+/*!
```
david-cortes-intel, line 1: Could this file be added through a build dependency instead of being committed into the repository?

Otherwise, could it be linked from a git module of the repository where it came from?

icfaust: That would be problematic if aspects change in the API which aren't reflected in the code which adapts to it, specifically DLManagedTensor (if removed). Other frameworks include copies: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/dlpack.h
https://github.com/IntelPython/dpctl/blob/master/dpctl/tensor/include/dlpack/dlpack.h
https://github.com/numpy/numpy/blob/main/numpy/_core/src/common/dlpack/dlpack.h

icfaust:
Failures in Intel CI unrelated to the PR

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    // Get pointer to the data following dlpack.h conventions.
+    // use static cast because data is known to be void*, and reintrepret_cast
+    // since we know that T* is of a set specific types that we will guarantee
+    // for the compiler.
+    const auto* const ptr =
+        reinterpret_cast<const T*>(static_cast<char*>(tensor.data) + tensor.byte_offset);
+
+    // get shape, if 1 dimensional, force col count to 1
+    std::int64_t row_count, col_count;
```
ahuber21, line 49: Why this C-style separation of declaration and definition?

onedal/datatypes/dlpack/dlpack.h
david-cortes-intel: > That would be problematic if aspects change in the API which aren't reflected in the code which adapts to it. Other frameworks include copies: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/dlpack.h https://github.com/IntelPython/dpctl/blob/master/dpctl/tensor/include/dlpack/dlpack.h https://github.com/numpy/numpy/blob/main/numpy/_core/src/common/dlpack/dlpack.h

The header file should not change automatically if you add it as a git submodule. It only gets updated if you manually update submodules.

But if let's say the source file changes, and there are no adaptations in sklearnex, wouldn't that potentially lead to segfaults, memory corruptions, and similar when the code compiled from here receives an object from a library compiled with a different header?

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -0,0 +1,183 @@
...
+    if (!tensor.data) {
+        return res;
+    }
+
+    // Get pointer to the data following dlpack.h conventions.
+    // use static cast because data is known to be void*, and reintrepret_cast
+    // since we know that T* is of a set specific types that we will guarantee
+    // for the compiler.
+    const auto* const ptr =
+        reinterpret_cast<const T*>(static_cast<char*>(tensor.data) + tensor.byte_offset);
```
ahuber21, line 46:
```suggestion
    auto raw_data = static_cast<char*>(tensor.data);
    const auto ptr = reinterpret_cast<const T*>(raw_data + tensor.byte_offset);
```

Not sure this is the best way, but I would consider a few things.

- Don't like the double `static_cast` in a single line
- Do we even need to convert to `char*` when all we do is adding an offset and casting again? The `operator+()` would produce the same target address, regardless of the base being `char*` or `void*`, wouldn't it?
- Is `const auto* const` really what you want? We don't often declare pointers themselves `const`, so I'm just wondering.

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -0,0 +1,183 @@
...
+        // from get_queue_by_device_id. Note, this behavior only occurs if on a oneAPI device.
+        sycl::queue queue;
+        queue = !q_obj.is(py::none()) ? get_queue_from_python(q_obj)
+                                      : get_queue_by_device_id(tensor.device.device_id);
+
+        if (readonly) {
+            MAKE_QUEUED_HOMOGEN(ptr);
+        }
+        else {
+            auto* const mut_ptr = const_cast<T*>(ptr);
```
ahuber21, line 89:
```suggestion
            auto* mut_ptr = const_cast<T*>(ptr);
```

Same thought about making the pointer `const` as above.
By the way, the data is provided non-`const`, you are making it `const` by choice above and `const`-casting it away again here.
At first glance, the non-`const` usage is dominant and I don't see a point in ever making the data `const`. Did I miss something?

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -0,0 +1,183 @@
...
+    // Extract and convert a DLpack data type into a oneDAL dtype.
+    dtype = convert_dlpack_to_dal_type(tensor.dtype);
+
+    // if there is a queue, check that the data matches the necessary precision.
+#ifdef ONEDAL_DATA_PARALLEL
+    if (!q_obj.is(py::none()) && !q_obj.attr("sycl_device").attr("has_aspect_fp64").cast<bool>() &&
+        dtype == dal::data_type::float64) {
+        // If the queue exists, doesn't have the fp64 aspect, and the data is float64
+        // then cast it to float32 (using reduce_precision)
+        py::object copy = reduce_precision(obj);
```
ahuber21, line 154: Do we need to issue a warning, similar to the python version. Is silently downcasting a problem? Or is there a check elsewhere that I didn't see yet?

ahuber21 (commented):
Great work, just a few questions from my side. Ping me for my approval once addressed.

onedal/datatypes/dlpack/dtype_conversion.hpp
```diff
@@ -0,0 +1,42 @@
...
+namespace py = pybind11;
+
+namespace oneapi::dal::python::dlpack {
+
+// DLDataType is only 64 bits in size. Not expensive
+dal::data_type convert_dlpack_to_dal_type(DLDataType dt);
+DLDataType convert_dal_to_dlpack_type(dal::data_type dt);
+
+template <typename Type>
+inline DLDataType make_data_type(Type tag = {}) {
```
Vika-F, line 37: It doesn't look like this function is used somewhere. All the occurrences have `detail::make_data_type` which refers to oneDAL's function with the same name.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: For point 2) https://stackoverflow.com/a/3524270
Otherwise good call, will separate the casts, and look at the constness more closely

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -0,0 +1,183 @@
+/*******************************************************************************
+* Copyright Contributors to the oneDAL project
```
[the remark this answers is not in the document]
icfaust: Request from above to do it this way when we switched to the uxlfoundation.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: Ahh good catch, we do emit a warning with numpy inputs for this issue, I will modify it to do the same.

onedal/datatypes/dlpack/dtype_conversion.hpp
icfaust: Oh yeah, that came from @KulikovNikita 's PR and I should have gone through it more closely than I did.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: Bad habits. Will change

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+#ifdef ONEDAL_DATA_PARALLEL
+#include "onedal/common/sycl_interfaces.hpp"
+#endif // ONEDAL_DATA_PARALLEL
+
+namespace oneapi::dal::python::dlpack {
+
+template <typename T, typename managed_t>
+inline dal::homogen_table convert_to_homogen_impl(managed_t* dlm_tensor,
+                                                  bool readonly,
```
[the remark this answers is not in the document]
icfaust: Ahh, so with the sycl interface they do something similar: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/datatypes/sycl_usm/data_conversion.cpp#L101 and in linear regression I guess we need to be very careful how we've defined the copy parameter. @david-cortes-intel @Vika-F can you verify that the linear regression modifies data in place for return? If not we may need to say what we do with ```copy_X```. I can remove it no problem, in fact that would simplify things. Let me know.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: ahh, I lifted this part from here, maybe we need to modify it there too? https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/datatypes/sycl_usm/data_conversion.cpp#L113

onedal/datatypes/dlpack/data_conversion.cpp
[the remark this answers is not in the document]
david-cortes-intel: @icfaust Linear regression does not ever modify data, despite the `copy_X` parameter. That parameter comes from scikit-learn.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: I double checked the status of warnings in pybind11, looks like it will show up in the next release pybind/pybind11#5291. I will just implement it like I did for numpy.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: Just as an update: with the discussion of the readonly, I removed the aspect entirely.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: For ```const auto* const```, dlpack data is immutable, and we don't want to move its location, so it a const pointer to const data. Luckily dal tables are okay with that out of the box.

onedal/datatypes/dlpack/dtype_conversion.hpp
icfaust: Ahhh, so I will have to re-add that later for dlpack support for ```from_table```, but that is a good thing.  I will remove it from this PR, but we should not close his PR (he save so much work for me btw), and I will make a follow-up ticket for it.

onedal/datatypes/dlpack/data_conversion.cpp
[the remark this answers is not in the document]
icfaust: I have stripped out mutable tensors entirely from the code, which actually is a really good thing (and thank you for discussions @Vika-F @david-cortes-intel) as we are now very dlpack tensor compliant (input data is expected to be immutable, but not enforced).

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: removed

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+#endif // ONEDAL_DATA_PARALLEL
+
+    // unusual data format found, try to make contiguous, otherwise throw error
+    // error throw located in regenerate_layout
+    if (get_dlpack_layout(tensor) == dal::data_layout::unknown) {
+        // NOTE: this attempts to make a contiguous deep copy of the data
+        // if possible, this is expected to be a special case
+        py::object copy = regenerate_layout(obj);
+        res = convert_to_table(copy, q_obj);
```
KulikovNikita, line 147: So this is some kind of a recursive call. Isn't it? Are there any limitations on the depth of recursion?

KulikovNikita: As a quick fix I would recommend to add an assertion on object being convertible right copying it.

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -0,0 +1,163 @@
...
+        return res;
+    }
+
+#define MAKE_HOMOGEN_TABLE(CType)                                                           \
+    res = versioned ? convert_to_homogen_impl<CType, DLManagedTensorVersioned>(dlmv, q_obj) \
```
KulikovNikita, line 152: Since you are assigning value to the already defined object I would suggest to use if statement instead of ternary

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -0,0 +1,163 @@
...
+                    : convert_to_homogen_impl<CType, DLManagedTensor>(dlm, q_obj);
+    SET_CTYPE_FROM_DAL_TYPE(dtype,
+                            MAKE_HOMOGEN_TABLE,
+                            throw std::invalid_argument("Found unsupported array type"));
```
KulikovNikita, line 156: Shouldn't an error message be declared in the separate file alongside other error messages?

onedal/datatypes/dlpack/dlpack_utils.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+    else {
+        throw std::runtime_error("unable to extract dltensor");
+    }
+}
+
+std::int32_t get_ndim(const DLTensor& tensor) {
+    // check if 1 or 2 dimensional, and return the number of dimensions
+    const std::int32_t ndim = tensor.ndim;
+    if (ndim != 2 && ndim != 1) {
```
KulikovNikita, line 51: What about the zero dimensional tensors (scalars)?

onedal/datatypes/dlpack/dlpack_utils.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+    else if (strides[0] == 1 && strides[1] == r_count) {
+        return dal::data_layout::column_major;
+    }
+    else {
+        return dal::data_layout::unknown;
+    }
+}
+
+bool check_dlpack_oneAPI_device(const DLDeviceType& device) {
```
KulikovNikita, line 78: Is this naming in line with the snake case?

sklearnex/tests/test_memory_usage.py
```diff
@@ -35,6 +35,7 @@
     get_dataframes_and_queues,
 )
 from onedal.tests.utils._device_selection import get_queues, is_dpctl_device_available
+from onedal.utils._array_api import _get_sycl_namespace
```
KulikovNikita, line 38: FYI, it's very uncommon in Python (and ML) world to name importance entities starting with underscore. Some Python interpreters and widely used libraries do not support this naming. Famously you can not use any _* entities in torch.jit context due to the inability to inspect these functions

icfaust: Yeah, I totally agree. Private functions should be private. Will follow up with changes to make this (wasn't my doing, but I guess I did allow it to happen).

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: Follows convention set out in the original numpy interfaces (https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/datatypes/numpy/data_conversion.cpp#L205). I'm not saying its right, but its consistent.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: Yeah, I was trying to keep it small/ for readability, I know it may cause a slight difference in the output assembly.

doc/third-party-programs-sklearnex.txt
```diff
@@ -421,4 +421,209 @@ Apache License
 
 -------------------------------------------------------------
 
+3. DLPack
```
Alexsandruss, line 424:
```suggestion
4. DLPack
```

onedal/datatypes/dlpack/dlpack_utils.cpp
icfaust: Will make a test explicitly for testing this scenario and handling it gracefully

sklearnex/tests/test_memory_usage.py
ahuber21: +1, but I do want to add that the rules are more relaxed in testing context.

onedal/datatypes/dlpack/dlpack_utils.cpp
```diff
@@ -0,0 +1,127 @@
...
+            "Data will be converted into float32 from float64 because device does not support it",
+            1);
+        const auto space = obj.attr("__array_namespace__")();
+        copy = space.attr("astype")(obj, space.attr("float32"));
+    }
+    else {
+        throw std::invalid_argument("Data has higher precision than the supported device");
+    }
+    return copy;
+}
```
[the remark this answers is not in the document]
icfaust: This was how it was done previously, but an explanation on why this is important can be found here in the PR description: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2217  We would have to replicate a bunch of checks we already do in C++ side, and the conversions are different calls each for numpy, sycl_usm, sparse and dlpack/array_api. Good question.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -152,7 +153,7 @@ inline csr_table_t convert_to_csr_impl(PyObject *py_data,
     return res_table;
 }
 
-dal::table convert_to_table(py::object inp_obj, py::object queue) {
+dal::table convert_to_table(py::object inp_obj, py::object queue, bool recursed) {
```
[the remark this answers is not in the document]
Vika-F: Sure, let's leave it for further imrpovements.

onedal/datatypes/dlpack/dlpack_utils.cpp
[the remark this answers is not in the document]
Vika-F: Oh, my memory is short. Thanks for the explanation.

onedal/datatypes/tests/test_data.py
```diff
@@ -473,3 +514,51 @@ class DummySyclDevice:
...
+
+    X_tens = _OnlyDLTensor(X[:, :3])
+
+    # give the _OnlyDLTensor the ability to copy
+    if can_copy:
+        X_tens.copy = lambda: _OnlyDLTensor(X.copy())
+        to_table(X_tens)
+    else:
+        with pytest.raises(RuntimeError, match="Wrong strides"):
+            to_table(X_tens)
```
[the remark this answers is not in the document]
icfaust: Good question, turns out there was a bug in the numpy ```from_table``` command that I've included in the PR, and the verification of the returned values is now complete.

onedal/datatypes/tests/test_data.py
```diff
@@ -91,6 +92,29 @@ class DummyEstimatorWithTableConversions:
...
+    This takes inspiration from sklearn's `_NotAnArray`."""
+
+    def __init__(self, data):
+        self.data = data
+
+    def __dlpack__(self):
+        return self.data.__dlpack__()
+
+
+def _to_table_supported(array):
```
ahuber21, line 109: Brief explanation here would be helpful.

onedal/datatypes/tests/test_data.py
```diff
@@ -434,21 +445,51 @@ class DummySyclDevice:
...
+
 @pytest.mark.parametrize("X", [None, 5, "test", True, [], np.pi, lambda: None])
 @pytest.mark.parametrize("queue", get_queues())
 def test_non_array(X, queue):
     # Verify that to and from table doesn't raise errors
     # no guarantee is made about type or content
+    error = ValueError
     err_str = ""
 
+    xp = X.__array_namespace__() if hasattr(X, "__array_namespace__") else np
```
ahuber21, line 474: Would it be helpful to write a generic helper that returns the namespace and use that everywhere? I think we're extracting it quite often and not always doing it in the same fashion.

sklearnex/tests/utils/base.py
```diff
@@ -415,8 +415,8 @@ def fit(self, X, y=None):
                 xp=xp,
             )
         else:
-            self.x_attr = from_table(X_table)
```
ahuber21, line 418: Why is this change necessary?

icfaust: There was a bug in the initial implementation, you can see how the assignment above under ```if sua_iface``` has the trailing underscore.  This is to follow scikit-learn standards https://scikit-learn.org/stable/developers/develop.html#estimated-attributes. It is done here in order to use the test infrastructure developed for dlpack dpnp for memory leak awareness for dlpack inputs.

onedal/datatypes/tests/test_data.py
icfaust: Sounds good, will do

onedal/datatypes/tests/test_data.py
icfaust: Good comment, its sneakily close to scikit-learns ```get_namespace``` which we have a version of, the problem with that code is that it is 1) varies by scikit-learn version causing problems in scikit-learn 1.0 2) it must be activated using ```config_context```, which is also painful going forward. We can have a minimal version (I think its a good idea), as long as we are okay with a tiny bit of duplication.

ahuber21 (approved):

david-cortes-intel:
@icfaust I'm again thinking that it might be better to have the dlpack header as a git submodule:
https://github.com/dmlc/dlpack/blob/main/include/dlpack/dlpack.h

It has already undergone changes since the time that you copied it here.

onedal/datatypes/dlpack/dlpack_utils.cpp
```diff
@@ -0,0 +1,128 @@
...
+    py::object copy;
+    if (py::hasattr(obj, "copy")) {
+        copy = obj.attr("copy")();
+    }
+    else if (py::hasattr(obj, "__array_namespace__")) {
+        const auto space = obj.attr("__array_namespace__")();
+        copy = space.attr("asarray")(obj, "copy"_a = true);
+    }
+    else {
+        throw std::runtime_error("Wrong strides");
```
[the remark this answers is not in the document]
icfaust: This is for conformance to raised errors in dpctl and dpnp. I'm not saying its right, but its consistent.

onedal/datatypes/dlpack/dlpack_utils.cpp
```diff
@@ -0,0 +1,128 @@
...
+    if (hasattr(obj, "__array_namespace__")) {
+        PyErr_WarnEx(
+            PyExc_RuntimeWarning,
+            "Data will be converted into float32 from float64 because device does not support it",
+            1);
+        const auto space = obj.attr("__array_namespace__")();
+        copy = space.attr("astype")(obj, space.attr("float32"));
+    }
+    else {
+        throw std::invalid_argument("Data has higher precision than the supported device");
```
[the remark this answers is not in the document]
icfaust: Thank you for the suggestion, changed to "Data has higher precision than supported by the device", as this error occurs when data is attempted to be loaded onto a GPU/device when it exists elsewhere (i.e. CPU).

onedal/datatypes/dlpack/dlpack_utils.cpp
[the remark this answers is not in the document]
david-cortes-intel: If I'm getting the logic correctly, this error could potentially be raised for inputs that are from packages unrelated to `dpctl` / `dpnp`, like [PyTensor](https://pytensor.readthedocs.io/en/latest/index.html).

PR #2296, opened 2025-02-02
[enhancement] array_api acceleration for sklearnex's ```validate_data``` and ```_check_sample_weight```

icfaust:
## Description

Continues dlpack work from #2275. ```validate_data``` and ```_check_sample_weight``` do not follow standard sklearnex offloading practice, namely they compute always wherever the data is (as the data movement could ruin any speedup provided by oneDAL, the algorithm is extraordinarily simple).  Therefore, they must be enabled separately for array_api support. Since they are to be included in every zero-copy array_api supported algorithm, it is a prerequisite for enabling every other estimator.

Previously this aspect was controlled by the looking for the ```flags``` attribute, which is not in the array_api standard. The array api standard does not include python-facing attributes or methods which can show if C-contiguous or F-contiguous. However, the array_api standard requires dlpack support. The attributes of from a DLPack tensor can be checked for the memory layout instead. This PR introduces a special onedal backend function which extracts and checks the necessary memory layout (without taking ownership of the tensor).  A python function is created which first checks and queries the ```flags``` or ```__dlpack__``` attributes. If neither are available, it will return False triggering the sklearn ```_assert_all_finite```.  This is done as ```to_table``` will attempt to convert to a contiguous memory layout, which again will ruin the performance gain.

---

**Testing**

**Performance**

### DIFF
--- a/onedal/utils/validation.py
+++ b/onedal/utils/validation.py
@@ -462,3 +462,12 @@ def assert_all_finite(
         allow_nan=allow_nan,
         input_name=input_name,
     )
+
+
+def is_contiguous(X):
+    if hasattr(X, "flags"):
+        return X.flags["C_CONTIGUOUS"] or X.flags["F_CONTIGUOUS"]
+    elif hasattr(X, "__dlpack__"):
+        return _backend.dlpack_memory_order(X.__dlpack__()) is not None
+    else:
+        raise TypeError(f"layout of {type(X)} object cannot be checked")
--- a/.github/.licenserc.yaml
+++ b/.github/.licenserc.yaml
@@ -65,7 +65,8 @@ header:
     # Specific files
     - 'setup.cfg'
     - 'LICENSE'
-
+    # External copies of copyrighted work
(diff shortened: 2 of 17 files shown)


onedal/utils/validation.py
```diff
@@ -462,3 +462,12 @@ def assert_all_finite(
         allow_nan=allow_nan,
         input_name=input_name,
     )
+
+
+def is_contiguous(X):
+    if hasattr(X, "flags"):
+        return X.flags["C_CONTIGUOUS"] or X.flags["F_CONTIGUOUS"]
```
ahuber21, line 469: Just making sure: Both keys `C_CONTIGUOUS` and `F_CONTIGUOUS` will always be defined, with the appropriate `True`/`False` values?

Alexsandruss (approved):

onedal/utils/validation.py
icfaust: Good question! This works for numpy/dpctl and dpnp, and is a numpy standard but not available in array_api.

PR #2301, opened 2025-02-06
feature: trigger doc build on release tag

yuejiaointel:
## Description
Hi all,
This PR intends to automate the process to release sklearnex doc. Main changes are in a new file called doc-release.yml. Automation will be triggered if a valid tag is pushed.
Best

---

**Testing**


.ci/pipeline/docs.yml
```diff
@@ -15,10 +15,10 @@
 #===============================================================================
 
 trigger:
-  branches:
+  #doc build trigger by a release tag
+  tags:
     include:
-    - main
-    - rls/*
+      - "[0-9]*.[0-9]*"
```
david-cortes-intel, line 21: Looks like the release tags currently have format year.month.minor - e.g. "2025.1.0".

.ci/pipeline/docs.yml
```diff
@@ -54,6 +54,38 @@ jobs:
     inputs:
       versionSpec: '3.11'
       addToPath: true
+  - script: |
+      # Extract the version from the triggering tag
+      DOC_VERSION=$(git describe --tags --abbrev=0)
+      echo "DOC_VERSION1: $DOC_VERSION"
+
+      # If no tag is found, fetch the latest version from PyPI
```
david-cortes-intel, line 62: I don't think we'd want to fetch something from PyPI for the purpose of building new docs. Better to let it fail.

.ci/pipeline/docs.yml
yuejiaointel: Hi David,
Thx for the info, I added extra .number in tag trigger 
Best,
Yue

.ci/pipeline/docs.yml
yuejiaointel: Sure, that make sense, I removed this part now it will fail if cannot find tag Best, Yue

.github/workflows/docs-release.yml
yuejiaointel, line 1: Was able to trigger by tag using github action instead of azure, so added a new yml file for doc release, but not sure if we still want to keep the old one

doc/sources/conf.py
```diff
@@ -44,9 +44,9 @@
 # The short X.Y version
 # Note: it should not have more than  two parts (year.month), otherwise the
 # version switcher will not be able to pick it.
-version = "2025.2"
+version = os.environ.get("DOC_VERSION", "")
```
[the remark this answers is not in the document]
yuejiaointel: thx for this! fixed Yue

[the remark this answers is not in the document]
david-cortes-intel: This file may still be used outside of this script, so it should still keep working without the variable `SHORT_DOC_VERSION` being defined. Perhaps it could default to "latest".

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,174 @@
...
+        pip install -r requirements-test.txt
+        pip list
+
+    - name: Build daal4py/sklearnex
+      run: |
+        export DALROOT=$(dirname $(dirname $(which python)))
+        export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
+        ./conda-recipe/build.sh
+
+    - name: Build daal4py Documentation
```
david-cortes-intel, line 84: These are not needed for sklearnex docs.

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,174 @@
...
+        else
+            echo "ERROR: Tag $DOC_VERSION does not exist."
+            exit 1
+        fi
+        git branch
+        echo "Can checkout release brance"
+    
+    - name: Install Python Dependencies
+      run: |
+        pip install daal-devel impi-devel
```
david-cortes-intel, line 72: These would have to be installed outside of 'pip', so as to ensure that the versions match. Perhaps it could pull it from some oneDAL job or similar.

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,174 @@
...
+
+    - name: Build scikit-learn-intelex Documentation
+      run: |
+        export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
+        cd doc
+        ./build-doc.sh --gh-pages
+        echo "doc build folder"
+        ls _build/scikit-learn-intelex
+
+    - name: Modify version json file
```
david-cortes-intel, line 103: Perhaps this could better be done by checking the versions that exist in the "gh-pages" branch. Maybe the file could be removed from the main branch and left only in "gh-pages" (in which case it might be easier to have a python script to re-generate it).

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,174 @@
...
+        python-version: "3.11"
+
+    - name: Install System Dependencies
+      run: sudo apt-get update && sudo apt-get install -y clang-format pandoc
+
+    - name: Extract Tag Version
+      run: |
+        echo "GITHUB_REF: $GITHUB_REF"
+        export DOC_VERSION="${GITHUB_REF#refs/tags/}"
+        # Error iif cannot find version
```
david-cortes-intel, line 45:
```suggestion
        # Error out if it cannot find version
```

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,174 @@
...
+        export SHORT_DOC_VERSION=$(echo "$DOC_VERSION" | awk -F'.' '{print $1"."$2}')
+        echo "DOC_VERSION: $DOC_VERSION $SHORT_DOC_VERSION"
+        echo "doc version can be found"
+        # export env var in other files
+        echo "DOC_VERSION=$DOC_VERSION" >> $GITHUB_ENV
+        echo "SHORT_DOC_VERSION=$SHORT_DOC_VERSION" >> $GITHUB_ENV
+        echo "can set short doc version"
+
+
+    - name: Checkout release brance
```
david-cortes-intel, line 59:
```suggestion
    - name: Checkout release branch
```

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,174 @@
...
+        fi
+        git branch
+        echo "Can checkout release brance"
+    
+    - name: Install Python Dependencies
+      run: |
+        pip install daal-devel impi-devel
+        pip install -r dependencies-dev
+        pip install -r requirements-doc.txt
+        pip install -r requirements-test.txt
```
david-cortes-intel, line 75: This one is not necessary for building the docs.

.ci/pipeline/docs.yml
icfaust: Running this only on push seems to be polluting the tags list (as is evidenced here: https://github.com/uxlfoundation/scikit-learn-intelex/actions/workflows/docs-release.yml)

@yuejiaointel do you have a strategy for preventing misuse of the tags?

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,174 @@
...
+        jq --arg version "$SHORT_DOC_VERSION" '. |= [{"name": $version, "version": $version, "url": ("/scikit-learn-intelex/" + $version + "/")}] + .' "$VERSIONS_FILE" > tmp.json && mv tmp.json "$VERSIONS_FILE"
+        jq --arg version "$SHORT_DOC_VERSION" '(.[] | select(.name == "latest")) .version = $version' "$VERSIONS_FILE" > tmp.json && mv tmp.json "$VERSIONS_FILE"
+
+        echo "Updated versions.json:"
+        cat "$VERSIONS_FILE"
+        ls
+
+    - name: Deploy Documentation to gh-pages
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
icfaust, line 124: It seems that unlike pull_request it seems like push is allowed secret access, and since the token that will go here will have elevated rights to the repo, we need to be extraordinarily careful with it.  Could you please describe how we will keep the docs-release from leaking the secret (say if someone were to modify docs-release.yml to do so)?

icfaust (commented):
Since the github token will get replaced by something with higher privileges, I'd like this answered before any changes to our secrets occurs @homksei . CC'ing @david-cortes-intel since he seems to be taking charge on the review-side.

.github/workflows/docs-release.yml
icfaust: Like is the secret only available on push to main? Otherwise this would look to be a security vulnerability.

.ci/pipeline/docs.yml
yuejiaointel: > That regex would also match tags like "1.2", "..1", etc. Better to make it more specific for what we're looking for, for example like this:
> 
> ```
> \d{3}\.\d{1,2}\.\d
> ```

Fixed

yuejiaointel: > Running this only on push seems to be polluting the tags list (as is evidenced here: https://github.com/uxlfoundation/scikit-learn-intelex/actions/workflows/docs-release.yml)
> 
> @yuejiaointel do you have a strategy for preventing misuse of the tags?

Hi Ian, I realized that now I am testing on my fork for tag testing

.github/workflows/docs-release.yml
yuejiaointel: Thx removed!

.github/workflows/docs-release.yml
yuejiaointel: Thx fixed typo!

.github/workflows/docs-release.yml
yuejiaointel: Thx, fixed the typo!

doc/sources/conf.py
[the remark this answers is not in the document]
yuejiaointel: thx, made the default latest

.github/workflows/docs-release.yml
yuejiaointel: Hi David, I am not sure if I understand it. Do you mean we don't modify version file in release branch, but modify it on gh-pages branch? Could you give a bit more hints on this? Thx

.ci/pipeline/docs.yml
david-cortes-intel: @yuejiaointel I tihnk you need to escape the "." there as otherwise it'd match any character.

.github/workflows/docs-release.yml
david-cortes-intel: The 'versions.json' file is only used by the docs on 'gh-pages'. It's not needed for local builds of docs like in the CI.

The issue with it is that it has the entries listed manually for each version, and those are not kept up to date as newer versions are added, nor as release branches are created.

It only should list the versions that are available to see in the 'gh-pages' so I think it'd be better to keep it there only. It'd have to be generated by listing all the versions for which there are docs (including the new version being added) plus the "latest" entry.

david-cortes-intel:
@yuejiaointel Looks like there's something wrong with the "latest" folder in the PR that the bot made in your fork. It should be an identical copy of the new version that got added, but it doesn't contain .html files.

Also, could you enable the "pages" functionality in your fork so that we can see the rendered docs? You can do so in the repository settings:
image

.github/workflows/docs-release.yml
yuejiaointel: Hi David,
I made changes to completely remove versions.json from the repo and will generate it dynamically on last step of doc release based on current version folders in gh-pages 
Best,
Yue

.github/workflows/docs-release.yml
yuejiaointel: Hi David, 
I removed the test one, but the dev one contains dependencies to build daal4py/sklearnex, and the doc one contains dependencies like sphinx that are needed to generate doc htmls, so I kept them
Best,
Yue

.github/workflows/docs-release.yml
yuejiaointel: Hi Ian,
Here is my understanding, can correct me if wrong:

Now this workflow is only triggered on tag pushes and tag pushes require write access to the repository. Only trusted maintainers can create and push tags.

yuejiaointel:
> @yuejiaointel Looks like there's something wrong with the "latest" folder in the PR that the bot made in your fork. It should be an identical copy of the new version that got added, but it doesn't contain .html files.
> 
> Also, could you enable the "pages" functionality in your fork so that we can see the rendered docs? You can do so in the repository settings: image

Hi David,
Fixed. Problem is the previous command is rf latest it deleted some files when cp, I made sure it delete first then copy. I also enabled the pages the rendered links should be visible at https://yuejiaointel.github.io/scikit-learn-intelex

doc/sources/conf.py
```diff
@@ -155,12 +154,10 @@
     "titles_only": False,
 }
 
-switcher_url = "/scikit-learn-intelex/versions.json"
 
 html_context = {
     "current_version": version,
     "project_name": "scikit-learn-intelex",
-    "switcher_url": switcher_url,
```
david-cortes-intel, line 163: @yuejiaointel  This line needs to be here for the version switcher to work. Perhaps it could be left with only version "latest" on the main branch, while being re-generated for the "gh-pages" branch.

.github/workflows/docs-release.yml
david-cortes-intel: Yes, that's what I meant.

.github/workflows/docs-release.yml
david-cortes-intel: Yes, that'd be a reasonable approach.

david-cortes-intel:
> > @yuejiaointel Looks like there's something wrong with the "latest" folder in the PR that the bot made in your fork. It should be an identical copy of the new version that got added, but it doesn't contain .html files.
> > Also, could you enable the "pages" functionality in your fork so that we can see the rendered docs? You can do so in the repository settings: image
> 
> Looking into it rn. Enabled the pages the rendered links should be visible at https://yuejiaointel.github.io/scikit-learn-intelex

Link throws a 404 by default, unless specifying a previous version like "2025.1":
image

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,132 @@
...
+        DEPLOY_DIR="/tmp/gh-pages-deploy"
+        mkdir -p "$DEPLOY_DIR"
+        cp -R "$BUILD_DIR"/* "$DEPLOY_DIR"
+        ls $DEPLOY_DIR
+        
+        # Checkout gh-pages branch
+        if git fetch origin gh-pages; then
+          git checkout gh-pages
+        else
+          git checkout -b gh-pages
```
david-cortes-intel, line 105: I think better to error out with an informative message in this situation. Otherwise it anyway wouldn't let you merge the commits.

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,132 @@
...
+        fi
+        
+        # Move the new versioned folder to the correct location
+        rm -rf latest
```
david-cortes-intel, line 109:
```suggestion
        rm -Rf latest
```

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,132 @@
...
+          echo '  {"name": "'$version'", "version": "'$version'", "url": "/scikit-learn-intelex/'$version'/"},'
+        done | sort -u >> versions.json
+        # Remove trailing comma and close array
+        sed -i '$ s/,$//' versions.json
+        echo "]" >> versions.json
+        cat versions.json
+        
+        # Commit and push changes
+        git add "$SHORT_DOC_VERSION"
+        git add latest
```
david-cortes-intel, line 129: Should also add the "version.json" and "index.html" files.

yuejiaointel: Added, tho they should work still since we already have these files

doc/sources/conf.py
david-cortes-intel: @yuejiaointel It does need to point to a JSON file - sorry about the earlier comment about removing it entirely.

.github/workflows/docs-release.yml
yuejiaointel: Fixed. Tho I think they are the same for rm

.github/workflows/docs-release.yml
david-cortes-intel: @yuejiaointel  The files are already committed, but they need to have entries for the last added versions, which is why they are re-generated.

.github/workflows/docs-release.yml
yuejiaointel: Good catch! Fixed now it will error out

doc/sources/conf.py
yuejiaointel: Thx, that make sense now, I was wondering why version switcher does not work.

doc/sources/_templates/layout.html
yuejiaointel, line 1: These styles are are applied automatic when I save, would it be okay if we keep them?

doc/sources/_static/version_switcher.js
yuejiaointel, line 1: So is this file, before some automatic headers are added but they are not correct comment in js and is causing layout.html to not recognize the load_versions function

yuejiaointel:
> > > @yuejiaointel Looks like there's something wrong with the "latest" folder in the PR that the bot made in your fork. It should be an identical copy of the new version that got added, but it doesn't contain .html files.
> > > Also, could you enable the "pages" functionality in your fork so that we can see the rendered docs? You can do so in the repository settings: image
> > 
> > 
> > Looking into it rn. Enabled the pages the rendered links should be visible at https://yuejiaointel.github.io/scikit-learn-intelex
> 
> Link throws a 404 by default, unless specifying a previous version like "2025.1": image

Fixed! The version_switcher.js is having some wrong style comment on top and is causing layout html to not recognize the load_versions function

doc/sources/_templates/layout.html
david-cortes-intel: Are you talking about the linting? Why are you saving some change on the file? You could reset it with `git checkout` if needed.

doc/sources/_static/version_switcher.js
david-cortes-intel: Yes, please make the modifications as needed. Curious that it was working before without issuing any kind of error.

yuejiaointel: Hi David,
I think before versions.json is in doc already, after built the htmls generated already have this part, so we don't need to dynamically load the versions.
Best,
Yue

doc/sources/_templates/layout.html
yuejiaointel: Yes it is the linting, I have js extensions. The modified style is better for read/normal html/js style, before the style is applied probably as python style

yuejiaointel:
> @yuejiaointel I notice in your fork the versions are ordered from oldest to newest, while in our docs they are from newest to oldest. Would be better to have the newest entry first as that's what users are most likely to want to see: image

that's a good catch! Fixed

doc/sources/conf.py
```diff
@@ -155,12 +153,11 @@
...
+    "switcher_url": "/scikit-learn-intelex/versions.json",
```
david-cortes-intel, line 160: @yuejiaointel Please remember to leave a file `/scikit-learn-intelex/versions.json` in the main branch so that the docs could still be built locally. Perhaps something having only the entry "latest" would do.

yuejiaointel: Right otherwise version switcher won't work locally, fixed and tested it locally. I had to add versions.json to the license yml tho because json cannot have comments so I cannot add license header

versions.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "name": "latest",
+        "version": "2025.1",
```
david-cortes-intel, line 4: This one I think should also be "latest", because that's the version string that would be found in `conf.py` if not set through an env. variable.

versions.json
```diff
@@ -0,0 +1,7 @@
+[
```
david-cortes-intel, line 1: Since this file is only for docs, better to leave it under folder `doc/`.

yuejiaointel: moved, but need to unify the versions.json location on gh-pages too so on gh-pages there is now also a doc folder with versions.json generated dynamically inside

versions.json
yuejiaointel: Good catch fixed!

.github/.licenserc.yaml
```diff
@@ -54,7 +54,7 @@ header:
     - 'doc/sources/_templates/versions.html'
     - 'doc/daal4py/third-party-programs.txt'
     - 'doc/third-party-programs-sklearnex.txt'
-    - 'doc/versions.json'
+    - 'versions.json'
```
david-cortes-intel, line 57:
```suggestion
    - 'doc/versions.json'
```

versions.json
david-cortes-intel: Yes, but you don't need to copy this one over to gh-pages. You could simply omit it, then re-generate it.

yuejiaointel: Done, now it is working both locally and deployment https://yuejiaointel.github.io/scikit-learn-intelex/latest/

.github/.licenserc.yaml
yuejiaointel: Good catch, fixed!

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,141 @@
...
+        done | sort -rV >> doc/versions.json
+        # Remove trailing comma and close array
+        sed -i '$ s/,$//' doc/versions.json
+        echo "]" >> doc/versions.json
+        cat doc/versions.json
+        
+        # Commit and push changes
+        git add -A "$SHORT_DOC_VERSION"
+        git add -A latest
+        git add doc/versions.json
```
david-cortes-intel, line 137: This file should go into the root folder of the 'gh-pages' branch.

yuejiaointel: Hi David, I think it has to be same location as in main as defined by switcher_url in conf.py, otherwise the switcher won't work.

david-cortes-intel: I don't think that's the case. Currently it's placed in the root of 'gh-pages', and it works as intended:
image

If you notice also, the docs get built inside `doc/`, so there's no subfolder `doc` inside docs that they'd reference.

yuejiaointel: Hi David, If I understand it correctly, this is a screenshot of main repo of gh-pages? I think originally we are defining switcher_url to be scikit-learn-intelex/versions.json, since gh-pages has versions.json in root that's why version switcher works. But now we changed the switcher_url to be doc/versions.json, which means htmls now looks for versions.json in doc folder both on main and gh-pages. I also tested it and it now verison switcher no longer works https://yuejiaointel.github.io/scikit-learn-intelex/latest/
image

david-cortes-intel: Yes, you're right, probably that's why it doesn't show up when building it locally.

But if you move it under `doc/`, would it still keep working after you switch to an older doc version that had it at the root?

yuejiaointel: yes, it will still work, reason is we are dynamically generating verisons.json now and as long as it keeps the same as defined in swicher_url it should work. Let me revert the last commit and it should be viewable

david-cortes-intel (approved):
LGTM, pending a reply back from @icfaust about https://github.com/uxlfoundation/scikit-learn-intelex/pull/2301#discussion_r1955467322

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,141 @@
...
+        ./conda-recipe/build.sh
+    - name: Build scikit-learn-intelex Documentation
+      run: |
+        export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
+        cd doc
+        ./build-doc.sh --gh-pages
+    - name: Deploy Documentation to gh-pages
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
```
icfaust, line 83: This run call is 57 lines long, ideally it should be a separate shell script located in .github/scripts/

doc/sources/_templates/layout.html
```diff
@@ -1,23 +1,24 @@
...
+</script>
 {% endblock %}
 
 {% block menu %}
-  {% include "versions.html" %}
-  {{ super() }}
-{% endblock %}
+{% include "versions.html" %}
+{{ super() }}
+{% endblock %}
```
icfaust, line 24: Endline here needs to be added (unless it breaks the html).

doc/versions.json
```diff
@@ -1,38 +1,7 @@
...
-    {
-        "version": "2024.3",
-        "url": "/scikit-learn-intelex/2024.3/"
-    },
-    {
-        "version": "2023.2",
-        "url": "/scikit-learn-intelex/2023.2/"
     }
-]
+]
```
icfaust, line 7: another newline here.

icfaust (commented):
overall good, just some nitpicks I noticed.

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,141 @@
...
+        # Set git auth for push changes 
+        git config --global user.name "GitHub Actions"
+        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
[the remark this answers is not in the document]
yuejiaointel: I searched git config and there is a .cache/per-commit/repondmp4e6s/workflows/main.yml so I tried to jsut copy it over, need this config to push the changes

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,141 @@
...
+        export SHORT_DOC_VERSION=$(echo "$DOC_VERSION" | awk -F'.' '{print $1"."$2}')
+        # export env var in other files
+        echo "DOC_VERSION=$DOC_VERSION" >> $GITHUB_ENV
+        echo "SHORT_DOC_VERSION=$SHORT_DOC_VERSION" >> $GITHUB_ENV
+    - name: Checkout release branch
+      run: |
+        if git checkout $DOC_VERSION 2>/dev/null; then
+          echo "Successfully checked out tag $DOC_VERSION."
+        else
+            echo "ERROR: Tag $DOC_VERSION does not exist."
```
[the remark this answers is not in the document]
yuejiaointel: Fixed thx!

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,141 @@
...
+
+    - name: Install System Dependencies
+      run: sudo apt-get update && sudo apt-get install -y clang-format pandoc
+
+    - name: Extract Tag Version
+      run: |
+        export DOC_VERSION="${GITHUB_REF#refs/tags/}"
+        # Error out if cannot find version
+        if [ -z "$DOC_VERSION" ]; then
+          echo "ERROR: Failed to determine documentation version."
```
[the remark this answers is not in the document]
yuejiaointel: Fixed thx!

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,141 @@
...
+        
+        # Copy built documentation to a temp location
+        DEPLOY_DIR="/tmp/gh-pages-deploy"
+        mkdir -p "$DEPLOY_DIR"
+        cp -R "$BUILD_DIR"/* "$DEPLOY_DIR"
+        ls $DEPLOY_DIR
+        
+        # Checkout gh-pages branch
+        if ! git checkout gh-pages; then
+            echo "ERROR: Could not checkout gh-pages branch!"
```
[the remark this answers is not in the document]
yuejiaointel: Fixed thx!

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,141 @@
...
+            exit 1
+        fi
+        
+        # Move the new versioned folder to the correct location
+        rm -Rf latest
+        cp -R "$DEPLOY_DIR/$SHORT_DOC_VERSION" "$SHORT_DOC_VERSION"
+        cp -R "$DEPLOY_DIR/$SHORT_DOC_VERSION" latest
+        cp "$DEPLOY_DIR/index.html" .
+        if ! diff -r "$SHORT_DOC_VERSION" latest > /dev/null; then
+            echo "ERROR: Content mismatch between $SHORT_DOC_VERSION and latest directories"
```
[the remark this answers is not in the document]
yuejiaointel: Fixed thx!

doc/sources/_templates/layout.html
yuejiaointel: Fixed thx!

doc/versions.json
yuejiaointel: Fixed thx!

.github/workflows/docs-release.yml
yuejiaointel: Moved now I have a doc_release.sh in scripts

david-cortes-intel: I had the impression that the scripts were meant to be executable both locally and on GH.

This script would mostly be about git commands which are specific to this particular workflow and not reusable elsewhere. I think it makes sense to keep it here.

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,141 @@
...
+    - name: Build daal4py/sklearnex
+      run: |
+        export DALROOT=$(dirname $(dirname $(which python)))
+        export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
+        ./conda-recipe/build.sh
+    - name: Build scikit-learn-intelex Documentation
+      run: |
+        export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
+        cd doc
+        ./build-doc.sh --gh-pages
```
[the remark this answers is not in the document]
david-cortes-intel: Yes, that's because they do slightly different things. This one with `--gh-pages` uses different build options and generates a versioned folder - simply reusing the result from the docs CI and renaming it results in docs that don't render when put under gh-pages.

But that shouldn't be a problem because in the end both of them are using the same software to build the docs, with the same pinned versions from `requirements-doc.txt`.

.github/workflows/docs-release.yml
icfaust: @Alexsandruss @homksei please weigh in on this, as we are both inferring about the rules of the codebase when it comes to this topic.

david-cortes-intel: Yes, that sounds like a reasonable split.

yuejiaointel: Thx for the suggestions! Moved git config back to yml

.github/scripts/doc_release.sh
```diff
@@ -0,0 +1,71 @@
...
+echo "]" >> doc/versions.json
+cat doc/versions.json
+
+# Commit and push changes
+git add -A "$SHORT_DOC_VERSION"
+git add -A latest
+git add doc/versions.json
+git add index.html
+git commit . -m "Automatic doc update for version $DOC_VERSION"
+git push origin gh-pages
```
icfaust, line 71: newline here

icfaust (approved):
Just add the endline to the one file and can merge.

.github/scripts/doc_release.sh
yuejiaointel: Fixed thx!

PR #2316, opened 2025-02-12
ENH: Add new parameters for linear regression

david-cortes-intel:
## Description

This PR adds the new parameters for linear regression that would be introduced to oneDAL in this PR:
https://github.com/uxlfoundation/oneDAL/pull/3058

These parameters allow controlling when to fall back to a non-batched route for the computation of cross products.

~Right now it should be failing as the PR hasn't been merged, but will leave it here for later.~ PR in oneDAL has been merged by now.

---

**Testing**

**Performance**

Not applicable.


david-cortes-intel:
Small note: the code route that is enabled/disabled through these parameters is tested for correctness in the oneDAL side.

Vika-F:
Please add at least the minimal amount of testing as it was done for cpu_macro_block, for example:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/linear_model/tests/test_linear.py#L66

david-cortes-intel:
> Please add at least the minimal amount of testing as it was done for cpu_macro_block, for example: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/linear_model/tests/test_linear.py#L66

Added a test like that.

david-cortes-intel:
Error raised by the CI is not related to this PR and affects estimators not touched here.

onedal/linear_model/linear_model.cpp
```diff
@@ -276,23 +276,52 @@ void init_train_hyperparameters(py::module_& m) {
...
-                   .def("get_gpu_macro_block", [](const train_hyperparams_t& self) {
-                       return self.get_gpu_macro_block();
-                   });
+    auto cls =
+        py::class_<train_hyperparams_t>(m, "train_hyperparameters")
+            .def(py::init())
+            .def("set_cpu_macro_block",
+                 [](train_hyperparams_t& self, int64_t cpu_macro_block) {
+                     self.set_cpu_macro_block(cpu_macro_block);
+                 })
```
icfaust, line 285: I know that you are just following convention of what was done before (so this is not a hard request that must be done, just a possible nice-to-have), and I haven't looked at the hyperparameter implementation before in sklearnex, but shouldn't the equivalent work and simplify the interface? Let me know what you think.

```suggestion
            .def("set_cpu_macro_block", &train_hyperparams_t::set_cpu_macro_block)
```

It worked for me when I did something equivalent here: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/common/sycl.cpp#L84

sklearnex/linear_model/tests/test_linear.py
```diff
@@ -29,13 +29,70 @@
 from sklearnex.tests.utils import _IS_INTEL
 
 
+# Note: this is arranged as a fixture with a finalizer instead of as a parameter
```
icfaust, line 32: Thank you so much for the note.

sklearnex/linear_model/tests/test_linear.py
```diff
@@ -29,13 +29,70 @@
...
 
+# Note: this is arranged as a fixture with a finalizer instead of as a parameter
+# 'True' / 'False' in order to undo the changes later so that it doesn't affect
+# other tests afterwards. It returns a function instead of making the change
+# directly, in order to avoid importing the estimator class before the import test
+# itself, but it still needs to import the class inside the the function that it
+# returns due to serialization logic in pytest causing differences w.r.t. current
+# closure where the function is called.
+@pytest.fixture(params=[False, True])
+def non_batched_route(request):
```
icfaust, line 40: Not to be done in this PR, but the general strategy seems sound, and other hyperparameter testing for other estimators likely needs similar control of hysteresis.  Maybe it would be worth putting a ticket in the backlog to generalize your idea here to other estimators? Let me know what you think.

icfaust (commented):
This should be a rather quick PR to get integrated. I had some questions before I approve and I didn't notice anything blocking it, just maybe some improvement recommendations.

sklearnex/linear_model/tests/test_linear.py
david-cortes-intel: Yes, definitely for a different PR. We already have a ticket for it.

onedal/linear_model/linear_model.cpp
david-cortes-intel: Looks like it works too. Modified all the setters to use that syntax.

sklearnex/linear_model/tests/test_linear.py
```diff
@@ -29,13 +29,70 @@
...
+    def change_parameters(queue, macro_block):
+        if not daal_check_version((2025, "P", 500)):
```
icfaust, line 42: It looks like this causes test_sklearnex_import_linear to skip the test for all conditions in azure pipelines CI https://dev.azure.com/daal/daal4py/_build/results?buildId=44302&view=results . Was this desired? Secondarily, It also may be worth to place the skips in the main test function to simplify understanding the test (just a recommendation).

david-cortes-intel: Yes, that's expected, since version 2025.5 is only available in the nightlies. The skip logic could be added to the test body, but then it'd need to have different return types and if-else logic in the test.

david-cortes-intel:
Error from the CI is just a timeout.

sklearnex/linear_model/tests/test_linear.py
icfaust: It seems that latest main commit runs the test in azure pipelines as it seemed relevant for <2025.5, but these changes disable the testing for <2025.5. Shouldn't the skip be moved to after line 52: for when ```if request.param``` is true? That way the test will run as normal when ```request.param``` is False

david-cortes-intel: Yes, you're right. Moved the skipping logic to inside of the the if condition.

icfaust (approved):

david-cortes-intel:
Error in the CI is rather curious:
```
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda/envs/build-env/lib/python3.11/site-packages/conda_build/os_utils/liefldd.py", line 381, in <listcomp>
    result = [e.name for e in dynamic_entries if e.tag == ELF_DYNAMIC_TAGS.SONAME]
                                                 ^^^^^
ValueError: 32212254717 is not a valid TAG.
```

I'm not sure how that happens. I assume it might be a sporadic thing.

PR #2326, opened 2025-02-25
fix: use utf-8 encoding

ahuber21:
## Description

In some cases non-ASCII characters are used in logging (e.g. the trademark sign) which breaks the current write to file.

---

**Testing**

**Performance**

### DIFF
--- a/tests/run_examples.py
+++ b/tests/run_examples.py
@@ -246,7 +246,7 @@ def run(exdir, logdir, args):
                     )
                 else:
                     logfn = jp(logdir, script.replace(".py", ".res"))
-                    with open(logfn, "w") as logfile:
+                    with open(logfn, "w", encoding="utf-8") as logfile:
                         print("\n##### " + jp(dirpath, script))
                         execute_string = get_exe_cmd(jp(dirpath, script), args)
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
@ahuber21 Why would you need to explicitly set a default instead of omitting the specific encoding? If you are concerned about windows, producing a file with an encoding that doesn't match with the system's encoding might later on result in issues with text editors not opening it correctly. Even then, the "(r)" character should be present in the default windows encoding for English language.

ahuber21:
> @ahuber21 Why would you need to explicitly set a default instead of omitting the specific encoding? If you are concerned about windows, producing a file with an encoding that doesn't match with the system's encoding might later on result in issues with text editors not opening it correctly. Even then, the "(r)" character should be present in the default windows encoding for English language.

No concerned. Just ran into a very specific failure when running the tests that was fixed by using "utf-8". I'm not attached to that at all. It looks like just using `out.decode()` without arguments also does the trick and I'm perfectly happy with that, too.

david-cortes-intel (approved):

Alexsandruss (approved):

PR #2340, opened 2025-03-05
chore: update documentation on config parameters

ahuber21:
## Description

Add documentation about `use_raw_input` parameter. Also adds a missing explanation of `allow_sklearn_after_onedal` in one function, and the doc of all config parameters in the dictionary of default values in `onedal/_config.py`.

---

**Testing**

**Performance**

### DIFF
--- a/sklearnex/_config.py
+++ b/sklearnex/_config.py
@@ -88,7 +88,7 @@ def set_config(
 
     Warnings
     --------
-    Using `use_raw_input=True` is not recommended for general use as it
+    Using ``use_raw_input=True`` is not recommended for general use as it
     bypasses data consistency checks, which may lead to unexpected behavior.
     """
 
--- a/sklearnex/_config.py
+++ b/sklearnex/_config.py
@@ -151,7 +151,7 @@ def config_context(**new_config):
 
     Warnings
     --------
-    Using `use_raw_input=True` is not recommended for general use as it
+    Using ``use_raw_input=True`` is not recommended for general use as it
     bypasses data consistency checks, which may lead to unexpected behavior.
     """
     old_config = get_config()
--- a/sklearnex/_config.py
+++ b/sklearnex/_config.py
@@ -26,10 +26,12 @@ from onedal._config import _get_config as onedal_get_config
 
 def get_config():
     """Retrieve current values for configuration set by :func:`set_config`
+
     Returns
     -------
     config : dict
         Keys are parameter names that can be passed to :func:`set_config`.
+
     See Also
     --------
     config_context : Context manager for global configuration.
@@ -48,6 +50,7 @@ def set_config(
     **sklearn_configs,
 ):
     """Set global configuration
+
     Parameters
     ----------
     target_offload : string or dpctl.SyclQueue, default=None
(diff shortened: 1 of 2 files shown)


sklearnex/_config.py
```diff
@@ -63,6 +63,11 @@ def set_config(
         If True, allows to fallback computation to sklearn after onedal
         backend in case of runtime error on onedal backend computations.
         Global default: True.
+    use_raw_input : bool, default=None
+        If True, uses the raw input data in the onedal backend computations
+        without any checks on data consistency or validity.
+        Note: This option is not recommended for general use.
+        Global default: False.
     See Also
```
david-cortes-intel, line 71: I think you need an empty line before starting a new section.

david-cortes-intel:
@ahuber21 Since now the options differ from those in regular sklearn, would be ideal to add these to the .rst docs through autofunction - otherwise, it's very hard to discover these options or to notice that they differ from sklearn's.

ahuber21:
> @ahuber21 Since now the options differ from those in regular sklearn, would be ideal to add these to the .rst docs through autofunction - otherwise, it's very hard to discover these options or to notice that they differ from sklearn's.

Hi @david-cortes-intel, yes these are valid concerns. We have decided to go that way because this functionality isn't intended for general use. In fact, we're hoping to remove it as quickly as possible. We don't want any adoption of this option, but rather implement array API, which makes all the raw input workarounds obsolete.

ahuber21:
@david-cortes-intel thank you! Let me know if you agree with the latest version.

david-cortes-intel:
Thanks for adding these. One detail though: 'Notes' is not a recognized section type, needs to be 'Note', otherwise the docs do not render correctly - see screenshot:
image

sklearnex/_config.py
```diff
@@ -48,25 +50,46 @@ def set_config(
...
         If True, allows to fallback computation to host device
         in case particular estimator does not support the selected one.
         Global default: False.
+
     allow_sklearn_after_onedal : bool, default=None
         If True, allows to fallback computation to sklearn after onedal
         backend in case of runtime error on onedal backend computations.
         Global default: True.
+
+    use_raw_input : bool, default=None
```
david-cortes-intel, line 72: Not entirely about this PR, but all of these should say 'bool or None', since 'None' is also supported.

sklearnex/_config.py
```diff
@@ -90,25 +113,46 @@ def set_config(
...
     their previous values when the context manager is exited.
+
     See Also
     --------
     set_config : Set global scikit-learn configuration.
     get_config : Retrieve current values of the global configuration.
+
+    Warnings
+    --------
+    Using `use_raw_input=True` is not recommended for general use as it
```
david-cortes-intel, line 154:
```suggestion
    Using ``use_raw_input=True`` is not recommended for general use as it
```

sklearnex/_config.py
```diff
@@ -48,25 +50,46 @@ def set_config(
...
+    their previous values when the context manager is exited.
+
     See Also
     --------
     config_context : Context manager for global configuration.
     get_config : Retrieve current values of the global configuration.
+
+    Warnings
+    --------
+    Using `use_raw_input=True` is not recommended for general use as it
```
david-cortes-intel, line 91:
```suggestion
    Using ``use_raw_input=True`` is not recommended for general use as it
```

sklearnex/_config.py
ahuber21: Yes I agree. We are not at all consistent here.

ahuber21:
> Thanks for adding these. One detail though: 'Notes' is not a recognized section type, needs to be 'Note', otherwise the docs do not render correctly - see screenshot: ...

Hope it's better now!

david-cortes-intel (approved):

PR #2346, opened 2025-03-06
Enhance sklearn 1.7 support and remove sklearn<1.0 code

Alexsandruss:
## Description

Changes:
- Fully apply deprecations and removals from sklearn 1.7 in sklearnex and daal4py by adding dispatchers based on used sklearn version in daal4py and sklearnex modules:
  - `BaseEstimator._check_feature_names` changed to `sklearn.utils.validation.check_feature_names` (dispatcher in `daal4py.sklearn.utils.validation` module)
  - `BaseEstimator._check_n_features` changed to `sklearn.utils.validation.check_n_features` (dispatcher in `daal4py.sklearn.utils.validation` module)
  - `BaseEstimator._validate_data` changed to `sklearn.utils.validation.validate_data` (dispatchers in `daal4py.sklearn.utils.validation` and `sklearnex.utils.validation` modules, sklearnex's one uses onedal4py finiteness checker)
  - `force_all_finite` argument changed to `ensure_all_finite` in `validate_data` and other functions
- Unify sklearn tags (`requires_y`, `array_api_support`) getting in few functions
- Remove numeric y_dtype conversion from `sklearnex.utils.validation.validate_data` for compatibility with sklearn which allows object types output
- Change `validate_date` design rule in `test_common.py` to correctly count sklearn's and sklearnex's calls
- Add `tol` to LinearRegression parameters (new in 1.7 version)
- Remove code branches for unsupported sklearn versions < 1.0

---

**Testing**

**Performance**

### DIFF
--- a/daal4py/sklearn/cluster/dbscan.py
+++ b/daal4py/sklearn/cluster/dbscan.py
@@ -26,6 +26,7 @@ import daal4py
 
 from .._n_jobs_support import control_n_jobs
 from .._utils import PatchingConditionsChain, getFPType, make2d, sklearn_check_version
+from ..utils.validation import check_feature_names
 
 if sklearn_check_version("1.1") and not sklearn_check_version("1.2"):
     from sklearn.utils import check_scalar
@@ -121,8 +122,7 @@ class DBSCAN(DBSCAN_original):
             if self.eps <= 0.0:
                 raise ValueError(f"eps == {self.eps}, must be > 0.0.")
 
-        if sklearn_check_version("1.0"):
-            self._check_feature_names(X, reset=True)
+        check_feature_names(self, X, reset=True)
 
         if sample_weight is not None:
             sample_weight = _check_sample_weight(sample_weight, X)
--- a/daal4py/sklearn/utils/validation.py
+++ b/daal4py/sklearn/utils/validation.py
@@ -49,6 +49,19 @@ from .._utils import (
     sklearn_check_version,
 )
 
+if sklearn_check_version("1.6"):
+    from sklearn.utils.validation import (
+        _check_feature_names as _sklearn_check_feature_names,
+    )
+    from sklearn.utils.validation import _check_n_features as _sklearn_check_n_features
+    from sklearn.utils.validation import validate_data as _sklearn_validate_data
+else:
+    from sklearn.base import BaseEstimator
+
+    _sklearn_validate_data = BaseEstimator._validate_data
+    _sklearn_check_feature_names = BaseEstimator._check_feature_names
+    _sklearn_check_n_features = BaseEstimator._check_n_features
+
 
 def _assert_all_finite(
     X, allow_nan=False, msg_dtype=None, estimator_name=None, input_name=""
@@ -394,13 +407,19 @@ def _daal_check_array(
 
     if sp.issparse(array):
         _ensure_no_complex_data(array)
+        kwargs = {
+            "accept_sparse": accept_sparse,
+            "dtype": dtype,
+            "copy": copy,
+            "accept_large_sparse": accept_large_sparse,
+        }
+        if sklearn_check_version("1.6"):
+            kwargs["ensure_all_finite"] = force_all_finite
+        else:
+            kwargs["force_all_finite"] = force_all_finite
         array = _ensure_sparse_format(
             array,
-            accept_sparse=accept_sparse,
-            dtype=dtype,
-            copy=copy,
-            force_all_finite=force_all_finite,
-            accept_large_sparse=accept_large_sparse,
+            **kwargs,
         )
     else:
         # If np.array(..) gives ComplexWarning, then we convert the warning
@@ -694,3 +713,29 @@ def _daal_num_features(X):
         return len(first_sample)
     except Exception as err:
         raise TypeError(message) from err
+
+
+def get_requires_y_tag(estimator):
+    if sklearn_check_version("1.6"):
+        requires_y = estimator.__sklearn_tags__().target_tags.required
+    else:
+        try:
+            requires_y = estimator._get_tags()["requires_y"]
+        except KeyError:
+            requires_y = False
+    return requires_y
+
+
+# simplified copy of similar function from sklearnex.utils.validation
+def validate_data(*args, **kwargs):
+    if not sklearn_check_version("1.6") and "ensure_all_finite" in kwargs:
+        kwargs["force_all_finite"] = kwargs.pop("ensure_all_finite")
+    return _sklearn_validate_data(*args, **kwargs)
+
+
+def check_feature_names(*args, **kwargs):
+    _sklearn_check_feature_names(*args, **kwargs)
+
+
+def check_n_features(*args, **kwargs):
+    _sklearn_check_n_features(*args, **kwargs)
--- a/daal4py/sklearn/utils/validation.py
+++ b/daal4py/sklearn/utils/validation.py
@@ -717,6 +717,8 @@ def _daal_num_features(X):
 
 
 def get_requires_y_tag(estimator):
+    """Gets the value of the 'requires_y' tag from the estimator
+    using correct code path depending on the scikit-learn version."""
     if sklearn_check_version("1.6"):
         requires_y = estimator.__sklearn_tags__().target_tags.required
     else:
--- a/daal4py/sklearn/utils/validation.py
+++ b/daal4py/sklearn/utils/validation.py
@@ -16,6 +16,7 @@
 
 import warnings
 from contextlib import suppress
+from functools import wraps
 
 import numpy as np
 import scipy.sparse as sp
@@ -726,16 +727,45 @@ def get_requires_y_tag(estimator):
     return requires_y
 
 
-# simplified copy of similar function from sklearnex.utils.validation
+def add_dispatcher_docstring(original_function):
+    """Adds a note about the dispatcher function to the docstring of the original function."""
+
+    def wrapper(dispatcher_function):
+        @wraps(
+            original_function,
+            ["__name__", "__doc__", "__annotations__", "__type_params__"],
+        )
+        def new_function(*args, **kwargs):
+            return dispatcher_function(*args, **kwargs)
+
+        new_function.__doc__ = (
+            f"Sklearnex dispatcher for '{original_function.__qualname__}' "
+            f"from '{original_function.__module__}' module supporting "
+            "it's multiple implementations from different scikit-learn versions.\n\n"
+            + original_function.__doc__
+        )
+
+        return new_function
+
+    return wrapper
+
+
+# simplified copy of similar function from sklearnex.utils.validation,
+# ensures that the correct finiteness check argument is used
+@add_dispatcher_docstring(_sklearn_validate_data)
 def validate_data(*args, **kwargs):
     if not sklearn_check_version("1.6") and "ensure_all_finite" in kwargs:
         kwargs["force_all_finite"] = kwargs.pop("ensure_all_finite")
     return _sklearn_validate_data(*args, **kwargs)
 
 
+# dispatcher functions which use correct `check_feature_names`/`check_n_features` import
+# depending on the scikit-learn version
+@add_dispatcher_docstring(_sklearn_check_feature_names)
 def check_feature_names(*args, **kwargs):
     _sklearn_check_feature_names(*args, **kwargs)
 
 
+@add_dispatcher_docstring(_sklearn_check_n_features)
 def check_n_features(*args, **kwargs):
     _sklearn_check_n_features(*args, **kwargs)
--- a/onedal/utils/validation.py
+++ b/onedal/utils/validation.py
@@ -131,6 +131,10 @@ def _validate_targets(y, class_weight, dtype):
 
 
 def get_finite_keyword():
+    """Gets the argument name for scikit-learn's validation functions compatible with
+    the current version of scikit-learn and using function inspection instead of
+    `sklearn_check_version` due to `onedal` design rule: sklearn versioning should occur
+    in `sklearnex` module."""
     if "ensure_all_finite" in inspect.signature(check_array).parameters:
         return "ensure_all_finite"
     return "force_all_finite"
--- a/onedal/utils/validation.py
+++ b/onedal/utils/validation.py
@@ -133,7 +133,7 @@ def _validate_targets(y, class_weight, dtype):
 def get_finite_keyword():
     """Gets the argument name for scikit-learn's validation functions compatible with
     the current version of scikit-learn and using function inspection instead of
-    `sklearn_check_version` due to `onedal` design rule: sklearn versioning should occur
+    version check due to `onedal` design rule: sklearn versioning should occur
     in `sklearnex` module."""
     if "ensure_all_finite" in inspect.signature(check_array).parameters:
         return "ensure_all_finite"
--- a/onedal/utils/validation.py
+++ b/onedal/utils/validation.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
+import inspect
 import warnings
 from collections.abc import Sequence
 from numbers import Integral
@@ -129,6 +130,12 @@ def _validate_targets(y, class_weight, dtype):
     return np.asarray(y, dtype=dtype, order="C"), class_weight_res, classes
 
 
+def get_finite_keyword():
+    if "ensure_all_finite" in inspect.signature(check_array).parameters:
+        return "ensure_all_finite"
+    return "force_all_finite"
+
+
 def _check_array(
     array,
     dtype="numeric",
@@ -138,6 +145,7 @@ def _check_array(
     force_all_finite=True,
     ensure_2d=True,
     accept_large_sparse=True,
+    _finite_keyword=get_finite_keyword(),
 ):
     if force_all_finite:
         if sp.issparse(array):
@@ -147,15 +155,19 @@ def _check_array(
         else:
             _daal4py_assert_all_finite(array)
             force_all_finite = False
+    check_kwargs = {
+        "array": array,
+        "dtype": dtype,
+        "accept_sparse": accept_sparse,
+        "order": order,
+        "copy": copy,
+        "ensure_2d": ensure_2d,
+        "accept_large_sparse": accept_large_sparse,
+    }
+    check_kwargs[_finite_keyword] = force_all_finite
+
     array = check_array(
-        array=array,
-        dtype=dtype,
-        accept_sparse=accept_sparse,
-        order=order,
-        copy=copy,
-        force_all_finite=force_all_finite,
-        ensure_2d=ensure_2d,
-        accept_large_sparse=accept_large_sparse,
+        **check_kwargs,
     )
 
     if sp.issparse(array):
--- a/sklearnex/_device_offload.py
+++ b/sklearnex/_device_offload.py
@@ -60,6 +60,16 @@ def _get_backend(obj, queue, method_name, *data):
     raise RuntimeError("Device support is not implemented")
 
 
+def get_array_api_support_tag(estimator):
+    if hasattr(estimator, "__sklearn_tags__"):
+        return estimator.__sklearn_tags__().array_api_support
+    elif hasattr(estimator, "_get_tags"):
+        tags = estimator._get_tags()
+        if "array_api_support" in tags:
+            return tags["array_api_support"]
+    return False
+
+
 def dispatch(obj, method_name, branches, *args, **kwargs):
     if get_config()["use_raw_input"] is False:
         with QM.manage_global_queue(None, *args) as queue:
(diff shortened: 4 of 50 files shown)


daal4py/sklearn/cluster/dbscan.py
```diff
@@ -121,8 +122,7 @@ def fit(self, X, y=None, sample_weight=None):
             if self.eps <= 0.0:
                 raise ValueError(f"eps == {self.eps}, must be > 0.0.")
 
-        if sklearn_check_version("1.0"):
```
david-cortes-intel, line 124: Curious about this part: was the sklearn version check not necessary before? I see the job with sklearn==1.0 passes though.

sklearnex/utils/validation.py
```diff
@@ -206,3 +212,11 @@ def _check_sample_weight(
         check_non_negative(sample_weight, "`sample_weight`")
 
     return sample_weight
+
+
+def check_feature_names(*args, **kwargs):
```
david-cortes-intel, line 217: These are also defined in `daal4py`. Could this one import it from `daal4py` instead of duplicating them?

daal4py/sklearn/utils/validation.py
```diff
@@ -49,6 +49,19 @@
...
+if sklearn_check_version("1.6"):
+    from sklearn.utils.validation import (
+        _check_feature_names as _sklearn_check_feature_names,
+    )
+    from sklearn.utils.validation import _check_n_features as _sklearn_check_n_features
+    from sklearn.utils.validation import validate_data as _sklearn_validate_data
+else:
+    from sklearn.base import BaseEstimator
+
+    _sklearn_validate_data = BaseEstimator._validate_data
```
david-cortes-intel, line 61: Question: wouldn't this kind of switch between method/function lead to problems if you serialize a model using something like `cloudpickle` or `dill` and then load it in a different version of sklearnex? Sklearn itself doesn't provide that kind of compatibility guarantees across versions, but it tends to work or otherwise show breakage right at the moment that the object is deserialized.

sklearnex/neighbors/common.py
```diff
@@ -308,3 +308,13 @@ def kneighbors_graph(self, X=None, n_neighbors=None, mode="connectivity"):
         return kneighbors_graph
 
     kneighbors_graph.__doc__ = KNeighborsMixin.kneighbors_graph.__doc__
+
+    def _get_requires_y(self):
```
icfaust, line 312: This needs documentation.  There is also tag usage of ```requires_y``` in SVM, but is not corrected there.

sklearnex/_device_offload.py
```diff
@@ -57,6 +57,16 @@ def _get_backend(obj, queue, method_name, *data):
     raise RuntimeError("Device support is not implemented")
 
 
+def get_array_api_support_tag(estimator):
+    if hasattr(estimator, "__sklearn_tags__"):
```
icfaust, line 61: This needs documentation.

sklearnex/tests/test_common.py
```diff
@@ -415,11 +415,9 @@ def call_validate_data(text, estimator, method):
     except ValueError:
         pytest.skip("onedal backend not used in this function")
 
-    validate_data = "validate_data" if sklearn_check_version("1.6") else "_validate_data"
-
     assert (
-        validfuncs.count(validate_data) == 1
-    ), f"sklearn's {validate_data} should be called"
+        validfuncs.count("validate_data") + validfuncs.count("_validate_data") == 2
```
icfaust, line 419: As far as I can read, this fundamentally breaks the purpose of the test. Unless you can provide justification for allowing 2 calls of validate data, this should be removed.  You can list which cases are causing this, at which point we can discuss if a temporary deselection is a good idea.

icfaust (changes requested):
In general there are a couple major themes:

1) documentation of the changes need to be improved in the codebase, especially by adding docstrings.
2) use of ```sklearn/utils/validation.py```'s ```validate_data``` must be benchmarked, as it relying on a different execution path (oneDAL) instead of daal4py (only when patched).  It has purposefully not rolled out to the sklearnex estimators for this reason.
3) Changes to the design rules must be discussed, as it is strict to prevent performance degradation of multiple validate_data/ check_array/ check_X_y calls.

Additionally, I would like the title to reflect the deprecation of sklearn_check_version("1.0") calls, which should be more prominent.

daal4py/sklearn/utils/validation.py
icfaust: Sounds like a good thing to test for if this is something we want to try and guarantee a smooth user experience for this circumstance.  As of now we support multiple versions of sklearn, but we make no guarantee of support between versions of sklearn.

sklearnex/svm/svc.py
```diff
@@ -147,99 +140,77 @@ def score(self, X, y, sample_weight=None):
             sample_weight=sample_weight,
         )
 
-    if sklearn_check_version("1.0"):
```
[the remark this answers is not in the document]
Alexsandruss: Changed title.

sklearn<1.0 code branches definitely should be removed:
- sklearn 0.24.2 was released 4 years ago
- 0.24.* downloads are 3-5% of 1.* ([stats](https://pepy.tech/projects/scikit-learn?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=line&versions=1.*%2C0.24.*))
- Sklearnex supports 8 versions of sklearn with current version of this PR

sklearnex/utils/validation.py
```diff
@@ -100,6 +105,7 @@ def validate_data(
     /,
     X="no_validation",
     y="no_validation",
+    skip_y_conversion=False,
```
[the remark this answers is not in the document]
Alexsandruss: Moved to `ensure_y_dtype=False` making conversion optional.
Previous dtype conversion enabled by default made sklearnex version of `validate_data` non-compliant with estimator's data handling of stock sklearn. For example, if `y` is object (string, particularly), previous version of function tried to pre-emptively convert it into numeric resulting in failures of sklearn tests.

daal4py/sklearn/utils/validation.py
Alexsandruss: Portability of estimator instances between sklearn versions is not in the scope of this PR.

sklearnex/svm/svc.py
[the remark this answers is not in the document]
icfaust: @syakov-intel @napetrov please confirm

sklearnex/tests/test_common.py
icfaust: Use text["modules"] in order to find one vs the other.  The test will pass if sklearn's validate_data is called twice.

sklearnex/utils/validation.py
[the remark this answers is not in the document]
icfaust: @Alexsandruss can you provide a concrete example of this occurring? That way I can judge if the approach is the correct one.

sklearnex/neighbors/common.py
Alexsandruss: Added comment and unified actual `requires_y` getting in one place.

sklearnex/utils/validation.py
```diff
@@ -206,3 +212,11 @@ def _check_sample_weight(
         check_non_negative(sample_weight, "`sample_weight`")
 
     return sample_weight
+
+
+def check_feature_names(*args, **kwargs):
+    _sklearn_check_feature_names(*args, **kwargs)
+
+
+def check_n_features(*args, **kwargs):
```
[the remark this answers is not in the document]
Alexsandruss: Added.

sklearnex/utils/validation.py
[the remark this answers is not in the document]
Alexsandruss: Here is a simple SVC example:
```python
from sklearn.svm import SVC

X = [[1, 2], [2, 3], [3, 4], [4, 5], [5, 6], [6, 7]]
y = ["class1", "class2", "class1", "class2", "class1", "class2"]

model = SVC(kernel="rbf", random_state=42)
model.fit(X, y)

y_pred = model.predict(X)
print(y_pred)
```
Output for `ensure_y_dtype=False`:
```
$ SKLEARNEX_VERBOSE=INFO python -m sklearnex svc_string_labels.py
Intel(R) Extension for Scikit-learn* enabled (https://github.com/uxlfoundation/scikit-learn-intelex)
INFO:sklearnex: sklearn.svm.SVC.fit: running accelerated version on CPU
INFO:sklearnex: sklearn.svm.SVC.predict: running accelerated version on CPU
['class1' 'class1' 'class1' 'class2' 'class2' 'class2']
```
Output for `ensure_y_dtype=True`:
```
$ SKLEARNEX_VERBOSE=INFO python -m sklearnex svc_string_labels.py
Intel(R) Extension for Scikit-learn* enabled (https://github.com/uxlfoundation/scikit-learn-intelex)
INFO:sklearnex: sklearn.svm.SVC.fit: running accelerated version on CPU
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/sklearnex/__main__.py", line 58, in <module>
    sys.exit(_main())
             ^^^^^^^
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/sklearnex/__main__.py", line 55, in _main
    runf(args.name, run_name="__main__")
  File "<frozen runpy>", line 287, in run_path
  File "<frozen runpy>", line 98, in _run_module_code
  File "<frozen runpy>", line 88, in _run_code
  File "svc_string_labels.py", line 7, in <module>
    model.fit(X, y)
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/daal4py/sklearn/_n_jobs_support.py", line 111, in n_jobs_wrapper
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/sklearnex/svm/svc.py", line 101, in fit
    dispatch(
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/sklearnex/_device_offload.py", line 85, in dispatch
    return branches[backend](obj, *hostargs, **hostkwargs, queue=q)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/sklearnex/svm/svc.py", line 280, in _onedal_fit
    X, _, weights = self._onedal_fit_checks(X, y, sample_weight)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/sklearnex/svm/_common.py", line 162, in _onedal_fit_checks
    X, y = validate_data(
           ^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/daal4py/sklearn/utils/validation.py", line 741, in new_function
    return dispatcher_function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/sklearnex/utils/validation.py", line 155, in validate_data
    outy = yp.asarray(outy, dtype=dtype[0])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/envs/build/lib/python3.12/site-packages/sklearn/utils/_array_api.py", line 407, in asarray
    return numpy.asarray(x, dtype=dtype)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: could not convert string to float: np.str_('class1')
```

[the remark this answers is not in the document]
icfaust: Thank you for the example. I think there are two things that need to be done 1) there needs to be testing added for this use case for classification algorithms. 2) Remove the entire y dtype checking and the keyword. There are two reasons for this, firstly the new keyword is not used by any algorithm that I see. Secondarily, ```validate_data``` has a ```validate_separately``` keyword which will be applied to the inputs individually, and this is the proper approach if we need to explicitly set the y dtype rather than baking up our own solution.

sklearnex/svm/svc.py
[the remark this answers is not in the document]
icfaust: That is a good point @david-cortes-intel, it comes down to the definition of active vs legacy support.  Previous instructions for development where that we explicitly support for certain versions (with testing etc.), and it is possible that we support older versions but do not guarantee it.  In this case we violate that by actively remove support for legacy versions.

sklearnex/tests/test_common.py
Alexsandruss: Changed rule to check order and number of calls in https://github.com/uxlfoundation/scikit-learn-intelex/pull/2346/commits/87c8580258081232149fb7d7e4381b7df35cd1ea

sklearnex/utils/validation.py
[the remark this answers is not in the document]
Alexsandruss: > 1) there needs to be testing added for this use case for classification algorithms.

Actually, sklearn tests already check that indirectly - some testing inputs belong to the same type as the example above or other applicable types.

> 2) Remove the entire y dtype checking and the keyword.

Done.

sklearnex/svm/svc.py
```diff
@@ -147,99 +140,77 @@ def score(self, X, y, sample_weight=None):
...
-
-        def _predict_log_proba(self, X):
-            xp, _ = get_namespace(X)
-            return xp.log(self.predict_proba(X))
-
-        predict_proba.__doc__ = _sklearn_SVC.predict_proba.__doc__
+    @available_if(_sklearn_SVC._check_proba)
+    def predict_proba(self, X):
+        """
+        Compute probabilities of possible outcomes for samples in X.
```
david-cortes-intel, line 146: I realize this all was already committed before, but I think it'd be better to assign to the docstring variable from the docstring of sklearn instead of copy-pastying the whole content here. Chances are that it might change on the sklearn side and then later on become out of synch. I'm pretty sure assignment to the `@available_if` could be worked around with a simple if-else if you make the assignment after the class declaration.

Alexsandruss:
@icfaust, performance changes of this PR:
image

sklearnex/svm/svc.py
Alexsandruss: Actually, there is a simpler solution:
```python
@available_if(_sklearn_SVC._check_proba)
@wraps(_sklearn_SVC.predict_proba, assigned=["__doc__"])
def predict_proba(self, X):
```

onedal/utils/validation.py
```diff
@@ -127,6 +128,12 @@ def _validate_targets(y, class_weight, dtype):
     return np.asarray(y, dtype=dtype, order="C"), class_weight_res, classes
 
 
+def get_finite_keyword():
```
icfaust, line 131: I know why this was done because I wrote the test for sklearn_check_version, please leave a note here for future developers as to why this is the way it is.

sklearnex/ensemble/_forest.py
```diff
@@ -375,10 +366,7 @@ def _estimators_(self):
             est_i.set_params(
                 random_state=random_state_checked.randint(np.iinfo(np.int32).max)
             )
-            if sklearn_check_version("1.0"):
-                est_i.n_features_in_ = self.n_features_in_
-            else:
-                est_i.n_features_ = self.n_features_in_
+            est_i.n_features_in_ = self.n_features_in_
```
icfaust, line 369: Given the change you made at line 404, a similar change must be made here.  Please also leave a not to deprecate this functionality when sklearn 1.2 support is dropped.

sklearnex/linear_model/linear.py
```diff
@@ -60,26 +56,46 @@ class LinearRegression(PatchableEstimator, _sklearn_LinearRegression):
...
-                n_jobs=n_jobs,
-                positive=positive,
-            )
+        if sklearn_check_version("1.7"):
+
+            def __init__(
+                self,
+                fit_intercept=True,
+                copy_X=True,
+                tol=1e-06,
```
icfaust, line 65: I do not see any change in the logic associated with the addition of the ```tol``` parameter, nor any change in the ```_onedal_cpu_supported``` or ```_onedal_gpu_supported``` class methods.  Please include documentation and a code comment as to the parameter's effect on oneDAL for clarity for our users and developers respectively.

sklearnex/neighbors/knn_classification.py
```diff
@@ -44,58 +41,28 @@ class KNeighborsClassifier(KNeighborsDispatchingBase, _sklearn_KNeighborsClassif
...
-                leaf_size=leaf_size,
-                metric=metric,
-                p=p,
-                metric_params=metric_params,
-                n_jobs=n_jobs,
-            )
-
-    else:
-
-        @_deprecate_positional_args
```
icfaust, line 74: please also remove the import

sklearnex/svm/svc.py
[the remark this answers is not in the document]
icfaust: @syakov-intel @napetrov ping for comment

sklearnex/svm/svc.py
```diff
@@ -147,99 +142,27 @@ def score(self, X, y, sample_weight=None):
...
-
-    else:
-
-        @property
-        def predict_proba(self):
-            self._check_proba()
-            check_is_fitted(self)
-            return self._predict_proba
+    @available_if(_sklearn_SVC._check_proba)
+    @wraps(_sklearn_SVC.predict_proba, assigned=["__doc__"])
```
icfaust, line 146: interesting, never seen this before and is a very good addition. Please make a ticket in the backlog to address other circumstances using ```available_if``` and docs in the codebase. @david-cortes-intel how do we go about verifying that this has been done correctly?

icfaust (commented):
Here is my next review.

sklearnex/svm/svc.py
david-cortes-intel: I cannot think of any way to verify that docstrings haven't been overwritten with exact copies defined in code.

icfaust:
@Alexsandruss There are changes in this PR that aren't mentioned in the description, could you add them for posterity?

sklearnex/neighbors/knn_classification.py
Alexsandruss: Done.

sklearnex/linear_model/linear.py
Alexsandruss: `tol` is the new parameter since 1.7 version for sparse data only. No effect for oneDAL side but required to comply with original LinearRegression.__init__ signature. I added note about it.

onedal/utils/validation.py
Alexsandruss: Done.

sklearnex/svm/svc.py
Alexsandruss: It makes sense to create a class decorator for an extension estimator later which adds docstring for each patched method, not only `available_if`-decorated. I created ticket for this.

sklearnex/ensemble/_forest.py
Alexsandruss: Actually, property from line 404 can be completely removed because it's handled by stock sklearn BaseForest which is a parent of sklearnex ensemble estimators:
https://github.com/scikit-learn/scikit-learn/blob/1.1.X/sklearn/ensemble/_forest.py#L652
I deleted it in https://github.com/uxlfoundation/scikit-learn-intelex/pull/2346/commits/d496d74a44f1fd9907ff7037d229b5a89dec7181

Base tree estimator works with `n_features_in_ ` internally and `n_features_` is deprecated forwarded property starting from sklearn 1.0:
https://github.com/scikit-learn/scikit-learn/blob/1.0.X/sklearn/tree/_classes.py#L189
https://github.com/scikit-learn/scikit-learn/blob/1.0.X/sklearn/tree/_classes.py#L1027

sklearnex/svm/nusvc.py
```diff
@@ -145,99 +138,77 @@ def score(self, X, y, sample_weight=None):
...
-            self._check_proba()
-            check_is_fitted(self)
-            return self._predict_proba
-
-        def _predict_log_proba(self, X):
-            xp, _ = get_namespace(X)
-            return xp.log(self.predict_proba(X))
-
-        predict_proba.__doc__ = _sklearn_NuSVC.predict_proba.__doc__
+    @available_if(_sklearn_NuSVC._check_proba)
```
icfaust, line 141: I assume that this will be addressed in the follow-up ticket, as the changes associated with available_if wrapped method documentation is not consistent in the PR.

icfaust (approved):
In the future, deprecation of sklearn version support should be done in separate PRs, with proper involvement and oversight by multiple maintainers.  It will make alignment of relevant parties simpler, easier for developers to understand and track changes, and simplify review. I must admit it slowed my review of the code, for which I apologize. What I recommend is that these PRs be separated into multiple PRs once they approach close to final form in the case that they have multiple themes (in this case sklearn1.7 support and sklearn <1.0 removal). For the sake of expediency, I believe this is ready to merge with the expectation that the follow up tickets are addressed soon.

PR #2364, opened 2025-03-17
[bugfix, enhancement] Address affinity bug by using threadpoolctl/joblib for n_jobs dispatching

icfaust:
## Description

This addresses an issue related to linux thread control using affinity.  New tests added which validated n_jobs when a reduced number of threads are available, set by the affinity. This requires 4 threads/CPU cores be made available to the main python pytest process in order to have proper functionality, meaning that this test does not run on azure pipelines runners, but only on intelci and github actions.

This change was needed in order to have scikit-learn-intelex properly work on thread-limited Kubernetes pods.

---

**Testing**

**Performance**


daal4py/sklearn/_n_jobs_support.py
```diff
@@ -79,59 +76,42 @@ def _run_with_n_jobs(method):
...
-        old_n_threads = get_n_threads()
-        if n_jobs == old_n_threads:
-            return method(self, *args, **kwargs)
 
-        try:
+        if not self.n_jobs:
+            n_jobs = cpu_count()
+        else:
+            n_jobs = (
+                self.n_jobs if self.n_jobs > 0 else max(1, cpu_count() + self.n_jobs + 1)
```
david-cortes-intel, line 101: I think a value of zero here would differ from the behavior in `joblib` used by sklearn.

daal4py/sklearn/_n_jobs_support.py
```diff
@@ -79,59 +76,42 @@ def _run_with_n_jobs(method):
...
-            else:
-                n_jobs = max(1, n_threads + n_jobs + 1)
-        # branch with set n_jobs
-        old_n_threads = get_n_threads()
-        if n_jobs == old_n_threads:
-            return method(self, *args, **kwargs)
 
-        try:
+        if not self.n_jobs:
+            n_jobs = cpu_count()
```
david-cortes-intel, line 98: Would this later on get limited to the number of physical cores from oneDAL side?

daal4py/sklearn/_n_jobs_support.py
```diff
@@ -19,53 +19,50 @@
...
 
+    def validate_parameter_constraints(n_jobs):
+        if n_jobs is not None and n_jobs.__class__ != int:
+            raise TypeError(
+                f"n_jobs must be an instance of int, not {n_jobs.__class__.__name__}."
+            )
+
+
+class oneDALLibController(threadpoolctl.LibController):
+    user_api = "oneDAL"
```
david-cortes-intel, line 44: These could be in lower case to be more consistent with `threadpoolctl`.

daal4py/sklearn/_n_jobs_support.py
david-cortes-intel: Per their docs:
https://joblib.readthedocs.io/en/stable/generated/joblib.Parallel.html
> If n_jobs = 0 then a ValueError is raised.

icfaust: Good catch. I'll redefine the param_constraint definition for this.

icfaust: Looks like that was the wrong way to go, scikit-learn lets joblib raise an error, not via param constraints, added a comment and added an ```elif self.n_jobs == 0:```.

daal4py/sklearn/_n_jobs_support.py
icfaust: I'll be honest, I'm not 100% sure yet.  The default in the threading.h in daal will set it to the number of CPUs, but with the affinity I didn't spend the full time to track the default setting there.

sklearnex/linear_model/incremental_ridge.py
```diff
@@ -112,7 +112,7 @@ class IncrementalRidge(IntelEstimator, MultiOutputMixin, RegressorMixin, BaseEst
             "fit_intercept": ["boolean"],
             "alpha": [Interval(numbers.Real, 0, None, closed="left")],
             "copy_X": ["boolean"],
-            "n_jobs": [Interval(numbers.Integral, -1, None, closed="left"), None],
```
icfaust, line 115: Reviewers: let me know if I should remove n_jobs from this and the other estimator entirely in order to minimize maintenance cost (all other estimators get the kwarg from _n_jobs_support.py, so far that I see).

daal4py/sklearn/_n_jobs_support.py
```diff
@@ -19,53 +19,50 @@
...
+from daal4py import _get__daal_link_version__, daalinit, num_threads
 
 from ._utils import sklearn_check_version
 
 if sklearn_check_version("1.2"):
     from sklearn.utils._param_validation import validate_parameter_constraints
+else:
 
+    def validate_parameter_constraints(n_jobs):
+        if n_jobs is not None and n_jobs.__class__ != int:
```
david-cortes-intel, line 37: Joblib and sklearn do not have the constraint that n_jobs needs to be a built-in `int` type. Sklearn for example will happily accept numpy scalars:
```python
import numpy as np
X = rng.random(size=(100,3))
y = rng.random(size=X.shape[0])
from sklearn.ensemble import RandomForestRegressor
model = RandomForestRegressor(n_jobs=np.int64(2))
model.fit(X, y)
```

You could try checking if it is a subdtype of `np.integer`.

icfaust: good catch. Thanks will change.

david-cortes-intel:
CI failures are from `test_memory_usage.py`. They might be just sporadic issues, but given the numbers it shows:
```
E           AssertionError: Size of extra allocated memory after using garbage collector is greater than 15.0% of input data
E             	Algorithm: IncrementalBasicStatistics(result_options=['min', 'max', 'sum', 'mean',
E                                                        'variance', 'variation',
E                                                        'sum_squares', 'standard_deviation',
E                                                        'sum_squares_centered',
E                                                        'second_order_raw_moment'])
E             	Input data size: 808000 bytes
E             	Extra allocated memory size: 5297461 bytes / 655.63 %
E           assert 5297461 < (0.15 * 808000)
```

It looks like this PR could be making the algorithms run with more threads than they were using before. Is that intended?

daal4py/sklearn/_n_jobs_support.py
david-cortes-intel: Looks like setting the number of threads like this would not result in that number later on getting limited to the number of physical cores. How about passing argument `only_physical_cores=True` here?

david-cortes-intel: Tried adding a line to print this value here: https://github.com/uxlfoundation/oneDAL/blob/31cafec9950f1db352b639dafad5875971ca00fe/cpp/daal/src/threading/threading.cpp#L267
.. and from what I see, it is indeed set to the result of `cpu_count(False)`.

david-cortes-intel: Although from some further testing, this behavior _also_ appears to be the same in the current main branch.

daal4py/sklearn/_n_jobs_support.py
```diff
@@ -19,53 +19,50 @@
...
+class oneDALLibController(threadpoolctl.LibController):
+    user_api = "onedal"
+    internal_api = "onedal"
+
+    filename_prefixes = ("libonedal_thread", "libonedal")
+
+    def get_num_threads(self):
+        return num_threads()
+
+    def set_num_threads(self, nthreads):
```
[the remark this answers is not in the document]
david-cortes-intel: Actually on a further look, it _does_ already have a mutex on the daal side. Still better to document this behavior being global.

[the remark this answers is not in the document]
icfaust: Sounds good, will do!

sklearnex/tests/test_run_to_run_stability.py
```diff
@@ -55,9 +55,6 @@
     sklearn_clone_dict,
 )
 
-# to reproduce errors even in CI
-d4p.daalinit(nthreads=100)
```
icfaust, line 59: Removing this causes all sorts of memory leak check test failures, not just in windows and not just with pandas.

icfaust:
This last run is very useful/interesting. it shows that:
1) linux and windows
2) both GitHub actions and azure pipelines 
3) for daal4py (logistic regression) and onedal
4) many different estimators.
5) Different input datatypes

im very worried about our malloc routines in daal. Thoughts @david-cortes-intel @Vika-F

david-cortes-intel:
> This last run is very useful/interesting. it shows that:
> 
> 1. linux and windows
> 2. both GitHub actions and azure pipelines
> 3. for daal4py (logistic regression) and onedal
> 4. many different estimators.
> 5. Different input datatypes
> 
> im very worried about our malloc routines in daal. Thoughts @david-cortes-intel @Vika-F

Perhaps there could be memory leaks, but I'm not sure that we can conclude anything from these tests alone.

I see that some are for cases where the input is from DPCTL - does it set flags like `SYCL_PI_LEVEL_ZERO_DISABLE_USM_ALLOCATOR` for example? Might also be a better idea to set `PYTHONMALLOC=malloc`, and maybe `LD_PRELOAD=jemalloc.so` on linux, given the way in which the test is structured.

But what'd be even better is to run them through valgrind and asan to see if there are any reports of leaks coming specifically from oneDAL or sklearnex. We unfortunately get a lot of false positives from numpy, scipy, pybind11, and others that make it hard to browse logs; but true leaks should definitely show up there.

Alexsandruss:
@Mergifyio update

icfaust:
@Alexsandruss I updated, let me know if there is anything I can do to help.

sklearnex/tests/test_memory_usage.py
```diff
@@ -114,7 +114,7 @@ def gen_functions(functions):
 
 data_shapes = [
     pytest.param((1000, 100), id="(1000, 100)"),
-    pytest.param((2000, 50), id="(2000, 50)"),
+    pytest.param((2000, 40), id="(2000, 40)"),
```
Alexsandruss, line 117: This change alone makes memory usage test failures. @icfaust, what is the reason for it?

icfaust: My goal was to see if I could tweak the parameters to make it either completely disappear or be consistent, giving a way to properly address the issue. Is this something you can now reproduce locally?

Alexsandruss: Test failures disappear locally if data shape change is reverted. Not sure if it will be the same in CI.
Lower data shapes are more likely to cause failures because base memory usage takes larger percentage of data size.

syakov-intel:
GitHub currently reports this branch as CONFLICTING with main. This work is directly relevant to current high-core Random Forest testing, so please rebase it onto current main before further validation. After rebasing, please rerun the affinity and memory suites and assert that the effective process CPU affinity exactly matches the requested CPU set; combining taskset and numactl in the wrong order can silently widen affinity to SMT siblings.

PR #2376, opened 2025-03-24
DOC: Correct information about MPI requirements and fix example in daal4py

david-cortes-intel:
## Description

The docs for SPMD mode of daal4py have some issues which this PR fixes:
* They mention MPI not being required, but this is incorrect - SPMD mode only works through MPI (e.g. no dask). I guess it meant to say that "mpi4py" is not required given the contents of the next paragraphs, which is more reasonable.
* The code snippet it gives is incorrect - e.g. it doesn't add the process ID to the file name from where it loads things.
* It doesn't mention the requirements on MPI software.

Note that this change unfortunately won't reach the deployed docs under github pages for the old domain.

---

**Testing**

**Performance**

Not applicable.

### DIFF
--- a/doc/daal4py/scaling.rst
+++ b/doc/daal4py/scaling.rst
@@ -23,10 +23,7 @@ Scaling on Distributed Memory (Multiprocessing)
 It's Easy
 ---------
 daal4py operates in SPMD style (Single Program Multiple Data), which means your
-program is executed on several processes (e.g. similar to MPI).  The use of MPI is
-not required for daal4py's SPMD-mode to work, all necessary communication and
-synchronization happens under the hood of daal4py. It is possible to use daal4py and
-mpi4py in the same program, though.
+program is executed on several processes, using MPI, optionally aided by ``mpi4py``.
 
 Only very minimal changes are needed to your daal4py code to allow daal4py to
 run on a cluster of workstations. Initialize the distribution engine::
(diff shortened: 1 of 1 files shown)


doc/daal4py/scaling.rst
```diff
@@ -54,19 +53,18 @@ Finally stop the distribution engine::
...
   daalfini()
 
 To actually get it executed on several processes use standard MPI mechanics,
 like::
 
   mpirun -n 4 python ./kmeans.py
 
-The binaries provided by Intel use the Intel® MPI library, but
-daal4py can also be compiled for any other MPI implementation.
+.. important:: SPMD mode will only work with the same MPI library with which ``daal4py`` was compiled. PyPI and conda distributions of ``daal4py`` both are built with Intel's MPI as backend (package ``impi_rt``), and will thus not work under other MPI libraries such as OpenMPI. Using SPMD mode with OpenMPI requires building ``daal4py`` from source with that library as backend. Same requirement applies to ``mpi4py`` (must be built with the same backend as ``daal4py``) if using it for SMPD mode. Note that using an incompatible MPI library will not result in an explicit error, but will rather result in all processes running separately as the first rank without any communication, thereby producing incorrect results.
```
icfaust, line 67:
```suggestion
.. important:: SPMD mode will only work with the same MPI library with which ``daal4py`` was compiled. PyPI and conda distributions of ``daal4py`` both are built with Intel's MPI as backend (package ``impi_rt``), and will thus not work under other MPI libraries such as OpenMPI. Using SPMD mode with OpenMPI requires building ``daal4py`` from source with that library as backend. The same requirement applies to ``mpi4py`` (must be built with the same backend as ``daal4py``) if using it for SMPD mode. Note that using an incompatible MPI library will not result in an explicit error, but will rather result in all processes running separately as the first rank without any communication, thereby producing incorrect results.
```

icfaust (approved):
small correction, otherwise good to go.

PR #2386, opened 2025-03-27
FIX: Fix incorrect scale of base_score being used for XGB regression objectives

david-cortes-intel:
## Description

This PR fixes an issue with the daal4py model builders using the intercepts from XGBoost models incorrectly when it comes to regression models that have a link function, such as Poisson or Gamma.

Compared to the other GBT libraries supported by model builders, XGBoost has a 'base_score' field which is in the scale of the response variable (i.e. inverse link function is applied to it), but the model builders assume that the intercept is on the link scale, so predictions will be wrong when there is a non-identity link function.

Along the way, it also fixes an issue of the intercept not being applied to objective `binary:logitraw` due to this same issue, which was previously being approach through workarounds that remove the intercept.

---

**Testing**

**Performance**

Not applicable.


Alexsandruss (approved):

razdoburdin (approved):
Great!

PR #2399, opened 2025-04-02
FIX: Respect user-supplied `$CXX`

david-cortes-intel:
## Description

Currently, the logic in the setup script propagates build-related environment variables correctly for the `daal4py` module which is built through `build_ext`, but it overrides those for the `onedal` module which is built through cmake.

Similar to https://github.com/uxlfoundation/scikit-learn-intelex/pull/2397 , it uses a flawed logic that checks for exact string matches in order to determine if it's using ICX, which would break for example if one tries something like the following:
```
export CXX="icpx -Wl,-rpath,..."
export CXX="ccache icpx"
```

This PR fixes it so that it doesn't override user-supplied compiler commands when not strictly needed.

---

**Testing**

**Performance**

Not applicable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -432,10 +432,8 @@ class onedal_build:
         elif n_threads is True:
             n_threads = os.cpu_count() or 1
 
-        cxx = os.getenv("CXX", "cl" if IS_WIN else "g++")
         build_onedal = lambda iface: build_backend.custom_build_cmake_clib(
             iface=iface,
-            cxx=cxx,
             onedal_major_binary_version=ONEDAL_MAJOR_BINARY_VERSION,
             no_dist=no_dist,
             use_parameters_lib=use_parameters_lib,
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -215,7 +215,9 @@ without reinstalling the package after a small change.
    python setup.py build
    ```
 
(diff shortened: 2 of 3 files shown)


david-cortes-intel:
Not sure what exactly the reported CI failure is. Only thing I can find is a warning about BasicStatistics.

setup.py
```diff
@@ -432,10 +432,8 @@ def onedal_run(self):
         elif n_threads is True:
             n_threads = os.cpu_count() or 1
 
-        cxx = os.getenv("CXX", "cl" if IS_WIN else "g++")
```
icfaust, line 435: Not that I agreed with this code in the first place (this is a good change), by removing it, how are the defaults decided? It will use the cmake default list, but does it prioritize cl over g++ on windows when both are available? (Not that this is a big thing, but it is a change in default state possibly and worth mentioning in the description)

icfaust (approved):
Just a small question/ possible addition to the description about changes in how the default compiler is chosen on host builds.

icfaust:
private CI issues unrelated

setup.py
david-cortes-intel: It leaves the decision to CMake. AFAIK CMake prefers MSVC on windows. And yes, it could potentially lead to compatibility problems due to their ABI for structs if MinGW is used together with ICX, but mixing them together like that without setting additional flags would be a misusage on the part of the user.

PR #2409, opened 2025-04-07
[enhancement] add ```__version__``` attribute to ```onedal``` and ```sklearnex```

icfaust:
## Description

A setup.py install sets a version in pip but does not include that information in the ```__version__``` attribute of the various packages. This adds a small script in the run step which accesses the copied source files in the build directory and modifies the hardcoded default version.  The default version is "2199.9.9" which matches oneDAL convention. If no intermediate build directory exists, this step will be skipped (possible in ```setup.py develop```).

---

**Testing**

**Performance**

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -422,6 +422,16 @@ class onedal_build:
         self.onedal_run()
         super(onedal_build, self).run()
         self.onedal_post_build()
+        if hasattr(self, "build_lib"):
+            # swap out __version__ before install
+            for p in ["onedal", "sklearnex"]:
+                loc = os.sep.join((self.build_lib, p, "__init__.py"))
+                if os.path.isfile(loc):
+                    with open(loc, "r+") as f:
+                        data = f.read().replace("2199.9.9", sklearnex_version)
+                        f.seek(0)
+                        f.write(data)
+                        f.truncate()
 
     def onedal_run(self):
         n_threads = self.parallel
--- a/onedal/__init__.py
+++ b/onedal/__init__.py
@@ -121,3 +121,6 @@ if _spmd_backend is not None:
         ]
     if daal_check_version((2023, "P", 200)):
         __all__ += ["spmd.cluster"]
+
+__version__ = "2199.9.9"
+
--- a/sklearnex/__init__.py
+++ b/sklearnex/__init__.py
@@ -52,6 +52,7 @@ __all__ = [
     "unpatch_sklearn",
     "utils",
 ]
+__version__ = "2199.9.9"
 onedal_iface_flag = os.environ.get("OFF_ONEDAL_IFACE", "0")
 if onedal_iface_flag == "0":
     from onedal import _spmd_backend


david-cortes-intel:
@icfaust I think this change would result in problems once one tries to build the extension locally, as it will overwrite files. How about adding these as new files that would get overwritten entirely, and then adding them to .gitignore?

icfaust:
> @icfaust I think this change would result in problems once one tries to build the extension locally, as it will overwrite files. How about adding these as new files that would get overwritten entirely, and then adding them to .gitignore?

It will change the value in the build directory, not in the repo.

icfaust:
> Also, there's `importlib.metadata.version` which perhaps could be an alternative, but that doesn't work when building the extensions locally.

True, you can see the discuss about that here. It comes at the cost of increasing load times. https://discuss.python.org/t/please-make-package-version-go-away/58501

david-cortes-intel:
> > @icfaust I think this change would result in problems once one tries to build the extension locally, as it will overwrite files. How about adding these as new files that would get overwritten entirely, and then adding them to .gitignore?
> 
> It will change the value in the build directory, not in the repo, changes caused by a develop build are unchanged because of the initial if statement (can look at the setuptools class defintion for more details).

... unless you run some command like `setup.py build` or `setup.py build_ext`, in which case it will modify them in the repository.

icfaust:
> > > @icfaust I think this change would result in problems once one tries to build the extension locally, as it will overwrite files. How about adding these as new files that would get overwritten entirely, and then adding them to .gitignore?
> > 
> > 
> > It will change the value in the build directory, not in the repo, changes caused by a develop build are unchanged because of the initial if statement (can look at the setuptools class defintion for more details).
> 
> ... unless you run some command like `setup.py build` or `setup.py build_ext`, in which case it will modify them in the repository.

https://github.com/pypa/distutils/blob/main/distutils/command/build.py#L100

david-cortes-intel:
Thanks, can confirm that `build` and `build_ext` did not overwrite things in the repository.

david-cortes-intel (approved):

setup.py
```diff
@@ -422,6 +422,16 @@ def run(self):
         self.onedal_run()
         super(onedal_build, self).run()
         self.onedal_post_build()
+        if hasattr(self, "build_lib"):
+            # swap out __version__ before install
+            for p in ["onedal", "sklearnex"]:
+                loc = os.sep.join((self.build_lib, p, "__init__.py"))
+                if os.path.isfile(loc):
+                    with open(loc, "r+") as f:
+                        data = f.read().replace("2199.9.9", sklearnex_version)
```
Alexsandruss, line 431: Does it assign `%Y%m%d.%H%M%S` to `__version__` for local non-release build?
If so, then it might be better to use something like `%Y.%m.%d+%commit_hash` as `__version__` and package version, for example: `2024.5.3+a1b2c3d`.

icfaust: Will be handled in #2413 in order to limit scope creep

PR #2414, opened 2025-04-08
[CI] upgrade dpctl/dpnp versions

icfaust:
## Description

This upgrades the version of dpnp and dpctl based off the recent release. This will notably add array_api support to dpnp.

This PR does the following:
- Readd Python 3.12 to conda environment linux dpctl/dpnp testing
- Makes clear the installation versions of dpctl, dpnp and the dpcpp runtime.
- Adds renovatebot hooks to prevent this circumstance from occurring again
- Update the DPC++ compiler to 2025.2
- Adds a special wrapper (```support_sycl_format```) to move sycl data to cpu before ```validate_data``` is called, as the ```__array__``` attribute has become blocked
- Upgrade dpctl to 0.20.1 and dpnp to 0.18.1
- Remove the special pytorch XPU install as the compiler upgrade breaks compatibility with it.
- simplify the conversion of oneDAL tables to sycl_usm datatypes, changes in dpnp prevent easy access of zero copy support.
- Fix lazy import ```to_dpnp``` function to use now that the previous method has become blocked for data zero copy of dpnp data, use it instead.
- Fix mistake in array_to_usm recursion
- Prevent return conversion of sparse datatypes, which occurs more often in lower priority estimator methods like ```kneighbors_graph```
- Add a test to make sure that dpctl and dpnp do not get silently removed from testing (which was the case before this update) making data framework testing intentional, even though data framework support is passive in the codebase
- Wrap all inherited methods with ```support_input_format``` or ```support_sycl_format``` in order to guarantee sycl support with and without array api support (because of the blocking of np.array)
- Move an incorrectly implemented validate_data call from ```EmpiricalCovariance.fit``` which should occur after dispatching.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -463,7 +463,7 @@ jobs:
       - name: Install PyTorch
         if: contains(matrix.FRAMEWORKS, 'pytorch')
         run: |
-          pip install torch-cpu
+          pip install torch index-url https://download.pytorch.org/whl/cpu
           python -c "import torch; _=[print(torch.xpu.get_device_name(i)) for i in range(torch.xpu.device_count())]" 
       - name: Install daal4py/sklearnex
         run: pip install *.whl
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -463,7 +463,7 @@ jobs:
       - name: Install PyTorch
         if: contains(matrix.FRAMEWORKS, 'pytorch')
         run: |
-          pip install torch --index-url https://download.pytorch.org/whl/xpu
+          pip install torch
           python -c "import torch; _=[print(torch.xpu.get_device_name(i)) for i in range(torch.xpu.device_count())]" 
       - name: Install daal4py/sklearnex
         run: pip install *.whl
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  DPCTL_VERSION: 0.20.2
+  DPCTL_VERSION: 0.20.1
   DPNP_VERSION: 0.18.1
   DPCTL_PY_VERSIONS: '3.9\|3.11'
   UXL_PYTHONVERSION: "3.12"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -463,7 +463,7 @@ jobs:
       - name: Install PyTorch
         if: contains(matrix.FRAMEWORKS, 'pytorch')
         run: |
-          pip install torch
+          pip install torch-cpu
           python -c "import torch; _=[print(torch.xpu.get_device_name(i)) for i in range(torch.xpu.device_count())]" 
       - name: Install daal4py/sklearnex
         run: pip install *.whl
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -463,7 +463,7 @@ jobs:
       - name: Install PyTorch
         if: contains(matrix.FRAMEWORKS, 'pytorch')
         run: |
-          pip install torch index-url https://download.pytorch.org/whl/cpu
+          pip install torch --index-url https://download.pytorch.org/whl/cpu
           python -c "import torch; _=[print(torch.xpu.get_device_name(i)) for i in range(torch.xpu.device_count())]" 
       - name: Install daal4py/sklearnex
         run: pip install *.whl
--- a/onedal/_device_offload.py
+++ b/onedal/_device_offload.py
@@ -173,32 +173,21 @@ def support_sycl_format(func):
     # This wrapper enables scikit-learn functions and methods to work with
     # all sycl data frameworks as they no longer support numpy implicit
     # conversion and must be manually converted. This is only necessary
-    # when array API support is not active, meaning unlike wrap_output_data
-    # and support_input_format, return values are matching sklearn (sparse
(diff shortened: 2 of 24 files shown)


icfaust:
This PR must wait until 2025.1 is available on apt for Linux. Then corrections to the codebase must be completed

onedal/_device_offload.py
```diff
@@ -171,3 +171,36 @@ def wrapper_impl(*args, **kwargs):
         return result
 
     return wrapper_impl
+
+
+def support_sycl_format(func):
+    # This wrapper enables scikit-learn functions and methods to work with
+    # all sycl data frameworks as they no longer support numpy implicit
+    # conversion and must be manually converted. This is only necessary
+    # when array API support is not active, meaning unlike wrap_output_data
```
david-cortes-intel: Does it mean that if sklearnex gets passed a GPU array and array API is not active, it will end up moving the array to CPU and perform the compute on CPU? Does this get used outside of tests?

icfaust: Yep, but this was already the case for inherited methods, so no regression. Some methods simply are too difficult to develop GPU enabled versions of : https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestRegressor.html#sklearn.ensemble.RandomForestRegressor.decision_path  Note that this only occurs when array API dispatch is disabled, when enabled either the method works, or fails as would be the case for the equivalent sklearn call.

david-cortes-intel: Could it instead raise an informative error, or at least give a warning when that forced data movement happens?

icfaust: We force move data to CPU for checks in all cases except for dispatched methods using ```use_raw_inputs``` or for array API supporting algorithms with ```array_api_dispatch``` enabled.  At the moment it can be assumed that forced data movement is standard, and can be made the exception at the completion of the array API rollout.

onedal/_device_offload.py
```diff
@@ -62,23 +62,23 @@ def wrapper(self, *args, **kwargs):
...
-            has_host_data = True
-        else:
-            has_host_data = True
-
-        mismatch_host_item = usm_iface is None and item is not None and has_usm_data
-        mismatch_usm_item = usm_iface is not None and has_host_data
 
-        if mismatch_host_item or mismatch_usm_item:
+        # set has_usm_data to boolean and use xor to see if they don't match
+        if (has_usm_data := bool(has_usm_data)) ^ usm_iface:
```
david-cortes-intel: Could use xor operator named more clearly like that: https://docs.python.org/3/library/operator.html#operator.xor

onedal/datatypes/_data_conversion.py
```diff
@@ -96,11 +96,19 @@ def return_type_constructor(array):
...
         xp = array.__array_namespace__()
-        func = lambda x: (
-            xp.asarray(x)
-            if hasattr(x, "__sycl_usm_array_interface__")
-            else xp.asarray(backend.from_table(x), device=device)
-        )
+        if is_dpctl_tensor(array):
+            func = lambda x: (
+                xp.asarray(x)
+                if hasattr(x, "__sycl_usm_array_interface__")
```
david-cortes-intel, line 104: Is this something that was introduced in newer dpctl versions? Would be helpful to leave a comment about when it can be expected to be there and when not to.

.github/workflows/ci.yml
```diff
@@ -463,7 +463,7 @@ jobs:
       - name: Install PyTorch
         if: contains(matrix.FRAMEWORKS, 'pytorch')
         run: |
-          pip install torch
+          pip install torch --index-url https://download.pytorch.org/whl/cpu
```
david-cortes-intel, line 466: Would the intel version come with support for sycl on CPU? Perhaps that could be a more comprehensive alternative for tests.

icfaust: It does not, we would also run into the shared library issue there as well.

onedal/datatypes/_data_conversion.py
icfaust: added a comment

onedal/_device_offload.py
icfaust: swapped to using operator

sklearnex/cluster/k_means.py
```diff
@@ -334,6 +334,7 @@ def _check_test_data(self, X):
             return X
 
         _transform = support_input_format(_sklearn_KMeans._transform)
+        transform = support_input_format(_sklearn_KMeans.transform)
```
david-cortes-intel, line 337: Was this meant to replace the line before?

icfaust: Unfortunately no ☹️ . This is an ugly solution based on how things have been implemented before this, KMeans.transform has its validate_data in a separate function, and then calls a function KMeans._transform .  It was done this way on their side to do ```fit_transform```, allowing for the second valdiation to be removed. We reuse ```_transform``` in our ```fit_transform``` implementation too, so I can't remove it there, but I need make ```transform``` work. I couldn't swap in a dispatch because we don't accelerate anything there with oneDAL in either case.  Direct use of KMeans.transform using dpctl 0.18 is likely to be extremely slow.  In general this is because its not a high-prority method, which we *technically* support but in a very bad way/ without benchmarking.

david-cortes-intel: Perhaps that could be left as a comment for when it comes the time to refactor again.

david-cortes-intel (approved):
LGTM, pending CI results.

PR #2419, opened 2025-04-11
FIX, MAINT: Migrate model builders tests to pytest + bugfixes

david-cortes-intel:
## Description

This PR migrates the model builders tests to pytest, increasing test coverage by using the combinatorial parameterizations offered by pytest. Despite the increase in number of tests being conducted, the running time is severely shortened thanks to only smaller problems being executed now.

Along the way, some bugs were uncovered with the increased test coverage which this PR is fixing, along with changing a bit the logic for SHAP support based on the results from the tests.

There still remain some strange cases which I wasn't sure what to do about - for example, when catboost models have no intercept, the first node will have a cover of exactly '0', which results in daal4py outputting NaNs, while catboost itself is still able to calculate shap values there.

---

**Testing**

**Performance**

Not applicable.


daal4py/mb/gbt_convertors.py
```diff
@@ -437,9 +439,9 @@ def get_gbt_model_from_lightgbm(model: Any, booster=None) -> Any:
     is_regression = False
     objective_fun = booster["objective"]
     if n_classes > 2:
-        if "multiclass" not in objective_fun:
+        if ("ova" in objective_fun) or ("ovr" in objective_fun):
             raise TypeError(
-                "multiclass (softmax) objective is only supported for multiclass classification"
+                "Only multiclass (softmax) objective is supported for multiclass classification"
```
ahuber21, line 444: Isn't checking for `multiclass` more aligned with the error message?

daal4py/mb/gbt_convertors.py
```diff
@@ -504,11 +513,6 @@ def get_gbt_model_from_xgboost(booster: Any, xgb_config=None) -> Any:
         if objective_fun == "binary:logitraw":
             # daal4py always applies a sigmoid for pred_proba, wheres XGBoost
             # returns raw predictions with logitraw
-            warn(
```
ahuber21, line 507: Why remove this?

daal4py/mb/gbt_convertors.py
```diff
@@ -567,6 +571,22 @@ def get_gbt_model_from_catboost(booster: Any) -> Any:
             "Categorical features are not supported in daal4py Gradient Boosting Trees"
         )
 
+    objective = booster.get_params().get("objective")
```
ahuber21, line 574:
```suggestion
    objective = booster.get_params().get("objective", "")
```

Otherwise the checks below might fail, no?

```bash
>>> "A" in None
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'NoneType' is not iterable
```

daal4py/mb/gbt_convertors.py
```diff
@@ -587,21 +607,37 @@ def get_gbt_model_from_catboost(booster: Any) -> Any:
                     {"feature_index": feature["feature_index"], "value": feature_border}
                 )
 
+    # Note: catboost models might have a 'bias' (intercept) which gets added
```
ahuber21, line 610: Great work!

daal4py/mb/gbt_convertors.py
```diff
@@ -840,16 +907,16 @@ def get_gbt_model_from_catboost(booster: Any) -> Any:
                     # Tree has only one node
                     mb.add_leaf(
                         tree_id=cur_tree_id,
-                        response=root_node.value[class_label],
+                        response=root_node.value[
+                            class_label
+                        ],  # <- root node already has scale and bias added
```
ahuber21, line 912: Sorry for being naggy, but can we put the comment in a separate line?

tests/test_model_builders.py
```diff
@@ -68,7 +72,7 @@
 # See https://github.com/catboost/catboost/issues/2556
 # Disable SHAP tests temporarily until it's next major version
 if cb_available:
-    catboost_skip_shap = tuple(map(int, cb.__version__.split("."))) < (1, 3, 0)
+    catboost_skip_shap = tuple(map(int, cb.__version__.split("."))) < (1, 4, 0)
```
ahuber21, line 75: Did you just bump that to the latest release? Reading this comment https://github.com/catboost/catboost/issues/2136#issuecomment-1914217777 I doubt there will be a fix and we rather have to add a ticket to our backlog that investigates using the default `predict()`.
Maybe update the logic to `XFAIL`?

ahuber21 (changes requested):
I didn't dive into all the changes in the test file because the diff is so hard to read, but given that you didn't reduce the number of tests and that all test categories are still there I trust it's correct. Let me know if you think anything in there needs closer inspection.

daal4py/mb/gbt_convertors.py
david-cortes-intel: Because now there's documentation that explains what happens with this objective. Also because previously the intercept wasn't being applied and that info was part of the warning, but that has been solved by now, so what's left to 'warn' about in my opinion doesn't feel like it should be a warning.

tests/test_model_builders.py
david-cortes-intel: No, I pushed it further to an unreleased version, precisely because it looks like the next minor release (1.3) will not contain a fix for it.

daal4py/mb/gbt_convertors.py
david-cortes-intel: Perhaps, but this not part of the publicly exported interface so all the logic for deciding when to interpret is as multi-class classification is internal, and the new error message is more reflective of what's being checked for as a whole.

daal4py/mb/gbt_convertors.py
david-cortes-intel: Moved to a separate line.

razdoburdin (approved):
lgtm

tests/test_model_builders.py
ahuber21: Okay, let's leave it at that. I will look into it.

ahuber21 (approved):

PR #2428, opened 2025-04-14
Fix multiclass SVM example failures on AVX2

Vika-F:
## Description

daal4py multiclass SVM example is modified to use float64 linear kernel to fix the failure that sporadically appears in oneDAL CI on AVX2 machines:
https://dev.azure.com/daal/DAAL/_build/results?buildId=45307&view=logs&j=a8904dfb-edc6-5d1f-0634-80e20164e44d&t=43785ab5-fe0c-5af6-408a-03e67dcd3abc&l=13649

---

**Testing**

**Performance**

not applicable.

### DIFF
--- a/examples/daal4py/svm_multiclass.py
+++ b/examples/daal4py/svm_multiclass.py
@@ -34,11 +34,12 @@ def main(readcsv=pd_read_csv):
     train_file = data_path / "svm_multi_class_train_dense.csv"
     train_data = readcsv(train_file, range(nFeatures))
     train_labels = readcsv(train_file, range(nFeatures, nFeatures + 1))
+    lin_kernel = d4p.kernel_function_linear()
 
     # Create and configure algorithm object
     algorithm = d4p.multi_class_classifier_training(
         nClasses=nClasses,
-        training=d4p.svm_training(method="thunder"),
+        training=d4p.svm_training(method="thunder", kernel=lin_kernel),
         prediction=d4p.svm_prediction(),
     )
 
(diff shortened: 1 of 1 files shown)


examples/daal4py/svm_multiclass.py
```diff
@@ -34,11 +34,12 @@ def main(readcsv=pd_read_csv):
     train_file = data_path / "svm_multi_class_train_dense.csv"
     train_data = readcsv(train_file, range(nFeatures))
     train_labels = readcsv(train_file, range(nFeatures, nFeatures + 1))
+    lin_kernel = d4p.kernel_function_linear()
 
     # Create and configure algorithm object
     algorithm = d4p.multi_class_classifier_training(
         nClasses=nClasses,
-        training=d4p.svm_training(method="thunder"),
+        training=d4p.svm_training(method="thunder", kernel=lin_kernel),
```
david-cortes-intel, line 42: From a look at the paper pointed in the oneDAL docs, isn't this method 'thunder' meant for non-linear kernels?

Vika-F: Both methods implemented in DAAL work with any types of kernels.

They both also implement what is called "kernel SVM" that uses dual form of the optimization task to find the support vectors; but this algorithm can be used with linear kernel as well.

david-cortes-intel: Wouldn't that be much more inefficient compared to primal methods specifically designed for linear models exploiting subgradients and similar? How about not passing the 'method' argument in that case?

Vika-F: We do not have what is called "linear SVM" implemented at all in oneDAL. Another method is SMO method, similar to libSVM, it is even less efficient on multi-core CPUs than "thunder" in most of the cases.

Vika-F: The only change I made here is I've changed the default float32 linear kernel to float64 linear kernel to pass the accuracy test that is failing in CI now on AVX2.

Alexsandruss (approved):
Assuming there is already a ticket for failures due to AVX2/AVX512 differences

PR #2430, opened 2025-04-16
[enhancement] add ```__dlpack__``` and ```__dlpack_device__``` attributes to python oneDAL table class

icfaust:
## Description

I missed something with respect to array_api support in scikit-learn, namely that the return types should match for stored attributed when ```array_api_dispatch``` is enabled.  This requires that the returned tables be convertible to array_api arrays rather than numpy arrays, which would be impossible for gpu inputs without dpctl at the current moment (requiring some significant complexity).  This enables the ```__dlpack__``` and ```__dlpack_device__``` attributes which allow for python oneDAL tables to be converted into array_api inputs via the ```from_dlpack``` method (https://data-apis.org/array-api/2023.12/API_specification/generated/array_api.from_dlpack.html). A follow-up PR will use this new C++ infrastructure to change the backend function ```from_table``` to directly return the table into the proper array type (and will be the next-step PR).  This is necessary for true array_api support in matching sklearn.

Additional tests must be written for enabled algorithms to verify matching return types and stored attributes to the submitted data.

These attributes aren't used in any workflows (yet), no performance benchmarks necessary

The following changes occur in this PR:
 - Add ```__dlpack__``` attribute to pybind11 oneDAL table interface for generating a capsule of a DLManagedTensor (Not versioned because of low support in other frameworks (see pytorch source code))
 - Add ```__dlpack_device__``` attribute to pybind11 oneDAL table interface (requirement for dpctl's ```from_dlpack```)
 - Add DLPack enums (cpu, oneAPI) to pybind11 backends to interface with other frameworks

---

**Testing**

**Performance**


onedal/datatypes/tests/test_data.py
```diff
@@ -574,3 +574,40 @@ def test_table_conversions_dlpack(dataframe, queue, order, data_shape, dtype):
...
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+
+    X_table = to_table(X_df)
+    if hasattr(X_df, "__dlpack_device__"):
+        assert X_df.__dlpack_device__() == X_table.__dlpack_device__()
+
+    if xp := getattr(X_df, "__array_namespace__", lambda: None)():
+        X_out = xp.from_dlpack(X_table)
+        X_temp = xp.asnumpy(X_out) if hasattr(xp, "asnumpy") else np.asarray(X)
+        assert_allclose(np.squeeze(X_temp), np.squeeze(X))
```
david-cortes-intel, line 603: Should't this test for exact equality if there's no dtype conversion?

onedal/datatypes/tests/test_data.py
```diff
@@ -574,3 +574,40 @@ def test_table_conversions_dlpack(dataframe, queue, order, data_shape, dtype):
...
+
+    X_df = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
+
+    X_table = to_table(X_df)
+    if hasattr(X_df, "__dlpack_device__"):
+        assert X_df.__dlpack_device__() == X_table.__dlpack_device__()
+
+    if xp := getattr(X_df, "__array_namespace__", lambda: None)():
+        X_out = xp.from_dlpack(X_table)
+        X_temp = xp.asnumpy(X_out) if hasattr(xp, "asnumpy") else np.asarray(X)
```
david-cortes-intel, line 602: Per the docstring at the top of this test, shouldn't it always check that `np.asarray` works?

onedal/datatypes/tests/test_data.py
```diff
@@ -574,3 +574,40 @@ def test_table_conversions_dlpack(dataframe, queue, order, data_shape, dtype):
...
+        assert_allclose(np.squeeze(X_temp), np.squeeze(X))
+    else:
+        # only some numpy versions support array_api and from_dlpack
+        pytest.skip(f"{dataframe} does not have an __array_namespace__ attribute")
+
+    # test capsule deletion, should have no impact on underlying memory
+    # important for testing ``dlpack::free_capsule``
+    capsule = X_table.__dlpack__()
+    assert_allclose(np.squeeze(from_table(X_table)), np.squeeze(X))
+    del capsule
```
david-cortes-intel, line 612: I think there's no guarantee that the deletor will be called right at this point. Here you could try to forcibly trigger garbage collection:
```python
import gc
gc.collect()
```

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -154,6 +154,105 @@ dal::table convert_to_table(py::object obj, py::object q_obj, bool recursed) {
     return res;
 }
 
+DLDevice get_dlpack_device(const dal::array<byte_t>& array) {
+    DLDevice device;
+#ifdef ONEDAL_DATA_PARALLEL
+    // std::optional<sycl::queue>
```
david-cortes-intel, line 160: Forgot to remove this?

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -154,6 +154,105 @@ dal::table convert_to_table(py::object obj, py::object q_obj, bool recursed) {
...
+    tensor.strides = &tensor.shape[2];
+    tensor.byte_offset = std::uint64_t(0);
+
+    return tensor;
+}
+
+static void free_capsule(PyObject* cap) {
+    DLManagedTensor* dlm = nullptr;
+    if (PyCapsule_IsValid(cap, "dltensor")) {
+        dlm = reinterpret_cast<DLManagedTensor*>(PyCapsule_GetPointer(cap, "dltensor"));
```
david-cortes-intel, line 214: I think this should be a static cast.

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -154,6 +154,105 @@ dal::table convert_to_table(py::object obj, py::object q_obj, bool recursed) {
...
+        auto stored_array = static_cast<dal::array<byte_t>*>(self->manager_ctx);
+        if (stored_array) {
+            delete stored_array;
+        }
+        delete[] self->dl_tensor.shape;
+        delete self;
+    };
+
+    // create capsule
+    py::capsule capsule(reinterpret_cast<void*>(dlm), "dltensor", free_capsule);
```
david-cortes-intel, line 252: Also static cast.

onedal/datatypes/tests/test_data.py
icfaust: Sorry, don't understand fully.  The ```__dlpack__``` table attribute is used on line 601, but it needs to be converted to numpy from these other frameworks for value testing with ```assert_allclose```. Line 602 is necessary for conversion from dpctl/dpnp to numpy, as np.asarray on dpnp/dpctl inputs will make very weird numpy object arrays rather than numeric ones.

onedal/datatypes/tests/test_data.py
icfaust: At first glance I agree with your assessment, the only problem is this test on reduced precision hardware (i.e. fp32-only support), where a downconversion may occur from fp64 and then it wouldn't be an exact comparison.  If you think it should if statement for that case let me know, and I'll do it.

onedal/datatypes/tests/test_data.py
icfaust: You are right, the only thing that let me know it was working was changes in the coverage afterward.  I'll do your suggestion as it is more guaranteed.

onedal/datatypes/tests/test_data.py
david-cortes-intel: Then how about checking for exact equality when dtypes are the same, and approximate when not? Or how about in the latter case downcasting the numpy array and checking for exact equality on that?

onedal/datatypes/tests/test_data.py
david-cortes-intel: Wouldn't that kind of behavior on their part (producing non-numeric numpy arrays) be non-compliant with the array API specification?

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: removed

icfaust:
@david-cortes-intel thanks for helping on this PR. let me know if its good to go, or if you'd like another reviewer to double check things.

david-cortes-intel (approved):

PR #2437, opened 2025-04-17
[CI] fix LinuxCondaEnv builds (install dpcpp-cpp-rt with testing requirements)

icfaust:
## Description

The dpcpp-cpp-rt is needed to get the code functional for testing, but is not required for building. Install it at the proper step, which will fix the missing library issue (libsvml.so)

---

**Testing**

**Performance**

### DIFF
--- a/.ci/pipeline/build-and-test-lnx.yml
+++ b/.ci/pipeline/build-and-test-lnx.yml
@@ -55,6 +55,7 @@ steps:
       . /usr/share/miniconda/etc/profile.d/conda.sh
       conda activate CB
       bash .ci/scripts/setup_sklearn.sh $(SKLEARN_VERSION)
+      conda install -q -y -c conda-forge dpcpp-cpp-rt=2025.0
       pip install --upgrade -r requirements-test.txt
       pip install $(python .ci/scripts/get_compatible_scipy_version.py)
       # dpep installation is set to pypi to avoid conflict of numpy versions from pip and conda


.ci/pipeline/build-and-test-lnx.yml
```diff
@@ -55,6 +55,7 @@ steps:
       . /usr/share/miniconda/etc/profile.d/conda.sh
       conda activate CB
       bash .ci/scripts/setup_sklearn.sh $(SKLEARN_VERSION)
+      conda install -q -y -c conda-forge dpcpp-cpp-rt=2025.0
```
david-cortes-intel, line 58: Why does it need to be versioned?

icfaust: Very good question, it was previously always pinned (and thought it was best practice). Thoughts @Alexsandruss ? I have no problem either way.

Alexsandruss: dpcpp releases might be unevenly rolled out in different distribution channels, version pinning is required to avoid random failures related to incompatibilities between different versions in that case. Maybe, it is possible to reduce dpcpp pin to major version only.

david-cortes-intel: But this har 'conda-forge' hard-coded into it.

icfaust: I think he doesn't mean conda exclusively by distribution channel, but also include all the possibilities found here: https://www.intel.com/content/www/us/en/developer/tools/oneapi/dpc-compiler-download.html

napetrov: in general we would like reproducibility of results in main pipeline.  with all changes of tooling done through commits. this is something that ideally should be tracked by automation, so might be putting this to environment.yml and using from there

icfaust:
#2440 is a much better approach

PR #2441, opened 2025-04-17
[WIP, enhancement] enable array_api return values from ```from_table```

icfaust:
## Description

NOTE: The initial support version for ```from_table``` is not general enough and highly coded for sycl usm inputs.  This changes the interface to mirror the type (and possibly queue) of the data.  Getting this integrated will begin the fight against the technical debt introduced with ```use_raw_inputs```. The core of this PR is simple, the changes needed associated with the technical debt will be brutal.

An open question exists for type persistence in incremental algorithms, which do not follow use_raw_inputs in the return types (probably for this reason).  Some sort of solution must be found and implemented, probably as a follow-up to this PR.  It is likely that a conversion function must be stored in the instance along with the queue for returning the proper value.

---

**Testing**

**Performance**

### DIFF
--- a/onedal/datatypes/_data_conversion.py
+++ b/onedal/datatypes/_data_conversion.py
@@ -34,84 +34,64 @@ def to_table(*args, queue=None):
     """Create oneDAL tables from scalars and/or arrays.
 
     Note: this implementation can be used with scipy.sparse, numpy ndarrays,
-    DPCTL/DPNP usm_ndarrays and scalars. Tables will use pointers to the
-    original array data. Scalars and non-contiguous arrays will be copies.
-    Arrays may be modified in-place by oneDAL during computation. This works
-    for data located on CPU and SYCL-enabled Intel GPUs. Each array may only
-    be of a single datatype (i.e. each must be homogeneous).
+    dpctl/dpnp usm_ndarrays, array API standard arrays, and scalars. Tables
+    will use pointers to the original array data. Scalars and non-contiguous
+    arrays will be copies. Arrays may be modified in-place by oneDAL during
+    computation. This works for data located on CPU and SYCL-enabled Intel GPUs.
+    Each array may only be of a single datatype (i.e. each must be homogeneous).
 
     Parameters
     ----------
-    *args : {scalar, numpy array, sycl_usm_ndarray, csr_matrix, or csr_array}
+    *args : scalar, numpy array, sycl_usm_ndarray, array API standard array,
+        csr_matrix, or csr_array
         arg1, arg2... The arrays should be given as arguments.
+
+    queue : SyclQueue or None, default=None
+        A dpctl or oneDAL backend python representation of a SYCL Queue or None
 
     Returns
     -------
-    tables: {oneDAL homogeneous tables}
+    tables: oneDAL homogeneous tables
     """
     return _apply_and_pass(_convert_one_to_table, *args, queue=queue)
 
 
-if backend.is_dpc:
+def from_table(*args, array=None):
+    """Create 2 dimensional arrays from oneDAL tables.
 
-    try:
-        # try/catch is used here instead of dpep_helpers because
-        # of circular import issues of _data_conversion.py and
-        # utils/validation.py. This is a temporary fix until the
-        # issue with dpnp is addressed, at which point this can
-        # be removed entirely.
-        import dpnp
+    Note: this implementation will convert any table to numpy ndarrays,
+    dpctl/dpnp usm_ndarrays, and array API standard arrays of designated
+    type. By default, from_table will return numpy arrays and can only
+    return other types when necessary object attributes exist (i.e.
+    ``__sycl_usm_array_interface__`` or ``__array_namespace__``).
 
-        def _table_to_array(table, xp=None):
-            # By default DPNP ndarray created with a copy.
-            # TODO:
-            # investigate why dpnp.array(table, copy=False) doesn't work.
-            # Work around with using dpctl.tensor.asarray.
-            if xp == dpnp:
-                return dpnp.array(dpnp.dpctl.tensor.asarray(table), copy=False)
-            else:
-                return xp.asarray(table)
+    Parameters
+    ----------
+    *args : single or multiple python oneDAL tables
+        arg1, arg2... The arrays should be given as arguments.
 
-    except ImportError:
+    array : array-like or None, default=None
+        python object representing return type. Accessed for conversion
+        namespace when sycl_usm_array type or array API standard type
 
-        def _table_to_array(table, xp=None):
-            return xp.asarray(table)
+    Returns
+    -------
+    arrays: numpy arrays, sycl_usm_ndarrays, or array API standard arrays
+    """
 
-    def convert_one_from_table(table, sycl_queue=None, sua_iface=None, xp=None):
-        # Currently only `__sycl_usm_array_interface__` protocol used to
-        # convert into dpnp/dpctl tensors.
-        if sua_iface:
-            if (
-                sycl_queue
-                and sycl_queue.sycl_device.is_cpu
-                and table.__sycl_usm_array_interface__["syclobj"] is None
-            ):
-                # oneDAL returns tables with None sycl queue for CPU sycl queue inputs.
-                # This workaround is necessary for the functional preservation
-                # of the compute-follows-data execution.
-                # Host tables first converted into numpy.narrays and then to array from xp
-                # namespace.
-                return xp.asarray(
-                    backend.from_table(table), usm_type="device", sycl_queue=sycl_queue
-                )
-            else:
-                return _table_to_array(table, xp=xp)
-
-        return backend.from_table(table)
-
-else:
-
-    def convert_one_from_table(table, sycl_queue=None, sua_iface=None, xp=None):
-        # Currently only `__sycl_usm_array_interface__` protocol used to
-        # convert into dpnp/dpctl tensors.
-        if sua_iface:
-            raise RuntimeError(
-                "SYCL usm array conversion from table requires the DPC backend"
-            )
-        return backend.from_table(table)
-
-
-def from_table(*args, sycl_queue=None, sua_iface=None, xp=None):
-    return _apply_and_pass(
-        convert_one_from_table, *args, sycl_queue=sycl_queue, sua_iface=sua_iface, xp=xp
-    )
+    func = backend.from_table
+    if isintance(array, np.ndarray):
+        pass
+    elif hasattr(array, "__sycl_usm_array_interface__"):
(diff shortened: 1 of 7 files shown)


onedal/datatypes/_data_conversion.py
```diff
@@ -34,84 +34,108 @@ def to_table(*args, queue=None):
...
+    dpctl/dpnp usm_ndarrays, array API standard arrays, and scalars. Tables
+    will use pointers to the original array data. Scalars and non-contiguous
+    arrays will be copies. Arrays may be modified in-place by oneDAL during
+    computation. This works for data located on CPU and SYCL-enabled Intel GPUs.
+    Each array may only be of a single datatype (i.e. each must be homogeneous).
 
     Parameters
     ----------
-    *args : {scalar, numpy array, sycl_usm_ndarray, csr_matrix, or csr_array}
+    *args : scalar, numpy array, sycl_usm_ndarray, array API standard array,
```
david-cortes-intel, line 45: Is `sycl_usm_ndarray` this specific class? https://intelpython.github.io/dpctl/latest/api_reference/dpctl/generated/dpctl.tensor.usm_ndarray.html

onedal/datatypes/_data_conversion.py
```diff
@@ -34,84 +34,108 @@ def to_table(*args, queue=None):
...
+        csr_matrix, or csr_array
         arg1, arg2... The arrays should be given as arguments.
 
+    queue : SyclQueue or None, default=None
+        A dpctl or oneDAL backend python representation of a SYCL Queue or None
+
     Returns
     -------
-    tables: {oneDAL homogeneous tables}
+    tables: oneDAL homogeneous tables
```
david-cortes-intel, line 54: Format here would need to be:
```
name_of_returned_variable : type_of_returned variable
    Description about it.
```

icfaust:
This has implications about array-api-compat. I need to rethink this strategy.

icfaust:
The situation is as follows. I didn't look thoroughly enough into the reasons behind array-api-compat and its inclusion into sklearn.  Even if we fully support array_api inputs, it will not mean we will support pytorch out of the box.  Pytorch support via array api comes via getting the namespace generating from array-api-compat, meaning manually searching for __array_namespace__ won't work (and the reason why sklearn has ```get_namespace``` to compensate for that.  Thus either we need to re-invent ```get_namespace``` in onedal, acquire array namespaces from sklearn and pass them around like we do for queues (not advised) having used sklearn's ```get_namespace``` or think of another way of doing it.  The logic is insufficient for pytorch support.

icfaust:
I'm going to have to modify the injestion scripts (```to_table```) for non-contiguous inputs and dtype changes which we have a special case trigger in sycl-usm, which we will also have to add for pytorch...

icfaust:
> The situation is as follows. I didn't look thoroughly enough into the reasons behind array-api-compat and its inclusion into sklearn. Even if we fully support array_api inputs, it will not mean we will support pytorch out of the box. Pytorch support via array api comes via getting the namespace generating from array-api-compat, meaning manually searching for **array_namespace** won't work (and the reason why sklearn has `get_namespace` to compensate for that. Thus either we need to re-invent `get_namespace` in onedal, acquire array namespaces from sklearn and pass them around like we do for queues (not advised) having used sklearn's `get_namespace` or think of another way of doing it. The logic is insufficient for pytorch support.

I ended up doing a final catch for types with a lazy load of ```array_api_compat``` which may or may not be installed with sklearn.  If someone tries to use sklearnex with other array types (like pytorch), then they will have installed compat anyway to interface with sklearn.  If someone uses the oneDAL estimators directly, then they will need the warning for pytorch in the case that the sklearn install doesn't exist.

onedal/linear_model/logistic_regression.py
```diff
@@ -209,20 +209,15 @@ def _infer(self, X):
...
-            xp=xp,
-        )
+        y = from_table(result.responses, like=X)
         y = xp.take(xp.asarray(self.classes_), xp.reshape(y, (-1,)), axis=0)
         return y
 
     def _predict_proba(self, X):
         result = result = self._infer(X)
         sua_iface, xp, _ = _get_sycl_namespace(X)
         queue = QM.get_global_queue()
```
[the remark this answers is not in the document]
icfaust: lol, oh copilot.

onedal/datatypes/_data_conversion.py
```diff
@@ -61,50 +65,93 @@ def to_table(*args, queue=None):
...
+        When set to None, will return numpy arrays or scipy csr arrays.
+
+    Returns
+    -------
+    func : callable
+        A function which takes in a single table input and returns an array.
+    """
+    if isinstance(array, np.ndarray) or array is None or sp.issparse(array):
+        func = backend.from_table
+    elif hasattr(array, "__sycl_usm_array_interface__"):
```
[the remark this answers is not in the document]
icfaust: If some data ever gets that far to fail in that way, I will shake their hand.

onedal/datatypes/_data_conversion.py
```diff
@@ -37,16 +38,19 @@ def to_table(*args, queue=None):
...
-    *args : {scalar, numpy array, sycl_usm_ndarray, csr_matrix, or csr_array}
+    *args : scalar, numpy array, sycl_usm_ndarray, csr_matrix, or csr_array
         Arguments to be individually converted to oneDAL tables.
 
     queue : SyclQueue or None, default=None
         SYCL Queue object to be associated with the oneDAL tables. Default
         value None causes no change in data location or queue.
 
+    queue : SyclQueue or None, default=None
+        A dpctl or oneDAL backend python representation of a SYCL Queue or None.
```
david-cortes-intel: Argument has two tocs.

david-cortes-intel:
@icfaust What happens if the 'X' and 'y' inputs are of different types, or if one is on device and another on host?

onedal/datatypes/_data_conversion.py
icfaust: shoot. good catch, i assume there was a bad main merge

icfaust:
> @icfaust What happens if the 'X' and 'y' inputs are of different types, or if one is on device and another on host?

Good question, this goes to some underlying open questions about the onedal module estimators and their direct use.  In sklearnex there are guards for this associated with ```dispatch()```  which can be seen/ are updated in #2569 (see ```_sycl_queue_manager.from_data```).  However, using the onedal estimators directly (which is the case for spmd as that code is currently completely unregulated), disregards finiteness, shape, input type, verification of the input queue, etc. In sklearnex these are handled by ```validate_data``` with some additional constraints as previously described.

With respect to the second question: if return data is on GPU, it will try to return GPU data to the array constructor.  If input data was with a cpu sycl queue, it will attempt to copy the cpu queue from the input data (only pertinent for  array types with ```__sycl_usm_ndarray_interface__``` attributes.

Hope this helps.

david-cortes-intel:
@icfaust Could you please add tests having (a) 'X' as numpy and 'y' as GPU array; (b) 'X' as GPU array and 'y' as numpy; (c) 'X' as sycl host array on torch and 'y' as sycl device array on torch.

I'm not sure if this is the most suitable PR though.

icfaust:
> @icfaust Could you please add tests having (a) 'X' as numpy and 'y' as GPU array; (b) 'X' as GPU array and 'y' as numpy; (c) 'X' as sycl host array on torch and 'y' as sycl device array on torch.
> 
> I'm not sure if this is the most suitable PR though.

Thank you for the suggestion, I will add testing to that effect in #2569 

EDIT: this mixing of types will be only problematic for ```array_api_dispatch=True```, but should be caught via the proper use of ```get_namespace```. Likely a new design rule must be added that ```get_namespace``` must be run for all args, as that occurs independently for each estimator and method (at the moment). @david-cortes-intel

onedal/datatypes/_data_conversion.py
```diff
@@ -37,16 +38,19 @@ def to_table(*args, queue=None):
...
+
     Returns
     -------
-    tables: {oneDAL homogeneous tables}
+    tables : oneDAL homogeneous tables
```
Vika-F: Shoulld it be?
```suggestion
    tables : oneDAL homogeneous tables for scalar, numpy array, sycl_usm_ndarray; oneDAL csr_table for csr_matrix, csr_array
```

onedal/covariance/covariance.py
```diff
@@ -102,29 +102,27 @@ def fit(self, X, y=None, queue=None):
...
         use_raw_input = _get_config()["use_raw_input"] is True
-        sua_iface, _, _ = _get_sycl_namespace(X)
+        sua_iface, xp, _ = _get_sycl_namespace(X)
         if use_raw_input and sua_iface:
             queue = X.sycl_queue
 
         if not use_raw_input:
             X = _check_array(X, dtype=[np.float64, np.float32])
-        X = to_table(X, queue=queue)
+        X_t = to_table(X, queue=queue)
```
Vika-F: It would be better to replace `X_t` with `X_table` as `X_t` has a misleading connotation related to X transposed.

onedal/ensemble/forest.py
```diff
@@ -390,14 +390,14 @@ def _predict(self, X, hparams=None):
 
         model = self._onedal_model
         queue = QM.get_global_queue()
-        X = to_table(X, queue=queue)
-        params = self._get_onedal_params(X)
+        X_t = to_table(X, queue=queue)
```
Vika-F: Same with `X_t` here.

onedal/covariance/covariance.py
icfaust: Thats a valid point, will switch.

onedal/cluster/dbscan.py
```diff
@@ -78,15 +78,17 @@ def fit(self, X, y=None, sample_weight=None, queue=None):
         params = self._get_onedal_params(X_table.dtype)
         result = self.compute(params, X_table, sample_weight_table)
 
-        self.labels_ = from_table(result.responses, sycl_queue=queue).ravel()
+        self.labels_ = xp.squeeze(from_table(result.responses, like=X))
```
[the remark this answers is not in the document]
icfaust: See line 68

onedal/ensemble/forest.py
icfaust: swapped.

onedal/datatypes/_data_conversion.py
icfaust: Good call, changed to reflect that.

Vika-F (approved):
The failures in CI (WindowsCondaEnv Python3.12_Sklearn1.6) are not related to these changes, as they appear in main as well.
Let's wait for a private CI and merge.

PR #2452, opened 2025-04-25
ENH: Enable model conversion from TreeLite

david-cortes-intel:
## Description

This PR extends the model builders module in daal4py to be able to convert from objects in TreeLite format.

By supporting this format, models from other libraries such as scikit-learn can be converted to daal4py using this as an intermediate, without having to support those libraries directly.

It also allows supporting dart boosters from xgboost and lightgbm along the way, since treelite already takes care of the right conversion into a format that is very similar to daal4py's.

---

**Testing**

**Performance**

Not applicable.

### DIFF
--- a/daal4py/mb/__init__.py
+++ b/daal4py/mb/__init__.py
@@ -29,8 +29,8 @@ def convert_model(model) -> "GBTDAALModel | LogisticDAALModel":
     prediction methods.
 
     It supports gradient-boosted decision tree ensembles (GBT) from the libraries
-    ``xgboost``, ``lightgbm``, and ``catboost``; and logistic regression (binary
-    and multinomial) models from scikit-learn.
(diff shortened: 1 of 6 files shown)


razdoburdin:
Is it possible to replace original xgboost and lgbm convertors by the new one for treelite?

david-cortes-intel:
> Is it possible to replace original xgboost and lgbm convertors by the new one for treelite?

Yes, but it would require adding treelite as a hard dependency. But why would you want to do that?

razdoburdin:
> Yes, but it would require adding treelite as a hard dependency.
Got it.
>  But why would you want to do that?
I look for some ways to generalize conversion (not high priority).

razdoburdin (approved):
great job.

PR #2459, opened 2025-04-30
fix: use upload github action to public site

yuejiaointel:
## Description
This PR use github action instead of gh-pages to deploy documentation. Also will create a new branch doc_archive that will store tar.gz file for each doc release and pull it to current dir before adding a new doc folder.
Workflow:
1. Get archived tar.gz doc from previous deployment or skip it first time
2. Check in gh-pages if we have any new folders we don't have yet
3. Generate a new doc (new release doc)
4. archive current dir as a tar.gz and push to a branch called doc_archive

---

**Testing**

### DIFF
--- a/.github/scripts/doc_release.sh
+++ b/.github/scripts/doc_release.sh
@@ -23,20 +23,32 @@ if [ ! -d "$BUILD_DIR" ]; then
     exit 1
 fi
 
+rm -rf _site
 mkdir -p _site
-        
-# # Check if gh-pages branch exists in the remote repository
-# if git ls-remote --heads origin gh-pages | grep -q gh-pages; then
-#     echo "gh-pages branch exists in remote, fetching it..."    
-#     mkdir -p gh-pages
-#     git fetch origin gh-pages:refs/remotes/origin/gh-pages
-#     git worktree add gh-pages origin/gh-pages
-# else
-#     mkdir -p gh-pages
-# fi
-# # sync only new version folders from gh-pages into _site
-# rsync -av --ignore-existing gh-pages/ _site/
 
+##### Get potential new docs from gh-pages #####
+if git ls-remote --heads origin gh-pages | grep -q gh-pages; then
+    echo "gh-pages branch exists, setting up worktree for sync..."
+    git fetch origin gh-pages:gh-pages
+    git worktree add gh-pages gh-pages
+    rsync -av --ignore-existing gh-pages/ _site/
+    git worktree remove gh-pages --force
+else
+    echo "gh-pages branch does not exist, skipping sync."
+fi
+
+##### Get archived docs #####
+if git ls-remote --heads origin doc_archive | grep -q doc_archive; then
+    echo "doc_archive branch exists, syncing archived versions..."
+    git fetch origin doc_archive:doc_archive
+    git worktree add archive_sync doc_archive
+    rsync -av --ignore-existing archive_sync/ _site/
+    git worktree remove archive_sync --force
+else
+    echo "doc_archive branch does not exist, skipping archive sync."
+fi
+
+##### Prepare new doc #####
 # Copy the new built version to _site
 mkdir -p _site/$SHORT_DOC_VERSION
 cp -R doc/_build/scikit-learn-intelex/$SHORT_DOC_VERSION/* _site/$SHORT_DOC_VERSION/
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-release.yml
```diff
@@ -77,17 +83,82 @@ jobs:
         export LD_LIBRARY_PATH=$(dirname $(dirname $(which python)))/lib:$LD_LIBRARY_PATH
         cd doc
         ./build-doc.sh --gh-pages
-    - name: Deploy Documentation to gh-pages
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    # - name: Deploy Documentation to gh-pages
```
david-cortes-intel, line 87: Please do not forget to remove these before merging.

.github/scripts/doc_release.sh
```diff
@@ -23,49 +23,46 @@ if [ ! -d "$BUILD_DIR" ]; then
...
+# Copy index.html
+cp doc/_build/scikit-learn-intelex/index.html _site/
 
-# Generate versions.json by scanning for year.month folders
-rm -f doc/versions.json
-mkdir -p doc
-echo "[" > doc/versions.json
+# Generate versions.json
+mkdir -p _site/doc
+echo "[" > _site/doc/versions.json
```
david-cortes-intel, line 54: Will need to be put one level above `doc/` now.

yuejiaointel: Thx, moved

.github/scripts/doc_release.sh
```diff
@@ -15,57 +15,114 @@
...
-if ! git checkout gh-pages; then
-    echo "::error:: Could not checkout gh-pages branch!"
-    exit 1
+##### Get archived docs from tar.gz artifacts #####
+if git ls-remote --heads origin $STORAGE_BRANCH | grep -q $STORAGE_BRANCH; then
+    echo "$STORAGE_BRANCH branch exists, checking for archived artifacts..."
+    git fetch origin $STORAGE_BRANCH:$STORAGE_BRANCH
+    git worktree add archive_sync $STORAGE_BRANCH
+
+    # Get most recent tar.gz in archive branch
```
david-cortes-intel, line 35: Am I understanding it correctly that this would generate one file for every deployment containing all the doc versions that it had? Wouldn't it be better to have one file per doc version instead?

.github/scripts/doc_release.sh
```diff
@@ -15,57 +15,114 @@
...
+    echo "gh-pages branch does not exist, skipping sync."
 fi
 
-# Generate versions.json by scanning for year.month folders
-rm -f doc/versions.json
-mkdir -p doc
-echo "[" > doc/versions.json
+##### Prepare new doc #####
+# Copy the new built version to _site
+mkdir -p _site/$SHORT_DOC_VERSION
```
david-cortes-intel, line 64: Is `_site` an arbitrary name? If so, perhaps it could be controlled through an env. variable `$TEMP_DOC_FOLDER` or similar.

yuejiaointel: That makes sense, I changed to create a TEMP_DOC_FOLDER at beginning and use it throughout the script

.github/scripts/doc_release.sh
david-cortes-intel: Yes, the new design is better.

david-cortes-intel (approved):
LGTM, but would be ideal to get a review from @homksei about the GH workflow credentials and config.

.github/workflows/docs-release.yml
```diff
@@ -64,7 +83,7 @@ jobs:
         git branch
     - name: Install Python Dependencies
       run: |
-        pip install daal-devel impi-devel
+        pip install daal-devel impi-devel scikit-learn
```
david-cortes-intel, line 86: Why would this be needed if it's under `dependencies-dev`? If there's a conflict between the dev dependencies and the doc dependencies, then perhaps it could use separate environments for building the package and for building the docs. It should work as long as the python version is the same for both and the runtime requirements are present in the docs build.

yuejiaointel: Good catch, it is something I forgot to remove after some testing

david-cortes-intel (approved):

PR #2465, opened 2025-05-05
Support for sklearn 1.6 conformance testing

ethanglaser:
## Description

- scikit-learn-intelex PR to support internal conformance checks on sklearn 1.6 for CPU and 1.4 for GPU
- deselected tests updates to handle xfail/xpass discrepancies and a handful of outlying tests - all were investigated and large-scale failures have been fixed
- updates to daal4py logistic_path logic to align with scikit-learn
- updates to ridge to align coefficient and intercept shapes with scikit-learn logic

---

PR should start as a draft, then move to ready for review state after CI is passed and all applicable checkboxes are closed.
This approach ensures that reviewers don't spend extra time asking for regular requirements.

You can remove a checkbox as not applicable only if it doesn't relate to this PR in any way.
For example, PR with docs update doesn't require checkboxes for performance while PR with any change in actual code should have checkboxes and justify how this code change is expected to affect performance (or justification should be self-evident).

Checklist to comply with **before moving PR from draft**:

**PR completeness and readability**

- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes or created a separate PR with update and provided its number in the description, if necessary.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have added a respective label(s) to PR if I have a permission for that.
- [x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [x] I have run it locally and tested the changes extensively.
- [x] All CI jobs are green or I have provided justification why they aren't.
- [x] I have extended testing suite if new functionality was introduced in this PR.

PR #2494, opened 2025-05-26
[CI,WIP, enhancement] add pytorch+gpu testing ci

icfaust:
## Description

---

**Testing**

**Performance**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
           name: __release_lnx
           github-token: ${{ github.token }}
           repository: ${{ env.ONEDAL_REPO }}
-          run-id: ${{ steps.get-run-id.outputs.run-id }}
+          run-id: ${{ needs.get-run-id.outputs.run-id }}
           path: ./__release_lnx
       - name: Download oneDAL environment artifact
         uses: actions/download-artifact@v4
@@ -92,7 +92,7 @@ jobs:
           name: oneDAL_env
           github-token: ${{ github.token }}
           repository: ${{ env.ONEDAL_REPO }}
-          run-id: ${{ steps.get-run-id.outputs.run-id }}
+          run-id: ${{ needs.get-run-id.outputs.run-id }}
           path: .ci/env
       - name: Set Environment Variables
         id: set-env
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,9 +33,30 @@ env:
   DPCTL_VERSION: 0.18.1
   DPNP_VERSION: 0.16.0
   DPCTL_PY_VERSIONS: '3.9\|3.11'
+  UXL_PYTHONVERSION: "3.12"
+  ONEDAL_REPO: "uxlfoundation/oneDAL"
 
 jobs:
+
+  onedal_nightly:
+    runs-on: ubuntu-24.04
+    name: Identify oneDAL nightly
+    timeout-minutes: 5
+
+    steps:
+      - name: Get run ID of "Nightly-build" workflow
+        id: get-run-id
+        run: |
+          WF_NAME="Nightly-build"
+          JQ_QUERY='map(select(.event == "workflow_dispatch" or .event == "schedule")) | .[0].databaseId'
+          RUN_ID=`gh run --repo ${{ env.ONEDAL_REPO }} list --workflow "${WF_NAME}" --json databaseId,event --status success --jq "${JQ_QUERY}"`
+          echo "Detected latest run id of ${RUN_ID} for workflow ${WF_NAME}"
+          echo "run-id=${RUN_ID}" >> "$GITHUB_OUTPUT"
+        env:
+          GH_TOKEN: ${{ github.token }}
+
   sklearn_lnx:
+    needs: onedal_nightly
     strategy:
       fail-fast: false
       matrix:
@@ -57,23 +78,12 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.PYTHON_VERSION }}  
-      - name: Get run ID of "Nightly-build" workflow
-        id: get-run-id
-        run: |
-          OTHER_REPO="uxlfoundation/oneDAL"
-          WF_NAME="Nightly-build"
-          JQ_QUERY='map(select(.event == "workflow_dispatch" or .event == "schedule")) | .[0].databaseId'
-          RUN_ID=`gh run --repo ${OTHER_REPO} list --workflow "${WF_NAME}" --json databaseId,event --status success --jq "${JQ_QUERY}"`
-          echo "Detected latest run id of ${RUN_ID} for workflow ${WF_NAME}"
-          echo "run-id=${RUN_ID}" >> "$GITHUB_OUTPUT"
-        env:
-          GH_TOKEN: ${{ github.token }}
       - name: Download oneDAL build artifact
         uses: actions/download-artifact@v4
         with:
           name: __release_lnx
           github-token: ${{ github.token }}
-          repository: uxlfoundation/oneDAL
+          repository: ${{ env.ONEDAL_REPO }}
           run-id: ${{ steps.get-run-id.outputs.run-id }}
           path: ./__release_lnx
       - name: Download oneDAL environment artifact
@@ -81,7 +91,7 @@ jobs:
         with:
           name: oneDAL_env
           github-token: ${{ github.token }}
-          repository: uxlfoundation/oneDAL
+          repository: ${{ env.ONEDAL_REPO }}
           run-id: ${{ steps.get-run-id.outputs.run-id }}
           path: .ci/env
       - name: Set Environment Variables
@@ -161,6 +171,7 @@ jobs:
           bash .ci/scripts/run_sklearn_tests.sh $CPU
 
   sklearn_win:
+    needs: onedal_nightly
     strategy:
       fail-fast: false
       matrix:
@@ -182,33 +193,21 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.PYTHON_VERSION }}
-      - name: Get run ID of "Nightly-build" workflow
-        id: get-run-id
-        shell: bash
-        run: |
-          OTHER_REPO="uxlfoundation/oneDAL"
-          WF_NAME="Nightly-build"
-          JQ_QUERY='map(select(.event == "workflow_dispatch" or .event == "schedule")) | .[0].databaseId'
-          RUN_ID=`gh run --repo ${OTHER_REPO} list --workflow "${WF_NAME}" --json databaseId,event --status success --jq "${JQ_QUERY}"`
-          echo "Detected latest run id of ${RUN_ID} for workflow ${WF_NAME}"
-          echo "run-id=${RUN_ID}" >> "$GITHUB_OUTPUT"
-        env:
-          GH_TOKEN: ${{ github.token }}
       - name: Download oneDAL build artifact
         uses: actions/download-artifact@v4
         with:
           name: __release_win
           github-token: ${{ github.token }}
-          repository: uxlfoundation/oneDAL
-          run-id: ${{ steps.get-run-id.outputs.run-id }}
+          repository: ${{ env.ONEDAL_REPO }}
+          run-id: ${{ needs.get-run-id.outputs.run-id }}
           path: ./__release_win
       - name: Download Intel BaseKit artifact
         uses: actions/download-artifact@v4
         with:
           name: intel_oneapi_basekit
           github-token: ${{ github.token }}
-          repository: uxlfoundation/oneDAL
-          run-id: ${{ steps.get-run-id.outputs.run-id }}
+          repository: ${{ env.ONEDAL_REPO }}
+          run-id: ${{ needs.get-run-id.outputs.run-id }}
       - name: Decompress Intel BaseKit
         shell: cmd
         run: |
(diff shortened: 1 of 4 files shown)


deselected_tests.yaml
```diff
@@ -448,3 +469,6 @@ gpu:
 
   # Introduced with RNG forest updates in oneDAL
   - ensemble/tests/test_voting.py::test_set_estimator_drop
+
+  # Deselection for public GPU runner
+  - model_selection/tests/test_validation.py::test_learning_curve_some_failing_fits_warning[42]
```
david-cortes-intel: Does this deselect the test for one specific random seed? Does it pass for other seeds?

daal4py/sklearn/utils/validation.py
```diff
@@ -72,21 +72,20 @@ def _assert_all_finite(
...
-                    msg_dtype=msg_dtype,
-                    estimator_name=estimator_name,
-                    input_name=input_name,
-                )
-            else:
-                _sklearn_assert_all_finite(X, allow_nan=allow_nan, msg_dtype=msg_dtype)
-            return
-
     is_df = is_DataFrame(X)
+    if not (is_df or isinstance(X, np.ndarray)) or X.size < 32768:
```
david-cortes-intel, line 76: Wouldn't this need the same pytorch workaround used in the tests for the size attribute?

.github/workflows/ci.yml
```diff
@@ -313,3 +317,175 @@ jobs:
...
+          echo "NUMPY_BUILD=$(python -m pip freeze | grep numpy)" >> "$GITHUB_ENV"
+          bash .ci/scripts/setup_sklearn.sh ${{ env.UXL_SKLEARNVERSION }}
+          pip install --upgrade -r requirements-test.txt
+          pip install $(python .ci/scripts/get_compatible_scipy_version.py ${{ env.UXL_SKLEARVERSION }}) pyyaml
+          pip list
+      - name: Download sklearnex wheel
+        uses: actions/download-artifact@v4
+        with:
+          name: sklearnex_build_${{ env.UXL_PYTHONVERSION }}
+      - name: Install PyTorch
```
david-cortes-intel, line 466: Do you think this could make a check to verify that the install recognizes the GPU? Like in here:
https://docs.pytorch.org/docs/stable/notes/get_start_xpu.html#check-availability-for-intel-gpu

david-cortes-intel:
@icfaust Could the jobs be renamed to have 'torch' in the names instead of generic 'gpu'?

daal4py/sklearn/utils/validation.py
icfaust: first part of the conditional should block that

david-cortes-intel: But the condition is 'if not', so a torch array would meet the first part (not df, not numpy) and the second one would then be evaluated on it.

icfaust: you are correct that the first condition will evaluate true and the following or command would not be evaluated. it would already know that sklearn must be used.

example
```python
if not (False or False) or 0/0:
    print("foobar")
```
should print 'foobar'

deselected_tests.yaml
icfaust: Good question, it looks like sklearn has a specific test fixutre for that value as its the ```global_random_seed``` (https://github.com/scikit-learn/scikit-learn/blob/1.4.X/sklearn/model_selection/tests/test_validation.py#2420) just as a remark, this is also being deselected in #2465 and some information on it is available there for from @ethanglaser

icfaust:
> @icfaust Could the jobs be renamed to have 'torch' in the names instead of generic 'gpu'?

I'll make sure to add the frameworks to the title. I'd like to also still keep the `cpu` vs `gpu` because of how it impacts sklearn conformance testing.  Its the only piece currently running on GPU in this PR. Let me know what you think.

.github/workflows/ci.yml
icfaust: Will do! Seems like a good thing to add

david-cortes-intel:
> > @icfaust Could the jobs be renamed to have 'torch' in the names instead of generic 'gpu'?
> 
> I'll make sure to add the frameworks to the title. I'd like to also still keep the `cpu` vs `gpu` because of how it impacts sklearn conformance testing. Its the only piece currently running on GPU in this PR. Let me know what you think.

torch x [cpu, gpu] sounds good.

deselected_tests.yaml
david-cortes-intel: @icfaust But that other PR is not putting a specific seed to exclude in this test.

david-cortes-intel:
@icfaust Is https://github.com/uxlfoundation/scikit-learn-intelex/pull/2465 meant to be merged before this?

deselected_tests.yaml
icfaust: Good question, both will deselect this test, as it is a single parameter single value test as they want it deterministic (specific details can be found here) https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/conftest.py#L318 where the global_random_seed default value is where it is defined.

icfaust:
> @icfaust Is #2465 meant to be merged before this?

Another good question. It isn't a requirement and are independent of one another. They are related as they are both testing sklearnex on GPU vs sklearn 1.4 in CI for the first time.

deselected_tests.yaml
david-cortes-intel: Then how about leaving it deselected as a whole instead of parameterizing the deselection? Leaving it like this gives the impression that there's some specific kind of input that is problematic for the test, whereas it would still fail with any other input.

.ci/scripts/run_sklearn_tests.sh
```diff
@@ -29,8 +29,11 @@ export DESELECT_FLAGS="--public ${DESELECT_FLAGS}"
 if [ -n "${SKLEARNEX_PREVIEW}" ]; then
     export DESELECT_FLAGS="--preview ${DESELECT_FLAGS}"
 fi
-export DESELECTED_TESTS=$(python ../.circleci/deselect_tests.py ../deselected_tests.yaml ${DESELECT_FLAGS})
```
ahuber21, line 32: I know we've `export`ed before, but is this even necessary? Looks like we're just providing the list as command line argument. Seems somehow inconsistent to me.

.github/workflows/ci.yml
```diff
@@ -313,3 +317,177 @@ jobs:
           if "${{ steps.set-env.outputs.DPCFLAG }}"=="" set CPU=cpu
           set SKLEARNEX_PREVIEW=YES
           bash .ci/scripts/run_sklearn_tests.sh %CPU%
+
+  build_uxl:
+    if: github.repository == 'uxlfoundation/scikit-learn-intelex'
+    needs: onedal_nightly
+    name: LinuxNightly build Python${{ needs.onedal_nightly.outputs.uxl-python }}
+    runs-on: uxl-xlarge
+    timeout-minutes: 30
```
ahuber21, line 326: How much headroom is 30 mins? I'd say we should give it at least 4x the expected runtime, and depending on how many cores we're using it could run some 10 or so minute?

.github/workflows/ci.yml
```diff
@@ -313,3 +317,177 @@ jobs:
...
+          path: .ci/env
+      - name: Set Environment Variables
+        id: set-env
+        run: |
+          # Disable SPMD testing
+          echo "NO_DIST=1" >> "$GITHUB_ENV"
+          # enable coverage report generation
+          echo "SKLEARNEX_GCOV=1" >> "$GITHUB_ENV"
+      - name: apt-get
+        run: sudo apt-get update && sudo apt-get install -y clang-format
```
ahuber21, line 363: It is totally out of scope, but `clang-format` can be installed through pip. Maybe next time when you're playing around with the setup, experiment with adding `clang-format` to the dev requirements?

.github/workflows/ci.yml
```diff
@@ -313,3 +317,177 @@ jobs:
...
+        id: set-env
+        run: |
+          echo "NO_DIST=1" >> "$GITHUB_ENV"
+          # enable coverage report generation
+          echo "COVERAGE_RCFILE=$(readlink -f .coveragerc)" >> "$GITHUB_ENV"
+          echo "ONEDAL_PYTEST_FRAMEWORKS=${{ matrix.FRAMEWORKS }}" >> "$GITHUB_ENV"
+          # reduce GPU driver/runner related memory issues
+          echo "NEOReadDebugKeys=1" >> "$GITHUB_ENV"
+          echo "EnableRecoverablePageFaults=1" >> "$GITHUB_ENV"
+          echo "GpuFaultCheckThreshold=0" >> "$GITHUB_ENV"
```
ahuber21, line 443: We could argue that those should be set in `run_sklearn_tests.sh`. Maybe that's neater?

onedal/_device_offload.py
```diff
@@ -126,10 +127,29 @@ def _transfer_to_host(*data):
                 order=order,
             )
             has_usm_data = True
-        elif array_api and not _is_numpy_namespace(array_api):
-            # `copy`` param for the `asarray`` is not setted.
-            # The object is copied only if needed.
-            item = np.asarray(item)
+        elif not isinstance(item, np.ndarray) and (
```
ahuber21, line 132: Since you're adding so many new lines to an already long function with many branches, can I request that you refactor `_transfer_to_host()`. I'd suggest to put the logic for each case into a dedicated function and only set `has_usm_data` and `has_host_data` directly in the if statements.

onedal/datatypes/_data_conversion.py
```diff
@@ -18,6 +18,9 @@
 
 from onedal import _default_backend as backend
 
+kDLCPU = backend.kDLCPU
```
ahuber21: Are those global symbols necessary?

sklearnex/utils/validation.py
```diff
@@ -70,7 +71,14 @@ def _sklearnex_assert_all_finite(
     # size check is an initial match to daal4py for performance reasons, can be
     # optimized later
     xp, _ = get_namespace(X)
-    if X.size < 32768 or not _onedal_supported_format(X, xp):
+    # this try-catch is a PyTorch-specific fix, as Tensor.size is a function.
+    # The try-catch minimizes changes to most common code path (numpy arrays).
+    try:
+        too_small = X.size < 32768
+    except TypeError:
+        too_small = math.prod(X.shape) < 32768
```
ahuber21: Wouldn't this work?

```suggestion
    too_small = (math.prod(X.shape) if xp is torch else X.size) < 32768
```

ahuber21 (dismissed):

deselected_tests.yaml
icfaust: Sounds good!

onedal/_device_offload.py
icfaust: You and I are on the same page, done in the follow up to this PR: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2509 which will simplify the ```__sycl_usm_array_interface__``` branch, and then an yet unstarted PR will separate out the ```__dlpack_device__``` branch to a new file under ```onedal/datatypes/``` as that will coincide with new queue extraction added for ```from_data``` in ```onedal/utils/_sycl_queue_manager.py``` (as a follow-up to #2509).

sklearnex/utils/validation.py
icfaust: Its definitely cleaner. The only problem is that we would have to have imported torch, which should be avoided unless absolutely necessary.  I could modify your suggestion to use array_api_compat's ```is_torch_array``` (https://github.com/data-apis/array-api-compat/blob/main/array_api_compat/common/_helpers.py#L149). Let me know what you think.

onedal/datatypes/_data_conversion.py
icfaust: You highlight something that I did which wasn't so perfect.  The dlpack device data should be on CPU, the most sure way to do that is to check if its the kDLCPU enum.  This will only be used in _transfer_to_host, where the kDLOneAPI will be used for recognizing if a queue can be extracted.  I need to get it in ```onedal/_device_offload.py``` do you think I should import it from the backend there directly? I put it in ```_data_conversion``` since its a data conversion topic, but it required importing it through 2-3 spots to get it in _device_offload. Let me know what you think.

.github/workflows/ci.yml
icfaust: The only worry is that it will only impact the driver settings for the sklearn conformance steps, will need to do it also for the sklearnex testing. We could move it to ```run_sklearn_tests.sh```, and in doing so add it to ```conda_recipe/run_test.sh``` and possibly ```conda_recipe/run_test.bat```. I am willing to do it, just let me know.

.github/workflows/ci.yml
icfaust: Honestly that's a great idea, half the time when I am building a new environment I run head long into this issue at the daal4py build step.  Having a proper versioned clang-format from pip would be great to have in dev-dependencies to simplify my life, especially because we are already referencing the pip clang-format in the pre-commit step anyway.

.github/workflows/ci.yml
icfaust: Never looked deeply into the value.  Last run was approx 10 min on the build step, so I guess its a 3x buffer, do you think I should bump to 40 min?

.ci/scripts/run_sklearn_tests.sh
icfaust: oh yeah, this stuff could be improved from the original design in #1199. Its re-acquired in the python script called in the shell script: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/.ci/scripts/run_sklearn_tests.py#L54 .  Truly should just be a command line argument into ```run_sklearn_tests.py```.  Would it be okay to assign to me as a follow up to re-jigger that aspect?

.github/workflows/ci.yml
ahuber21: I guess it's fine

onedal/_device_offload.py
ahuber21: Cool.

onedal/datatypes/_data_conversion.py
ahuber21: I would always import it straight from the original source. Let me know if that causes any issues.

sklearnex/utils/validation.py
ahuber21: Yeah, so we do this?

```python
is_torch = is_torch_array(X)
too_small = (math.prod(X.shape) if is_torch else X.size) < 32768
```

.github/workflows/ci.yml
ahuber21: No strong opinion, it was just a thought. Let's leave it like this. Keeping multiple script files in sync isn't any better.

.github/workflows/ci.yml
```diff
@@ -313,3 +317,177 @@ jobs:
...
+        run: |
+          # This CI system yields oneAPI dependencies from the oneDAL repository
+          bash .ci/env/apt.sh dpcpp
+      - name: describe system
+        run: |
+          source /opt/intel/oneapi/setvars.sh
+          bash .ci/scripts/describe_system.sh
+      - name: Install test requirements
+        run: |
+          pip install -r dependencies-dev
```
ethanglaser: I believe installation of dependencies-dev be skipped for test env

icfaust: Oh interesting. You are absolutely right, if we have implemented requirements-test.txt properly, then dependencies-dev shouldn't be necessary.  I'll remove it and see what happens, if it works, now we would have a way of testing publicly that they work as intended separately (which we otherwise don't currently do).

.github/workflows/ci.yml
icfaust: Started in #2543

.github/workflows/ci.yml
icfaust: Done! looks like it works.

david-cortes-intel:
> Something broke with the new coverage stats and need to investigate it.

@icfaust It's a general world-wide GCP outage.

sklearnex/utils/validation.py
icfaust: Done

onedal/datatypes/_data_conversion.py
icfaust: Done

icfaust:
Had to switch to just using ```math.prod(X.shape)``` just like sklearn due to private CI infrastructure issues.  This removes the array_api_compat dependency.

PR #2506, opened 2025-06-03
[maintenance] use standard ```_is_numpy_namespace``` for numpy namespace checking in PCA

icfaust:
## Description

```_is_numpy_namespace``` was introduced in #1861 which is used for overall numpy namespace checking.  This swaps a custom check with our standard one in order to reduce maintenance and make general conformance across the codebase.  This was also done in ```IncrementalEmpericalCovariance```.

---

**Testing**

### DIFF
--- a/sklearnex/decomposition/pca.py
+++ b/sklearnex/decomposition/pca.py
@@ -45,6 +45,7 @@ if daal_check_version((2024, "P", 100)):
     from sklearn.decomposition import PCA as _sklearn_PCA
 
     from onedal.decomposition import PCA as onedal_PCA
+    from onedal.utils._array_api import _is_numpy_namespace
 
     @control_n_jobs(decorated_methods=["fit", "transform", "fit_transform"])
     class PCA(oneDALEstimator, _sklearn_PCA):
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
@icfaust What would happen in this case if the input falls into numpy namespace, but the means/components are not on CPU?

icfaust:
> @icfaust What would happen in this case if the input falls into numpy namespace, but the means/components are not on CPU?

Interesting question.  The means and components would only be non numpy arrays currently if they successfully completed a PCA fit using array_api_dispatch with sklearn's fit. It would fail, but this would conform to what would occur with array_api_dispatch using sklearn's PCA, which means the failure is conforming to sklearn's failure.

However, our return values for array_api_dispatch using the onedal fit with non-numpy inputs should also create non-numpy array attributes, but currently only returns non-numpy arrays and sets numpy attributes. This is violation with sklearn array_api conformance, and when we change that, this would become a problem here.

david-cortes-intel (approved):
Ok, guess that's not in the scope of this PR.

PR #2521, opened 2025-06-09
fix: minor basic stats quality fixes

ethanglaser:
## Description

A few small corrections

---

**Testing**

### DIFF
--- a/examples/sklearnex/basic_statistics_spmd.py
+++ b/examples/sklearnex/basic_statistics_spmd.py
@@ -48,7 +48,7 @@ size = comm.Get_size()
 
 params_spmd = {"ns": 19, "nf": 31}
 
-data, weights = generate_data(params_spmd, rank)
+data, weights = generate_data(params_spmd, size, seed=rank)
 weighted_data = np.diag(weights) @ data
 
 dpt_data = dpt.asarray(data, usm_type="device", sycl_queue=q)
--- a/examples/sklearnex/basic_statistics_spmd.py
+++ b/examples/sklearnex/basic_statistics_spmd.py
@@ -48,7 +48,7 @@ size = comm.Get_size()
 
 params_spmd = {"ns": 19, "nf": 31}
 
-data, weights = generate_data(params_spmd, size)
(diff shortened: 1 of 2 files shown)


onedal/basic_statistics/basic_statistics.py
```diff
@@ -157,7 +157,7 @@ def fit(self, data, sample_weight=None, queue=None):
         data_table, weights_table = to_table(data, sample_weight, queue=queue)
 
         dtype = data_table.dtype
-        raw_result = raw_result = self._compute_raw(
+        raw_result = self._compute_raw(
```
icfaust: yikes

icfaust (commented):
requires a 'black' fix. otherwise good to go

examples/sklearnex/basic_statistics_spmd.py
```diff
@@ -48,7 +48,7 @@ def generate_data(par, size, seed=777):
 
 params_spmd = {"ns": 19, "nf": 31}
 
-data, weights = generate_data(params_spmd, size)
+data, weights = generate_data(params_spmd, rank)
```
david-cortes-intel: Shouldn't this be made to generate different data for each rank?

ethanglaser: Yes - that was the original mistake here. `size` is the same for every rank, so the same data is generated. `rank` is different on every rank, so different data is generated.

david-cortes-intel: But it's still being generated in a loop where each rank contains the data from the previous one.

ethanglaser: Although maybe that would be reflected in the `seed` parameter and this should be tweaked further. The data generation function here is pretty wonky. I'll take a closer look tomorrow.

ethanglaser:
```suggestion
data, weights = generate_data(params_spmd, size, seed=rank)
```

Alexsandruss (approved):

PR #2523, opened 2025-06-10
MAINT: Remove unused arguments to LogisticRegression SPMD

david-cortes-intel:
## Description

This PR removes unused arguments to the SPMD version of logistic regression, as it ignores some of them.

---

**Testing**


david-cortes-intel:
Failures do not seem to be blamable on this PR.

avolkov-intel:
I think we can hard-code method "dense_batch" in this line here https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/linear_model/logistic_regression.py#L66 and avoid having this argument at all

david-cortes-intel:
> I think we can hard-code method "dense_batch" in this line here https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/linear_model/logistic_regression.py#L66 and avoid having this argument at all

Would it require removing the argument from some other function signature?

avolkov-intel:
> > I think we can hard-code method "dense_batch" in this line here https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/linear_model/logistic_regression.py#L66 and avoid having this argument at all
> 
> Would it require removing the argument from some other function signature?

I'm pretty sure it only should be removed from class declaration in this file, it shouldn't be used in any other places

https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/linear_model/logistic_regression.py

david-cortes-intel:
> This is impacting beyond spmd, it is impacting all onedal GPU use of logisticregression please reflect that in the title and description.

I don't follow. The non-SPMD GPU estimator from sklearnex would already check for conditions that wouldn't result in passing what's removed here. Only the SPMD estimator would allow calling it with all the arguments it offers that it then ignores.

> As per the evolving design documents, these would be sklearn conformance additions, and are in the wrong module. Thanks for removing them.

Which design documents do you mean?

david-cortes-intel:
Looks like these arguments were actually needed for some reason for the non-SPMD GPU version to work, so not sure what would be the best way to remove them from SPMD instead of accepting but ignoring.

PR #2537, opened 2025-06-12
FIX: Don't forcibly add `-s` to debug builds

david-cortes-intel:
## Description

This PR avoids adding the `-s` flag when `-g` is used, or when the modules are built with absolute rpaths for oneDAL, which is only in internal non-release builds.

Flag `-s` strips debug information, which makes it harder to investigate issues. There is currently a mechanism to detect `-g` flag and avoid adding `-s` if so, but it doesn't work correctly. This PR extends it to be able to work in more situations.

---

**Testing**

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -309,9 +309,13 @@ def get_build_options():
         ela.append("-Wl,-rpath,@loader_path/../../../")
     elif IS_WIN:
         ela.append("-IGNORE:4197")
-    elif IS_LIN and not any(
-        x in os.environ and "-g" in os.environ[x]
-        for x in ["CPPFLAGS", "CFLAGS", "LDFLAGS"]
+    elif (
+        IS_LIN
+        and not any(
+            x in os.environ and "-g" in os.environ[x]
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -309,9 +309,13 @@ def get_build_options():
         ela.append("-Wl,-rpath,@loader_path/../../../")
     elif IS_WIN:
         ela.append("-IGNORE:4197")
-    elif IS_LIN and not any(
-        x in os.environ and "-g" in os.environ[x]
-        for x in ["CPPFLAGS", "CFLAGS", "LDFLAGS"]
+    elif (
```
Vika-F: It might simplify the code's logic if this `elif` is combined with the next if as `elif IS_LIN: ...`.

david-cortes-intel: Moved under the other condition.

david-cortes-intel:
> Seems reasonable, quick question though, why avoid `USE_ABS_RPATH`?

It is not avoiding it, it's making it not add `-s` (strip debug symbols, meant to reduce size of binaries in releases) when it sees the option to use an absolute rpath (which is only used in non-release builds).

PR #2555, opened 2025-06-17
[bugfix] fix array type issues in ```pca.inverse_transform``` without ```array_api_dispatch``` enabled 

icfaust:
## Description

Sklearn does not input check their ```PCA.inverse_transform``` method, making array_api support with and without ```array_api_dispatch``` enablement difficult. Unlike forcing all data to the cpu for evaluation, this takes the necessary components and matches it to the input data.  It will then use the native matrix multiplication infrastructure of the data framework to evaluate on device. This is a temporary fix as it will change with the rollout of array api support (namely #2106).

This was uncovered with an update to array_api_strict from 2.3.1 to 2.4.  ```wrap_output_data``` is  removed as it has no impact based on how the method is written.

This would also be solved easily using scikit-learn/scikit-learn#29310 had been reviewed and merged given it would add shape and type conformance to the estimator.

This should solve issues in LinuxCondaRecipe and WindowsCondaRecipe returning green CI.

---

**Testing**

**Performance**

### DIFF
--- a/sklearnex/decomposition/pca.py
+++ b/sklearnex/decomposition/pca.py
@@ -233,7 +233,7 @@ if daal_check_version((2024, "P", 100)):
                 # same device as the data (compute follows data).
                 components = xp.asarray(components, device=X.device)
                 mean = xp.asarray(mean, device=X.device)
-            print(type(mean),type(components))
+            print(type(mean), type(components))
             return X @ components + mean
 
         def _onedal_supported(self, method_name, X):
--- a/sklearnex/decomposition/pca.py
+++ b/sklearnex/decomposition/pca.py
@@ -224,7 +224,7 @@ if daal_check_version((2024, "P", 100)):
                 )
             else:
                 components = self.components_
-
+            print(_is_numpy_namespace(xp), type(X), type(mean), type(components))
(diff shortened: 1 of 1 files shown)


sklearnex/decomposition/pca.py
```diff
@@ -213,9 +213,17 @@ def fit_transform(self, X, y=None):
                 # Scikit-learn PCA["covariance_eigh"] was fit
                 return self._transform(X_fit, xp, x_is_centered=x_is_centered)
 
-        @wrap_output_data
```
david-cortes-intel, line 216: Is it required to remove `wrap_output_data`? How about deselecting the sklearn test instead?

icfaust: It's a sklearnex test which is failing. ```wrap_output_data``` was superfluous (i.e. we could leave it, but its something that isn't needed)

Alexsandruss: Although it is obvious from function content why `wrap_output_data` is not required, the reason might be useful as a small comment before function.

icfaust:
Pre-commit failure is due to explained memmap issue, otherwise green.

PR #2561, opened 2025-06-18
Add the ability to reset hyperparameters

Vika-F:
## Description

Static `reset_hyperparameters(op)` method that resets the hyperparameters of the algorithm to the default settings was added to the estimators that have hyperparameters.

---

**Testing**

**Performance**

not applicable


onedal/common/hyperparameters.py
```diff
@@ -117,9 +139,18 @@ def get_methods_with_prefix(obj, prefix):
...
-            algorithm, op, setters, getters, hyperparameters
+            algorithm, op, setters, getters, hyperparameters_backend
         )
 
 
 def get_hyperparameters(algorithm, op):
     return hyperparameters_map.get((algorithm, op), None)
+
+
+def reset_hyperparameters(algorithm, op):
```
david-cortes-intel:
```suggestion
def reset_hyperparameters(algorithm: str, op: str): -> None
```

onedal/common/hyperparameters.py
```diff
@@ -88,27 +112,25 @@ def __setattr__(self, __name, __value):
...
-
-    hyperparameters_backend: Dict[Tuple[str, str], Any] = {
-        (
-            "linear_regression",
-            "train",
-        ): backend.linear_model.regression.train_hyperparameters(),
-        ("covariance", "compute"): backend.covariance.compute_hyperparameters(),
-    }
+    hyperparameters_backend_items = [
+        ("linear_regression", "train"),
```
david-cortes-intel: What about 'partial_train'?

onedal/common/hyperparameters.py
```diff
@@ -35,6 +35,30 @@
         "to_dict",
     ]
 
+    def get_hyperparameters_backend(algorithm, op):
```
david-cortes-intel: Could this be put into a dictionary or similar, so that later on it would be easier to add more entries?

For example, could be joined conditionally like this:
```python
dct | (
{<other dict>}
if daal_check_version(...)
else {} # <- empty dict
)
```

Vika-F: In main branch it is currently a dictionary `hyperparameters_backend: Dict[Tuple[str, str], Any]`.
I have replaced it with a function, because in the dictionary the backend object is created once and then re-used causing those problems with inability to truly reset the hyperparameters.
I have replaced the dictionary with this function to be able to create a new backend on each reset of hyperparameters object.
I will think about how to create a dict that allows to create a new backend object on every reset (maybe with lambdas?), but for now it is not obvious to me how I can achieve this.

onedal/common/hyperparameters.py
Vika-F: there was no backend for partial_train in main branch. Anyway the work is in progress, I will check that, thanks.

onedal/common/hyperparameters.py
david-cortes-intel: I meant the backend part, where it looks for entries by key like 'linear_regression+fit'.

onedal/common/hyperparameters.py
Vika-F: I've checked. It uses the same ("linear_regression", "train") pair to access hyperparameters object.
A testcase for `IncrementalLinearRegression.partial_fit()` was added into test_hyperparameters.py.

onedal/common/hyperparameters.py
Vika-F: Is it better now?

david-cortes-intel: Yes, thanks.

david-cortes-intel:
@Vika-F Failure looks legit:
```

svd_solver = 'full'

    @pytest.mark.parametrize("svd_solver", PCA_SOLVERS)
    def test_pca_dtype_preservation(svd_solver):
>       check_pca_float_dtype_preservation(svd_solver)

decomposition\tests\test_pca.py:518: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
decomposition\tests\test_pca.py:538: in check_pca_float_dtype_preservation
    assert_allclose(pca_64.components_, pca_32.components_, rtol=2e-4)
utils\_testing.py:290: in assert_allclose
    np_assert_allclose(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (<function assert_allclose.<locals>.compare at 0x000001BC11C137F0>, array([[-0.62022375, -0.15983497,  0.38316965,  0....4071014,  0.9080595 , -0.21966814],
       [ 0.12497696,  0.8811203 , -0.16712739,  0.42435375]],
      dtype=float32))
kwds = {'equal_nan': True, 'err_msg': '', 'header': 'Not equal to tolerance rtol=0.0002, atol=0', 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Not equal to tolerance rtol=0.0002, atol=0
E           
E           Mismatched elements: 4 / 12 (33.3%)
E           Max absolute difference: 0.00014655
E           Max relative difference: 0.00087689
E            x: array([[-0.620224, -0.159835,  0.38317 ,  0.66555 ],
E                  [ 0.263179,  0.240851,  0.908007, -0.21966 ],
E                  [ 0.124977,  0.881087, -0.167274,  0.424366]])
E            y: array([[-0.620258, -0.159886,  0.38311 ,  0.66554 ],
E                  [ 0.263121,  0.24071 ,  0.908059, -0.219668],
E                  [ 0.124977,  0.88112 , -0.167127,  0.424354]], dtype=float32)
```

It appears that something might have assumed that it was running with different parameters and needs its tolerance adjusted. Either that, or the parameters when it reaches that test might not have been reset.

sklearnex/decomposition/tests/test_pca.py
```diff
@@ -113,3 +70,7 @@ def test_sklearnex_import(
     assert_allclose([6.30061232, 0.54980396], _as_numpy(pca.singular_values_))
     assert_allclose(X_transformed_expected, _as_numpy(X_transformed), rtol=tol)
     assert_allclose(X_transformed_expected, _as_numpy(X_fit_transformed), rtol=tol)
+
+    # Reset hyperparameters to the default values
+    if daal_check_version((2025, "P", 700)):
+        PCA.reset_hyperparameters("fit")
```
david-cortes-intel: To ensure this is called, better to put it into a fixture cleanup function:
https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization

david-cortes-intel: Like in this PR for example: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2568

Vika-F:
> @Vika-F Failure looks legit:
> 
> ...
> It appears that something might have assumed that it was running with different parameters and needs its tolerance adjusted. Either that, or the parameters when it reaches that test might not have been reset.

But we do not use any hyperparameters with SVD solver. I.e. the changes should not affect this test case at all.
I also see that this step fails on non-Intel runners, maybe this is the cause of the difference.

david-cortes-intel:
> > @Vika-F Failure looks legit:
> > ...
> > It appears that something might have assumed that it was running with different parameters and needs its tolerance adjusted. Either that, or the parameters when it reaches that test might not have been reset.
> 
> But we do not use any hyperparameters with SVD solver. I.e. the changes should not affect this test case at all. I also see that this step fails on non-Intel runners, maybe this is the cause of the difference.

I think there's some logic that modifies the solver for older version of sklearn:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/7585bf84d9acaef054fa1d386b353ddd21a476e3/sklearnex/decomposition/pca.py#L323

Does this bug happen in others PRs? If not, perhaps one thing to do would be to verify which solver from oneDAL is being called in those failing environments.

david-cortes-intel:
@Vika-F The bug doesn't appear to affect other recent PRs, so probably something here is having an effect.

david-cortes-intel:
I'm now seeing the same error in a different PR:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/2577

So probably not caused by the changes here, but odd that it happens only sporadically.

david-cortes-intel:
That's a curious error in the CI:
```
>       with pytest.raises(ValueError, match=msg):
E       AssertionError: Regex pattern did not match.
E        Regex: 'X has 2 features, but \\w+ is expecting 10 features as input'
E        Input: 'Incompatible dimension for X and Y matrices: X.shape[1] == 2 while Y.shape[1] == 10'
```

It looks like the error is what was expected, just with a different message.

Is sklearnex perhaps copy-pastying an outdated error message from sklearn?

david-cortes-intel:
@Vika-F Perhaps you could hard-reset this to commit `ad92a22e0ec8fd86ab941200ce5064a4835ad7ca` and then merge main.

Vika-F:
> Perhaps you could hard-reset this to commit `ad92a22e0ec8fd86ab941200ce5064a4835ad7ca` and then merge main.
@david-cortes-intel I will try to do it. Thanks for the suggestion.

david-cortes-intel:
@Vika-F The errors do not look related to the changes in this PR. Looks like you need to merge main for the import errors about sklearn.

sklearnex/preview/covariance/tests/test_covariance.py
```diff
@@ -23,73 +23,41 @@
...
-    from sklearnex.preview.covariance import EmpiricalCovariance
-
-    X = np.array([[0, 1], [0, 1], [0, 1], [0, 1], [0, 1], [0, 1]])
+    X = [[0, 1], [0, 1], [0, 1], [0, 1], [0, 1], [0, 1]]
 
     X = _convert_to_dataframe(X, sycl_queue=queue, target_df=dataframe)
     empcov = EmpiricalCovariance(assume_centered=assume_centered)
 
-    hyperparameters_route(queue, macro_block, grain_size)
+    if daal_check_version((2025, "P", 7)):
```
david-cortes-intel:
```suggestion
    if daal_check_version((2025, "P", 700)):
```

sklearnex/preview/covariance/covariance.py
```diff
@@ -90,11 +89,15 @@ def _onedal_supported(self, method_name, *data):
...
+        return self
+
+    @wrap_output_data
+    def _fit(self, X):
         if sklearn_check_version("1.2"):
             self._validate_params()
-        X = validate_data(self, X, ensure_all_finite=False)
 
-        dispatch(
+        return dispatch(
```
icfaust, line 100: Why is this done?

sklearnex/preview/covariance/covariance.py
```diff
@@ -90,11 +89,15 @@ def _onedal_supported(self, method_name, *data):
     _onedal_gpu_supported = _onedal_supported
 
     def fit(self, X, y=None):
+        self._fit(X)
+        return self
+
+    @wrap_output_data
```
icfaust: Wrap output data should at least be removed, this must cause problems minimally.  I think the easiest thing to do here is to move the validate_data call into ```_onedal_fit``` and leave it how it was.

Vika-F: Oh, I think this was a copy-paste from IncrementalCovariance as one of the attempts to fix CI error.
It looks like it made them even worse. I will revert those changes.

sklearnex/preview/covariance/tests/test_covariance.py
```diff
@@ -23,73 +23,41 @@
...
 @pytest.mark.parametrize("grain_size", [None, 2])
 @pytest.mark.parametrize("assume_centered", [True, False])
 def test_sklearnex_import_covariance(
-    dataframe, queue, macro_block, grain_size, assume_centered, hyperparameters_route
+    hyperparameters, dataframe, queue, macro_block, grain_size, assume_centered
 ):
-    from sklearnex.preview.covariance import EmpiricalCovariance
-
-    X = np.array([[0, 1], [0, 1], [0, 1], [0, 1], [0, 1], [0, 1]])
+    X = [[0, 1], [0, 1], [0, 1], [0, 1], [0, 1], [0, 1]]
```
david-cortes-intel: Does something break if they are passed as numpy arrays?

Vika-F: It was breaking something at some point. But maybe now it does not.
Does this really important to have this as numpy array?

david-cortes-intel: I don't know, but if it was working before and it's not working now, it sounds like it might be a problem with the changes here.

PR #2577, opened 2025-06-30
FIX: Fix logic to check for existence of different oneDAL libraries

david-cortes-intel:
## Description

This PR fixes the logic in the setup file to be able to detect DPC and SPMD shared objects of oneDAL when it receives `$DALROOT` but that path is not added to `$LD_LIBRARY_PATH`.

---

**Testing**

### DIFF
--- a/scripts/version.py
+++ b/scripts/version.py
@@ -17,7 +17,6 @@
 
 import os
 import re
-import sys
 from ctypes.util import find_library
 from os.path import isfile
 from os.path import join as jp


scripts/version.py
```diff
@@ -70,6 +72,19 @@ def get_onedal_version(dal_root, version_type="release"):
     return version
 
 
+def find_library_custom_paths(alias, dal_root):
+    if find_library(alias):
+        return True
+    IS_WIN = sys.platform in ["win32", "cygwin"]
```
icfaust: Instead of redefining ```IS_WIN``` here, why not just make it an argument as its being determined in setup.py anyway.

david-cortes-intel: Changed to reuse the variable, along with other fixes.

icfaust (approved):
I do like the changes for improved english as well.

scripts/version.py
```diff
@@ -15,7 +15,9 @@
 # limitations under the License.
 # ===============================================================================
 
+import os
 import re
+import sys
```
icfaust:
```suggestion
```
I think sys is no longer necessary.

scripts/version.py
```diff
@@ -15,7 +15,9 @@
 # limitations under the License.
 # ===============================================================================
 
+import os
 import re
+import sys
 from ctypes.util import find_library
 from os.path import isfile
 from os.path import join as jp
```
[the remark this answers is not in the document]
icfaust: Hey Copilot, I'm already on it. https://github.com/uxlfoundation/scikit-learn-intelex/pull/2577#discussion_r2175107648

scripts/version.py
david-cortes-intel: Removed.

PR #2583, opened 2025-07-02
ENH: Allow compiling with LLD as linker

david-cortes-intel:
## Description

This PR adds an option to the setup file which avoids adding linkage arguments that are not supported by the LLD linker, alongside with docs about it.

---

**Testing**


setup.py
```diff
@@ -167,10 +168,13 @@ def get_sdl_cflags():
 
 def get_sdl_ldflags():
     if IS_LIN:
-        return [
-            "-Wl,-z,noexecstack,-z,relro,-z,now,-fstack-protector-strong,"
-            "-fno-strict-overflow,-fno-delete-null-pointer-checks,-fwrapv"
-        ]
+        if not USING_LLD:
+            return [
+                "-Wl,-z,noexecstack,-z,relro,-z,now,-fstack-protector-strong"
```
icfaust: Copilot did correctly spot a drop of a comma between ```-fstack-protector-strong``` and ```-fno-strict-overflow```

david-cortes-intel: Added back.

icfaust: @david-cortes-intel  After the private CI runs through, I'll give it a last look. I think the english is fine, and the changes for the ```is_win``` etc. centralization are okay.  If you could, just mention that last bit in the description, and I'll approve it after looking at the private CI run (assuming everything is fine, which is highly likely).

david-cortes-intel:
Failures are all due to timeouts and issues that are not connected with this PR.

scripts/CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ message(STATUS "ADD_ONEDAL_RPATH:" ${ADD_ONEDAL_RPATH})
 option(SKLEARNEX_GCOV "Compile with gcov" OFF)
 message(STATUS "SKLEARNEX_GCOV:" ${SKLEARNEX_GCOV})
 
+option(USING_LLD "Using LLD as linker" OFF)
+message(STATUS "USING_LLD:" ${USING_LLD})
```
Vika-F, line 37: Can it be handled with `CMAKE_CXX_LINK_EXECUTABLE` option?
It looks like this is a more common choice to provide linker and it's options to CMake.

david-cortes-intel: @Vika-F This PR is not setting the linker, it's only making the scripts avoid forcibly adding arguments that are not compatible with LLD when using this new option.

Or are you suggesting that it should additionally set the linker? There's the issue that it also needs to compile daal4py, which is not built with cmake.

Vika-F: I meant it could be better to do something like this instead of defining a custom flag:
```
if(CMAKE_CXX_COMPILER_LINKER_ID STREQUAL "LLD")
...
else()
...
endif()
```
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LINKER_ID.html#variable:CMAKE_%3CLANG%3E_COMPILER_LINKER_ID

david-cortes-intel: Yes, could be done for the cmake part. But it would not apply to the cython part which doesn't use cmake.

PR #2589, opened 2025-07-07
[CI] Add design rule which enforces all validate_data inputs in get_namespace when array api enabled

icfaust:
## Description

This adds a rule which parses the trace to verify that the inputs to ```get_namepace``` exist in ```validate_data``` and possibly ```_check_sample_weight```.  ```get_namespace``` verifies that all inputs are of the same array type, which is important for setting the return values and estimator attributes properly.

It skips unless the estimator is array API enabled, as otherwise the input is converted to numpy via ```validate_data```
---

**Testing**

**Performance**


david-cortes-intel:
I realize that this is adding to an existing system, but there's a more idiomatic way to construct these tests that would leave more understandable error messages and better traces:
https://pytest-mock.readthedocs.io/en/latest/usage.html#spy

david-cortes-intel:
@icfaust I'm unable to make the test here run when running 'test_common.py'. What would be required to make this test execute and reach the section with the assertion about the inputs?

sklearnex/tests/test_common.py
```diff
@@ -545,11 +547,92 @@ def fit_check_before_support_check(text, estimator, method):
...
+        get_valid_inputs = "".join(validate_data_call.split()).split(",")
+
+        assert "self" == get_valid_inputs[0]
+        for idx, inp in enumerate(get_valid_inputs[1:]):
+            # this is specifically written for validate_data definition and
+            # how we lint sklearnex. Generally X and y should be passed not
+            # as keyword arguments. See:
+            # https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.validate_data.html
+            if ("=" not in inp and idx < 2) or inp.startswith(("X=", "y=")):
+                assert (
```
david-cortes-intel, line 615: @icfaust Could you add an intentional failure right here to verify that at least one CI job is able to reach this section?

icfaust: The reason it doesn't run is because none of the estimators do native array api dispatching yet.  This ```get_namespace``` type issue is only relevant once we start introducing that into the estimators.  Its a bit of a chicken and the egg problem.

david-cortes-intel: Then I guess it'd be better to postpone this PR until we have some estimator with native array API dispatching.

daal4py/sklearn/linear_model/logistic_path.py
```diff
@@ -902,75 +857,35 @@ def daal4py_predict(self, X, resultsToEvaluate):
...
-        dual=dual,
-        penalty=penalty,
-        intercept_scaling=intercept_scaling,
-        multi_class=multi_class,
-        random_state=random_state,
-        check_input=check_input,
-        max_squared_sum=max_squared_sum,
-        sample_weight=sample_weight,
-        l1_ratio=l1_ratio,
+def logistic_regression_path(*args, **kwargs):
```
david-cortes-intel, line 860: Since this is an internal function, I think there would be no harm in leaving it with an extra unused argument instead of splitting the signature like this.

PR #2592, opened 2025-07-08
FIX: Install scikit-learn in doc deployment CI job

david-cortes-intel:
## Description

This PR manually installs scikit-learn as a necessary dependency in the CI job that builds and deploys docs for sklearnex, since it is not listed as a dev requirement, nor as a doc requirement.

---

**Testing**

### DIFF
--- a/.github/workflows/docs-release.yml
+++ b/.github/workflows/docs-release.yml
@@ -83,7 +83,7 @@ jobs:
         git branch
     - name: Install Python Dependencies
       run: |
-        pip install daal-devel impi-devel
+        pip install daal-devel impi-devel scikit-learn
         pip install -r dependencies-dev
         pip install -r requirements-doc.txt
     - name: Build daal4py/sklearnex


.github/workflows/docs-release.yml
```diff
@@ -83,7 +83,7 @@ jobs:
         git branch
     - name: Install Python Dependencies
       run: |
-        pip install daal-devel impi-devel
+        pip install daal-devel impi-devel scikit-learn
         pip install -r dependencies-dev
         pip install -r requirements-doc.txt
```
icfaust, line 88:
```suggestion
        pip install daal-devel impi-devel
        pip install -r dependencies-dev
        pip install -r requirements-doc.txt
        pip install -r requirements-test.txt
```

to match the docs CI

david-cortes-intel: I don't think this would be beneficial, because it pulls docstrings from sklearn (such as the metadata routing calls), so it's better for it to have always the latest.

icfaust: gotcha, then you should probably change it here so we can get a preview of what occurrs: https://github.com/david-cortes-intel/scikit-learn-intelex/blob/install_sklearn_for_docs/.ci/pipeline/docs.yml#L71

yuejiaointel (approved):
Thx for fixing this!

.github/workflows/docs-release.yml
david-cortes-intel: Addressed in a follow-up PR: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2603

PR #2484, opened 2025-05-20
fix: add cstdint include to transceiver for public CI

ethanglaser:
## Description

Fixes public CI Linux Conda Recipe error: 'uint8_t' was not declared in this scope (amongst others). Exact origin of this is unclear, but this fixes.

---

Checklist to comply with **before moving PR from draft**:

**PR completeness and readability**

- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes or created a separate PR with update and provided its number in the description, if necessary.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have added a respective label(s) to PR if I have a permission for that.
- [x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [x] I have run it locally and tested the changes extensively.
- [x] All CI jobs are green or I have provided justification why they aren't.
- [x] I have extended testing suite if new functionality was introduced in this PR.

PR #2595, opened 2025-07-08
FIX: Specify namespace for std types

david-cortes-intel:
## Description

This PR adds the necessary `std::` namespace prefix for references to types from this namespace in the daal4py source files. It looks like some update in the compiler somehow ended up making these not be in the global namespace by the time it tries to compile these files.

---

**Testing**

### DIFF
--- a/src/daal4py.cpp
+++ b/src/daal4py.cpp
@@ -229,20 +229,20 @@ PyObject * make_nda(daal::data_management::NumericTablePtr * ptr)
         if ((res = _make_nda_from_csr<float, NPY_FLOAT32>(ptr)) != NULL) return res;
         break;
     case daal::data_management::data_feature_utils::DAAL_INT32_S:
-        if ((res = _make_nda_from_homogen<int32_t, NPY_INT32>(ptr)) != NULL) return res;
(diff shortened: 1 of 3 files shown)


david-cortes-intel:
> Could you detail this a bit more? I am thinking its related to #2484 but its not clear to me.

Including `<cstdint>` does not put the types into the global namespace the same way including `<stdint.h>` would.

> Is this specific to one compiler, or to g++? Of course the code looks fine etc.

I got a compilation error with the latest version of ICX. I understand it uses headers from elsewhere so not sure what specific compiler has the issue. But either way: it's not guaranteed that the types will be in the global namespace just from the numpy includes.

icfaust:
> > Could you detail this a bit more? I am thinking its related to #2484 but its not clear to me.
> 
> Including `<cstdint>` does not put the types into the global namespace the same way including `<stdint.h>` would.
> 
> > Is this specific to one compiler, or to g++? Of course the code looks fine etc.
> 
> I got a compilation error with the latest version of ICX. I understand it uses headers from elsewhere so not sure what specific compiler has the issue. But either way: it's not guaranteed that the types will be in the global namespace just from the numpy includes.

gotcha. May be worth writing the compiler team if this popped up using something newer than 2025.1

PR #2604, opened 2025-07-10
FIX: Sort indices of sparse matrices

david-cortes-intel:
## Description

ref https://github.com/uxlfoundation/scikit-learn-intelex/issues/1880

This PR forcibly sorts the indices of sparse matrices before passing them to any routines in oneDAL if they aren't already sorted, as some algorithms assume that they come sorted and produce incorrect results when they aren't.

It's unclear yet which algorithms require this and which ones do not. For efficiency reasons, this additional step should be skipped in algorithms that don't need it, but that would need to be done over different PRs later on.

~It's also unclear yet if this is something that should be done on the oneDAL side instead, so leaving it as a draft PR for now.~ From internal discussions, this is a reasonable approach to follow, so moving out of draft.

---

**Testing**

**Performance**

### DIFF
--- a/onedal/datatypes/numpy/data_conversion.cpp
+++ b/onedal/datatypes/numpy/data_conversion.cpp
@@ -153,7 +153,10 @@ inline csr_table_t convert_to_csr_impl(PyObject *py_data,
     return res_table;
 }
 
-dal::table convert_to_table(py::object inp_obj, py::object queue, bool recursed) {
+dal::table convert_to_table(py::object inp_obj,
+                            py::object queue,
+                            bool recursed,
+                            bool require_sparse_with_sorted_indices) {
     dal::table res;
 
     PyObject *obj = inp_obj.ptr();
@@ -213,6 +216,11 @@ dal::table convert_to_table(py::object inp_obj, py::object queue, bool recursed)
     }
     else if (strcmp(Py_TYPE(obj)->tp_name, "csr_matrix") == 0 ||
              strcmp(Py_TYPE(obj)->tp_name, "csr_array") == 0) {
+        if (require_sparse_with_sorted_indices) {
+            if (!py::getattr(obj, "has_sorted_indices").cast<bool>()) {
+                py::reinterpret_borrow<py::object>(obj).attr("sort_indices")();
+            }
+        }
         PyObject *py_data = PyObject_GetAttrString(obj, "data");
         PyObject *py_column_indices = PyObject_GetAttrString(obj, "indices");
         PyObject *py_row_indices = PyObject_GetAttrString(obj, "indptr");
--- a/sklearnex/utils/validation.py
+++ b/sklearnex/utils/validation.py
@@ -102,7 +102,6 @@ def validate_data(
     /,
     X="no_validation",
     y="no_validation",
-    require_sparse_with_sorted_indices=True,
     **kwargs,
 ):
     # force finite check to not occur in sklearn, default is True
@@ -135,14 +134,6 @@ def validate_data(
             assert_all_finite(next(arg), allow_nan=allow_nan, input_name="X")
         if check_y:
             assert_all_finite(next(arg), allow_nan=allow_nan, input_name="y")
-
-    if (
-        require_sparse_with_sorted_indices
-        and sp.issparse(X)
-        and hasattr(X, "has_sorted_indices")
-        and not X.has_sorted_indices
-    ):
-        X.sort_indices()
 
     return out
 
--- a/onedal/datatypes/numpy/data_conversion.hpp
+++ b/onedal/datatypes/numpy/data_conversion.hpp
@@ -32,6 +32,7 @@ namespace py = pybind11;
 PyObject *convert_to_pyobject(const dal::table &input);
 dal::table convert_to_table(py::object inp_obj,
                             py::object queue = py::none(),
-                            bool recursed = false);
+                            bool recursed = false,
+                            bool require_sparse_with_sorted_indices = true);
 
 } // namespace oneapi::dal::python::numpy


david-cortes-intel:
CI error is just a timeout.

sklearnex/utils/validation.py
```diff
@@ -135,6 +136,14 @@ def validate_data(
         if check_y:
             assert_all_finite(next(arg), allow_nan=allow_nan, input_name="y")
 
+    if (
+        require_sparse_with_sorted_indices
+        and sp.issparse(X)
+        and hasattr(X, "has_sorted_indices")
+        and not X.has_sorted_indices
+    ):
+        X.sort_indices()
```
Vika-F: sort_indices() does the sorting inplace.
Is it Ok to alter the user-provided data during the compuattions?

david-cortes-intel: I guess it should be, because that's just an implementation detail of SciPy which doesn't alter the results of operations after modifying.

@icfaust Any comments here?

david-cortes-intel: For reference, it looks like sklearn also modifies sparse data in-place like that even when not having a `copy_X` argument:
https://github.com/scikit-learn/scikit-learn/blob/9b7a86fb6d45905eec7b7afd01d3ae32643c8180/sklearn/svm/_base.py#L359

.. so I guess it should be fine to do the same.

Vika-F: Yes, I understand that the matrix stays the same in some sense, but the initial order of the indices and values might be important to the user for some reason, and after X.sort_indices() the order would be altered.

sklearnex/utils/validation.py
```diff
@@ -135,6 +136,14 @@ def validate_data(
         if check_y:
             assert_all_finite(next(arg), allow_nan=allow_nan, input_name="y")
 
+    if (
+        require_sparse_with_sorted_indices
+        and sp.issparse(X)
+        and hasattr(X, "has_sorted_indices")
```
ahuber21: Don't quite understand the `if` chain here. Wouldn't it break if we don't have the attribute `has_sorted_indices`, but the indices are unsorted? I wouldn't unnecessarily add the assumption of getting passed `scipy` matrices (which provide this attribute).

In other words, shouldn't we rather use `not getattr(X, "has_sorted_indices", False)`?

david-cortes-intel: It is only an issue with matrices in compressed format. Other formats like COO are converted to that and produce something with sorted indices as a result.

ahuber21: And those matrices don't have `has_sorted_indices`, but implicitly always have sorted indices? Then it all makes sense.

david-cortes-intel: They don't have the attribute, and are not guaranteed to have sorted indices either, but the conversion functions from SciPy to compressed formats from other formats always output something with sorted indices.

david-cortes-intel:
Moved the sorting part inside the `to_table` function as suggested instead of having it in the validator.

david-cortes-intel:
Failures appear to be due to timeouts. Waiting for a review from @icfaust after moving the sorting to `to_table`.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -213,6 +216,11 @@ dal::table convert_to_table(py::object inp_obj, py::object queue, bool recursed)
     }
     else if (strcmp(Py_TYPE(obj)->tp_name, "csr_matrix") == 0 ||
              strcmp(Py_TYPE(obj)->tp_name, "csr_array") == 0) {
+        if (require_sparse_with_sorted_indices) {
+            if (!py::getattr(obj, "has_sorted_indices").cast<bool>()) {
+                py::reinterpret_borrow<py::object>(obj).attr("sort_indices")();
```
icfaust, line 221: An absolute nitpick: inp_obj may have the info you want instead of needing to use a reinterpret_borrow.

sklearnex/svm/tests/test_svm.py
```diff
@@ -91,3 +96,29 @@ def test_sklearnex_import_nusvr(dataframe, queue):
         _as_numpy(svc.dual_coef_), [[-1.0, 0.611111, 1.0, -0.611111]], rtol=1e-3
     )
     assert_allclose(_as_numpy(svc.support_), [1, 2, 3, 5])
+
+
+# https://github.com/uxlfoundation/scikit-learn-intelex/issues/1880
```
icfaust, line 101: Thank you for this, this will help guide devs in the future a lot.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -153,7 +153,10 @@ inline csr_table_t convert_to_csr_impl(PyObject *py_data,
     return res_table;
 }
 
-dal::table convert_to_table(py::object inp_obj, py::object queue, bool recursed) {
+dal::table convert_to_table(py::object inp_obj,
+                            py::object queue,
+                            bool recursed,
+                            bool require_sparse_with_sorted_indices) {
```
icfaust, line 159: I'm debating whether require_sparse_with_sorted_indices should go before recursed, but I think that could be done in a follow-up when we run into a scenario when someone would want to set it as false, doesn't need to be done now.

onedal/datatypes/numpy/data_conversion.cpp
david-cortes-intel: The input is `PyObject`. Getting the attribute out of it without `py::reinterpret` would require using python C api methods instead of pybind11.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -213,6 +216,11 @@ dal::table convert_to_table(py::object inp_obj, py::object queue, bool recursed)
     }
     else if (strcmp(Py_TYPE(obj)->tp_name, "csr_matrix") == 0 ||
              strcmp(Py_TYPE(obj)->tp_name, "csr_array") == 0) {
+        if (require_sparse_with_sorted_indices) {
+            if (!py::getattr(obj, "has_sorted_indices").cast<bool>()) {
```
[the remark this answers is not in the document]
icfaust: Not a problem CoPilot, statement before does a guarantee on that (though I am not a fan that it checks based on the type name, probably could be something more exact).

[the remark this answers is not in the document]
david-cortes-intel: Comment is wrong - code here is only reached if the input is a CSR matrix, which is guaranteed to have this attribute.

PR #2610, opened 2025-07-14
FIX: Fix double division of logistic model regularization by sum of weights

david-cortes-intel:
## Description

This PR fixes an issue in which the regularization of logistic regression models passed to oneDAL is incorrect due to being divided by the sum of weights (or number of rows when no weights are present) twice when used from `sklearnex.linear_model.LogisticRegression`.

Note: I am not sure here if the function being modified is otherwise also accessible from some daal4py class or similar or if this might break something in daal4py outside of sklearnex.

---

**Testing**

### DIFF
--- a/sklearnex/linear_model/tests/test_logreg.py
+++ b/sklearnex/linear_model/tests/test_logreg.py
@@ -19,7 +19,8 @@ import pytest
 from numpy.testing import assert_allclose, assert_array_equal
 from scipy.sparse import csr_matrix
 from sklearn.datasets import load_breast_cancer, load_iris, make_classification
-from sklearn.metrics import accuracy_score
+from sklearn.linear_model import LogisticRegression as origLogisticRegression
+from sklearn.metrics import accuracy_score, log_loss
 from sklearn.model_selection import train_test_split
 
 from daal4py.sklearn._utils import daal_check_version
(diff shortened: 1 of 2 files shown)


sklearnex/linear_model/tests/test_logreg.py
```diff
@@ -19,7 +19,8 @@
 from numpy.testing import assert_allclose, assert_array_equal
 from scipy.sparse import csr_matrix
 from sklearn.datasets import load_breast_cancer, load_iris, make_classification
-from sklearn.metrics import accuracy_score
+from sklearn.linear_model import LogisticRegression as origLogisticRegression
```
icfaust:
```suggestion
from sklearn.linear_model import LogisticRegression as _sklearn_LogisticRegression

```
I'm not certain, just worried about how this interacts with sklearn's estimator discovery (```all_estimators```) which would see this as public, and would require an underscore (may be fine that its in a test file, not sure).

sklearnex/linear_model/tests/test_logreg.py
```diff
@@ -132,3 +133,31 @@ def test_csr(queue, dtype, dims):
         assert_allclose(prob, prob_sp, rtol=rtol)
         assert_allclose(model.coef_, model_sp.coef_, rtol=rtol)
         assert_allclose(model.intercept_, model_sp.intercept_, rtol=rtol)
+
+
+def test_logistic_regression_is_correct():
```
icfaust, line 145: This is technically a conformance test. Thoughts @ethanglaser ?

sklearnex/linear_model/tests/test_logreg.py
david-cortes-intel: Renamed.

sklearnex/linear_model/tests/test_logreg.py
david-cortes-intel: It is taken from one of the conformance tests that was failing, but since we don't expect the test to pass due to the way that sklearnex approaches logistic regression, it tests the function value instead (oneDAL's result is oftentimes better than sklearn's).

icfaust: Ahhhh, interesting.  Tbh, I am not opposed to bringing it in as it is, though if you could comment to the related sklearn conformance test that would be good.

david-cortes-intel: Added a comment about it.

ethanglaser: In the event that we have a meaningful conformance test that must be deselected due to some numerical/minor differences in approach I think its a good idea to add a variant to our own testing

PR #2614, opened 2025-07-16
MAINT: Deselect failing tests for logistic regression

david-cortes-intel:
## Description

As a follow up from https://github.com/uxlfoundation/scikit-learn-intelex/pull/2465

There is a tests file from sklearn for logistic regression which runs unpatched in conformance tests due to the way that it imports the estimators:
https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/linear_model/tests/test_logistic.py

A change in a different test made the pytest collector override the global variables from that file when running the full conformance tests, which leads to them now running patched and throwing lots of errors. The previous PR deselected the test that brought the override, and it worked in the sense that the other tests for logistic regression still run unpatched and thus pass, but this is not an ideal scenario and could break again due to global variables being overridden.

This PR instead enables back the unrelated test and deselects the tests that would fail when they run patched.

Requires the following PR for the remainder of the tests to pass:
https://github.com/uxlfoundation/oneDAL/pull/3293

Note that the last deselection here currently shows totally different test results, but the differences will become just small numeric differences after this PR:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/2639

---

**Testing**


deselected_tests.yaml
```diff
@@ -398,6 +395,26 @@ deselected_tests:
   - tests/test_common.py::test_estimators[NuSVC()-check_class_weight_classifiers]
   - tests/test_common.py::test_estimators[CalibratedClassifierCV(estimator=LogisticRegression(C=1))-check_sample_weights_invariance(kind=ones)]
 
+  # Logistic regression is not expected to give exact matches due to small differences in numerical tolerances
```
icfaust, line 398: Could you elaborate on this in the description? Is it a setting issue, or something numerical in the loss function, as we see a lot of warnings associated with convergence in testing both in public and private CI.

david-cortes-intel: OneDAL clips numbers when they get too close to zero or one. Scikit-learn used to do that, but now has a different logic with many if-elses that differs from flat clipping.

icfaust:
Just to be clear @david-cortes-intel #2639 doesn't solve any of these?

david-cortes-intel:
> Just to be clear @david-cortes-intel #2639 doesn't solve any of these?

It "solves" them in the sense that afterwards, the differences will be very small, but still not conforming to the default `rtol=1e-7`, whereas currently the differences are very big and the numbers being compared in tests look totally off.

icfaust:
> > Just to be clear @david-cortes-intel #2639 doesn't solve any of these?
> 
> It "solves" them in the sense that afterwards, the differences will be very small, but still not conforming to the default `rtol=1e-7`, whereas currently the differences are very big and the numbers being compared in tests look totally off.

The reason I ask is to see if the PR was properly tested with these deselections removed, which it seems you have done. So at the end, because of the nature of the sklearn implementation and its weak precision, we are hitting a moving target. Sort of curious what is necessary to bring it in line, or what you envision as the next steps for this algo (doesn't impact the mergability of this PR).

deselected_tests.yaml
icfaust: From a maintenance perspective, your comment here is more actionable (just as an FYI).

david-cortes-intel:
> > > Just to be clear @david-cortes-intel #2639 doesn't solve any of these?
> > 
> > 
> > It "solves" them in the sense that afterwards, the differences will be very small, but still not conforming to the default `rtol=1e-7`, whereas currently the differences are very big and the numbers being compared in tests look totally off.
> 
> The reason I ask is to see if the PR was properly tested with these deselections removed, which it seems you have done. So at the end, because of the nature of the sklearn implementation and its weak precision, we are hitting a moving target. Sort of curious what is necessary to bring it in line, or what you envision as the next steps for this algo (doesn't impact the mergability of this PR).

Making them match to more decimal places would require following the same logic that they use for handling probabilities close to zero and to one in their computations:
https://github.com/scikit-learn/scikit-learn/blob/9b7a86fb6d45905eec7b7afd01d3ae32643c8180/sklearn/_loss/_loss.pyx.tp#L728

But doing that in oneDAL would make it slower, since those are row-by-row calls to 'exp'/'log' (thus not vectorizable) and oneDAL cannot call `std::exp`, requiring instead to use one function call to MKL for each 'exp'/'log' call. If oneDAL used their logic, it would have to do one or two MKL function calls per row, which would end up being slower than sklearn's.

PR #2637, opened 2025-07-30
[fix] Add capability with DLManagedTensorVersioned

icfaust:
## Description

This PR updates dlpack support to include modern capabilities including:

 - Output conversion of oneDAL tables to DLManagedTensorVersioned objects, which is now default in NumPy, and will be in the next PyTorch release
 - Add support for keywords made available in the array API 2023.12 standard
 - Modify default behavior to make output data not read only (will be modified for all output table conversions in the future, but this makes all conversions consistent)
 - Allow for the ability to convert to host when requested by the ```dl_device``` keyword argument.
 - Small refactor to allow for re-use of code
 - use ```is_none``` call instead of ```.is(py::none())``` to simplify code
 - update dlpack.h from v1.0 to v1.1
 
No performance benchmarks necessary, tests added to cover new conditions.

---

**Testing**

**Performance**


onedal/datatypes/_data_conversion.py
```diff
@@ -111,7 +111,9 @@ def return_type_constructor(array):
                 else xp.asarray(backend.from_table(x), device=device)
             )
     elif hasattr(array, "__array_namespace__"):
-        func = array.__array_namespace__().from_dlpack
+        xp = array.__array_namespace__()
+        device = array.device
+        func = lambda inp: xp.from_dlpack(inp, device=device)
```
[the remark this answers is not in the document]
icfaust: Can be addressed as a follow up if considered to be an expected  failure mode as we would need to modify more in this function in that case.

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -178,77 +179,131 @@ DLDevice get_dlpack_device(const dal::table& input) {
...
+py::capsule construct_dlpack(const dal::table& input,
+                             py::object max_version,
+                             py::object dl_device,
+                             py::object copyobj) {
     // check table type and expose oneDAL array
     if (input.get_kind() != dal::homogen_table::kind())
-        throw pybind11::type_error("Unsupported table type for dlpack conversion");
+        throw py::type_error("Unsupported table type for dlpack conversion");
+
+    bool copy = !copyobj.is(py::bool_(false));
```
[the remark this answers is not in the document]
icfaust: It is intended behavior, will be swapped in a future PR.

onedal/datatypes/_data_conversion.py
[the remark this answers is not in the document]
icfaust: no

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -178,77 +179,131 @@ DLDevice get_dlpack_device(const dal::table& input) {
...
-    dlm->deleter = [](struct DLManagedTensor* self) -> void {
-        auto stored_array = static_cast<dal::array<byte_t>*>(self->manager_ctx);
-        if (stored_array) {
-            delete stored_array;
+    // oneDAL tables are by definition immutable and must be made mutable via a copy.
+    if (copy)
+        array.need_mutable_data();
+
+    if (max_version.is_none() ||
+        max_version.cast<py::tuple>()[0].cast<int>() < DLPACK_MAJOR_VERSION) {
```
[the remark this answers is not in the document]
icfaust: The current logic is sufficient.

onedal/datatypes/common.hpp
```diff
@@ -0,0 +1,47 @@
...
+*******************************************************************************/
+
+#include "oneapi/dal/array.hpp"
+
+namespace oneapi::dal::python {
+
+template <typename T>
+dal::array<T> transfer_to_host(const dal::array<T> &array) {
+#ifdef ONEDAL_DATA_PARALLEL
+    auto opt_queue = array.get_queue();
```
david-cortes-intel, line 24: Will GPU arrays always necessarily have an associated queue?

onedal/datatypes/tests/test_data.py
```diff
@@ -616,3 +616,44 @@ def test_table___dlpack__(dataframe, queue, order, data_shape, dtype):
...
+
+
+@pytest.mark.skipif(
+    not hasattr(np, "from_dlpack"), reason="no dlpack support in installed numpy"
+)
+@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues("dpctl", "cpu,gpu"))
+@pytest.mark.parametrize("order", ["F", "C"])
+@pytest.mark.parametrize("data_shape", data_shapes)
+@pytest.mark.parametrize("dtype", [np.float32, np.float64, np.int32, np.int64])
+def test_table_cvt_to_host_dlpack(dataframe, queue, order, data_shape, dtype):
```
david-cortes-intel: Does 'cvt' mean 'convert'? If so, perhaps the longer name could be used.

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -178,77 +179,131 @@ DLDevice get_dlpack_device(const dal::table& input) {
...
+                                                     homogen_input.get_column_count(),
+                                                     homogen_input.get_metadata().get_data_type(0),
+                                                     homogen_input.get_data_layout(),
+                                                     copy);
+
+        // create capsule
+        capsule = py::capsule(static_cast<void*>(dlm), "dltensor", free_capsule);
+    }
+    else {
+        // assume RVO
```
david-cortes-intel: What does 'RVO' mean?

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -178,77 +179,131 @@ DLDevice get_dlpack_device(const dal::table& input) {
...
+
+        // by definition for oneDAL tables, unless a copy of the array is made, it is readonly.
+        if (copy) {
+            dlmv->flags = DLPACK_FLAG_BITMASK_IS_COPIED;
         }
-        delete[] self->dl_tensor.shape;
-        delete self;
-    };
+        else {
+            dlmv->flags = DLPACK_FLAG_BITMASK_READ_ONLY;
```
david-cortes-intel, line 300: Are there tests checking for this code branch and property?

onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -178,77 +179,131 @@ DLDevice get_dlpack_device(const dal::table& input) {
...
-    if (PyCapsule_IsValid(cap, "dltensor")) {
-        dlm = static_cast<DLManagedTensor*>(PyCapsule_GetPointer(cap, "dltensor"));
-        if (dlm->deleter) {
-            dlm->deleter(dlm);
+static inline void move_dlpack_data(dal::array<byte_t>& array, py::tuple dl_device, bool copy) {
+    DLDevice requested{ dl_device[0].cast<DLDeviceType>(), dl_device[1].cast<std::int32_t>() };
+#ifdef ONEDAL_DATA_PARALLEL
+    DLDevice current = get_dlpack_device(array);
+    // only allow transfers to host, dlpack implementation does not provide sufficient
+    // fine-grained control for SYCL devices.
```
david-cortes-intel, line 230: Is this because dlpack doesn't include the queue object?

icfaust: That is correct!

EDIT: To be fully explicit, since oneDAL being context agnostic via the queues (which limits context handling and other specific things), it would be a burden and require some development to create an allocation on another device without assuming the default context, which we wouldn't know about from the dl_device info given.  Ideally the user should handle movement from sycl devices to other sycl devices on their own, since we in oneDAL only handle from host to device and back. There some tool like dpctl may be better, or using the native conversions in the external library itself.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: My bad, a small note if someone looks at why I am not creating the object and then passing it in by reference. https://en.wikipedia.org/wiki/Copy_elision

onedal/datatypes/common.hpp
icfaust: get_queue returns a ```std::optional``` when used in a dpc build which its own pain. Good catch.

onedal/datatypes/tests/test_data.py
icfaust: Sounds good. Will change.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: Added.

onedal/datatypes/dlpack/data_conversion.cpp
david-cortes-intel: Could you add a comment in the code?

onedal/datatypes/common.hpp
david-cortes-intel: Then wouldn't this function produce something that will later on cause errors if the array is on GPU but has no queue?

icfaust: That would never be the case, oneDAL gpu data arrays always have a queue. @Vika-F @Alexandr-Solovev can comment more

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: To be able to fulfill your request, please let me know in more detail what beyond what is already written in the comment is necessary. @david-cortes-intel

david-cortes-intel: It says "assume RVO". What does that mean? Why can it be assumed in this 'if' branch but not the other?

onedal/datatypes/common.hpp
```diff
@@ -0,0 +1,47 @@
...
+    if (opt_queue.has_value()) {
+        auto device = opt_queue->get_device();
+        if (!device.is_cpu()) {
+            const auto *device_data = array.get_data();
+
+            auto memory_kind = sycl::get_pointer_type(device_data, opt_queue->get_context());
+            if (memory_kind == sycl::usm::alloc::unknown) {
+                throw std::runtime_error("[convert_to_numpy] Unknown memory type");
+            }
+            if (memory_kind == sycl::usm::alloc::device) {
```
Alexandr-Solovev, line 34: it could be also shared allocation. Do you handle this case?

icfaust: Was not in the original implementation in ```onedal/datatypes/numpy/data_conversion.cpp``` but is now added to call ```wait``` on the queue to ensure data synchronization.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: I realized now writing it out that since its just a pointer its not a big deal, it was previously part a function which returned a struct, so I had been worried about the copy of it, especially with respect to C++ arrays vs the C pointer used in dlpack.h. After the development, the discussion of return value optimization is a bit moot, but added it anyway. @david-cortes-intel

david-cortes-intel: I still don't get what can be assumed to be a return value optimization here, since the return value is a `py::capsule`, which behind the scenes is just a `PyObject*` (i.e. a pointer) AFAIK.

But since now the same comment appears under both if-else branches, maybe it could be moved to the beginning, or removed if it's not applicable.

icfaust: Comment has been removed.

david-cortes-intel (approved):
LGTM, pending CI results after merging the latest main, hopefully with this other PR included: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2656

icfaust:
Private CI failures unrelated (numpy install issue?)

PR #2639, opened 2025-07-31
FIX, MAINT: Offload LogisticRegression to sklearn when oneDAL is not used, fix various issues

david-cortes-intel:
## Description

This PR refactors the custom code for logistic regression, solving some issues along the way and adding comments where appropriate.

Currently, the code for logistic regression appears to be a copy-paste from an old version of scikit-learn, with some sections adapted to call functions from oneDAL. Lots of what it does is just re-implementing the exact same code as in sklearn without any acceleration from oneDAL side, so this PR changes the logic to call sklearn directly instead of reimplementing code when there won't be any calls to oneDAL involved.

Along the way, two issues in cases where oneDAL gets used are fixed here:
* Solver newton-cg outputs the coefficients multiplied by 2, when it shouldn't.
* Warm starts with more than one class reinterpret row-major coefficients as column-major.

I've also updated the documentation to reflect the actual situation, since it ways that all parameters are supported but in lots of cases it doesn't end up doing anything differently from sklearn.

---

**Testing**


daal4py/sklearn/linear_model/logistic_path.py
```diff
@@ -682,6 +685,8 @@ def _func_(x, *args):
                     maxiter=max_iter,
                     tol=tol,
                 )
+                if _dal_ready and C_daal_multiplier == 2:
```
icfaust: Are there other values beyond 2? (i.e. any idea why they sorta hacked this in?)

daal4py/sklearn/linear_model/tests/test_logreg.py
```diff
@@ -128,15 +135,339 @@ def test_multinomial_logistic_regression_is_correct_with_weights(
...
+    if n_classes == 2:
+        np.testing.assert_allclose(
+            model_sklearnex_refitted.intercept_, model_sklearn.intercept_, atol=1e-3
+        )
+    else:
+        np.testing.assert_allclose(
+            model_sklearnex_refitted.intercept_
+            - model_sklearnex_refitted.intercept_.sum(),
+            model_sklearn.intercept_ - model_sklearn.intercept_.sum(),
+            atol=1e-2,
```
icfaust: I guess with respect of the comment of the test these tolerances may be fine, but in general 1e-2 is eye opening.  I there a reason why you are using ```sum``` here?

daal4py/sklearn/linear_model/tests/test_logreg.py
```diff
@@ -128,15 +135,339 @@ def test_multinomial_logistic_regression_is_correct_with_weights(
...
+        solver == "newton-cholesky"
+        and multi_class == "multinomial"
+        and not sklearn_check_version("1.6")
+    ):
+        pytest.skip("Functionality introduced in a later scikit-learn version.")
+
+    if (
+        solver == "newton-cholesky"
+        and multi_class == "multinomial"
+        and not sklearn_check_version("1.9")
```
icfaust: Is this a poison pill for later like in the previous test, it was set at sklearn 1.8, should they be the same or are the different on purpose?

daal4py/sklearn/linear_model/tests/test_logreg.py
```diff
@@ -128,15 +135,339 @@ def test_multinomial_logistic_regression_is_correct_with_weights(
...
+    if (
+        solver == "newton-cholesky"
+        and multi_class == "multinomial"
+        and not sklearn_check_version("1.6")
+    ):
+        pytest.skip("Functionality introduced in a later scikit-learn version.")
+
+    # Note: scikit-learn itself has bugs that would make it fail this test
+    # under some versions (starting with 1.6) but not others.
+    # TODO: Revisit whether this gets fixed in scikit-learn==1.9 once it gets
```
icfaust: If there is a linked issue on sklearn, this would be a great place to add a link.

daal4py/sklearn/linear_model/tests/test_logreg.py
```diff
@@ -128,15 +135,339 @@ def test_multinomial_logistic_regression_is_correct_with_weights(
...
+        # in the calculations, so when comparing against sklearn, we want
+        # to avoid ending up with large numbers in the computations.
+        class_sep=0.25,
+    )
+
+    # Note1: these will throw warnings due to reaching the maximum
+    # number of iterations without converging, which is expected
+    # given that those are being severely limited for the tests.
+    # Note2: this will first compare the results after one iteration, and
+    # if those already differ too much (which can be the case given numerical
```
icfaust: This would suggest that the loss calculation isn't deterministic given the random_state setting.  Is that the case (if so probably by multithreading?)?  While failures aren't sporadic, the test itself will be sporadic meaning that warm_start isn't working right.

daal4py/sklearn/linear_model/logistic_path.py
david-cortes-intel: No, it's hard coded in this same file.

daal4py/sklearn/linear_model/tests/test_logreg.py
david-cortes-intel: Yes, this should be `.mean()`. Fixed.

> but in general 1e-2 is eye opening

Yes, it can indeed produce large differences, because the solver has low precision in both sklearn and sklearnex.

daal4py/sklearn/linear_model/tests/test_logreg.py
david-cortes-intel: Should be the same version. Corrected the other to be at 1.9.

daal4py/sklearn/linear_model/tests/test_logreg.py
david-cortes-intel: Added.

daal4py/sklearn/linear_model/tests/test_logreg.py
david-cortes-intel: The loss calculation is deterministic, but some of the solvers that use it are not. The 's' in 'sag'/'saga' stands for 'stochastic'.

david-cortes-intel:
> Thank you for this intense amount of labor. With all the changes going into logisticregression, how far off is the l-BFGS solver from being usable vs. the maintenance that this has required? I am wondering if its just worth us doing the proper thing and getting the logisticRegression in DAAL functional, since at some point we should go that route. (Not that it has any bearing on this quality or nature of this PR). Some questions, and the amount of testing added is good.

The L-BFGS solver in particular will be in a usable state after all these PRs and using patching behind the scenes in most cases, just not expected to match sklearn results to many decimal places when running patched, and not providing much of a speed up. But it is indeed quite a maintenance burden (I don't know if there's other details that differ between the two codebases, they are quite hard to compare) and even if there's no DAAL C++ version, maybe it could be done better by calling sklearn's own classes with monkeypatched internal functions instead.

But if a DAAL version were to be introduced, chances are that it would match against sklearn's results to even fewer decimal places, since sklearn's implementation by default is not as precise as other libraries.

daal4py/sklearn/linear_model/tests/test_logreg.py
icfaust: Fair enough, then why does the test fully utilize an unsupported solver (I see a comment in Note 2 related, but doesn't directly answer this), since we don't do any offloading to DAAL in either 'sag' or 'saga'. https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/daal4py/sklearn/linear_model/logistic_path.py#L329 (It also looks like the documentation here is wrong: https://uxlfoundation.github.io/scikit-learn-intelex/2025.7/algorithms.html#classification

icfaust: I went and looked at the docs, ideally these stochastic solvers should be made deterministic via the random_state parameter. https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html

david-cortes-intel: In order to verify that it offloads it correctly to sklearn under all those cases. The only completely direct offload happens here:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/7d88e6eb8796f9cf50c2723b085fca810b479c0b/daal4py/sklearn/linear_model/logistic_path.py#L267

But that part is not reached in all of these test cases, even when using an unsupported solver.

Although it looks like the check could be moved earlier so that it would instead follow that direct offload.

david-cortes-intel: > I went and looked at the docs, ideally these stochastic solvers should be made deterministic via the random_state parameter. https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html

In this file, they are supplied `random_state` in all of the cases where a stochastic solver is used.

david-cortes-intel:
@icfaust I've rewritten the file to offload early to sklearn when there's no oneDAL involvement, and updated the PR description. Please re-review.

daal4py/sklearn/linear_model/logistic_path.py
```diff
@@ -57,20 +57,31 @@
...
         _multinomial_loss_grad,
-        _multinomial_grad_hess,
-        _LOGISTIC_SOLVER_CONVERGENCE_MSG,
-        LogisticRegression as LogisticRegression_original,
     )
 
+if sklearn_check_version("1.7.1"):
+    from sklearn.utils.fixes import _get_additional_lbfgs_options_dict
+else:
+    # From https://github.com/scikit-learn/scikit-learn/blob/760edca5fb5cc3538b98ebc55171806e2a6e3e84/sklearn/utils/fixes.py#L408
```
icfaust, line 79: wonderful comment.

david-cortes-intel:
I just discovered that there's a custom decorator `@pytest.mark.allow_sklearn_fallback` that can be used in the tests. I've moved all the tests for logistic regression under `sklearnex` now.

david-cortes-intel:
CI error appears to be in PCA, so not related to the changes here.

Alexsandruss (approved):

doc/sources/algorithms.rst
```diff
@@ -69,8 +69,13 @@ Classification
...
    * - :obj:`sklearn.linear_model.LogisticRegression`
-     - All parameters are supported
-     - No limitations
+     - All parameters are supported except:
+
+       - ``solver`` not in [``'lbfgs'``, ``'newton-cg'``]
+       - ``penalty`` in [``'l1'``, ``'elasticnet'``]
+       - ``sample_weight`` != ``None``
+       - ``class_weight`` != ``None``
+     - Sparse data is not supported.
```
ethanglaser, line 78: I thought we do support sparse for logreg? @david-cortes-intel @avolkov-intel

david-cortes-intel: Only on GPU as far as I can tell.

PR #2647, opened 2025-08-05
Enable common verbosity arguments for pytest

Alexsandruss:
## Description

Changes:
- Enable common verbosity arguments for pytest calls in sklearnex testing scripts and store them in one plain text file
- Enable test duration tracking (top-100 tests with >50ms runtime)

---

**Testing**

**Performance**

N/A.

### DIFF
--- a/.ci/scripts/run_sklearn_tests.py
+++ b/.ci/scripts/run_sklearn_tests.py
@@ -48,15 +48,7 @@ if __name__ == "__main__":
     if sklearn_check_version("1.6"):
         os.environ["SCIPY_ARRAY_API"] = "1"
 
-    pytest_verbosity_file = os.path.join(
-        os.path.dirname(os.path.dirname(os.path.dirname(__file__))),
-        ".pytest-verbosity-args",
-    )
-    with open(pytest_verbosity_file, "r") as f:
-        verbosity_args = f.read().strip()
-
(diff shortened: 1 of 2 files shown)


.ci/scripts/run_sklearn_tests.py
```diff
@@ -48,8 +48,15 @@
...
         os.environ["SCIPY_ARRAY_API"] = "1"
 
+    pytest_verbosity_file = os.path.join(
+        os.path.dirname(os.path.dirname(os.path.dirname(__file__))),
+        ".pytest-verbosity-args",
+    )
+    with open(pytest_verbosity_file, "r") as f:
+        verbosity_args = f.read().strip()
+
     pytest_args = (
```
david-cortes-intel, line 51: This file now allows forwarding extra arguments to pytest by passing those arguments to this file when executing it. These verbosity arguments could be moved to the .sh file instead - that way, they wouldn't always be added unconditionally when this file is used locally outside of the CI scripts.

Alexsandruss: You are right, it also makes passing args much shorter.

david-cortes-intel (approved):
LGTM. But is there a way to propagate these results to the JSON files? They are quite hard to locate otherwise given the length of the logs.

PR #2654, opened 2025-08-06
[ArrayAPI] Refactor KMeans estimator to follow oneDAL estimator design pattern

KateBlueSky:
## Refactor KMeans estimator to follow oneDAL estimator design pattern

This PR refactors the KMeans estimator to align with the standardized design pattern used across oneDAL estimators, such as DummyEstimator outlined in #2534 . The main goal is to make the estimator consistent, maintainable, and compatible with future extensions (e.g., other algorithms or backends).

Changes Made

- Reorganized KMeans and _BaseKMeans classes to follow the oneDAL estimator model pattern.

- Added backend bindings using @bind_default_backend decorators:

- train() → kmeans.clustering

- infer() → kmeans.clustering

- _is_same_clustering() → kmeans_common (no policy)

- Centralized creation of the params dictionary in _get_onedal_params(...).

- Ensured type dispatch and method dispatch use fptype and method respectively (e.g., by_default, lloyd_csr).

- Wrapped all input/output in to_table / from_table.

- Separated backend logic from estimator logic using _fit_backend() and _predict_backend().

- Deferred model creation and attribute assignment to follow a minimalistic and clean init/finalization process.

- Applied Sycl queue support via @supports_queue decorators for fit/predict/score.

- Removed redundant or sklearn-only attributes that aren’t required by oneDAL estimators.

- Preserved full feature parity (e.g., init modes, scoring, CSR support, random_state handling, etc.).

---

**Testing**

**Performance**


onedal/cluster/kmeans.py
```diff
@@ -84,52 +104,33 @@ def _validate_center_shape(self, X, centers):
...
-        # TODO(1.4): Remove
+        # n_init resolution (kept from your logic)
         self._n_init = self.n_init
         if self._n_init == "warn":
             warnings.warn(
                 (
                     "The default value of `n_init` will change from "
-                    f"{default_n_init} to 'auto' in 1.4. Set the value of `n_init`"
-                    " explicitly to suppress the warning"
+                    f"{default_n_init} to 'auto' in 1.4. Set `n_init` explicitly "
```
david-cortes-intel: Since we have a utility to check the sklearn version, this could be placed under an if-else, or removed altogether considering we appear to not support versions 1.1 through 1.3.

onedal/cluster/kmeans.py
```diff
@@ -180,123 +193,112 @@ def _init_centroids_onedal(
...
-                random_state=random_state,
-            )
+            centers, _ = _kmeans_plusplus(X, self.n_clusters, random_state=random_state)
         elif isinstance(init, str) and init == "random":
             seeds = random_state.choice(n_samples, size=self.n_clusters, replace=False)
             centers = X[seeds]
         elif callable(init):
             cc_arr = init(X, self.n_clusters, random_state)
-            cc_arr = np.ascontiguousarray(cc_arr, dtype=dtype)
+            if hasattr(cc_arr, "__array_namespace__"):
```
david-cortes-intel: Isn't the function `get_namespace` supposed to be doing these checks?

david-cortes-intel:
@yuejiaointel ~Not sure if this is an issue with the PR or with the GPU implementation of KMeans, but this hangs:~
```python
import os

import numpy as np
import dpnp

X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
Xd = dpnp.array(X, device="gpu")

from sklearnex import config_context
from sklearnex.cluster import KMeans

with config_context(array_api_dispatch=True):
    km = KMeans().fit(Xd)
    cl = km.predict(Xd)
```

Snippet above only hangs with some verbosity levels set on oneDAL. Runs fine under default settings.

david-cortes-intel:
Operations with data frames are also not working:
```python
import os

import numpy as np

X = rng.standard_normal(size=(1000, 20), dtype=np.float32)

from sklearnex import config_context
from sklearnex.cluster import KMeans

import polars as pl
Xdf = pl.DataFrame(X)
with config_context(array_api_dispatch=True, transform_output="polars"):
    km = KMeans().fit(Xdf)
    cl = km.transform(Xdf)
```

onedal/cluster/kmeans.py
david-cortes-intel: Bumping up comment.

david-cortes-intel:
@yuejiaointel Also having issues with torch:
```python
import os

import numpy as np
import torch

X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
Xt = torch.tensor(X, device="xpu")

from sklearnex import config_context
from sklearnex.cluster import KMeans

with config_context(array_api_dispatch=True):
    km = KMeans(n_clusters=3).fit(Xt)
    cl = km.predict(Xt)
```
```
TypeError: var() received an invalid combination of arguments - got (ddof=int, dtype=NoneType, out=NoneType, axis=int, ), but expected one of:
 * (tuple of ints dim, bool unbiased = True, bool keepdim = False)
 * (tuple of ints dim = None, *, Number correction = None, bool keepdim = False)
      didn't match because some of the keywords were incorrect: ddof, dtype, out, axis
 * (bool unbiased = True)
 * (tuple of names dim, bool unbiased = True, bool keepdim = False)
 * (tuple of names dim, *, Number correction = None, bool keepdim = False)
      didn't match because some of the keywords were incorrect: ddof, dtype, out, axis
```

onedal/cluster/kmeans.py
yuejiaointel: thx, removed!

onedal/cluster/kmeans.py
yuejiaointel: fixed, array_namespace check is removed. Init on GPU will fall back to sklearn via onedal_gpu_supported, and init_centroids_sklearn will only be called with np data

yuejiaointel:
> @yuejiaointel ~Not sure if this is an issue with the PR or with the GPU implementation of KMeans, but this hangs:~
> 
> ```python
> import os
> os.environ["SCIPY_ARRAY_API"] = "1"
> 
> import numpy as np
> import dpnp
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> Xd = dpnp.array(X, device="gpu")
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=True):
>     km = KMeans().fit(Xd)
>     cl = km.predict(Xd)
> ```
> 
> Snippet above only hangs with some verbosity levels set on oneDAL. Runs fine under default settings.

fixed
python -c "
> import os
> os.environ['SCIPY_ARRAY_API'] = '1'
> import numpy as np
> import dpnp
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> Xd = dpnp.array(X, device='gpu')
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> with config_context(array_api_dispatch=True):
>     km = KMeans().fit(Xd)
>     cl = km.predict(Xd)
> print('SUCCESS')
> print('labels:', type(cl).__name__, 'device:', cl.device, 'shape:', cl.shape)
> print('centers:', km.cluster_centers_.shape)
> print('inertia:', km.inertia_, 'n_iter:', km.n_iter_)
> "
SUCCESS
labels: dpnp_array device: Device(level_zero:gpu:0) shape: (1000,)
centers: (8, 20)
inertia: 16750.884765625 n_iter: 30

yuejiaointel:
> Operations with data frames are also not working:
> 
> ```python
> import os
> os.environ["SCIPY_ARRAY_API"] = "1"
> 
> import numpy as np
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> import polars as pl
> Xdf = pl.DataFrame(X)
> with config_context(array_api_dispatch=True, transform_output="polars"):
>     km = KMeans().fit(Xdf)
>     cl = km.transform(Xdf)
> ```



> Operations with data frames are also not working:
> 
> ```python
> import os
> os.environ["SCIPY_ARRAY_API"] = "1"
> 
> import numpy as np
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> import polars as pl
> Xdf = pl.DataFrame(X)
> with config_context(array_api_dispatch=True, transform_output="polars"):
>     km = KMeans().fit(Xdf)
>     cl = km.transform(Xdf)
> ```

Fixed

python -c "
> import os
> os.environ['SCIPY_ARRAY_API'] = '1'
> 
> import numpy as np
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> import polars as pl
> Xdf = pl.DataFrame(X)
> with config_context(array_api_dispatch=True, transform_output='polars'):
>     km = KMeans().fit(Xdf)
>     cl = km.transform(Xdf)
> 
> print('SUCCESS')
> print('type:', type(cl))
> print('shape:', cl.shape)
> print(cl.head(3))
> "
SUCCESS
type: <class 'polars.dataframe.frame.DataFrame'>
shape: (1000, 8)
shape: (3, 8)
┌──────────┬──────────┬──────────┬──────────┬──────────┬──────────┬──────────┬──────────┐
│ kmeans0  ┆ kmeans1  ┆ kmeans2  ┆ kmeans3  ┆ kmeans4  ┆ kmeans5  ┆ kmeans6  ┆ kmeans7  │
│ ---      ┆ ---      ┆ ---      ┆ ---      ┆ ---      ┆ ---      ┆ ---      ┆ ---      │
│ f64      ┆ f64      ┆ f64      ┆ f64      ┆ f64      ┆ f64      ┆ f64      ┆ f64      │
╞══════════╪══════════╪══════════╪══════════╪══════════╪══════════╪══════════╪══════════╡
│ 4.345361 ┆ 5.46025  ┆ 5.337861 ┆ 5.939007 ┆ 5.620585 ┆ 5.668931 ┆ 5.013967 ┆ 5.511316 │
│ 5.155623 ┆ 4.896421 ┆ 5.842982 ┆ 4.684233 ┆ 5.069546 ┆ 5.160526 ┆ 5.22758  ┆ 4.282856 │
│ 4.347491 ┆ 4.516921 ┆ 3.722448 ┆ 4.031341 ┆ 4.236632 ┆ 4.852192 ┆ 4.252797 ┆ 3.040989 │

yuejiaointel:
> ```python
> import os
> os.environ["SCIPY_ARRAY_API"] = "1"
> 
> import numpy as np
> import torch
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> Xt = torch.tensor(X, device="xpu")
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=True):
>     km = KMeans(n_clusters=3).fit(Xt)
>     cl = km.predict(Xt)
> ```



> @yuejiaointel Also having issues with torch:
> 
> ```python
> import os
> os.environ["SCIPY_ARRAY_API"] = "1"
> 
> import numpy as np
> import torch
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> Xt = torch.tensor(X, device="xpu")
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=True):
>     km = KMeans(n_clusters=3).fit(Xt)
>     cl = km.predict(Xt)
> ```
> 
> ```
> TypeError: var() received an invalid combination of arguments - got (ddof=int, dtype=NoneType, out=NoneType, axis=int, ), but expected one of:
>  * (tuple of ints dim, bool unbiased = True, bool keepdim = False)
>  * (tuple of ints dim = None, *, Number correction = None, bool keepdim = False)
>       didn't match because some of the keywords were incorrect: ddof, dtype, out, axis
>  * (bool unbiased = True)
>  * (tuple of names dim, bool unbiased = True, bool keepdim = False)
>  * (tuple of names dim, *, Number correction = None, bool keepdim = False)
>       didn't match because some of the keywords were incorrect: ddof, dtype, out, axis
> ```

fixed

 python -c "
> import os
> os.environ['SCIPY_ARRAY_API'] = '1'
> 
> import numpy as np
> import torch
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> Xt = torch.tensor(X, device='xpu')
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=True):
>     km = KMeans(n_clusters=3).fit(Xt)
>     cl = km.predict(Xt)
> 
> print('SUCCESS')
> print('labels:', type(cl).__name__, 'device:', cl.device, 'shape:', cl.shape)
> print('centers:', type(km.cluster_centers_).__name__, 'device:', km.cluster_centers_.device, 'shape:', km.cluster_centers_.shape)
> print('inertia:', km.inertia_, 'n_iter:', km.n_iter_)
> "
SUCCESS
labels: Tensor device: xpu:0 shape: torch.Size([1000])
centers: Tensor device: xpu:0 shape: torch.Size([3, 20])
inertia: 18364.755859375 n_iter: 20

onedal/cluster/kmeans.py
```diff
@@ -183,132 +136,137 @@ def _init_centroids_onedal(
...
-        dtype = X_table.dtype
+    # --- public API matched to the pattern ---
 
-        self._check_params_vs_input(X_table, is_csr, dtype=dtype)
+    @supports_queue
+    def fit(self, X, y=None, queue=None):
+        is_csr = _is_csr(X)
+        X_table = to_table(X, queue=queue)
+        dtype = X_table.dtype
+        self._input_type = X  # store for from_table(like=...) in properties
```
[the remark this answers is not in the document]
yuejiaointel: changed to return type constructor

onedal/cluster/kmeans.py
```diff
@@ -183,132 +136,137 @@ def _init_centroids_onedal(
...
+            default_n_init = 10
+            if isinstance(init, str) and init == "k-means++":
+                n_init = 1
+            elif isinstance(init, str) and init == "random":
+                n_init = default_n_init
+            elif callable(init):
+                n_init = default_n_init
+            else:  # array-like
+                n_init = 1
+
```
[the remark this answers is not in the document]
yuejiaointel: "warn" is resolved in sklearnex _resolve_n_init() before reaching onedal

sklearnex/cluster/k_means.py
```diff
@@ -153,21 +153,98 @@ def fit(self, X, y=None, sample_weight=None):
...
+            if not get_config()["use_raw_input"]:
+                X = validate_data(
+                    self,
+                    X,
+                    accept_sparse="csr",
+                    dtype=[xp.float64, xp.float32],
+                    order="C",
+                    copy=self.copy_x,
+                    accept_large_sparse=False,
+                    ensure_all_finite=False,
```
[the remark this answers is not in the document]
yuejiaointel: Removed — sklearn's KMeans uses default True; no reason to skip NaN/Inf check

sklearnex/cluster/k_means.py
```diff
@@ -86,7 +86,7 @@ def _initialize_onedal_estimator(self):
                 "init": self.init,
                 "max_iter": self.max_iter,
                 "tol": self.tol,
-                "n_init": self.n_init,
+                "n_init": self._n_init,
```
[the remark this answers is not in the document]
yuejiaointel: Changed to getattr(self, "_n_init", self._resolve_n_init())

sklearnex/cluster/k_means.py
```diff
@@ -332,8 +436,41 @@ def _check_test_data(self, X):
             )
             return X
 
-        _transform = support_input_format(_sklearn_KMeans._transform)
-        transform = support_input_format(_sklearn_KMeans.transform)
+        @wrap_output_data
+        def transform(self, X):
+            check_is_fitted(self)
+            return dispatch(
+                self,
```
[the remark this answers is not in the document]
yuejiaointel: not blocker

sklearnex/cluster/k_means.py
```diff
@@ -153,21 +153,98 @@ def fit(self, X, y=None, sample_weight=None):
...
+            # Validate init parameter if array-like (before X validation so n_features_in_ not set yet)
+            if _is_arraylike_not_scalar(self.init):
+                init = validate_data(
+                    self,
+                    self.init,
+                    dtype=[xp.float64, xp.float32],
+                    accept_sparse="csr",
+                    copy=True,
+                    order="C",
+                    reset=False,
```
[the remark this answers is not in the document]
yuejiaointel: Intentional — reset=False doesn't require n_features_in_ to exist

onedal/cluster/kmeans.py
```diff
@@ -52,15 +50,29 @@ def __init__(
         verbose,
         random_state,
         n_local_trials=None,
+        algorithm="lloyd",
     ):
+        # __init__ only stores user-visible params
```
david-cortes-intel: This contradicts the following lines like assigning to _tol.

sklearnex/cluster/k_means.py
```diff
@@ -153,21 +153,117 @@ def fit(self, X, y=None, sample_weight=None):
...
+                if isinstance(self.init, str) and self.init == "k-means++":
+                    n_init = 1
+                elif isinstance(self.init, str) and self.init == "random":
+                    n_init = default_n_init
+                elif callable(self.init):
+                    n_init = default_n_init
+                else:  # array-like
+                    n_init = 1
+
+            if _is_arraylike_not_scalar(self.init) and n_init != 1:
```
david-cortes-intel, line 178: Is `self.init` actually used by oneDAL? Is this part reachable after checking the patching conditions?

onedal/cluster/kmeans.py
yuejiaointel: Good catch these are old comment, removed

sklearnex/cluster/k_means.py
yuejiaointel: We are using it in onedal wrapper to run kmeans n_init times in for init_idx in range(n_init), for array like init we are setting it to 1. It's not needed by backend tho. We are passing n_init to onedal wrapper in onedal_params

onedal/cluster/kmeans.py
```diff
@@ -102,7 +92,7 @@ def _compute_tolerance(self, X_table, is_csr, dtype):
         self._tol = self._tolerance(X_table, self.tol, is_csr, dtype)
 
     def _get_onedal_params(self, is_csr=False, dtype=np.float32, result_options=None):
-        thr = self._tol if self._tol is not None else self.tol
+        thr = self._tol if hasattr(self, "_tol") else self.tol
```
david-cortes-intel, line 91: This kind of design wouldn't work after a user calls `obj.set_params` to adjust tolerances.

yuejiaointel: I think it is old code, main also has
 thr = self._tol if hasattr(self, "_tol") else self.tol
 rebased
 when users call obj.set_params it is setting [self.tol ](https://github.com/uxlfoundation/scikit-learn-intelex/blob/604ed45974d0b772480c9b7f066b0d5ffc193da9/onedal/cluster/kmeans.py#L59)

 instead of self._tol which is set [somewhere else](https://github.com/uxlfoundation/scikit-learn-intelex/blob/604ed45974d0b772480c9b7f066b0d5ffc193da9/onedal/cluster/kmeans.py#L122) so it should be safe

ethanglaser:
Please rebase

onedal/cluster/kmeans.py
david-cortes-intel: @yuejiaointel What happens if calling `.fit()` twice with a call to `.set_params(tol=...)` between them?

david-cortes-intel:
@yuejiaointel This is not working, please investigate:
```python
import os

import numpy as np
import torch

X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
Xd = torch.tensor(X, device="xpu")

from sklearnex import config_context, set_config
from sklearnex.cluster import KMeans

with config_context(array_api_dispatch=True, transform_output="polars"):
    model = KMeans().fit(Xd)
    res = model.transform(Xd)
```
```
TypeError: can't convert xpu:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

sklearnex/cluster/k_means.py
```diff
@@ -484,6 +482,19 @@ def _onedal_transform(self, X, queue=None):
...
+                XX = xp.sum(X * X, axis=1)[:, None]
+                CC = xp.sum(centers * centers, axis=1)[None, :]
+                XC = X @ centers.T
+                distances_sq = XX + CC - 2 * XC
+                distances_sq = xp.where(
+                    distances_sq < 0, xp.zeros_like(distances_sq), distances_sq
+                )
+                return xp.sqrt(distances_sq)
+
+            from sklearn.metrics.pairwise import euclidean_distances
```
david-cortes-intel: Could give it a name like `_sklearn_euclidean_distances` to make it more clear what it is.

sklearnex/cluster/k_means.py
```diff
@@ -484,6 +482,19 @@ def _onedal_transform(self, X, queue=None):
                     accept_large_sparse=False,
                 )
 
+            if is_array_api:
+                centers = xp.asarray(self.cluster_centers_)
+                XX = xp.sum(X * X, axis=1)[:, None]
+                CC = xp.sum(centers * centers, axis=1)[None, :]
+                XC = X @ centers.T
+                distances_sq = XX + CC - 2 * XC
+                distances_sq = xp.where(
```
david-cortes-intel: There's `xp.clip`.

sklearnex/cluster/k_means.py
```diff
@@ -332,8 +453,52 @@ def _check_test_data(self, X):
...
+                    self,
+                    X,
+                    accept_sparse="csr",
+                    reset=False,
+                    dtype=[xp.float64, xp.float32],
+                    order="C",
+                    accept_large_sparse=False,
+                )
+
+            if is_array_api:
```
david-cortes-intel, line 471: Also if this was meant as a fix for the cases where `transform` requires the input in a different device, looks like it didn't work.

david-cortes-intel:
@yuejiaointel Also please ensure that the example code above also works when using `.set_output(...)` instead of a config context.

david-cortes-intel:
Reminder from https://github.com/uxlfoundation/scikit-learn-intelex/pull/2971#discussion_r2977375062

Calling `.predict()` and `.fit_predict()` should produce the same results. Please remove that special casing from the recently added tests.

yuejiaointel:
> @yuejiaointel This is not working, please investigate:
> 
> ```python
> import os
> os.environ["SCIPY_ARRAY_API"] = "1"
> 
> import numpy as np
> import torch
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> Xd = torch.tensor(X, device="xpu")
> 
> from sklearnex import config_context, set_config
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=True, transform_output="polars"):
>     model = KMeans().fit(Xd)
>     res = model.transform(Xd)
> ```
> 
> ```
> TypeError: can't convert xpu:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
> ```

fixed looks like kmeans need the same change as in neighbors in wrap_output_data for torch inputs

sklearnex/cluster/k_means.py
yuejiaointel: thx renamed

sklearnex/cluster/k_means.py
yuejiaointel: thx, changed

david-cortes-intel:
> > @yuejiaointel This is not working, please investigate:
> > ```python
> > import os
> > os.environ["SCIPY_ARRAY_API"] = "1"
> > 
> > import numpy as np
> > import torch
> > 
> > rng = np.random.default_rng(seed=123)
> > X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> > Xd = torch.tensor(X, device="xpu")
> > 
> > from sklearnex import config_context, set_config
> > from sklearnex.cluster import KMeans
> > 
> > with config_context(array_api_dispatch=True, transform_output="polars"):
> >     model = KMeans().fit(Xd)
> >     res = model.transform(Xd)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > TypeError: can't convert xpu:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
> > ```
> 
> fixed looks like kmeans need the same change as in neighbors in wrap_output_data for torch inputs

@yuejiaointel Please add a test for those to ensure that the logic doesn't break with the changes that the logistic regression PR is trying to make to that function:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/2941

Note that you'll need the torch input to be on CPU and the test to be conditionally skipped according to torch availability in order for the CIs to execute and skip when appropriate.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -155,3 +155,40 @@ def test_dense_vs_sparse(queue, init, algorithm, dims):
         kmeans_dense.cluster_centers_,
         kmeans_sparse.cluster_centers_,
     )
+
+
+torch = pytest.importorskip("torch")
```
david-cortes-intel: I think you would need to put this inside the test for it to work.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -155,3 +155,40 @@ def test_dense_vs_sparse(queue, init, algorithm, dims):
...
+
+
+torch = pytest.importorskip("torch")
+
+
+@pytest.mark.parametrize("output_format", ["set_output", "config_context"])
+@pytest.mark.parametrize("transform_output", ["polars", "pandas"])
+def test_transform_output_torch(output_format, transform_output):
+    try:
+        __import__(transform_output)
```
david-cortes-intel: These should always be installed, since they are now test dependencies.

onedal/cluster/kmeans.py
david-cortes-intel: @yuejiaointel Pinging about this issue again.

sklearnex/cluster/k_means.py
```diff
@@ -332,8 +453,52 @@ def _check_test_data(self, X):
...
+                centers = xp.asarray(self.cluster_centers_)
+                XX = xp.sum(X * X, axis=1)[:, None]
+                CC = xp.sum(centers * centers, axis=1)[None, :]
+                XC = X @ centers.T
+                distances_sq = XX + CC - 2 * XC
+                distances_sq = xp.clip(distances_sq, 0, None)
+                return xp.sqrt(distances_sq)
+
+            from sklearn.metrics.pairwise import (
```
david-cortes-intel: Perhaps you could move the import to the top.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, changed

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, removed

sklearnex/cluster/k_means.py
yuejiaointel: thx moved

david-cortes-intel: @yuejiaointel They are still inside the test.

david-cortes-intel:
> ```python
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> Xt = torch.tensor(X, device="xpu")
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=True):
>     km = KMeans(n_clusters=3).fit(Xt)
>     cl = km.predict(Xt)
> ```

@yuejiaointel This one broke again, please investigate.

sklearnex/_device_offload.py
```diff
@@ -193,13 +194,20 @@ def wrapper(self, *args, **kwargs) -> Any:
...
-                if input_array_api and not _is_numpy_namespace(input_array_api):
-                    input_array_api_device = data.device
-                    result = _asarray(
-                        result, input_array_api, device=input_array_api_device
-                    )
+            transform_output = get_config().get("transform_output", "default")
+            est_output = getattr(self, "_sklearn_output_config", {}).get(
+                "transform", "default"
+            )
+            needs_host_output = transform_output not in (
```
david-cortes-intel: Please discuss with @avolkov-intel how to handle this across PRs. The modifications for logistic regression appear to be targeting a similar case.

yuejiaointel:
> > ```python
> > rng = np.random.default_rng(seed=123)
> > X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> > Xt = torch.tensor(X, device="xpu")
> > 
> > from sklearnex import config_context
> > from sklearnex.cluster import KMeans
> > 
> > with config_context(array_api_dispatch=True):
> >     km = KMeans(n_clusters=3).fit(Xt)
> >     cl = km.predict(Xt)
> > ```
> 
> @yuejiaointel This one broke again, please investigate.

fixed

david-cortes-intel:
@yuejiaointel This is not working, please investigate:
```python
import os
os.environ["SKLEARNEX_VERBOSE"] = "INFO"

import numpy as np

X = rng.standard_normal(size=(1000, 20), dtype=np.float32)

import dpnp
Xd = dpnp.array(X, device="gpu")

from sklearnex import config_context
from sklearnex.cluster import KMeans

with config_context(array_api_dispatch=False, transform_output="polars"):
    model = KMeans().fit(Xd)
    res = model.transform(Xd)
```

sklearnex/tests/test_patching.py
```diff
@@ -288,7 +288,12 @@ def _check_output_type(result, y, method, estimator_name, caplog, X, est=None):
     ("LogisticRegression", "n_iter_"): {"attr_type", "attr_device", "attr_dtype"},
     ("LogisticRegression", "classes_"): {"attr_type", "attr_device", "attr_dtype"},
     ("LogisticRegression", "classes_"): {"attr_type", "attr_device", "attr_dtype"},
-    ("KMeans", "cluster_centers_"): {"attr_type", "attr_device", "attr_dtype"},
+    ("KNeighborsClassifier", "classes_"): {"attr_type", "attr_device", "attr_dtype"},
```
ethanglaser: Should these be removed?

sklearnex/cluster/k_means.py
```diff
@@ -293,32 +399,50 @@ def predict(
...
 
         def _onedal_predict(self, X, sample_weight=None, queue=None):
-            X = validate_data(
-                self,
-                X,
-                accept_sparse="csr",
-                reset=False,
-                dtype=[np.float64, np.float32],
-            )
+            from .._config import get_config
```
ethanglaser: Let's keep it at top of file instead of inside function for consistency

sklearnex/cluster/k_means.py
```diff
@@ -153,21 +156,119 @@ def fit(self, X, y=None, sample_weight=None):
...
 
-            if sklearn_check_version("1.2"):
-                self._check_params_vs_input(X)
-            else:
-                self._check_params(X)
+            xp, _ = get_namespace(X)
+
+            # Validate init parameter if array-like (before X validation so n_features_in_ not set yet)
+            if _is_arraylike_not_scalar(self.init):
+                init = validate_data(
```
ethanglaser: this technically should also have raw input added to condition

sklearnex/cluster/k_means.py
```diff
@@ -185,14 +286,19 @@ def _validate_sample_weight(self, sample_weight, X):
...
-                if np.all(sample_weight == sample_weight[0]):
-                    return True
-                else:
+                try:
+                    sample_weight = _check_sample_weight(
+                        sample_weight,
+                        X,
+                        dtype=X.dtype if hasattr(X, "dtype") else None,
+                    )
+                    if np.all(sample_weight == sample_weight[0]):
```
ethanglaser: Can it be xp?

sklearnex/cluster/k_means.py
```diff
@@ -293,32 +399,50 @@ def predict(
...
             raise RuntimeError(
                 f"Unknown method {method_name} in {self.__class__.__name__}"
             )
 
-        _onedal_gpu_supported = _onedal_supported
         _onedal_cpu_supported = _onedal_supported
 
+        def _onedal_gpu_supported(self, method_name, *data):
+            # callable init uses numpy ops in _init_centroids_sklearn
+            # which don't work on GPU arrays; fall back to sklearn
```
ethanglaser, line 414: We have kmeans init for GPU/spmd though, I thought?

sklearnex/tests/test_patching.py
david-cortes-intel: Yes, it needs to be removed here.

yuejiaointel:
> ```python
> import os
> os.environ["SCIPY_ARRAY_API"] = "1"
> os.environ["SKLEARNEX_VERBOSE"] = "INFO"
> 
> import numpy as np
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> 
> import dpnp
> Xd = dpnp.array(X, device="gpu")
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=False, transform_output="polars"):
>     model = KMeans().fit(Xd)
>     res = model.transform(Xd)
> ```



> @yuejiaointel This is not working, please investigate:
> 
> ```python
> import os
> os.environ["SCIPY_ARRAY_API"] = "1"
> os.environ["SKLEARNEX_VERBOSE"] = "INFO"
> 
> import numpy as np
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> 
> import dpnp
> Xd = dpnp.array(X, device="gpu")
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=False, transform_output="polars"):
>     model = KMeans().fit(Xd)
>     res = model.transform(Xd)
> ```

fixed, now returning numpy res

sklearnex/tests/test_patching.py
yuejiaointel: some neighbors code got added during merge, removed

sklearnex/cluster/k_means.py
yuejiaointel: thx, fixed

sklearnex/cluster/k_means.py
yuejiaointel: thx, added

sklearnex/cluster/k_means.py
yuejiaointel: good catch, changed, tho it is still numpy from the skleran method, but it's good to have consistency

sklearnex/cluster/k_means.py
yuejiaointel: Yes k-means++ and random works for gpu, it is only blocking the callable case, where user would pass some numpy ops in a function and these does not work on gpu

david-cortes-intel:
> > ```python
> > import os
> > os.environ["SCIPY_ARRAY_API"] = "1"
> > os.environ["SKLEARNEX_VERBOSE"] = "INFO"
> > 
> > import numpy as np
> > 
> > rng = np.random.default_rng(seed=123)
> > X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> > 
> > import dpnp
> > Xd = dpnp.array(X, device="gpu")
> > 
> > from sklearnex import config_context
> > from sklearnex.cluster import KMeans
> > 
> > with config_context(array_api_dispatch=False, transform_output="polars"):
> >     model = KMeans().fit(Xd)
> >     res = model.transform(Xd)
> > ```
> 
> > @yuejiaointel This is not working, please investigate:
> > ```python
> > import os
> > os.environ["SCIPY_ARRAY_API"] = "1"
> > os.environ["SKLEARNEX_VERBOSE"] = "INFO"
> > 
> > import numpy as np
> > 
> > rng = np.random.default_rng(seed=123)
> > X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> > 
> > import dpnp
> > Xd = dpnp.array(X, device="gpu")
> > 
> > from sklearnex import config_context
> > from sklearnex.cluster import KMeans
> > 
> > with config_context(array_api_dispatch=False, transform_output="polars"):
> >     model = KMeans().fit(Xd)
> >     res = model.transform(Xd)
> > ```
> 
> fixed, now returning numpy res

@yuejiaointel This now broke with `array_api_dispatch=True`, please investigate.

david-cortes-intel:
@yuejiaointel This also broke, please investigate:
```python
import os
os.environ["SKLEARNEX_VERBOSE"] = "INFO"

import numpy as np

X = rng.standard_normal(size=(1000, 20), dtype=np.float32)

import dpnp
Xd = dpnp.array(X, device="gpu")

from sklearnex import config_context
from sklearnex.cluster import KMeans

with config_context(array_api_dispatch=False):
    model = KMeans().fit(Xd).set_output(transform="polars")
    res = model.transform(Xd)
```

yuejiaointel:
> @yuejiaointel This also broke, please investigate:
> 
> ```python
> import os
> os.environ["SKLEARNEX_VERBOSE"] = "INFO"
> 
> import numpy as np
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> 
> import dpnp
> Xd = dpnp.array(X, device="gpu")
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=False):
>     model = KMeans().fit(Xd).set_output(transform="polars")
>     res = model.transform(Xd)
> ```

thx for the reproducer, made another fix, root cause is still in device offload, where this specific case will enter dpnp/dpctl check and convert back to dpnp/dpctl instead of convert to numpy, so when sklearn calls set_ouput it cannot handle these inputs, so fix is move the transform check in front of the dpnp/dpctl chekc

david-cortes-intel:
@yuejiaointel This is producing numpy outputs when it should produce torch, please investigate:
```python
import os
os.environ["SKLEARNEX_VERBOSE"] = "INFO"

import numpy as np

X = rng.standard_normal(size=(1000, 20), dtype=np.float32)

import torch
Xd = torch.tensor(X, device="cpu")

from sklearnex import config_context
from sklearnex.cluster import KMeans

with config_context(array_api_dispatch=True, transform_output="polars"):
    model = KMeans().fit(Xd)
    pred = model.predict(Xd)
```

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
-                centers = init.toarray()
-            else:
-                centers = np.asarray(init)
-            assert centers.shape[0] == n_clusters
-            assert centers.shape[1] == X_table.column_count
-            # KMeans is implemented on both CPU and GPU for Dense and CSR data
-            # The original policy can be used here
-            centers_table = to_table(centers, queue=QM.get_global_queue())
+                init = init.toarray()
+            return to_table(init, queue=QM.get_global_queue())
```
avolkov-intel, line 130: I think you shouldn't use QM get_global queue in this file, you should pass the queue that you got from fit function with @support_queue function down to hidden functions to avoid the case where this queue is different.

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
+            )
 
-        return centers_table
+        alg = self._get_kmeans_init(
+            cluster_count=n_clusters,
+            seed=random_seed,
+            algorithm=algorithm,
+            is_csr=is_csr,
+        )
+        centers = alg.compute_raw(X_table, dtype, queue=QM.get_global_queue())
```
avolkov-intel, line 143: Same comment here, it's better to pass queue from fit function instead of using QM

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
         elif _is_arraylike_not_scalar(init):
             centers = init
         else:
             raise ValueError(
                 f"init should be either 'k-means++', 'random', a ndarray or a "
-                f"callable, got '{ init }' instead."
+                f"callable, got '{init}' instead."
             )
 
         return to_table(centers, queue=getattr(QM.get_global_queue(), "_queue", None))
```
avolkov-intel, line 176: Same comment here, it's better to pass queue from fit function instead of using QM

onedal/cluster/kmeans.py
```diff
@@ -318,89 +269,78 @@ def is_better_iteration(inertia, labels):
...
     def cluster_centers_(self):
-        del self._cluster_centers_
-
-    def _predict(self, X, result_options=None):
-        is_csr = _is_csr(X)
+        self._cluster_centers_ = None
 
+    @supports_queue
+    def predict(self, X, queue=None):
         X_table = to_table(X, queue=QM.get_global_queue())
```
avolkov-intel: Why do you need QM.get_global_queue here, you can use the argument queue

onedal/cluster/kmeans.py
```diff
@@ -318,89 +269,78 @@ def is_better_iteration(inertia, labels):
...
-
-    def _score(self, X):
-        result_options = "compute_exact_objective_function"
-
-        return self._predict(
-            X,
-            result_options,
+    @supports_queue
+    def score(self, X, queue=None):
+        X_table = to_table(X, queue=QM.get_global_queue())
```
avolkov-intel: Same comment here, please use argument queue instead of QM

onedal/cluster/kmeans.py
```diff
@@ -318,89 +269,78 @@ def is_better_iteration(inertia, labels):
...
 
-        # Types without conversion
+        # assign learned attributes (pattern)
         self.model_ = best_model
-
-        # Simple types
         self.n_iter_ = best_n_iter
         self.inertia_ = best_inertia
+        # from_table without like= returns numpy (needed for np.unique check below)
+        _labels_np = from_table(best_labels).ravel()
```
avolkov-intel: I'm against converting output to numpy here as this violates zero copy rule and negatively affect performance on GPU. 
I suggest you to move this check to sklearnex file and use xp.unique_values there. This way you can avoid redundant copy of data from device to host.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -155,3 +155,30 @@ def test_dense_vs_sparse(queue, init, algorithm, dims):
         kmeans_dense.cluster_centers_,
         kmeans_sparse.cluster_centers_,
     )
+
+
+@pytest.mark.parametrize("output_format", ["set_output", "config_context"])
+@pytest.mark.parametrize("transform_output", ["polars", "pandas"])
+def test_transform_output_torch(output_format, transform_output):
```
avolkov-intel, line 165: I think it worth extending this test with gpu inputs if gpu queue is available. Also I would add additional test where you check that dpnp inputs on GPU works fine with with config_context(array_api_dispatch=True) since that is a core functionality and it's better to have it properly tested

sklearnex/cluster/k_means.py
```diff
@@ -185,14 +283,20 @@ def _validate_sample_weight(self, sample_weight, X):
...
+                        sample_weight,
+                        X,
+                        dtype=X.dtype if hasattr(X, "dtype") else None,
+                    )
+                    if xp.all(sample_weight == sample_weight[0]):
+                        return True
+                    else:
+                        return False
+                except TypeError:
+                    # GPU arrays (dpctl/dpnp) cannot be implicitly converted
```
avolkov-intel: Does _check_sample_weight function have support for array api inputs?

Am I correct that if you provide a dpnp input on GPU with sample_weights being an array of equal numbers we will fallback to stock sklearn and transfer args to host? 

In this case I would suggest either stating this in the docs or trying to add support for this use case. Probably we can avoid using the _check_sample_weight function since our restrictions are more strict and avoid copy of data to host.

sklearnex/_device_offload.py
```diff
@@ -183,6 +183,17 @@ def wrapper(self, *args, **kwargs) -> Any:
...
+            if (
+                get_config().get("transform_output")
+                not in (
+                    "default",
+                    None,
+                )
+                or getattr(self, "_sklearn_output_config", {}).get("transform", "default")
+                != "default"
+            ):
+                _, (result,) = _transfer_to_host(result)
```
avolkov-intel, line 216: Could you, please, add comment here explaining why do we need this condition. For me it's also suspicious since we are forcing transfer to host, while transform might not force us to do this

sklearnex/_device_offload.py
```diff
@@ -204,7 +215,12 @@ def wrapper(self, *args, **kwargs) -> Any:
                 if hasattr(data, "dtype"):
                     xp, is_array_api = get_namespace(data)
                     if is_array_api and not _is_numpy_namespace(xp):
-                        result = _asarray(result, xp, device=data.device)
+                        if isinstance(result, tuple):
+                            result = tuple(
+                                xp.asarray(r, device=data.device) for r in result
```
avolkov-intel: let's use a safer option getattr(data, "device", None)

avolkov-intel:
Good job! In general the structure of the algorithm seems good, although I have few concerns about whether zero copy principle is followed. I suspect that we might have some problems with GPU performance. 

Please address the comments that I left, double check all the places where you force transfer to host and check if it can be avoided. 

Also there's a risk of silent fallback to host (e.g. sample weights is a dpnp array with equal weights) which is not expected by user.

Also you might want to add more GPU tests to check that core functionality works as expected and we can deprecate raw inputs without worries

yuejiaointel:
> @yuejiaointel This is producing numpy outputs when it should produce torch, please investigate:
> 
> ```python
> import os
> os.environ["SKLEARNEX_VERBOSE"] = "INFO"
> 
> import numpy as np
> 
> rng = np.random.default_rng(seed=123)
> X = rng.standard_normal(size=(1000, 20), dtype=np.float32)
> 
> import torch
> Xd = torch.tensor(X, device="cpu")
> 
> from sklearnex import config_context
> from sklearnex.cluster import KMeans
> 
> with config_context(array_api_dispatch=True, transform_output="polars"):
>     model = KMeans().fit(Xd)
>     pred = model.predict(Xd)
> ```

fixed, needed to add guard for transform methods

onedal/cluster/kmeans.py
yuejiaointel: This is unchanged from main. Also it seems we are just setting the global queue in fit once then inner calls only read global queue

onedal/cluster/kmeans.py
yuejiaointel: At this point just want minimal change, current approach does not break anything yet, don't want to risk it

onedal/cluster/kmeans.py
yuejiaointel: same reply above

onedal/cluster/kmeans.py
yuejiaointel: changed, main also use global queue, but since we already have the queue passed in, we can just use it

onedal/cluster/kmeans.py
yuejiaointel: thx, changed

onedal/cluster/kmeans.py
yuejiaointel: good catch, changed

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: added a few more tests

sklearnex/_device_offload.py
yuejiaointel: thx, changed

sklearnex/_device_offload.py
yuejiaointel: mostly for the transform
for example:
```python
 import os                                                                                                                                                                                                      
                                                                                                                                                                                                                 
  import numpy as np                                                                                                                                                                                             
  import dpnp                                                                                                                                                                                                    
  from sklearnex import config_context                                                                                                                                                                           
  from sklearnex.cluster import KMeans                                                                                                                                                                           
                                                                                                                                                                                                                 
  X = dpnp.array(np.random.rand(100, 5).astype(np.float32), device="gpu")                                                                                                                                        
                                                                                                                                                                                                                 
  with config_context(array_api_dispatch=True, transform_output="polars"):                                                                                                                                       
      km = KMeans(n_clusters=3, n_init=1).fit(X)            
      km.transform(X)                                                                                                                                                                                            
  # TypeError: DataFrame constructor called with unsupported type 'dpnp_array'   
```

would error out because we would hit the early return now(array api enabled) and just return dpnp that sklearn cannot handle, so we need to convert it to numpy in this case

avolkov-intel:
Check that private CI doesn't have failures (especially that spmd kmeans example doesn't fail)

sklearnex/cluster/k_means.py
```diff
@@ -153,21 +157,115 @@ def fit(self, X, y=None, sample_weight=None):
 
             return self
 
+        def _resolve_n_init(self, default_n_init=10):
+            """Resolve n_init parameter from 'auto'/'warn' to integer value."""
+            n_init = self.n_init
```
ethanglaser, line 165: nit: is there a reason we set n_init = self.n_init but not init = self.init? Would make sense to do both or neither of these

sklearnex/cluster/k_means.py
```diff
@@ -153,21 +157,115 @@ def fit(self, X, y=None, sample_weight=None):
...
+            if isinstance(n_init, str) and n_init in ("auto", "warn"):
+                if isinstance(self.init, str) and self.init == "k-means++":
+                    n_init = 1
+                elif isinstance(self.init, str) and self.init == "random":
```
ethanglaser, line 169: These isinstance string checks seem redundant, no?
```suggestion
            if n_init in ("auto", "warn"):
                if self.init == "k-means++":
                    n_init = 1
                elif self.init == "random":
```

ethanglaser: I see now this is copied from sklearn https://github.com/scikit-learn/scikit-learn/blob/fe2edb3cd/sklearn/cluster/_kmeans.py#L874, probably worth commenting this in the docstring for posterity

sklearnex/cluster/k_means.py
```diff
@@ -153,21 +157,115 @@ def fit(self, X, y=None, sample_weight=None):
...
+                        f"n_init={n_init}."
+                    ),
+                    RuntimeWarning,
+                    stacklevel=2,
+                )
+                n_init = 1
+
+            return n_init
+
+        def _validate_center_shape(self, X, centers):
```
ethanglaser: Is there a reason we are reimplementing https://github.com/scikit-learn/scikit-learn/blob/fe2edb3cd/sklearn/cluster/_kmeans.py#L936? Looks to be the exact same

sklearnex/cluster/k_means.py
```diff
@@ -180,20 +278,12 @@ def _onedal_fit(self, X, _, sample_weight, queue=None):
...
-                    sample_weight,
-                    X,
-                    dtype=X.dtype if hasattr(X, "dtype") else None,
-                )
-                if np.all(sample_weight == sample_weight[0]):
-                    return True
-                else:
-                    return False
+            xp, _ = get_namespace(sample_weight)
+            return bool(xp.all(sample_weight == sample_weight[0]))
```
ethanglaser: So we are bypassing the _check_sample_weight call - are we adding in any validation to make up for this?

onedal/cluster/kmeans.py
```diff
@@ -52,15 +50,19 @@ def __init__(
...
+        self.n_init = n_init
         self.max_iter = max_iter
         self.tol = tol
-        self.n_init = n_init
         self.verbose = verbose
         self.random_state = random_state
         self.n_local_trials = n_local_trials
+        self.algorithm = algorithm
+
+    # --- pybind11 backends (thin proxies) ---
```
ethanglaser: Comment not needed

onedal/cluster/kmeans.py
```diff
@@ -69,99 +71,32 @@ def _is_same_clustering(self, labels, best_labels, n_clusters): ...
...
+    def infer(self, params, model, X_table): ...
 
-    def _get_kmeans_init(self, cluster_count, seed, algorithm, is_csr):
-        return KMeansInit(
-            cluster_count=cluster_count,
-            seed=seed,
-            algorithm=algorithm,
-            is_csr=is_csr,
-        )
+    # --- helpers matching the pattern ---
```
ethanglaser: comment not needed

sklearnex/cluster/k_means.py
ethanglaser: Update: I see now that this is copied from onedal/ - but I dont think we need it defined here

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
-            assert centers.shape[1] == X_table.column_count
-            # KMeans is implemented on both CPU and GPU for Dense and CSR data
-            # The original policy can be used here
-            centers_table = to_table(centers, queue=QM.get_global_queue())
+                init = init.toarray()
+            return to_table(init, queue=QM.get_global_queue())
         else:
-            raise TypeError("Unsupported type of the `init` value")
+            raise ValueError(
+                "init should be either 'k-means++', 'random', a ndarray "
```
ethanglaser, line 133: Would an error be raised if callable init passed into this function?

onedal/cluster/kmeans.py
ethanglaser: This is fine with me for now, it is indeed how it works currently in main

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
+        return centers
 
-    def _init_centroids_sklearn(self, X, init, random_state, dtype=np.float32):
-        # For oneDAL versions < 2023.2 or callable init,
-        # using the scikit-learn implementation
-        logging.getLogger("sklearnex").info("Computing KMeansInit with Stock sklearn")
-        n_samples = X.shape[0]
+    def _init_centroids_sklearn(self, X, init, random_state, dtype=None):
+        if dtype is None:
+            dtype = np.float32
```
ethanglaser:
```suggestion
    def _init_centroids_sklearn(self, X, init, random_state, dtype=np.float32):
```
is there a reason this was changed? and why delete logging stuff?

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
         else:
             raise ValueError(
                 f"init should be either 'k-means++', 'random', a ndarray or a "
-                f"callable, got '{ init }' instead."
+                f"callable, got '{init}' instead."
             )
 
         return to_table(centers, queue=getattr(QM.get_global_queue(), "_queue", None))
 
+    # --- core train/infer wrappers in the estimator pattern ---
```
ethanglaser: comment not needed

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
-        if _get_config()["use_raw_input"] is False:
-            X = _check_array(
-                X,
-                dtype=[np.float64, np.float32],
-                accept_sparse="csr",
-                force_all_finite=False,
-            )
-        X_table = to_table(X, queue=QM.get_global_queue())
-        dtype = X_table.dtype
+    # --- public API matched to the pattern ---
```
ethanglaser: comment not needed

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
 
-        self._check_params_vs_input(X_table, is_csr, dtype=dtype)
+    @supports_queue
+    def fit(self, X, y=None, queue=None):
+        is_csr = _is_csr(X)
+        X_table = to_table(X, queue=queue)
+        dtype = X_table.dtype
+        self._input_type = return_type_constructor(X)  # lightweight callable
```
ethanglaser, line 199: what does this do?

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
                 return True
-            else:
-                better_inertia = inertia < best_inertia
-                return better_inertia and not self._is_same_clustering(
-                    labels, best_labels, self.n_clusters
-                )
+            better = inertia < best_inertia
+            return better and not self._is_same_clustering(
+                labels, best_labels, self.n_clusters
+            )
```
ethanglaser: seems like nothing changed here?

onedal/cluster/kmeans.py
```diff
@@ -183,132 +126,140 @@ def _init_centroids_onedal(
...
+            warnings.warn(
+                (
+                    "Explicit initial center position passed: performing only "
+                    f"one init in {self.__class__.__name__} instead of "
+                    f"n_init={n_init}."
+                ),
+                RuntimeWarning,
+                stacklevel=2,
+            )
+            n_init = 1
```
ethanglaser, line 245: Isn't all of this already handled on sklearnex side? Seems redundant

sklearnex/cluster/k_means.py
```diff
@@ -332,8 +439,47 @@ def _check_test_data(self, X):
...
+                self,
+                "transform",
+                {
+                    "onedal": self.__class__._onedal_transform,
+                    "sklearn": _sklearn_KMeans.transform,
+                },
+                X,
+            )
+
+        def _onedal_transform(self, X, queue=None):
```
ethanglaser, line 456: Is this really "_onedal_transform" if it does not call the onedal backend?

onedal/cluster/kmeans.py
```diff
@@ -318,89 +269,65 @@ def is_better_iteration(inertia, labels):
...
+    @supports_queue
+    def score(self, X, queue=None):
+        X_table = to_table(X, queue=queue)
+        result = self._predict_backend(
+            X_table, result_options="compute_exact_objective_function"
         )
+        return -1 * result.objective_function_value
 
-    def _transform(self, X):
+    def transform(self, X):
```
ethanglaser, line 319: This function will never be called by sklearnex

ethanglaser:
Additionally looks like 2 conformance fails in private CI

sklearnex/cluster/k_means.py
yuejiaointel: n_init is being modified so we are using a local var, self.n_init is "auto"/"warn"/integer, self.init is the the init function name

sklearnex/cluster/k_means.py
yuejiaointel: _check_sample_weight  does not support array api inputs and will error out, and we just need it to know if weights are uniform, if uniform treat it as none, if not uniform fallback to sklearn, onedal acceleration only works when no weights

sklearnex/cluster/k_means.py
yuejiaointel: Good call, removed

sklearnex/cluster/k_means.py
yuejiaointel: added a comment

onedal/cluster/kmeans.py
yuejiaointel: removed

onedal/cluster/kmeans.py
yuejiaointel: removed

onedal/cluster/kmeans.py
yuejiaointel: a value error should be raised, we also have a block for callable in fit

onedal/cluster/kmeans.py
yuejiaointel: good catch reverted, logging is removed since it is not used

onedal/cluster/kmeans.py
yuejiaointel: removed

onedal/cluster/kmeans.py
yuejiaointel: it was a suggestion from bot, self._input_type = X which stored entire training array just for type info, and we changed it to a more lightweighted callable

onedal/cluster/kmeans.py
ethanglaser: But the error message indicates that callable is allowed, but its not in any of the if statements

onedal/cluster/kmeans.py
yuejiaointel: reverted, was just function name change

sklearnex/cluster/k_means.py
ethanglaser: Ah this makes sense now

onedal/cluster/kmeans.py
ethanglaser: What do you mean by not used? Wouldn't the logger print a message to the user if the situation occurred?

onedal/cluster/kmeans.py
yuejiaointel: when using onedal as standalone like currently in onedal test

david-cortes-intel:
@yuejiaointel Please merge the main branch here.

PR #2665, opened 2025-08-14
FIX: Avoid re-usage of non-reusable daal4py objects

david-cortes-intel:
## Description

This PR fixes an issue in the model builders for logistic regression that was reusing daal4py objects whose internal C++ objects are destructed after their first use, by re-creating the objects every time.

Along the way, it also adds a big warning in the docs for daal4py that the objects that provide `compute` and similar are not meant to support repeated calls to that method.

Fixes the segfault observed here: https://github.com/uxlfoundation/oneDAL/pull/3315

---

**Testing**

### DIFF
--- a/daal4py/mb/logistic_regression_builders.py
+++ b/daal4py/mb/logistic_regression_builders.py
@@ -97,21 +97,6 @@ class LogisticDAALModel:
         )
         builder.set_beta(coefs, intercepts)
         self._model = builder.model
-        self._alg_pred_class = logistic_regression_prediction(
-            nClasses=self.n_classes_,
-            fptype=self._fptype,
-            resultsToEvaluate="computeClassLabels",
-        )
-        self._alg_pred_prob = logistic_regression_prediction(
-            nClasses=self.n_classes_,
-            fptype=self._fptype,
-            resultsToEvaluate="computeClassProbabilities",
-        )
-        self._alg_pred_logprob = logistic_regression_prediction(
-            nClasses=self.n_classes_,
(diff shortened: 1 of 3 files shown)


david-cortes-intel:
CI errors do not look related to the changes here.

daal4py/mb/logistic_regression_builders.py
```diff
@@ -133,7 +118,12 @@ def predict(self, X) -> np.ndarray:
             The most probable class, as integer indexes
         """
         return (
-            self._alg_pred_class.compute(X, self._model)
+            logistic_regression_prediction(
+                nClasses=self.n_classes_,
+                fptype=self._fptype,
+                resultsToEvaluate="computeClassLabels",
+            )
```
Alexsandruss: Better to wrap d4p.logistic_regression_prediction calls:
```python
def _logistic_regression_prediction(self, X, resultsToEvaluate):
    return logistic_regression_prediction(
        nClasses=self.n_classes_,
        fptype=self._fptype,
        resultsToEvaluate=resultsToEvaluate,
    ).compute(X, self._model)

def predict(self, X):
    return self._logistic_regression_prediction(
        self, X, "computeClassLabels"
    ).prediction.reshape(-1).astype(int)

def predict_proba(self, X):
    ...
    .probabilities

def predict_log_proba(self, X):
    ...
    .logProbabilities
```

david-cortes-intel: Moved into an internal method wrapping the prediction object.

PR #2687, opened 2025-09-09
[enh] generalize pytest configuration files

icfaust:
## Description

This enhances changes from #2647 to follow pytest standards for configuration files, rather than using a custom solution.  This will allow for us to centrally extend changes to pytest calls in a central way following pytest standards. This will reduce maintenance and simplify future enhancements.

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/conda-recipe/run_test.bat
+++ b/conda-recipe/run_test.bat
@@ -33,7 +33,7 @@ if "%PYTHON%"=="python" (
 %PYTHON% -c "from sklearnex import patch_sklearn; patch_sklearn()" || set exitcode=1
 
 set "PYTEST_ARGS= "
-set /p PYTEST_CONFIG=<"-c %1setup.cfg"
+set "PYTEST_CONFIG=-c %1setup.cfg"
 
 IF DEFINED COVERAGE_RCFILE (set "PYTEST_ARGS=--cov=onedal --cov=sklearnex --cov-config=%COVERAGE_RCFILE% --cov-append --cov-branch --cov-report= %PYTEST_ARGS%")
 


.ci/scripts/run_sklearn_tests.py
```diff
@@ -59,7 +69,7 @@
             "--cov-report=",
         )
     if json_file := os.getenv("JSON_REPORT_FILE"):
-        pytest_args += ["--json-report", f"--json-report-file={json_file}"]
+        pytest_args += ("--json-report", f"--json-report-file={json_file}")
```
david-cortes-intel: I think a list would be a more appropriate type if this is going to be extended dynamically.

.ci/scripts/run_sklearn_tests.py
```diff
@@ -35,6 +35,13 @@
         help="device name",
         choices=["none", "cpu", "gpu"],
     )
+    parser.add_argument(
```
david-cortes-intel: Since this now forwards unrecognized arguments and this is an argument that pytest would still take on its own, wouldn't it still work without having to add this explicitly in this file?

icfaust: Please read the documentation of the pytest.ini standard and its use in pytest.  To access it within a python terminal requires explicit passing of the argument in python unless specifically initialized in the repo root, which we avoid because of the daal4py extension module.

.ci/scripts/run_sklearn_tests.py
icfaust: This violated consistency with the other additions.  Adding a tuple to a list does not swap the list to a tuple.

david-cortes-intel: Yes. Could it be initialized straight away as a list instead?

.ci/scripts/run_sklearn_tests.py
david-cortes-intel: The docs mention it as a command line option too:
https://docs.pytest.org/en/stable/reference/reference.html#command-line-flags

.. which is how it's being passed to this python script from the shell scripts.

.ci/scripts/run_sklearn_tests.py
icfaust: In other words, you would request that all of the other additions to ```pytest_args``` should be done as lists instead to follow your precedent set in #2608?

david-cortes-intel: Looks like it's only 3 places, or are there more? Not a big deal, but a tuple is meant to be immutable while a list isn't.

icfaust: I have modified as requested.

.ci/scripts/run_sklearn_tests.py
icfaust: In line with #2629 (which this is in reference to) I have added documentation to the file to guide development (even though there is documentation in the scikit-learn-tests.md file). I have also simplified the logic, as ```extra_args``` checking is superfluous and can be set to the pytest_args variable directly (since when empty will default to an empty list).

.ci/scripts/run_sklearn_tests.py
```diff
@@ -45,28 +71,26 @@
...
+        ]
+
     if json_file := os.getenv("JSON_REPORT_FILE"):
         pytest_args += ["--json-report", f"--json-report-file={json_file}"]
 
     while "" in pytest_args:
         pytest_args.remove("")
 
-    if extra_args:
-        pytest_args += extra_args
```
david-cortes-intel: This isn't entirely equivalent, because currently, passing additional options can override previous ones, but with this one, additional options get overridden by builtin ones if they conflict.

icfaust: Code comments as to development patterns are helpful down the line.

icfaust:
private CI failure due to unrelated mpi issue.

PR #2697, opened 2025-09-23
[enh] add array API enabled ```compute_class_weight``` function for use in classifier estimators

icfaust:
## Description

Scikit-learn does not yet support any classifiers for array API.  As such, there are core functions like ```compute_class_weight``` which must be adapted locally for array API.  This adds a private function ```_compute_class_weight``` for use in sklearnex classifier estimators in order to enable them for array API.  It uses an initial check using ```get_namespace``` and replicates line by line the functionality of sklearn written using the array API standard if necessary.

This will be important for #2201 and #2209, as well as changes for KNeighborsClassifier.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/utils/tests/test_class_weight.py
+++ b/sklearnex/utils/tests/test_class_weight.py
@@ -46,10 +46,8 @@ def test_compute_class_weight_array_api(class_weight, dataframe, queue):
     # support of sample_weights added in sklearn 1.7
     set_sample_weight = class_weight == "balanced" and sklearn_check_version("1.7")
 
-    sample_weight = (
-        rng.random(y.shape).astype(np.float64) if set_sample_weight else None
-    )
-    
+    sample_weight = rng.random(y.shape).astype(np.float64) if set_sample_weight else None
+
     if class_weight == "ramp":
         class_weight = {int(i): int(i) for i in np.unique(y)}
 


sklearnex/utils/class_weight.py
```diff
@@ -0,0 +1,99 @@
...
+        weight = xp.ones((classes.shape[0],), dtype=xp.float64, device=classes.device)
+    elif class_weight == "balanced":
+        if not sklearn_check_version("1.6"):
+            raise RuntimeError(
+                "array API support with 'balanced' keyword not supported for sklearn <1.6"
+            )
+        # Find the weight of each class as present in y.
+        le = _sklearn_LabelEncoder()
+        y_ind = le.fit_transform(y)
+
```
Vika-F, line 63: sklearn has a check here:
https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/class_weight.py#L76

Do we need a similar check in sklearnex?

sklearnex/utils/class_weight.py
```diff
@@ -0,0 +1,99 @@
...
+        # nature of the LabelEncoder. Therefore only the maximum is found, and
+        # then core logic of bincount is replicated:
+        # https://github.com/numpy/numpy/blob/main/numpy/_core/src/multiarray/compiled_base.c
+        weighted_class_counts = xp.zeros(
+            (xp.max(y_ind) + 1,), dtype=sample_weight.dtype, device=y.device
+        )
+        for idx, val in enumerate(sample_weight):
+            weighted_class_counts[y_ind[idx]] += val
+
+        recip_freq = xp.sum(weighted_class_counts) / (
```
Vika-F, line 80: should it be something like:
`xp.divide(xp.sum(weighted_class_counts), xp.multiply(len(le.classes_), weighted_class_counts))` ?

sklearnex/utils/tests/test_class_weight.py
```diff
@@ -0,0 +1,68 @@
...
+
+    _, y = load_iris(return_X_y=True)
+    classes = np.unique(y)
+
+    y_xp = _convert_to_dataframe(y, target_df=dataframe, device=queue)
+    classes_xp = _convert_to_dataframe(classes, target_df=dataframe, device=queue)
+
+    sample_weight = (
+        np.ones(y.shape, dtype=np.float64) if class_weight == "balanced" else None
+    )
```
Vika-F: I think it would be better to test with different weights, not all ones. The accuracy threshold might be needed in this case.

Vika-F (commented):
The changes look good, I just have a couple of questions.
Also `check_class_weight` should be changed to `compute_class_weight` in the description.

And just for the sake of general understanding it would be good to know why this functionality is implemented as a hidden function in case of array API?

sklearnex/utils/class_weight.py
david-cortes-intel, line 80: If using float32, this might get enough roundoff error that the result might not sum to 1. Better to do another standardization round afterwards.

sklearnex/utils/class_weight.py
```diff
@@ -0,0 +1,99 @@
...
+        # min and max search, only erroring when a value < 0. Replicating this
+        # exactly via array API would cause another O(n) evaluation (by doing
+        # min and max separately). However this check can be removed due to the
+        # nature of the LabelEncoder. Therefore only the maximum is found, and
+        # then core logic of bincount is replicated:
+        # https://github.com/numpy/numpy/blob/main/numpy/_core/src/multiarray/compiled_base.c
+        weighted_class_counts = xp.zeros(
+            (xp.max(y_ind) + 1,), dtype=sample_weight.dtype, device=y.device
+        )
+        for idx, val in enumerate(sample_weight):
```
david-cortes-intel: For a more GPU-friendly function, could it perhaps first take the unique values of 'y', and then to a loop subsetting for those instead?
E.g.
```python
for y in y_unique:
    weighted_class_counts[y] = xp.sum(sample_weight[y_ind == y])
```

Or maybe multiplying by `(y_ind == y).astype(sample_weight.dtype)` could be faster.

sklearnex/utils/class_weight.py
icfaust: You caught me. I've added it back in the next commit.

sklearnex/utils/class_weight.py
icfaust: Don't worry, some aspects of the array API standard allows for use of the default division and multiplication operators: https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.array.__truediv__.html#array_api.array.__truediv__

icfaust: Added with a comment.

sklearnex/utils/class_weight.py
icfaust: Swapped to a new version based on your suggestion.

sklearnex/utils/tests/test_class_weight.py
icfaust: Swapped to a random array

sklearnex/utils/tests/test_class_weight.py
```diff
@@ -0,0 +1,72 @@
...
+def test_compute_class_weight_array_api(class_weight, dataframe, queue):
+    # This verifies that array_api functionality matches sklearn
+
+    _, y = load_iris(return_X_y=True)
+    classes = np.unique(y)
+
+    y_xp = _convert_to_dataframe(y, target_df=dataframe, device=queue)
+    classes_xp = _convert_to_dataframe(classes, target_df=dataframe, device=queue)
+
+    np.random.seed(42)
```
david-cortes-intel: This is deprecated. The recommended approach from numpy is generators - e.g.:
```python
rng = np.random.default_rng(seed=...)
rng.random(...)
```

icfaust:
@Vika-F you are absolutely right. I will change the title.

sklearnex/utils/tests/test_class_weight.py
icfaust: done.

sklearnex/utils/class_weight.py
icfaust: So there is some weirdness here in that that the sum of recip_freq is in the range [# of classes, inf) where the sum of its inverse must be guaranteed to =# of classes.  Those operations to ensure that the inverse is properly valued is likely also to induce roundoff errors in their calculations, so it could be that some sort of normalization to # of classes isn't possible. Do you have any recommendations here?

david-cortes-intel: The same standardization could be done on the inverses:
```
recip_freq := (recip_freq / length(recip_freq)) * sum(1/recip_freq)
```

david-cortes-intel: Although on second thought, that's probably going to be just as inexact. Maybe it could be removed altogether, or done in a three pass way where first the proportions are calculated and standardized, and then their reciprocals calculated.

PR #2113, opened 2024-10-15
FIX: update functional support fallback logic for a DPNP/DPCTL ndarray inputs

samir-nasibli:
## Description

Stock scikit-learn only support such Array API inputs as `cupy.array_api, array-api-strict, cupy, and PyTorch` [[scikit-learn's Array API support](https://scikit-learn.org/1.5/modules/array_api.html)]. Host numpy copies of the input data will be used for the fallback cases, since stock scikit-learn doesn't support DPCTL usm_ndarray and DPNP ndarray

Aligning with [Documentation PR](https://github.com/intel/scikit-learn-intelex/pull/1918)

### Proposed changes
* update the logic for scikit-learn fallback condition
* Logger enhanced for patch message when data transferred to host

---

Checklist to comply with **before moving PR from draft**:

**PR completeness and readability**

- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes or created a separate PR with update and provided its number in the description, if necessary.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/intel/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have added a respective label(s) to PR if I have a permission for that.
- [x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [x] I have run it locally and tested the changes extensively.
- [x] All CI jobs are green or I have provided justification why they aren't.
- [x] I have extended testing suite if new functionality was introduced in this PR.

**Performance**

- [x] I have measured performance for affected algorithms using [scikit-learn_bench](https://github.com/IntelPython/scikit-learn_bench) and provided at least summary table with measured data, if performance change is expected.
- [x] I have provided justification why performance has changed or why changes are not expected.
- [x] I have provided justification why quality metrics have changed or why changes are not expected.
- [x] I have extended benchmarking suite and provided corresponding scikit-learn_bench PR if new measurable functionality was introduced in this PR.

PR #2698, opened 2025-09-25
Described data copy cases

olegkkruglov:
## Description

This PR adds description for cases where data are implicitly copied to documentation.
It solves issue https://github.com/uxlfoundation/scikit-learn-intelex/issues/4

<details>
<summary>Checklist:</summary>

</details>

### DIFF
--- a/doc/sources/input-types.rst
+++ b/doc/sources/input-types.rst
@@ -41,11 +41,12 @@ here - when receiving an unsupported class, estimators will fall back to stock S
 handle it, so make sure to convert them to a supported type when using Extension.
 
 .. warning::
-  In certain cases data passed to estimators might be copied/duplicated during calls to methods such as fit/predict under some circumstances.
+  In some cases data passed to estimators might be copied/duplicated during calls to methods such as fit/predict.
   The affected cases are listed below.
 
   - Non-contiguous NumPy array - i.e. where strides are wider than one element across both rows and columns
   - For SciPy CSR matrix / CSR array index array is always copied.
   - Heterogeneous NumPy array
+  - If SYCL queue is provided for device without `float64` support but data are `float64`, data are copied with reduced precision.
   - If :ref:`Array API <array_api>` is not enabled then data from GPU devices are always copied to the host device and then result table 
     (for applicable methods) is copied to the source device.


doc/sources/input-types.rst
```diff
@@ -39,3 +39,13 @@ like `Polars <https://pola.rs>`__.
 Extension currently does not offer accelerated routines for input types not listed
 here - when receiving an unsupported class, estimators will fall back to stock Scikit-Learn to
 handle it, so make sure to convert them to a supported type when using Extension.
+
+.. warning::
+  In certain cases data could be copied during transition to oneDAL tables.
```
david-cortes-intel: I think this might not sound very descriptive to an external user reading the documentation. OneDAL tables are an internal implementation detail -  they aren't interfaced directly when using sklearnex.

Perhaps it could be reworded along the lines of "data passed to estimators might be copied/duplicated during calls to methods such as fit/predict under some circumstances".

doc/sources/input-types.rst
```diff
@@ -39,3 +39,13 @@ like `Polars <https://pola.rs>`__.
...
+  It would affect performance. The affected cases are listed below.
+
+  - Non-contiguous NumPy array (C or F)
```
david-cortes-intel:
```suggestion
  - Non-contiguous NumPy array - i.e. where strides are wider than one element across both rows and columns
```

doc/sources/input-types.rst
```diff
@@ -39,3 +39,13 @@ like `Polars <https://pola.rs>`__.
...
+  - SciPy CSR matrix / CSR array
```
david-cortes-intel: Here it could mention that all sparse matrices are copied, but CSR ones copy only the indices.

doc/sources/input-types.rst
```diff
@@ -39,3 +39,13 @@ like `Polars <https://pola.rs>`__.
...
+  - Heterogeneous NumPy array
+  - If :ref:`Array API <array_api>` is not enabled then data from GPU devices are always copied to the host device and then result table
```
david-cortes-intel, line 51: This is not actually the case - for example, passing DPNP arrays that are on GPU to an estimator with GPU support results in the data being used on-device without copying.

doc/sources/input-types.rst
```diff
@@ -39,3 +39,13 @@ like `Polars <https://pola.rs>`__.
...
 here - when receiving an unsupported class, estimators will fall back to stock Scikit-Learn to
 handle it, so make sure to convert them to a supported type when using Extension.
+
+.. warning::
+  In certain cases data could be copied during transition to oneDAL tables.
+  It would affect performance. The affected cases are listed below.
+
+  - Non-contiguous NumPy array (C or F)
+  - SciPy CSR matrix / CSR array
+  - Heterogeneous NumPy array
```
david-cortes-intel, line 49: Would masked arrays also get copied? If so, perhaps could add it as another entry.

doc/sources/input-types.rst
```diff
@@ -39,3 +39,13 @@ like `Polars <https://pola.rs>`__.
...
+  - If :ref:`Array API <array_api>` is not enabled then data from GPU devices are always copied to the host device and then result table 
+    (for applicable methods) is copied to the source device.
```
icfaust, line 52: i agree with @david-cortes-intel , if we want to keep this version, we also are missing fp64 -> fp32 for low precision devices

doc/sources/input-types.rst
olegkkruglov: I thought that this `_transfer_to_host` occurs here always if `onedal_array_api == False`. Could you please correct if I'm missing something?

https://github.com/uxlfoundation/scikit-learn-intelex/blob/fada56a2a44351c49e1034fecad083ae69af39e9/sklearnex/_device_offload.py#L138

david-cortes-intel: @icfaust It does look like @olegkkruglov 's comment is right: if passing something like a DPNP array on GPU, there is a data transfer from GPU to host here:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/2874cb1fa43bf0090f32a8b53295ae04ecb22bcf/sklearnex/_device_offload.py#L140
.. even though the log is called with `transferred_to_host=False`:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/2874cb1fa43bf0090f32a8b53295ae04ecb22bcf/sklearnex/_device_offload.py#L156

Any comments about that?

doc/sources/input-types.rst
```diff
@@ -39,3 +39,13 @@ like `Polars <https://pola.rs>`__.
 Extension currently does not offer accelerated routines for input types not listed
 here - when receiving an unsupported class, estimators will fall back to stock Scikit-Learn to
 handle it, so make sure to convert them to a supported type when using Extension.
+
+.. warning::
+  In certain cases data passed to estimators might be copied/duplicated during calls to methods such as fit/predict under some circumstances.
```
david-cortes-intel:
```suggestion
  In some cases data passed to estimators might be copied/duplicated during calls to methods such as fit/predict.
```

doc/sources/input-types.rst
icfaust: There are a couple things at play there:

1. Array API support is not the default, meaning the ```_onedal_cpu_supported``` and ```_oneda_gpu_supported``` may not properly evaluate with the given input. ```_transfer_to_host``` enables for those previously mentioned methods to operate properly regardless of the circumstances.
2. Until 0.18, dpnp lacked array API support. This means that falling back to scikit-learn required special consideration. This should be updated and I'll make a ticket for it. (I may be able to address this in November, but seems unlikely)
3.  This is behavior was introduced in #2113. Give it a look. I sort of understood it's utility came from how the data approaches oneDAL or scikit-learn. In this branch the data is not *hostargs **hostkwargs, but *args and **kwargs.

david-cortes-intel: @icfaust Sorry, I linked the wrong line.

What I meant was this other line here:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/2874cb1fa43bf0090f32a8b53295ae04ecb22bcf/sklearnex/_device_offload.py#L151

If passing a GPU DPNP array, by the moment it reaches that part:
* The GPU data would be transferred to CPU as a NumPy array already by this point: https://github.com/uxlfoundation/scikit-learn-intelex/blob/2874cb1fa43bf0090f32a8b53295ae04ecb22bcf/sklearnex/_device_offload.py#L140
* The log will nevertheless be called with `transferred_to_host=False`.
* The data will be passed to sklearnex directly, since it has been established that the op will be supported.
* Next, the data will be transferred again from CPU to GPU, and the compute will happen on GPU.

icfaust: Looks a like a bug, good catch. I look forward to your PR.

doc/sources/input-types.rst
olegkkruglov: If you mean numpy masked arrays then they are handled as numpy arrays here
https://github.com/uxlfoundation/scikit-learn-intelex/blob/cda899f4500da545090143c602714eb71dc72962/onedal/datatypes/numpy/data_conversion.cpp#L168
so, no special cases are necessary. If you mean something else, could you please clarify?

david-cortes-intel: Thanks. Looks like it ends up simply ignoring the mask. OTOH looks like sklearn does the same for these classes, so nothing to worry about for now.

david-cortes-intel (approved):

PR #2713, opened 2025-10-07
FIX: Don't prevent oneDAL usage with non-PSD elasticnet/lasso

david-cortes-intel:
## Description

For some reason, there were some checks that avoided calling oneDAL versions of enet and lasso when there are more columns than rows, even though this is actually supported on the C++ side and is the most common reason for wanting to run these algorithms in the first place.

This PR removes those checks, and adds tests to verify that the results match with sklearn.

Along the way, it also makes other corrections and suggestions in the support table for variants of linear regression, in order to accurately reflect the current status.

Note that the implementations match only up to some loose numerical tolerance, unless setting the tolerances to something very tight that would consume a large amount of time.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


david-cortes-intel:
CI failure for docs appears to be a sporadic rate limiting from medium.com.

PR #2723, opened 2025-10-13
FIX: Fix error when calling forest classifiers on single-class data

david-cortes-intel:
## Description

fixes https://github.com/uxlfoundation/scikit-learn-intelex/discussions/1232

This PR updates the checks in forest classifiers to fall back to sklearn when the input data contains a single class, which oddly does not trigger any kind of failure in sklearn even though it doesn't really make sense to do.

Along the way, it also fixes a bug in which the testing util `_as_numpy` fails with pandas inputs due to calling the wrong method.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/ensemble/_forest.py
+++ b/sklearnex/ensemble/_forest.py
@@ -594,11 +594,8 @@ class ForestClassifier(BaseForest, _sklearn_ForestClassifier):
 
             if self.n_outputs_ == 1:
                 xp, is_array_api_compliant = get_namespace(y)
-                if xp is np and not np.__version__.startswith("1"):
-                    num_classes = np.unique_counts(y)
-                else:
-                    sety = xp.unique_values(y) if is_array_api_compliant else np.unique(y)
-                    num_classes = sety.shape[0]
+                sety = xp.unique_values(y) if is_array_api_compliant else np.unique(y)
+                num_classes = sety.shape[0]
                 patching_status.and_conditions(
                     [
                         (


onedal/tests/utils/_dataframes_support.py
```diff
@@ -134,7 +134,7 @@ def _as_numpy(obj, *args, **kwargs):
     if dpctl_available and isinstance(obj, dpt.usm_ndarray):
         return dpt.to_numpy(obj, *args, **kwargs)
     if isinstance(obj, pd.DataFrame) or isinstance(obj, pd.Series):
-        return obj.to_array(*args, **kwargs)
+        return obj.to_numpy(*args, **kwargs)
```
icfaust, line 137: Needs to be described in the description.

sklearnex/ensemble/_forest.py
```diff
@@ -58,6 +58,7 @@
 from onedal.ensemble import RandomForestClassifier as onedal_RandomForestClassifier
 from onedal.ensemble import RandomForestRegressor as onedal_RandomForestRegressor
 from onedal.primitives import get_tree_state_cls, get_tree_state_reg
+from onedal.utils._array_api import _get_sycl_namespace
```
icfaust: This should be `get_namespace` instead.

sklearnex/ensemble/_forest.py
```diff
@@ -592,6 +593,23 @@ def _onedal_fit_ready(self, patching_status, X, y, sample_weight):
             )
             # TODO: Fix to support integers as input
 
+            if self.n_outputs_ == 1:
+                _, xp, _ = _get_sycl_namespace(y)
+
+                if xp is np and np.__version__.startswith("1"):
+                    sety = np.unique(y)
+                else:
+                    sety = xp.unique_values(y)
```
icfaust:
```suggestion
                xp, is_array_api_compliant = get_namespace(X, y)

                sety = xp.unique_values(y) if is_array_api_compliant else xp.unique(y)
```

sklearnex/ensemble/tests/test_forest.py
```diff
@@ -153,3 +155,42 @@ def test_sklearnex_import_et_regression(dataframe, queue):
     # Check that the trees aren't just empty nodes predicting the mean
     for estimator in rf.estimators_:
         assert estimator.tree_.children_left.shape[0] > 1
+
+
+@pytest.mark.allow_sklearn_fallback
```
icfaust, line 160: Since this is testing a fallback case, it may be good to inspect the logs and verify that the proper message is output.

sklearnex/ensemble/tests/test_forest.py
```diff
@@ -153,3 +155,42 @@ def test_sklearnex_import_et_regression(dataframe, queue):
     # Check that the trees aren't just empty nodes predicting the mean
     for estimator in rf.estimators_:
         assert estimator.tree_.children_left.shape[0] > 1
+
+
+@pytest.mark.allow_sklearn_fallback
+@pytest.mark.parametrize("dataframe, queue", get_dataframes_and_queues())
+def test_classifiers_work_on_single_class(dataframe, queue):
```
icfaust, line 162: Due to the naming scheme of the other test which specifically addresses `non array api`, I assume this supports array API, at which point it probably should use a `config_context(array_api_dispatch=True)` or use the array_api fixture.

sklearnex/utils/class_weight.py
```diff
@@ -43,9 +44,15 @@ def _compute_class_weight(class_weight, *, classes, y, sample_weight=None):
...
-    if sety.shape[0] != xp.unique_values(xp.concat((sety, setclasses))).shape[0]:
+    if xp is np and np.__version__.startswith("1"):
+        sety = xp.unique(y)
+        setclasses = xp.unique(classes)
+        ncomb = xp.unique(xp.concat((sety, setclasses))).shape[0]
+    else:
+        sety = xp.unique_values(y)
+        setclasses = xp.unique_values(classes)
+        ncomb = xp.unique_values(xp.concat((sety, setclasses))).shape[0]
+    if sety.shape[0] != ncomb:
```
icfaust: The if statement True branch will never occur (due to `is_array_api_compliant`, https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_array_api.py#L416).  The code change then becomes irrelevant and can be removed.

onedal/tests/utils/_dataframes_support.py
david-cortes-intel: This is just a bug fix for a code path that wasn't being executed.

sklearnex/ensemble/_forest.py
david-cortes-intel: But that requires array API support. This one would work without array API dependencies and without enabling it.

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: I think the logger itself is tested in the daal4py tests, and in the patching tests.

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: Renamed.

sklearnex/utils/class_weight.py
david-cortes-intel: Removed.

david-cortes-intel:
> How do `KNeighborsClassifier`, `LogisticRegression`, `SVC` and `NuSVC` respond to single class data?

Most classifier classes error out in scikit-learn if passed single-class data. `KNeighborsClassifier` already falls back to sklearn on single-class data, and the others that you mention error out with or without patching.

sklearnex/ensemble/_forest.py
icfaust: I don't understand your point. Please clarify. Our version of `get_namespace` works without array API support for sklearn 1.0 purposefully.

sklearnex/ensemble/tests/test_forest.py
icfaust: I don't mean you should test the logger, I mean you should use the logger to verify that the codepath you expect to execute is properly evaluated.

sklearnex/ensemble/_forest.py
david-cortes-intel: Looks like it does work. Changed to `get_namespace`.

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: I don't think that really matters, because the end result is tested in terms of expected predictions which is what we care about.

sklearnex/ensemble/_forest.py
```diff
@@ -592,6 +592,19 @@ def _onedal_fit_ready(self, patching_status, X, y, sample_weight):
             )
             # TODO: Fix to support integers as input
 
+            if self.n_outputs_ == 1:
+                xp, is_array_api_compliant = get_namespace(y)
+                sety = xp.unique_values(y) if is_array_api_compliant else np.unique(y)
```
icfaust: This ends up duplicating logic later down the line (https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/ensemble/forest.py#L321) is there any way we can change this?

icfaust: I guess I am wondering if all the classifiers fail both on sklearn and in sklearnex side, why do we need to have a special check for it?

david-cortes-intel: Not all classifiers fail on sklearn when passed a single class.

onedal/tests/utils/_dataframes_support.py
icfaust: That's why it should be added to the description. If someone does a git blame or attempts to understand why and when something was done it's good to give them an explanation somewhere why it was associated with a thematically separate PR. Two principles I would like to see in PRs: separability - keep the thematic nature of a PR focused, traceability - give maintainers a chance to understand decision making to speed and simplify maintenance.

david-cortes-intel:
Regarding the other comment about 'unique_values' being called elsewhere during fit: it only happens when passing `use_raw_input`, which is not the default and is deprecated either way.

onedal/tests/utils/_dataframes_support.py
david-cortes-intel: Updated the description.

icfaust:
> Regarding the other comment about 'unique_values' being called elsewhere during fit: it only happens when passing `use_raw_input`, which is not the default and is deprecated either way.

are you sure? what does `_validate_targets` do?

sklearnex/ensemble/_forest.py
icfaust: Can you list which of the ones we support pass on single class for me?

david-cortes-intel: From onedal side, none that I am aware of supports single class. But from sklearn side, forests and neighbors do. Current KNN in sklearnex falls back to sklearn on single class, so it works if you pass it a single class under patching.

david-cortes-intel:
> > Regarding the other comment about 'unique_values' being called elsewhere during fit: it only happens when passing `use_raw_input`, which is not the default and is deprecated either way.
> 
> are you sure? what does `_validate_targets` do?

It does more than just return unique values: it also encodes 'y' as integers matching to those classes, whereas this just checks the number of uniques. But I actually see that there's a more specialized function in newer numpy that could be used.

david-cortes-intel:
> > > Regarding the other comment about 'unique_values' being called elsewhere during fit: it only happens when passing `use_raw_input`, which is not the default and is deprecated either way.
> > 
> > 
> > are you sure? what does `_validate_targets` do?
> 
> It does more than just return unique values: it also encodes 'y' as integers matching to those classes, whereas this just checks the number of uniques. But I actually see that there's a more specialized function in newer numpy that could be used.

Actually no, it doesn't behave like the same function in DF libraries.

Issue #2312, opened 2025-02-10
Add polynomial_kernel Function


icfaust:
The polynomial kernel converts data via a polynomial into a new space. This is easy difficulty, but requires significant benchmarking to find when
the scikit-learn-intelex implementation provides better performance. This project will focus on the public API and including the benchmarking 
results for a seamless, high-performance user experience. Combines with the other kernel projects to a medium time commitment.

SnigdhaSarkar16:
Hi, I’d like to work on this issue for Hacktoberfest.

PR #2738, opened 2025-10-20
Add polynomial_kernel wrapper (#2312)

SnigdhaSarkar16:
This PR adds a wrapper for the polynomial kernel in the scikit-learn-intelex extension. 

- Implements the polynomial kernel function similar to linear, rbf, and sigmoid kernels.
- Supports gamma, coef0, degree, and optional second array Y.
- Uses _compute_kernel with the backend.polynomial_kernel.
- Includes input validation via _check_inputs.

Fixes / relates to issue #2312

### DIFF
--- a/onedal/primitives/kernel_functions.py
+++ b/onedal/primitives/kernel_functions.py
@@ -201,6 +201,7 @@ def sigmoid_kernel(X, Y=None, gamma=1.0, coef0=0.0, queue=None):
         {"method": "dense", "scale": gamma, "shift": coef0}, backend.sigmoid_kernel, X, Y
     )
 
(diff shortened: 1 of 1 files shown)


icfaust (changes requested):
Thank you for your contribution! The Polynomial kernel already exists as `poly_kernel` in the same file. If we are to add a sklearnex interface, would need to be in the `sklearnex` module.

SnigdhaSarkar16:
Thank you! I understand that the polynomial kernel already exists as poly_kernel. I’ll work on adding a sklearnex interface in the appropriate module instead.

PR #2741, opened 2025-10-21
MAINT: Updates for XGBoost3.1 model structures

david-cortes-intel:
## Description

This PR updates the model builders from xgboost objects to be able to process multi-class classification models with the new structure introduced in version 3.1.0, which has vector-valued intercepts.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/tests/test_model_builders.py
+++ b/tests/test_model_builders.py
@@ -90,10 +90,6 @@ catboost_skip_shap_msg = (
     "See https://github.com/catboost/catboost/issues/2556."
 )
 
-# TODO: remove checks using these once treelite becomes compatible with xgboost>=3.1.0
-xgb_is_31_version = xgb.__version__.startswith("3.1")
-tl_xgb_incompat_msg = "Incompatibilities between treelite and xgboost."
-
 
 # Note: models have an attribute telling whether SHAP calculations
 # are supported for it or not. When that attribute is 'False', attempts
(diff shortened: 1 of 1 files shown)


tests/test_model_builders.py
```diff
@@ -233,7 +236,10 @@ def test_xgb_regression(
 
     xgb_model = make_xgb_model(objective, base_score, sklearn_class, empty_trees)
     if from_treelite:
-        xgb_model = treelite.frontend.from_xgboost(xgb_model)
+        try:
+            xgb_model = treelite.frontend.from_xgboost(xgb_model)
+        except Exception:
```
icfaust: Any way you can catch a more specific exception here?

david-cortes-intel: It can fail in multiple ways which throw different errors, and it looks like the first error encountered here will be different in the next treelite release. It could be done, but this is just a temporary workaround that should be removed soon.

tests/test_model_builders.py
```diff
@@ -90,6 +90,10 @@
     "See https://github.com/catboost/catboost/issues/2556."
 )
 
+# TODO: remove checks using these once treelite becomes compatible with xgboost>=3.1.0
+xgb_is_31_version = xgb.__version__.startswith("3.1")
```
razdoburdin: The checking for version >= 3.1 would be more robust.

david-cortes-intel: Yes, but this is just a temporary workaround that should hopefully get removed before version 3.2 arrives.

tests/test_model_builders.py
david-cortes-intel: This is not being done anymore.

tests/test_model_builders.py
david-cortes-intel: Removed the check as treelite supports the latest xgboost models by now.

david-cortes-intel:
Issue with treelite incompatibilities was fixed with their latest update, so removed the workarounds from this PR.

PR #2752, opened 2025-10-27
ENH: Increase correction pairs used for quasi Newton approximations

david-cortes-intel:
## Description

Logistic regression calls the L-BFGS-B solver from SciPy with mostly default parameters mimicking scikit-learn. The parameters for the solver were meant as default choices for a general solver that's aimed at solving non-convex and bound-constrained problems, but for a comparatively easier problem like L2-regularized logistic regression, they are not optimal - instead, better approximations to the true Hessian (through more corrections in the underlying quasi-Newton approximations) are typically beneficial for high-dimensional datasets.

Benchmarks on an Ice Lake machine (note: comparison is main branch against this, so relative numbers below 1 mean improvement in this PR):
image

image

**Note:** In many of these benchmarks, the solver is not reaching convergence, so the comparison is not always apples-to-apples. Cases where neither branch reached convergence are highlighted in yellow in the screenshots above:
```
Increase the number of iterations to improve the convergence (max_iter=200).
You might also want to scale the data as shown in:
    https://scikit-learn.org/stable/modules/preprocessing.html
Please also refer to the documentation for alternative solver options:
    https://scikit-learn.org/stable/modules/linear_model.html#logistic-regression
  n_iter_i = _check_optimize_result(
/export/users/dcortes/repos/sklex-benchenv/daal4py/sklearn/linear_model/logistic_path.py:266: ConvergenceWarning: lbfgs failed to converge after 200 iteration(s) (status=1):
STOP: TOTAL NO. OF ITERATIONS REACHED LIMIT
````

Hence, even though some cases might look like performance degradations here, they are actually reaching better solutions, so it's not an exact comparison.

**Note2:** the docs CI job is failing due to a rate limiting issue with medium.com.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/daal4py/sklearn/linear_model/logistic_path.py
+++ b/daal4py/sklearn/linear_model/logistic_path.py
@@ -249,6 +249,11 @@ def __logistic_regression_path(
             iprint = [-1, 50, 1, 100, 101][
                 np.searchsorted(np.array([0, 1, 2, 3]), verbose)
             ]
+            # Note: this uses more correction pairs than the implementation in scikit-learn,
+            # which means better approximation of the Hessian at the expense of slower updates.
+            # This is beneficial for high-dimensional convex problems without bound constraints
+            # like the logistic regression being fitted here. For larger problems with sparse
+            # data (currently not supported), it might benefit from increasing the number further.
             opt_res = optimize.minimize(
                 func,
                 w0,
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
> Approving, as this is really an improvement. But need a follow up about the benchmarking methodology: Should we fix the tolerance in order to have equally accurate solution in both cases, when comparing to stock scikit, for example?

Yes, I'm working on a fix.

david-cortes-intel:
PR for the benchmarks: https://github.com/IntelPython/scikit-learn_bench/pull/190

PR #2781, opened 2025-11-18
FIX: Make detection of nested parallelism look only at what joblib modifies

david-cortes-intel:
## Description

Updates the logic in sklearnex that detects when it is running under parallel joblib jobs to look only at what joblib will always set.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


daal4py/sklearn/_n_jobs_support.py
```diff
@@ -46,16 +46,24 @@ def get_suggested_n_threads(n_cpus):
...
+    # (e.g. if installing SciPy from pip with its bundled openblas, but not if
+    # installing it from conda-forge with MKL as BLAS backend), and might be set
+    # by the user to something that doesn't match with the number of parallel
+    # jobs from joblib - hence this looks at the openmp configuration, even
+    # though openmp is not used by oneDAL.
     n_threads_map = {
         lib_ctl.internal_api: lib_ctl.get_num_threads()
         for lib_ctl in threadpool_controller.lib_controllers
-        if lib_ctl.internal_api != "mkl"
+        if lib_ctl.internal_api == "openmp"
```
olegkkruglov: does it make sense in this scenario to keep the map? single variable must be sufficient.

david-cortes-intel: Updated.

david-cortes-intel:
From a look at the failing jobs that report memory leaks: it seems the leak is something that happens with global data during thread initialization, and dependent on the number of threads with which oneDAL is initialized. It doesn't show up as 'direct leak', so likely something with the teardown, possibly related to the pointer corruptions here:
https://github.com/uxlfoundation/oneDAL/pull/3304

PR is safe to merge, it's just that the tests are not well designed.

david-cortes-intel:
Sporadically failing tests should be fixed with this PR that removes them: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2887

PR #2788, opened 2025-11-21
MAINT: Update skipped tests for sklearn1.8

david-cortes-intel:
## Description

As part of the updates for changes in sklearn1.8's logistic regression, some tests appear to have been adapted only partially. This PR removes a failing test for functionality that's being removed in sklearn1.8 and which is currently causing CI failures on the nightly jobs.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


david-cortes-intel:
Looks like it fixed the issue, but note that the CI jobs still fail due to other unrelated reasons.

icfaust:
At that point, a `pytest.skipif(not sklearn_check_version("1.8"), reason="no longer relevant in sklearn >1.8")` may be clearer

david-cortes-intel:
> At that point, a `pytest.skipif(not sklearn_check_version("1.8"), reason="no longer relevant in sklearn >1.8")` may be clearer

Modified to use skipif.

tests/test_model_builders.py
```diff
@@ -2172,21 +2172,23 @@ def test_logreg_builder_sequential_calls():
...
-        [
-            (
-                LogisticRegression(multi_class="ovr"),
-                3,
+    [
+        pytest.param(
+            LogisticRegression(multi_class="ovr"),
+            3,
+            marks=pytest.mark.skipif(
+                not sklearn_check_version("1.8"),
```
avolkov-intel: We should do the opposite skip if sklearn_check_version("1.8")

david-cortes-intel: Yes, fixed.

david-cortes-intel:
> At that point, a `pytest.skipif(not sklearn_check_version("1.8"), reason="no longer relevant in sklearn >1.8")` may be clearer

Actually not possible, because it requires instantiating the model objects with the removed arguments before the test is skipped.

tests/test_model_builders.py
```diff
@@ -2184,7 +2184,7 @@ def test_logreg_builder_sequential_calls():
             ),
         ]
         if not sklearn_check_version("1.8")
-        else [(LogisticRegression(solver="liblinear"), 2)]
```
avolkov-intel: Why do we need to remove this test, I think it's not deprecated

david-cortes-intel: The test is checking that it fails to convert the model. This would work fine.

david-cortes-intel:
Same change was made by another PR that got merged before this: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2847

PR #2793, opened 2025-11-24
MAINT: Update TSNE for sklearn1.8

david-cortes-intel:
## Description

This PR modifies the logic in the TSNE class to offload to base sklearn as early as possible when it receives unsupported inputs. With these changes, it now allows using the stock version for cases that they now support but oneDAL doesn't, such as PCA initialization with sparse inputs.

Along the way, it also makes a couple necessary changes that appear to have been scheduled for sklearn1.2 but were not updated here, and it updates the documentation about what is and isn't supported for this algorithm.

Note that a lot of the code added here mirrors scikit-learn, since for the most part this class is a copy-paste of it that's been getting out of synch:
https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/manifold/_t_sne.py

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/daal4py/sklearn/manifold/_t_sne.py
+++ b/daal4py/sklearn/manifold/_t_sne.py
@@ -164,7 +164,14 @@ class TSNE(BaseTSNE):
                 (
                     (
                         isinstance(self.init, str)
-                        and self.init in ["random", "pca", "warn"]
+                        and self.init
+                        in ["random", "pca"]
+                        + (
+                            ["warn"]
+                            if sklearn_check_version("1.0")
+                            and not sklearn_check_version("1.2")
+                            else []
+                        )
                     )
                     or isinstance(self.init, np.ndarray),
                     "'init' must be 'exact', 'pca', or a numpy array.",
@@ -175,7 +182,7 @@ class TSNE(BaseTSNE):
         if not _dal_ready:
             return super()._fit(X, skip_num_points)
 
-        if not sklearn_check_version("1.2"):
+        if sklearn_check_version("1.0") and not sklearn_check_version("1.2"):
             if isinstance(self.init, str) and self.init == "warn":
                 warnings.warn(
                     "The default initialization in TSNE will change "
@@ -188,7 +195,7 @@ class TSNE(BaseTSNE):
         else:
             self._init = self.init
 
-        if not sklearn_check_version("1.2"):
+        if sklearn_check_version("1.0") and not sklearn_check_version("1.2"):
             if self.learning_rate == "warn":
                 warnings.warn(
                     "The default learning rate in TSNE will change "


daal4py/sklearn/manifold/_t_sne.py
```diff
@@ -122,36 +122,80 @@ def _daal_tsne(self, P, n_samples, X_embedded):
...
-            )
-            self._learning_rate = 200.0
+        if not sklearn_check_version("1.2"):
+            if self.learning_rate == "warn":
+                warnings.warn(
+                    "The default learning rate in TSNE will change "
+                    "from 200.0 to 'auto' in 1.2.",
+                    FutureWarning,
+                )
+                self._learning_rate = 200.0
```
Vika-F, line 205: Looks like more versioning is needed here, as it fails on older versions of scikit:
AttributeError: 'TSNE' object has no attribute '_learning_rate'

david-cortes-intel: Fixed.

daal4py/sklearn/manifold/_t_sne.py
```diff
@@ -122,36 +122,82 @@ def _daal_tsne(self, P, n_samples, X_embedded):
...
+                # out to the base class, wherever in the process they might be encountered.
+                (
+                    np.isscalar(self.angle) and self.angle > 0.0 and self.angle < 1.0,
+                    "'angle' must be between 0.0 - 1.0",
+                ),
+                (self.early_exaggeration >= 1.0, "early_exaggeration must be at least 1"),
+                (
+                    (
+                        isinstance(self.init, str)
+                        and self.init in ["random", "pca", "warn"]
```
avolkov-intel: What is the meaning of "warn" initialization method? Is it documented somewhere, because this is not present in stock sklearn?

daal4py/sklearn/manifold/_t_sne.py
```diff
@@ -122,36 +122,82 @@ def _daal_tsne(self, P, n_samples, X_embedded):
...
+                    )
+                    or isinstance(self.init, np.ndarray),
+                    "'init' must be 'exact', 'pca', or a numpy array.",
```
avolkov-intel, line 177: I think there's a mistake in error message

daal4py/sklearn/manifold/_t_sne.py
david-cortes-intel: It was the default parameter in sklearn1.0:
https://github.com/scikit-learn/scikit-learn/blob/baf828ca126bcb2c0ad813226963621cafe38adb/sklearn/manifold/_t_sne.py#L750

It means that it will issue a future warning if not changed from that default.

daal4py/sklearn/manifold/_t_sne.py
david-cortes-intel: 'warn' is no longer allowed in newer sklearn versions, so it's not referenced here.

avolkov-intel: but init can't be exact? Also it can't be numpy array according to this condition

david-cortes-intel: No, there's no init 'exact'; and there's an 'or' condition where it allows numpy arrays.

doc/sources/guide/acceleration.rst
```diff
@@ -34,9 +34,9 @@ The overall acceleration of TSNE depends on the acceleration of each of these al
   - ``metric`` != `'euclidean'` or `'minkowski'` with ``p`` != `2`
 - The Gradient Descent part of the algorithm supports all parameters except:
 
-  - ``n_components`` = `3`
-  - ``method`` = `'exact'`
-  - ``verbose`` != `0`
+  - ``n_components`` > ``2``
+  - ``method`` = ``'exact'``
```
avolkov-intel, line 38: But in case of exact method we don't fallback to sklearn, should we add it as supported?

daal4py/sklearn/manifold/_t_sne.py
```diff
@@ -377,40 +422,11 @@ def _fit(self, X, skip_num_points=0):
...
-        return self._tsne(
-            P,
-            degrees_of_freedom,
-            n_samples,
-            X_embedded=X_embedded,
-            neighbors=neighbors_nn,
-            skip_num_points=skip_num_points,
-        )
+        X_embedded = check_array(X_embedded, dtype=[np.float32, np.float64])
+        return self._daal_tsne(P, n_samples, X_embedded=X_embedded)
```
avolkov-intel, line 429: Is it correct that in case method == exact we would still call this function?

doc/sources/guide/acceleration.rst
david-cortes-intel: It falls back on the first condition in the patching chain.

sklearnex/manifold/tests/test_tsne.py
```diff
@@ -161,8 +163,12 @@ def test_tsne_functionality_and_edge_cases(
         assert np.any(embedding != 0)
 
 
+# Note: since sklearn1.2, the PCA initialization divides by standard deviations of components.
```
avolkov-intel, line 166: Do we need to add another test case in the future instead of removed one?

daal4py/sklearn/manifold/_t_sne.py
david-cortes-intel: This code would not be reached with method 'exact'.

sklearnex/manifold/tests/test_tsne.py
david-cortes-intel: No, what other case would you add?

Vika-F (approved):
Lets wait for a (semi)green Pre-Commit and LGTM.

PR #2803, opened 2025-11-27
MAINT: Skip conformance tests running with too old array API packages

david-cortes-intel:
## Description

Some tests introduced in sklearn1.8 require versions of array API packages that are not available for older python versions.

Along with bumping the python versions used for the CI jobs that test against the main branch from sklearn, it's also required to skip these problematic tests once sklearn1.8 is released.

This PR tries to deselect the tests based on python versions, but I'm not sure that it will work as intended. Currently, it seems that all of the deselections work only on versions of scikit-learn, and I'm not sure what would be the logic for deselecting based on python versions or versions of other packages instead.

Note that the actual requirement is `array-api-strict>=2.4.1`, which in turn requires `python>=3.12`.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/deselected_tests.yaml
+++ b/deselected_tests.yaml
@@ -388,10 +388,11 @@ deselected_tests:
   # These fail due to not printing everything that scikit-learn prints on verbose mode
   - cluster/tests/test_k_means.py::test_kmeans_verbose
 
-  # These require version numbers of array API packages that are not available on older Python versions
-  - metrics/tests/test_classification.py::test_probabilistic_metrics_array_api python<3.12
-  - metrics/tests/test_common.py::test_array_api_compliance python<3.12
-  - utils/tests/test_stats.py::test_weighted_percentile_array_api_consistency python<3.12
+  # These require version numbers of array API packages that are not available on older Python versions.
+  # Since it's not possible to select or deselect based on Python version, they are removed entirely.
+  - metrics/tests/test_classification.py::test_probabilistic_metrics_array_api
+  - metrics/tests/test_common.py::test_array_api_compliance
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
This kind of syntax didn't work - tests were not skipped.

@ethanglaser Any ideas about how these conditional skips could be added?

ethanglaser:
I don't believe its possible to deselect based on python, probably makes sense to just deselect entirely and add comment about constraint

deselected_tests.yaml
```diff
@@ -388,6 +388,11 @@ deselected_tests:
   # These fail due to not printing everything that scikit-learn prints on verbose mode
   - cluster/tests/test_k_means.py::test_kmeans_verbose
 
+  # These require version numbers of array API packages that are not available on older Python versions
+  - metrics/tests/test_classification.py::test_probabilistic_metrics_array_api python<3.12
+  - metrics/tests/test_common.py::test_array_api_compliance python<3.12
+  - utils/tests/test_stats.py::test_weighted_percentile_array_api_consistency python<3.12
```
david-cortes-intel:
```suggestion
  # These require version numbers of array API packages that are not available on older Python versions.
  # Since it's not possible to select or deselect based on Python version, they are removed entirely.
  - metrics/tests/test_classification.py::test_probabilistic_metrics_array_api
  - metrics/tests/test_common.py::test_array_api_compliance
  - utils/tests/test_stats.py::test_weighted_percentile_array_api_consistency
```

david-cortes-intel:
> I don't believe its possible to deselect based on python, probably makes sense to just deselect entirely and add comment about constraint

Changed to deselect unconditionally.

PR #2814, opened 2025-12-01
MAINT: More deselections due to small numeric differences in logistic regression

david-cortes-intel:
## Description

Deselects another test that can fail due to small numeric differences that are expected from the implementation of logistic regression in sklearnex.

Note that this deselected test is not currently causing any failure in any CI job, but the failure can be observed when testing things locally in some setups.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/deselected_tests.yaml
+++ b/deselected_tests.yaml
@@ -369,6 +369,7 @@ deselected_tests:
   - linear_model/tests/test_logistic.py::test_newton_cholesky_fallback_to_lbfgs
   - linear_model/tests/test_logistic.py::test_logistic_cv_sparse[csr_array]
   - linear_model/tests/test_logistic.py::test_logistic_cv_sparse[csr_matrix]
+  - linear_model/tests/test_logistic.py::test_logistic_cv_multinomial_score
   - tests/test_common.py::test_estimators[LogisticRegressionCV(cv=3,max_iter=5)-check_sample_weight_equivalence_on_dense_data]
   - ensemble/tests/test_bagging.py::test_estimators_samples >=1.0,<1.1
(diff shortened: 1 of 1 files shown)


ethanglaser:
Does it happen starting in sklearn 1.8 or all versions?

david-cortes-intel:
> Does it happen starting in sklearn 1.8 or all versions?

Happens in older versions too, but is platform-dependent, so we don't see it on the CI runners.

deselected_tests.yaml
```diff
@@ -369,6 +369,7 @@ deselected_tests:
   - linear_model/tests/test_logistic.py::test_newton_cholesky_fallback_to_lbfgs
   - linear_model/tests/test_logistic.py::test_logistic_cv_sparse[csr_array]
   - linear_model/tests/test_logistic.py::test_logistic_cv_sparse[csr_matrix]
+  - linear_model/tests/test_logistic.py::test_logistic_cv_multinomial_score
```
ethanglaser, line 372: Can we move this to the public section?

ethanglaser: Also are all tests failing or just linear_model/tests/test_logistic.py::test_logistic_cv_multinomial_score[42-neg_log_loss-multiclass_agg_list3]

david-cortes-intel: As far as I can tell, all of the parameterizations fail in the platforms where this failure is seen.

david-cortes-intel: > Can we move this to the public section?

I initially observed this on an internal machine that might be used for non-public CI jobs. Although I see it's now showing up in some public jobs too.

PR #2821, opened 2025-12-03
FIX: Fix KNN predictions on `X=None`

david-cortes-intel:
## Description

Sklearn allows passing `X=None` to all prediction-related methods of KNN estimators, but sklearnex takes that as an all-nan data, which is not the same. This PR fixes it.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/deselected_tests.yaml
+++ b/deselected_tests.yaml
@@ -31,6 +31,9 @@ deselected_tests:
   - linear_model/tests/test_ridge.py::test_ridge_shapes_type
   - linear_model/tests/test_ridge.py::test_ridge_cv_results_predictions[2-False-False]
   - linear_model/tests/test_ridge.py::test_ridge_cv_results_predictions[2-False-True]
+  - neighbors/tests/test_neighbors.py::test_nan_euclidean_support[KNeighborsClassifier-params0]
+  - neighbors/tests/test_neighbors.py::test_nan_euclidean_support[KNeighborsRegressor-params1]
+  - neighbors/tests/test_neighbors.py::test_nan_euclidean_support[LocalOutlierFactor-params6]
   # sklearn 1.7 unsupported features
   - tests/test_common.py::test_estimators[LinearRegression()-check_sample_weight_equivalence_on_dense_data]
 


sklearnex/neighbors/common.py
```diff
@@ -151,6 +151,17 @@ def _onedal_supported(self, device, method_name, *data):
...
             f"sklearn.neighbors.{class_name}.{method_name}"
         )
+        patching_status.and_conditions(
+            [
+                (
+                    not (
+                        data[0] is None
+                        and method_name in ["predict", "predict_proba", "score"]
+                    ),
+                    "Predictions on 'None' data are handled by internal sklearn methods.",
```
icfaust: Shouldn't this instead look for the `_fit_X` attribute if `X` is None? I'm not entirely sure I follow the solution of this PR (I do see the problem however).

david-cortes-intel: No, that wouldn't give the same results. The sklearn docs mention:
> In this case, the query point is not considered its own neighbor
https://github.com/scikit-learn/scikit-learn/blob/4a10d0ed8d85e6ed24a647bd28a65c0c64b101ef/sklearn/neighbors/_base.py#L767C13-L767C77

And one can see that forwarding the same fit data doesn't generate the same results:
```python
import numpy as np
from sklearn.neighbors import KNeighborsClassifier
X = rng.standard_normal(size=(5,3))
y = rng.integers(2, size=X.shape[0])
knn = KNeighborsClassifier(n_neighbors=2).fit(X, y)
knn.predict_proba(None)
```
```
array([[0.5, 0.5],
       [1. , 0. ],
       [0. , 1. ],
       [0.5, 0.5],
       [0.5, 0.5]])
```
```python
knn.predict_proba(X)
```
```
array([[1. , 0. ],
       [0.5, 0.5],
       [0.5, 0.5],
       [0.5, 0.5],
       [0. , 1. ]])
```

icfaust: But this is why we have the following logic: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/neighbors/neighbors.py#L297-L306 Just to full block it is a loss of functionality seems a bit heavy handed. At a minimum the explanation text for debugging needs to be made clearer, it needs to be discussed with someone with KNN knowledge on oneDAL side about this case @ethanglaser ?

david-cortes-intel: If I'm understanding the logic correctly, that would either way produce incorrect results in cases where there are observations with the same 'X' features but different 'y'.

david-cortes-intel: Actually no, turns out that sklearn suffers from the same problem.

david-cortes-intel: Either way, even if this ends up offloading the operation to sklearn, since the class overrides the method `kneighbors` which sklearn ends up calling, the code from `onedal.neighbors` ends up being used instead of sklearn's.

david-cortes-intel: I've now modified it to use the sklearnex methods for `predict_proba` so that the log entry would be reflective of what happens, but it looks like there's no implementations of `predict` and `score` for `X=None` on either the onedal or sklearnex side, so we'll still need to fall back on those, unless they were to be reimplemented on the python side by calling `kneighbors` internally like sklearn does for theirs. Note that it still ends up calling oneDAL since it overrides the methods from the sklearn class that are called internally.

Vika-F:
Can you please clarify how the testing results change after this fix?
Were there some tests which failed previously with `X=None`, and now they are passing?

I understand the problem, but how can I understand that the changes actually fix it?

david-cortes-intel:
> Can you please clarify how the testing results change after this fix? Were there some tests which failed previously with `X=None`, and now they are passing?
> 
> I understand the problem, but how can I understand that the changes actually fix it?

Yes, but those tests are not being executed in any CI job.

david-cortes-intel:
> Can you please clarify how the testing results change after this fix? Were there some tests which failed previously with `X=None`, and now they are passing?
> 
> I understand the problem, but how can I understand that the changes actually fix it?

Turns out the tests were deselected. I've reincluded them in this PR.

PR #2822, opened 2025-12-03
FIX: Allow KNN with missing values

david-cortes-intel:
## Description

KNN algorithms on sklearn support data with missing values, but sklearnex makes a check for 'all finite'. This PR makes it fall back to sklearn when receiving data that has missing values.

Disclaimer: I'm not 100% sure about whether oneDAL supports KNN with missing values or not.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/neighbors/common.py
+++ b/sklearnex/neighbors/common.py
@@ -70,7 +70,10 @@ class KNeighborsDispatchingBase(oneDALEstimator):
                 X,
                 dtype=[np.float64, np.float32],
                 accept_sparse=True,
-                force_all_finite=not self.effective_metric_.startswith("nan"),
+                force_all_finite=not (
+                    isinstance(self.effective_metric_, str)
+                    and self.effective_metric_.startswith("nan")
+                ),
             )
             self.n_samples_fit_ = _num_samples(self._fit_X)
             self.n_features_in_ = _num_features(self._fit_X)


sklearnex/neighbors/common.py
```diff
@@ -254,6 +257,21 @@ def _onedal_supported(self, device, method_name, *data):
                 ]
             ):
                 return patching_status
+        X = data[0]
+        if X is not None and not sp.issparse(X):
+            xp, _ = get_namespace()
+            try:
+                has_nan = xp.any(xp.isnan(X))
```
icfaust: This should instead be done with the optimized `assert_all_finite` as this has performance implications and impacts follow-on finite checks. I recommend looking at the ensemble algos and @Alexsandruss 's solution for this.

david-cortes-intel: Switch to use the optimized internal version.

icfaust:
scikit-learn/scikit-learn#25319 @david-cortes-intel . Can you detail the missing value use case with an example? As of now we support these metrics: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/neighbors/common.py#L221 which does not include `nan_euclidean` which would be the use case of this missing value support.

david-cortes-intel:
> [scikit-learn/scikit-learn#25319](https://github.com/scikit-learn/scikit-learn/issues/25319) @david-cortes-intel . Can you detail the missing value use case with an example? As of now we support these metrics: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/neighbors/common.py#L221 which does not include `nan_euclidean` which would be the use case of this missing value support.

It still fails this test:
https://github.com/scikit-learn/scikit-learn/blob/4a10d0ed8d85e6ed24a647bd28a65c0c64b101ef/sklearn/neighbors/tests/test_neighbors.py#L2356

icfaust:
@david-cortes-intel then I think there is a bug in the metrics check. I think that should be the avenue taken to fix the issue.

icfaust:
Just wanted to note that I know that its a bit convoluted and annoying on the neighbors side of things here, I just want to make sure that if we can solve it without a separate finite check, then we should do that.

david-cortes-intel:
> @david-cortes-intel then I think there is a bug in the metrics check. I think that should be the avenue taken to fix the issue.

Switched it to use the metric checks instead.

PR #2827, opened 2025-12-05
FIX: Allow sparse data on TSNE

david-cortes-intel:
## Description

The docs say that TSNE does not support sparse inputs, and there are checks against them.

But it turns out that TSNE doesn't use the actual data - instead, it has some preliminary stages where it calculates distances or neighbors or PCA, and the results from those are then used to generate arrays that are passed to the actual TSNE algorithm. Those arrays are always dense, regardless of how the data comes in originally, so in way sklearnex has partial support for sparse data on TSNE.

This PR corrects the docs and removes the checks for sparsity.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/daal4py/sklearn/manifold/_t_sne.py
+++ b/daal4py/sklearn/manifold/_t_sne.py
@@ -149,7 +149,7 @@ class TSNE(BaseTSNE):
                 ),
                 # Scikit-learn didn't support sparse PCA initialization before 1.8.
                 # This nevertheless offloads it to sklearn because it produces a different
-                # error message than what would be throws by simply passing the input to PCA.
+                # error message than what would be thrown by simply passing the input to PCA.
                 (
                     sklearn_check_version("1.8")
                     or not (


Vika-F:
sklearn's `test_t_sne.py` doesn't have a case for sparse PCA initialization testing, only for sparse precomputed distances testing.
I think it worth adding the test on our side to test this branch.

david-cortes-intel:
> sklearn's `test_t_sne.py` doesn't have a case for sparse PCA initialization testing, only for sparse precomputed distances testing. I think it worth adding the test on our side to test this branch.

It does have a test like that as of 1.8: https://github.com/scikit-learn/scikit-learn/blob/bd0dd87f792ee88c8ec70248299aae779b51ac16/sklearn/manifold/tests/test_t_sne.py#L320

david-cortes-intel:
CI failure is unrelated to these changes:
```
libgomp: Thread creation failed: Resource temporarily unavailable
```

daal4py/sklearn/manifold/_t_sne.py
```diff
@@ -147,11 +147,15 @@ def _fit(self, X, skip_num_points=0):
                     daal_check_version((2021, "P", 600)),
                     "oneDAL version is lower than 2021.6.",
                 ),
+                # Scikit-learn didn't support sparse PCA initialization before 1.8.
+                # This nevertheless offloads it to sklearn because it produces a different
+                # error message than what would be throws by simply passing the input to PCA.
```
Vika-F:
```suggestion
                # error message than what would be thrown by simply passing the input to PCA.
```

PR #2846, opened 2025-12-10
FIX: Make global queue objects thread-local

david-cortes-intel:
## Description

Sycl queues on GPU are managed through global objects, but since the introduction of array API, some operations might be offloaded to other libraries that raise the GIL during their operations, and might thus lead to conflicts if sklearnex objects are used in parallel through python threads instead of being in separate processes.

This PR fixes potential thread racing issues by making the queues thread-local, with a helper class as suggested by @ahuber21 .

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/tests/test_parallel.py
+++ b/sklearnex/tests/test_parallel.py
@@ -61,9 +61,9 @@ def test_queue_is_thread_local():
         with QM.manage_global_queue(queue_arg):
             barrier.wait()
             if is_cpu:
-                assert not QM.__global.queue.sycl_device.is_gpu
+                assert not QM.__globals.queue.sycl_device.is_gpu
             else:
-                assert QM.__global.queue.sycl_device.is_gpu
+                assert QM.__globals.queue.sycl_device.is_gpu
             barrier.wait()
 
     t1 = Thread(target=fn, args=("cpu", True))


david-cortes-intel:
CI error on CPU appears to happen elsewhere too, likely not related to the changes here.

ahuber21:
Any quick test than we can add for verification? Maybe accessing the globals from two threads and checking the queue addresses differ?

david-cortes-intel:
> Any quick test than we can add for verification? Maybe accessing the globals from two threads and checking the queue addresses differ?

The address shouldn't differ if I understand python's logic correctly, only the attributes should differ. But the attribute would usually be `None` for queues, and otherwise it is usually a temporary variable within the exection.

Do you have some example for how to write a test like that?

ahuber21:
I don't have a working build of sklearnex, so nothing concrete. But sketching the idea.

```python

def foo(queue, tid):
  from onedal.utils import _sycl_queue_manager as QM
  
  with QM.manage_global_queue(queue):
     # do something with it? or wait?
     print(f"Thread {tid}: {QM.__global=}")
     print(f"Thread {tid}: {QM.__global.queue=}")

     # idle to ensure GIL-passing? or run GIL-free? In any case, make sure that both threads run concurrently

from threading import Thread

t1 = Thread(target=foo, args=('cpu', 1))
t1.start()
t2 = Thread(target=foo, args=('gpu', 2))
t2.start()

t1.join()
t2.join()
```

The test would be asserting that `QM.__global.queue` differs while ensuring that both threads are alive at the same time. It should probably be extended to the other attributes.
Maybe comparing addresses of attributes makes more sense than values ...

david-cortes-intel:
> I don't have a working build of sklearnex, so nothing concrete. But sketching the idea.
> 
> ```python
> def foo(queue, tid):
>   from onedal.utils import _sycl_queue_manager as QM
>   
>   with QM.manage_global_queue(queue):
>      # do something with it? or wait?
>      print(f"Thread {tid}: {QM.__global=}")
>      print(f"Thread {tid}: {QM.__global.queue=}")
> 
>      # idle to ensure GIL-passing? or run GIL-free? In any case, make sure that both threads run concurrently
> 
> from threading import Thread
> 
> t1 = Thread(target=foo, args=('cpu', 1))
> t1.start()
> t2 = Thread(target=foo, args=('gpu', 2))
> t2.start()
> 
> t1.join()
> t2.join()
> ```
> 
> The test would be asserting that `QM.__global.queue` differs while ensuring that both threads are alive at the same time. It should probably be extended to the other attributes. Maybe comparing addresses of attributes makes more sense than values ...

In that example, the GIL is not raised by the function executed in either thread, so they wouldn't execute concurrently.

david-cortes-intel:
Added a test with `time.sleep` to try to trigger concurrency, but not sure it's desirable to have.

ahuber21:
Could you try if this barrier approach works?

```python
import threading
from threading import Thread


def test_qm_global_queue_is_thread_local():
    from onedal.utils import _sycl_queue_manager as QM

    barrier = threading.Barrier(2)
    results = {}

    def foo(queue, tid):
        with QM.manage_global_queue(queue):
            # Ensure both threads are inside the context concurrently
            barrier.wait()

            q = QM.__global.queue
            print(f"Thread {tid}: QM.__global.queue = {q}")

            results[tid] = q

            # Keep both threads alive until values are captured
            barrier.wait()

    t1 = Thread(target=foo, args=('cpu', 1), name="worker-1")
    t2 = Thread(target=foo, args=('gpu', 2), name="worker-2")

    t1.start()
    t2.start()

    t1.join()
    t2.join()

    assert 1 in results
    assert 2 in results
    assert results[1] != results[2], (
        f"QM global queue leaked across threads: {results}"
    )
```

david-cortes-intel:
Changed to barrier.

ahuber21 (approved):
Cool! Thanks for adding the test. Approving but pending green CI of course.

PR #2856, opened 2025-12-17
Extend modelbuilders test set

avolkov-intel:
## Description

Since we plan an update to internal modelbuilders layout by introducing hybrid layout (https://github.com/uxlfoundation/oneDAL/pull/3455) we need to extend test set.
In the new layouts we would create full binary tree layout for levels <= 10 but after that we will store nodes in dense format. To check, that this implementation doesn't have any bugs we need to add tests where deep trees are used.

This test introduces new tests where trees with depth >=12 or >=15 created.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/tests/test_model_builders.py
+++ b/tests/test_model_builders.py
@@ -17,12 +17,12 @@
 import contextlib
 import gc
 import itertools
+import json
 import pickle
 import unittest
 import warnings
 from collections.abc import Callable
 from datetime import datetime
-import json
 
 import lightgbm as lgb
 import numpy as np
@@ -113,10 +113,10 @@ def force_shap_predict(model, X):
 def compute_xgb_tree_depth(booster: xgb.Booster) -> int:
     """
     Compute the maximum depth across all trees in an XGBoost model.
-    
+
     Returns the maximum depth found among all trees.
     """
-    
+
     max_depth = 0
     dump = booster.get_dump(dump_format="json")
     for i, tree_json in enumerate(dump):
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
Wouldn't it make more sense to have tests like this on the oneDAL repository?

tests/test_model_builders.py
```diff
@@ -106,81 +107,174 @@ def force_shap_predict(model, X):
         return model._predict_classification(X, fptype, "computeClassLabels", True, False)
 
 
+# ============================================================================
+# Tree depth utility functions for deep trees testing
+# ============================================================================
+def compute_xgb_tree_depth(booster: xgb.Booster) -> int:
```
david-cortes-intel, line 113: Since the idea is to test some internal detail of the implementation that doesn't anyhow depend on the input converter, how about checking only a single GBT library? Perhaps a random forest from sklearn converted with treelite could be used.

avolkov-intel: We actually have some existing tests related to xgboost and catboost including SHAP tests, it would require much more effort to create tests using RandomForest and TreeLite compared to updating existing tests

david-cortes-intel: Then how about leaving serialized model objects instead of re-creating them every time? These tests are already having issues in CIs due to taking too long to execute, and some are currently skipped due to timings.

ethanglaser:
Please address linter fail

tests/test_model_builders.py
avolkov-intel: I think this would make the testing more complicated, not sure if we use serialized models in other tests. Although, I agree that modelbuilders tests take quite a lot of time, the relative slowdown after introducing tests in this PR is small. That's why let's create a follow up for speeding up this set of tests

david-cortes-intel: From what I can see in the CI jobs here compared to those of recent commits, it adds roughly 1 additional minute in the CI servers.

PR #2873, opened 2025-12-30
FIX: Add missing dependency on `packaging`

david-cortes-intel:
## Description

There is some logic that tries to import the package `packaging` and falls back to `distutils` if not available. Since python3.13, distutils is no longer bundled in a base python install, and neither is `packaging`, so sklearnex fails to import without additional dependencies.

This PR adds an additional dependency on `packaging`.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -18,4 +18,4 @@ treelite==4.6.1
 array-api-compat==1.13.0
 array-api-strict==2.3.1 ; python_version <= '3.11'
 array-api-strict==2.4.1 ; python_version >= '3.12'
-packaging
+packaging==25.0


requirements-test.txt
```diff
@@ -18,3 +18,4 @@ treelite==4.6.1
 array-api-compat==1.13.0
 array-api-strict==2.3.1 ; python_version <= '3.11'
 array-api-strict==2.4.1 ; python_version >= '3.12'
+packaging
```
ethanglaser:
```suggestion
packaging==25.0
```
Test requirements should be versioned for reproducibility, seems like all python versions are installing 25.0

PR #2887, opened 2026-01-13
MAINT: Remove memory usage tests

david-cortes-intel:
## Description

This PR removes the memory usage tests, by deleting the whole file.

Motivation:
* These tests do not actually find memory leaks unless they are very big. We've previously had some that these CI jobs didn't catch.
* The logic behind the tests is flawed - it tests memory usage before/after executing some algorithm with the assumption that it should be close, but that's not how memory allocators work in general - i.e. if a program allocates memory and then frees it, the OS won't necessarily be returned the full amount of memory, unless using some specialized allocator with custom settings like 'jemalloc'. This is even less likely to happen when it comes to Python, which manages memory through its own allocators instead. Further, even if oneDAL were to use a different allocator and work in the way that the tests expect, they depend on other third-party libraries like arrow and pandas which do not.
* The tests are constantly generatic CI failures that need to be investigated, then ignored. They also cause issues in conda-forge where the builds need to complete the test suite successfully in order to be uploaded to conda-forge. In the last update, it took too many CI retries to get all the tests to pass, so these tests had to be temporarily deselected through a conda-specific patch: https://github.com/conda-forge/scikit-learn-intelex-feedstock/pull/75
* The tests fail even if running them unpatched (i.e. with stock scikit-learn) and removing the custom estimators from the tests.

I took a look at potential memory leaks through more specialized tools like leaksan and valgrind.

The "leaks" that I found were:
* Small global objects that come from PyBind11. PyBind11 by design leaks memory on initialization->exit of libraries. These take only 24 bytes per module, and are global objects that get created only once in a given process, so nothing to worry about: https://pybind11.readthedocs.io/en/stable/advanced/misc.html#common-sources-of-global-interpreter-lock-errors
image

* Memory leaks from small global objects in oneDAL, which should be solved with this PR: https://github.com/uxlfoundation/oneDAL/pull/3304
* Other leaks appear to come from python allocations. For example, this is what leaksan reports:

```
Direct leak of 37888 byte(s) in 148 object(s) allocated from:
    #0 0x7f6707d5f509 in malloc /usr/local/src/conda/compiler-rt-packages-21.1.8/compiler-rt/lib/asan/asan_malloc_linux.cpp:67:3
    #1 0x55b14c639d74 in PyObject_Malloc /usr/local/src/conda/python-3.12.11/Objects/obmalloc.c:801:12
    #2 0x55b14c639d74 in gc_alloc /usr/local/src/conda/python-3.12.11/Modules/gcmodule.c:2330:17
    #3 0x55b14c639d74 in _PyObject_GC_New /usr/local/src/conda/python-3.12.11/Modules/gcmodule.c:2345:20
```

But the same report is produced from running the tests unpatched:
```
Direct leak of 37888 byte(s) in 148 object(s) allocated from:
    #0 0x7fb8de55f509 in malloc /usr/local/src/conda/compiler-rt-packages-21.1.8/compiler-rt/lib/asan/asan_malloc_linux.cpp:67:3
    #1 0x55c0f97c2d74 in PyObject_Malloc /usr/local/src/conda/python-3.12.11/Objects/obmalloc.c:801:12
    #2 0x55c0f97c2d74 in gc_alloc /usr/local/src/conda/python-3.12.11/Modules/gcmodule.c:2330:17
    #3 0x55c0f97c2d74 in _PyObject_GC_New /usr/local/src/conda/python-3.12.11/Modules/gcmodule.c:2345:20
```

The source of the leak does not suggest anything with onedal/sklearnex in the paths, so it's probably just an issue of Python managing its own memory in a way that is not compatible with leak detectors.

---

<details>
<summary>Checklist:</summary>

</details>


sklearnex/tests/test_memory_usage.py
```diff
@@ -1,339 +0,0 @@
...
-from onedal.utils._array_api import _get_sycl_namespace
-from sklearnex import config_context
-from sklearnex.tests.utils import (
-    PATCHED_FUNCTIONS,
-    PATCHED_MODELS,
-    SPECIAL_INSTANCES,
-    DummyEstimator,
-)
-
-CPU_SKIP_LIST = (
```
avolkov-intel, line 46: We need another place to store all these TODO before removing this file

david-cortes-intel: From my investigations, these TODOs are not current issues, with the only exception of logistic regression, which will be dealt with in the C++ side.

icfaust:
@Vika-F @david-cortes-intel Just as a note for posterity, it went unmentioned that SYCL GPU memory leak checking was removed, of which leaksan and valgrind are unlikely to properly check. @david-cortes-intel did you do similar analyses for GPU algorithms?

I had sort of thought this was a stopgap measure we wanted to have until a better CI system was in place (when we talked about removing this in the past), which doesn't seem to have occurred...

david-cortes-intel:
> @Vika-F @david-cortes-intel Just as a note for posterity, it went unmentioned that SYCL GPU memory leak checking was removed, of which leaksan and valgrind are unlikely to properly check. @david-cortes-intel did you do similar analyses for GPU algorithms?
> 
> I had sort of thought this was a stopgap measure we wanted to have until a better CI system was in place (when we talked about removing this in the past), which doesn't seem to have occurred...

I have not checked the GPU cases. But be aware that ICX does have a SYCL version of ASAN which includes leaksan:
https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-2/device-side-compiler-sanitizers.html

PR #2893, opened 2026-01-16
MAINT: Move LogisticRegressionCV to preview

david-cortes-intel:
## Description

As a follow up from https://github.com/uxlfoundation/scikit-learn-intelex/pull/2871

This PR moves the recently introduced `LogisticRegressionCV` class to preview mode. While the PR that added this new estimator worked in the sense of behaving correctly and passing all the tests, it turned out that the performance of this new estimator ended up being worse than the one from stock sklearn, even though it's just reusing the same code path as `LogisticRegression` which runs substantially faster.

Hence, this PR moves it to preview until the performance issues have been investigated and addressed.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/preview/linear_model/logistic_regression.py
+++ b/sklearnex/preview/linear_model/logistic_regression.py
@@ -43,7 +43,7 @@ if daal_check_version((2024, "P", 1)):
 
         if sklearn_check_version("1.6"):
 
-            def __sklearn_tags__(self) -> "Tags":
+            def __sklearn_tags__(self) -> Tags:
                 tags = super().__sklearn_tags__()
                 tags.onedal_array_api = False
                 return tags


sklearnex/dispatcher.py
```diff
@@ -46,13 +46,19 @@ def get_patch_map_core(preview=False):
...
             import sklearn.decomposition as decomposition_module
+            import sklearn.linear_model as linear_model_module
 
             # Preview classes for patching
             from .preview.covariance import (
                 EmpiricalCovariance as EmpiricalCovariance_sklearnex,
             )
             from .preview.decomposition import IncrementalPCA as IncrementalPCA_sklearnex
 
+            if daal_check_version((2024, "P", 1)):
```
ethanglaser, line 57: Is the daal_check_version here necessary if we have another one in sklearnex.preview.linear_model logistic_regression.py already? Generally I think we don't check for that in this file - looks like ridge does but it may be a mistake

sklearnex/preview/linear_model/logistic_regression.py
```diff
@@ -0,0 +1,49 @@
...
+        fit = support_input_format(_daal4py_LogisticRegressionCV.fit)
+        predict_proba = _sklearnex_LogisticRegression.predict_proba
+        predict_log_proba = _sklearnex_LogisticRegression.predict_log_proba
+        decision_function = _sklearnex_LogisticRegression.decision_function
+
+        __doc__ = _daal4py_LogisticRegressionCV.__doc__
+
+        if sklearn_check_version("1.6"):
+
+            def __sklearn_tags__(self) -> "Tags":
```
ethanglaser: Is there a reason for the quotes around Tags? Based on other files this may make more sense:
```suggestion
            def __sklearn_tags__(self) -> Tags:
```

sklearnex/dispatcher.py
david-cortes-intel: Yes, because the one in non-preview logistic regression has an 'else' part which would make it define something different under the same class name.

sklearnex/preview/linear_model/logistic_regression.py
david-cortes-intel: This class is not available in earlier sklearn versions.

ethanglaser: I'm not sure that I understand, why is it different here compared to https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/base.py or https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/utils/_array_api.py

david-cortes-intel: Actually I now see that it's using a custom class `Tags` that's not from scikit-learn. I guess it should work.

ethanglaser:
At this point the PR looks good to me. But do we have any validation of LogisticRegressionCV? I don't believe conformance runs for preview algorithms, and we don't have any tests defined from our end

david-cortes-intel:
> At this point the PR looks good to me. But do we have any validation of LogisticRegressionCV? I don't believe conformance runs for preview algorithms, and we don't have any tests defined from our end

They are run under a separate job:
image

david-cortes-intel:
CI errors are in estimators not touched by this PR.

PR #2899, opened 2026-01-20
FIX, DOC, MAINT: Separate patching maps from sklearnex and daal4py, use unique IDs for patched objects

david-cortes-intel:
## Description

Follow up from https://github.com/uxlfoundation/scikit-learn-intelex/pull/2895 (includes the commits from that PR)

This PR makes some changes in the patching logic:
* It separates the patching map objects between daal4py and sklearnex, so that each one would define its own patching list without inheriting or modifying each other.
* It changes the structure of the patching maps to have a single tuple instead of a nested list with only one element. What I think happened here: probably, some classes required patching multiple things from sklearn in order to work (none of the current ones have such logic), but by this point, _lots_ of classes require patching functions outside of the class in question, and the functions they need are the same across estimators.
    * In this same line, it also adds a note in the docs that some functionalities strictly require patching to work correctly instead of being imported from `sklearnex`.
    * Note that if some estimator required a mechanism like this to work, it would not be possible to use it when importing from ``sklearnex`` without patching, so it doesn't make sense to allow for this in the design either way.
* It changes the structure of this tuple to have 4 elements, with the last one being the original sklearn implementation, instead of modifying a list dynamically upon patching. This allows:
    *  Making the patching mechanism work when used in Python threads - currently, the logic it uses can make `unpatch_sklearn` leave things patched if there are data races.
    * Fixing an issue in which estimators from sklearnex that are not part of sklearn are left as entries in the sklearn module even after unpatching.
* It expands the list of functions that get patched when patching a single estimator.
    * In this same regard, it also removes patching for function `_assert_all_finite`, which the sklearnex patching map was inheriting from the daal4py patching map. This function is internal to `sklearn`, and estimators from `sklearnex` should do the validation on their own end if needed. If they rely on this function being patched in order to work and aren't using the `raw_inputs` functionality which is meant to be removed, then I think they are doing something wrong if we go by the design logic of `sklearnex`. Otherwise, if a patched `_assert_all_finite` were required for some sklearnex estimator to work, then said estimator would not work if imported directly from `sklearnex` without patching, because in that scenario there is no replacement.
* It changes the structure of the patching maps to have a single ID for each object instead of being repeated under multiple names (like 'logreg' and 'logisticregression'), and in order to avoid clashes and to make the patching-related public functions that return names more useful, it uses the full path of the objects instead of a short name (e.g. `"sklearn.linear_model.LinearRegression"` instead of `"linreg"`).

**Note for reviewers:** please pay special attention to the change of not patching `_assert_all_finite`. I'm pretty sure it should't be patched under the current state of the library, but there might be some corner cases that I'm missing.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/doc/sources/config-contexts.rst
+++ b/doc/sources/config-contexts.rst
@@ -40,7 +40,7 @@ both libraries.
 .. important::
 
     In order for configuration contexts from the |sklearnex| to work correctly with
-    estimators imported from the ``sklearnex`` module used inside metaestimators from the ``sklearn``
+    estimators imported from the ``sklearnex`` module used inside meta-estimators from the ``sklearn``
     module (such as :obj:`sklearn.model_selection.GridSearchCV`), it is necessary to apply
     :doc:`patching <patching>` to the ``sklearn`` module. Importing estimators from ``sklearnex``
     without patching and passing them to metaestimators from ``sklearn`` might not


david-cortes-intel:
Failure in doc CI is due to a timeout.

sklearnex/dispatcher.py
```diff
@@ -59,439 +74,374 @@ def get_patch_map_core(preview=False):
...
+    ----------
+    name : str, list of str, or None
+        Names of the desired estimators to patch. Can pass a single instance name (e.g.
+        ``"sklearn.linear_model.LogisticRegression"``), or a list of names (e.g. ``["sklearn.linear_model.LogisticRegression", "sklearn.decomposition.PCA"]``).
+
+        If ``None``, will patch all the supported estimators.
+
+        See the :doc:`algorithm support table <algorithms>` for more information.
+
+        Note that functions related to :doc:`config contexts <config-context>` are
```
Vika-F: Doc build fails due to this reference:
:doc:`config contexts <config-context>`

Is it expected?

david-cortes-intel: It was a typo. Fixed now.

sklearnex/tests/test_monkeypatch.py
```diff
@@ -267,3 +273,11 @@ def get_estimators():
         assert "sklearnex" in svc.__module__
     finally:
         sklearnex.unpatch_sklearn()
+
+
+def test_extension_estimators_are_not_left_over():
+    from sklearn import linear_model
+
+    from sklearnex import patch_sklearn, unpatch_sklearn
```
david-cortes-intel, line 281:
```suggestion
    from sklearnex import patch_sklearn, unpatch_sklearn

    patch_sklearn()
    unpatch_sklearn()
```

icfaust:
@david-cortes-intel thanks for doing this. The naming scheme and some of the aspects of the patching map were not good, and this makes a very good step in the right direction.

daal4py/sklearn/monkeypatch/dispatcher.py
```diff
@@ -56,129 +81,176 @@
...
+# dict key: sklearn name
+# dict value: tuple entries:
+# - module from sklearn
+# - name of class/function within module
+# - sklearnex/daal4py replacement
+# - sklearn original if any
+PatchMap = dict[str, tuple[ModuleType, str, object, Optional[object]]]
+
 
+# Note: the keys of this dict are only used as internal IDs to keep
```
Vika-F, line 93: Is my understanding correct that this map should be removed completely when the 'OFF_ONEDAL_IFACE' option is removed?
Maybe it worth adding a note in order nor to forget to remove it?

david-cortes-intel: No, because there is still a patching function from daal4py that can be used independently of the one in sklearnex. What should be removed is the logic on the sklearnex side that falls back to daal4py patching instead of sklearnex patching.

doc/sources/config-contexts.rst
```diff
@@ -37,6 +37,15 @@ the configuration contexts and global settings of the |sklearnex| will also affe
 option is supported by it - meaning: the same context manager or global option setter is used for
 both libraries.
 
+.. important::
+
+    In order for configuration contexts from the |sklearnex| to work correctly with
+    estimators imported from the ``sklearnex`` module used inside metaestimators from the ``sklearn``
```
Vika-F: In Sklearn it is written like this, it would be better to align:
```suggestion
    estimators imported from the ``sklearnex`` module used inside meta-estimators from the ``sklearn``
```

daal4py/sklearn/monkeypatch/dispatcher.py
Vika-F: But do we really need that functionality in daal4py?
I.e. is there anything on daal4py side (i.e. what can be patched) which is not yet covered by sklearnex?
Or maybe daal4py patching provide some performance benefits in some algorithms comparing to sklearnex [I assume we do not test that]?

david-cortes-intel: No, but it would be removing functionality that is currently there. It's possible to use daal4py as a stand-alone module too (can even be used without compiling the `onedal` module), in which case there's less of an issue with version dependencies of sklearn.

But perhaps we could do that as a separate thing, since `daal4py` is either way not a separate package anymore.

PR #3101, opened 2026-04-08
DOC: Clarify that GPU package is required for array API on CPU

david-cortes-intel:
## Description

Turns out that using DPNP arrays on CPU with array API requires the DPC modules, as those objects contain sycl queues and the logic for handling them is in `_onedal_py_dpc`.

This PR updates the documentation to clarify that the GPU package is required for array API also on CPU.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/doc/sources/array_api.rst
+++ b/doc/sources/array_api.rst
@@ -33,7 +33,7 @@ on GPU without moving the data from host to device.
     global settings or using a :doc:`config_context <config-contexts>`.
 
 .. hint::
-    Executing computations on GPUs has additional dependencies, particularly on package ``scikit-learn-intelex-gpu`` - see
+    Executing computations on array API inputs (whether on CPUs or GPUs) requires additional dependencies, particularly on package ``scikit-learn-intelex-gpu`` - see
     :doc:`oneapi-gpu` for details.
 
 When passing array API inputs whose data is on a SYCL-enabled device (e.g. an Intel GPU), as


icfaust:
isnt pytorch cpu also array api? im worried you are conflating cpu sycl queues with array api by this PR.

david-cortes-intel:
> isnt pytorch cpu also array api? im worried you are conflating cpu sycl queues with array api by this PR.

Yes, and PyTorch is probably going to work, but not dpnp.

icfaust:
@david-cortes-intel this note should explicitly mention dpnp cpu support because it is otherwise misleading, array api support does not require gpu support unless the data is on a oneapi device, which is the case for pytorch gpu, dpnp cpu and dpnp gpu, otherwise not. array api strict, pytorch cpu and many of the mentioned array api compat frameworks will work with a host only build

david-cortes-intel:
> @david-cortes-intel this note should explicitly mention dpnp cpu support because it is otherwise misleading, array api support does not require gpu support unless the data is on a oneapi device, which is the case for pytorch gpu, dpnp cpu and dpnp gpu, otherwise not. array api strict, pytorch cpu and many of the mentioned array api compat frameworks will work with a host only build

@icfaust There's also more to it as the upcoming sklearn1.9 will have functions like `move_estimator_to` which, once we support, are likely to require the GPU package, due to working with string-valued `device` argument which on our side we handle by creating a queue out of it and moving data, which happens in the DPC module.

icfaust:
with that additional info that we should table or close this PR until more clarity can be added, and when sklearn 1.9 release occurs we can then update the docs here. if you want you can explicitly discuss oneapi device support, otherwise the info here is wrong

david-cortes-intel:
> with that additional info that we should table or close this PR until more clarity can be added, and when sklearn 1.9 release occurs we can then update the docs here. if you want you can explicitly discuss oneapi device support, otherwise the info here is wrong

That scikit-learn1.9 release is scheduled for mid-year, and we're aiming to be compatible with it in our 2026.1 release ahead of their release, which is the branch where this PR will go. PRs for 'move_estimator_to' support should follow soon.

icfaust:
This PR can be tabled until those changes are made, while i wont put a changes requested on it, it does need to be changed from its current form. do realize what you are suggesting would amount to a feature loss as array api support occurs on host builds for all current sklearn array api releases (1.2-1.8). additionally this may require an internal discussion as this would be deleterious to arm and risc-v sklearnex support which require host builds (depending on its status) @syakov-intel

Vika-F (approved):

PR #3112, opened 2026-04-16
MAINT: Reuse scikit-learn methods for LogisticRegression whenever it supports array API

david-cortes-intel:
## Description

There is some pure-Python code in logistic regression working with array API namespaces that implements methods that sklearn offers but oneDAL doesn't, such as '.decision_function()'.

By now, scikit-learn has implemented array API support for all of these methods, so there's no need for these pure-Python replacements which do roughly the same.

This PR makes the definition of those methods conditional on scikit-learn version, in order to use sklearn's when available but still be compatible with older sklearn versions that lack them.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


yuejiaointel (approved):
LGTM

PR #3114, opened 2026-04-17
ENH: Create logistic regression oneDAL object on the fly for predictions after fallbacks

david-cortes-intel:
## Description

This PR changes the logic of logistic regression on GPU to use oneDAL for predictions even if the model was fitted through scikit-learn, by creating a oneDAL object on-the-fly from the fitted coefficients.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


Vika-F:
From the description it is not clear which problem does this PR solve? I.e. what's wrong with the current approach?
Is it also related to 'everything follows X' issue?

david-cortes-intel:
> From the description it is not clear which problem does this PR solve? I.e. what's wrong with the current approach? Is it also related to 'everything follows X' issue?

Scikit-learn also supports array API, so there can be cases where `.fit()` falls back to scikit-learn, but then `.predict()` still meets the conditions to offload to oneDAL, other than there not being a oneDAL object.

Apart from that, this should also make it easier later on to implement `move_estimator_to`.

ethanglaser:
Looks like there are some GPU test fails to be addressed in private CI

ethanglaser:
Looks like there is still a private CI test fail (AttributeError: 'numpy.ndarray' object has no attribute 'device') for py3.10 GPU tests - at least the error message there coincides with the changes here

sklearnex/linear_model/logistic_regression.py
```diff
@@ -467,12 +472,13 @@ def _onedal_predict(self, X, queue=None):
                 self,
                 X,
                 reset=False,
-                accept_sparse=_sparsity_enabled,
+                accept_sparse="csr" if _sparsity_enabled else False,
```
avolkov-intel, line 476: Why do we need this update here? Did something change in sklearn 1.9?

onedal/linear_model/logistic_regression.py
```diff
@@ -88,6 +89,33 @@ def fit(self, X, y, queue=None):
...
+                    (coef_.shape[0], coef_.shape[1] + 1), device=coef_.device
+                )
+            else:
+                # Note: on numpy<2, these objects and functions do not have
+                # the 'device' attribute/argument.
+                packed_coefficients = xp.zeros((coef_.shape[0], coef_.shape[1] + 1))
+            packed_coefficients[:, 1:] = coef_
+
+        model.packed_coefficients = to_table(
+            packed_coefficients, queue=QM.get_global_queue()
```
avolkov-intel: I think we should extract the queue from coefs here, not just use the global queue here

sklearnex/linear_model/logistic_regression.py
david-cortes-intel: No, it just wasn't being tested with other formats which result in errors.

onedal/linear_model/logistic_regression.py
david-cortes-intel: Fixed.

onedal/linear_model/logistic_regression.py
```diff
@@ -88,6 +88,31 @@ def fit(self, X, y, queue=None):
 
         return self
 
+    def _create_model(self, coef_, intercept_, xp):
+        model = self.model()
```
avolkov-intel, line 92: Let's add the comment, clarifying the shapes
```suggestion
        model = self.model()
        # Packed coefficients have shape (1, n_features_ + 1), the first element is a placeholder for bias. 
        # If fit_intercept is set to False the first element is set to 0.
```

sklearnex/linear_model/logistic_regression.py
```diff
@@ -387,16 +387,21 @@ def _onedal_cpu_supported(self, method_name, *data):
             )
             return patching_status
 
-        def _onedal_gpu_initialize_estimator(self):
+        def _onedal_gpu_initialize_estimator(self, override_solver: bool = False):
             onedal_params = {
```
avolkov-intel, line 393: Let's add comment here (e.g. like this)
```suggestion
            # We need to override solver in case the model is only used for inference, since newton-cg is the only supported solver for GPU
            # For example, if we trained model with lbfgs solver (with fall back to stock sklearn) and we want to run the inference on GPU using sklearnex code
            onedal_params = {
```

avolkov-intel (approved):

PR #3117, opened 2026-04-21
FIX, MAINT: Implement 'everything follows X' for decision forests

david-cortes-intel:
## Description

This PR updates the decision forest classes to follow the new scikit-learn array API design of 'everything follows X', for both regressors and classifiers, so that now one can pass for example a 'y' that contains strings together with an 'X' that is on GPU.

Along the way, it fixes a couple bugs in internal functions related to sample weights processing that were not catched by tests.

Note that while this is introducing tests with array_api_strict, not everything here would work with torch right now when the devices of 'X' and 'y' are mixed, due to issues with scikit-learn internal functions like 'move_to'.

This PR touches some common functions that are shared with SVMs, so it will need to be merged before others to avoid conflicts while implementing this logic for all array API estimators.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


sklearnex/ensemble/_forest.py
```diff
@@ -104,6 +108,18 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
             y_numeric=not is_classifier(self),  # trigger for Regressors
         )
 
+        if not is_classifier(self) and sklearn_check_version("1.9"):
+            # Note: if 'y' is read-only, trying to get a dlpack object
```
icfaust: be careful, this error comes from numpy and array_api_strict mismstches and are unrelated to our code.

david-cortes-intel: This error happens with the current latest versions of both numpy and array-api-strict.

sklearnex/ensemble/tests/test_forest.py
```diff
@@ -194,3 +195,60 @@ def test_classifiers_work_on_single_class_non_numeric():
...
+# new scikit-learn methodology and sklearnex doesn't fall back.
+@pytest.mark.skipif(
+    not sklearn_check_version("1.9"),
+    reason="Functionality introduced in later scikit-learn versions.",
+)
+@pytest.mark.parametrize("X_xp", [np, pd, array_api_strict])
+@pytest.mark.parametrize("y_xp", [np, pd, array_api_strict])
+@pytest.mark.parametrize("class_weight", [None, "balanced"])
+@pytest.mark.parametrize("n_classes", [0, 2, 3])  # 0 == regression
+def test_mixed_array_namespaces(X_xp, y_xp, class_weight, n_classes):
```
avolkov-intel: Should we also add tests with GPU inputs here?

david-cortes-intel: Problem is: the number of combinations would grow exponentially.

david-cortes-intel:
CI failures are not caused by these changes.

sklearnex/ensemble/_forest.py
icfaust: link me the failing CI

icfaust: edit: please provide a reproducer of this. (including array_api_strict and numpy versions).

david-cortes-intel: Reproducer:
```python
import numpy as np
x = np.arange(10)
x.flags.writeable = False
x.__dlpack__()
```

icfaust: i think the numpy and array api strict versions would also be necessary

david-cortes-intel: This happens with numpy>=2.4.0 or so. All array-api-strict versions use the dlpack capsule, which results in numpy throwing this error. For example: numpy==2.4.4 and array-api-strict==2.5.

david-cortes-intel: On a deeper look, this is a more general problem that is currently affecting all estimator classes that support array API.

I created a more general workaround here, would need to be merged before this PR in order for the tests to pass:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/3132

sklearnex/ensemble/tests/test_forest.py
Vika-F: @david-cortes-intel Maybe it is at least possible to check 2 cases when y follows X: y (GPU) -> X (CPU) and y (CPU) -> X (GPU) ?

david-cortes-intel: Currently not possible with DPNP due to issues in scikit-learn:
https://github.com/scikit-learn/scikit-learn/issues/33870
https://github.com/scikit-learn/scikit-learn/issues/33822

Could be done with torch, but those tests would not currently execute in any CI.

david-cortes-intel:
@Vika-F Added tests for mixed devices too. Turns out some test helpers also required fixes so it's likely that they aren't being used elsewhere, and it required some change in the dispatching logic which perhaps could have effects elsewhere, so @ethanglaser please review.

sklearnex/ensemble/tests/test_forest.py
avolkov-intel: But sklearnex case should work with dpnp, shouldn't it?

avolkov-intel: As an alternative you can use array_api_strict with device1 for tests, but it's not sufficient I guess

david-cortes-intel: The current inter-device tests here (which do not cover as much as the others) pass locally with both dpnp and torch.

sklearnex/ensemble/tests/test_forest.py
```diff
@@ -194,3 +204,109 @@ def test_classifiers_work_on_single_class_non_numeric():
...
+        pred_is_correct = y_xp.astype(y_xp.asarray(pred == y), y_xp.float32)
+        assert y_xp.sum(pred_is_correct) >= (0.25 * int(X.shape[0]))
+
+
+@pytest.mark.skipif(
+    not is_sycl_device_available("gpu"), reason="Test checks GPU-specific functionality."
+)
+@pytest.mark.parametrize(
+    "X_xp, X_device",
+    ([(torch, "xpu"), (torch, "cpu")] if torch_available else [])
```
avolkov-intel: Even though pytorch is available, installed version might not have support for xpu devices. I think we need an additional check here

sklearnex/ensemble/tests/test_forest.py
```diff
@@ -194,3 +204,109 @@ def test_classifiers_work_on_single_class_non_numeric():
...
+    + ([(dpnp, "gpu"), (dpnp, "cpu")] if dpnp_available else []),
+)
```
avolkov-intel, line 288: Why not to add pandas to X_xp type?

sklearnex/ensemble/tests/test_forest.py
```diff
@@ -194,3 +204,109 @@ def test_classifiers_work_on_single_class_non_numeric():
...
+@pytest.mark.parametrize(
+    "y_xp, y_device",
+    ([(torch, "xpu"), (torch, "cpu")] if torch_available else [])
+    + ([(dpnp, "gpu"), (dpnp, "cpu")] if dpnp_available else [])
+    + [(pd, None)],
+)
+@pytest.mark.parametrize(
+    "estimator_class", ["RandomForestRegressor", "RandomForestClassifier"]
```
avolkov-intel: Do we plan to test ExtraTreesClassifier/Regressor as well?

sklearnex/ensemble/tests/test_forest.py
```diff
@@ -194,3 +204,109 @@ def test_classifiers_work_on_single_class_non_numeric():
...
+    pred = model.predict(X)
+    if is_regressor(model):
+        assert pred.__class__ == X.__class__
+    else:
+        if y_xp is pd:
+            assert isinstance(pred, np.ndarray)
+        else:
+            assert pred.__class__ == y.__class__
+    _ = model.score(X, y)
+    _ = model.score(X, y)
```
avolkov-intel, line 338: For classifiers let's also add check for classes_ attribute and predict_proba method

sklearnex/ensemble/tests/test_forest.py
```diff
@@ -194,3 +204,109 @@ def test_classifiers_work_on_single_class_non_numeric():
...
+    pred = model.predict(X)
+    _ = model.score(X, y)
+
+    if n_classes == 0:
+        assert pred.__class__ == (X.__class__ if X_xp is not pd else np.ndarray)
+
+    # Note: this is a quick check to ensure that the result has the same
+    # kind of values as the input. There's no particular justification
+    # behind requiring 25% classification accuracy.
+    if n_classes != 0:
```
avolkov-intel, line 270: Since you only have at maximum 3 classes then random predictions would already give you ~33.3%. Maybe it worth improving the data generation and raise the threshold for this check to verify that model quality is sufficient (e.g. ~70%). Now this check looks a little strange, if you only want to check the resulting types do it in a manual way

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: There is a check for availability of GPU before those lines.

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: It's also to check that the possible values conform to what's expected.

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: Problem is: between this and that other comment:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/3117#discussion_r3159835199

The number of combinations starts growing exponentially.

The case when X is a DataFrame is tested in the case above where they are in the same device.

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: Added.

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: Increased to 60%.

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: Added.

sklearnex/ensemble/tests/test_forest.py
david-cortes-intel: Added a check for XPU specifically within torch.

avolkov-intel:
Is it planned to add functionality related to move_estimator_to here to support the case when during fit we provided cpu array and during infrerence we provide gpu array?

david-cortes-intel:
> Is it planned to add functionality related to move_estimator_to here to support the case when during fit we provided cpu array and during infrerence we provide gpu array?

Yes, but not in this PR.

avolkov-intel:
Could you, please, check that when provided X on CPU and y on GPU, training is actually done on CPU and not on GPU. I have concern that it might not be the case because of this line: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/_device_offload.py#L126

david-cortes-intel:
> Could you, please, check that when provided X on CPU and y on GPU, training is actually done on CPU and not on GPU. I have concern that it might not be the case because of this line: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/_device_offload.py#L126

According to the logs, it should be executing on CPU:
image
image

sklearnex/ensemble/_forest.py
```diff
@@ -72,7 +72,8 @@
 
 if sklearn_check_version("1.2"):
     from sklearn.utils._param_validation import Interval
-
+if sklearn_check_version("1.9"):
```
avolkov-intel, line 75: Maybe it worth downgrading the version in this check to 1.8 ? move_to and get_namespace_and_device are already available there, that would make testing much easier and will provide broader coverage

david-cortes-intel: The behavior of "everything follows X" starts in scikit-learn1.9.

sklearnex/ensemble/tests/test_forest.py
avolkov-intel: Yes, but it's not tested for the case when y is on GPU, but I guess coverage is fine

sklearnex/utils/validation.py
```diff
@@ -276,26 +281,33 @@ def _check_sample_weight_internal(
     allow_all_zero_weights=False,
 ):
     n_samples = _num_samples(X)
-    xp, _ = get_namespace(X)
+    if sklearn_check_version("1.9"):
+        xp, _, device = get_namespace_and_device(X)
+    else:
+        xp, _ = get_namespace(X)
+        if hasattr(X, "device"):
```
avolkov-intel: I think you can use getattr(X, "device", None) for cleaner code, but that's a minor comment

avolkov-intel:
I think you also need additional tests covering sample weights, located on various devices, since you significantly updated the logic of sample_weight handling. You may want to add it as a separate test to reduce the number of potential combinations

david-cortes-intel:
> I think you also need additional tests covering sample weights, located on various devices, since you significantly updated the logic of sample_weight handling. You may want to add it as a separate test to reduce the number of potential combinations

Those will be handled in separate PRs for each class, like this:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/3142

avolkov-intel (approved):

sklearnex/utils/validation.py
david-cortes-intel: Updated.

PR #3127, opened 2026-04-24
FIX, MAINT: Implement 'everything follows X' and namespace checks for KNN

david-cortes-intel:
## Description

This PR:
* Implements the logic of 'everything follows X' for KNN classes.
* Implements the logic where class predictions from classifiers follow the 'y' namespace (e.g. so that it can return strings when fitting on GPU).
* Adds array API namespace and device checks for KNN methods that come after `.fit()` in order to throw informative Python exceptions instead of segfaults, the same way scikit-learn would do.
* Fixes some issues with internal functions that were not working with array_api_strict.

Includes the changes from https://github.com/uxlfoundation/scikit-learn-intelex/pull/3117 since they are also necessary for KNN classes.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/neighbors/_lof.py
+++ b/sklearnex/neighbors/_lof.py
@@ -32,9 +32,6 @@ from sklearnex.neighbors.knn_unsupervised import NearestNeighbors
 
 from ..utils._array_api import enable_array_api, get_namespace
 
-if sklearn_check_version("1.9"):
-    from sklearn.utils._array_api import check_same_namespace
-
 
 @enable_array_api
 @control_n_jobs(decorated_methods=["fit", "kneighbors", "_kneighbors"])
--- a/sklearnex/neighbors/_lof.py
+++ b/sklearnex/neighbors/_lof.py
@@ -31,7 +31,9 @@ from sklearnex.neighbors.common import KNeighborsDispatchingBase
 from sklearnex.neighbors.knn_unsupervised import NearestNeighbors
 
 from ..utils._array_api import enable_array_api, get_namespace
-from ..utils.validation import validate_data
+
+if sklearn_check_version("1.9"):
+    from sklearn.utils._array_api import check_same_namespace
 
 
 @enable_array_api
@@ -141,6 +143,10 @@ class LocalOutlierFactor(KNeighborsDispatchingBase, _sklearn_LocalOutlierFactor)
             self._fit_X = xp.asarray(self._fit_X, device=device)
         return self
 
+    # Note: this is overriding an internal method from scikit-learn with
+    # the same signature. In this case, 'validate_data' is called during
+    # 'decision_function', which calls '.kneighbors()'. Hence, it doesn't
+    # need to validate the namespace of 'X' with '_fit_X' here.
     def _predict(self, X=None):
         check_is_fitted(self)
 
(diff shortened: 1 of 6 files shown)


david-cortes-intel:
CI failures in BasicStatistics are unrelated to the changes here and should be solved with this PR:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/3128

sklearnex/neighbors/common.py
```diff
@@ -164,11 +179,16 @@ def _compute_class_probabilities(
...
+        if sklearn_check_version("1.9"):
+            xp, _, device = get_namespace_and_device(y_train)
+            neigh_dist = move_to(neigh_dist, xp=xp, device=device)
+            neigh_ind = move_to(neigh_ind, xp=xp, device=device)
+        else:
+            xp, _ = get_namespace(y_train)
             device = getattr(y_train, "device", None)
-            neigh_dist = xp.asarray(neigh_dist, device=device)
-            neigh_ind = xp.asarray(neigh_ind, device=device)
+            if not _is_numpy_namespace(xp):
```
avolkov-intel, line 189: What is the logic of this if-statement? As I understand it previously it was that neigh_dist and neigh_ind originally have numpy type and we only need to convert them if y is not a numpy. Is it correct? If yes do we need the same logic after the array-api update?

david-cortes-intel: This is how it was before if you look at the changes. I guess the purpose is to have them work with the other arrays.

sklearnex/neighbors/_lof.py
```diff
@@ -31,7 +31,9 @@
 from sklearnex.neighbors.knn_unsupervised import NearestNeighbors
 
 from ..utils._array_api import enable_array_api, get_namespace
-from ..utils.validation import validate_data
+
+if sklearn_check_version("1.9"):
+    from sklearn.utils._array_api import check_same_namespace
```
avolkov-intel: Did we use this import somewhere in this file?

sklearnex/neighbors/common.py
```diff
@@ -84,11 +93,19 @@ def _compute_weighted_prediction(self, neigh_dist, neigh_ind, weights_param, y_t
...
```
avolkov-intel, line 106: I think we also need some explanation about why do we need this numpy check

sklearnex/neighbors/knn_classification.py
```diff
@@ -86,7 +98,6 @@ def fit(self, X, y):
         # Ensure _fit_X matches the input namespace so that
         # kneighbors(X=None) can use get_namespace(self._fit_X).
         if is_array_api and not _is_numpy_namespace(xp):
```
avolkov-intel, line 100: Can we only keep namespace and device of _fit_X, not to store the reference to an object with data here?

avolkov-intel:
Left a few comments mostly related to clarification about the current logic. Also PR needs to be rebased to fix some CI failures

sklearnex/neighbors/knn_classification.py
david-cortes-intel: No, because the data itself is used after fit to determine nearest neighbors.

sklearnex/neighbors/common.py
david-cortes-intel: It has a different codepath for numpy with operations that are not supported by array API.

sklearnex/neighbors/_lof.py
david-cortes-intel: Removed.

avolkov-intel (approved):

PR #3128, opened 2026-04-24
TEST: Don't test CPU sycl queues when DPC module is not available

david-cortes-intel:
## Description

Some nightly jobs recently started failing with a message like this:
```
E       RuntimeError: oneDAL GPU/DPC++ support is not available in the current installation.
E         Reason: No module named 'onedal._onedal_py_dpc'
E         To enable SYCL/GPU acceleration, install the GPU extras:
E           pip install scikit-learn-intelex-gpu
E         or via conda:
E           conda install scikit-learn-intelex-gpu -c https://software.repos.intel.com/python/conda

onedal/__init__.py:180: RuntimeError
```

This comes from `_ensure_dpc_available`, which gets called during tests in BasicStatistics when they are passed a CPU queue.

The error makes sense, since the logic for handling queues is in the DPC module, but I don't know why it only started popping up recently.

Example of the error in CI: https://github.com/uxlfoundation/scikit-learn-intelex/actions/runs/24894507572/job/72895505497?pr=3127#step:13:102411

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/onedal/datatypes/tests/test_data.py
+++ b/onedal/datatypes/tests/test_data.py
@@ -648,6 +648,7 @@ def test_table_convert_to_host_dlpack(dataframe, queue, order, data_shape, dtype
     assert X_out.flags.writeable
 
 
+@pytest.mark.skipif(_dpc_backend is None, reason="Functionality requires DPC module")
 @pytest.mark.parametrize("queue", get_queues())
 def test_table_writable_dlpack(queue):
     """Test if __dlpack__ attribute can be properly consumed by moving data
--- a/onedal/tests/utils/_device_selection.py
+++ b/onedal/tests/utils/_device_selection.py
@@ -19,6 +19,7 @@ from collections.abc import Iterable
 
 import pytest
 
+from onedal import _dpc_backend
 from onedal.utils._third_party import SyclQueue, dpctl_available
 
 if dpctl_available:
@@ -54,6 +55,8 @@ def get_queues(filter_: str = "cpu,gpu") -> list[SyclQueue]:
         or `pytest.xfail` instead.
     """
     queues = [None] if "cpu" in filter_ else []
+    if _dpc_backend is None:
+        return queues
 
     for i in filter_.split(","):
         try:
--- a/onedal/tests/utils/_device_selection.py
+++ b/onedal/tests/utils/_device_selection.py
@@ -15,6 +15,7 @@
 # ==============================================================================
 
 import functools
+import warnings
 from collections.abc import Iterable
 
 import pytest
@@ -56,6 +57,10 @@ def get_queues(filter_: str = "cpu,gpu") -> list[SyclQueue]:
     """
     queues = [None] if "cpu" in filter_ else []
     if _dpc_backend is None:
+        if "gpu" in filter_:
+            warnings.warn(
+                "Attempting to get a GPU queue, but DPC backend is not available."
+            )
         return queues
 
     for i in filter_.split(","):
--- a/sklearnex/tests/test_config.py
+++ b/sklearnex/tests/test_config.py
@@ -145,7 +145,7 @@ def test_host_backend_target_offload(target):
             with sklearnex.config_context(target_offload=target):
                 est.fit(np.eye(5, 8))
     elif target != "auto":
-        with pytest.raises((ValueError, RuntimeError), match=err_msg):
+        with pytest.raises(Exception, match=err_msg):
             with sklearnex.config_context(target_offload=target):
                 est.fit(np.eye(5, 8))
     else:
--- a/sklearnex/tests/test_config.py
+++ b/sklearnex/tests/test_config.py
@@ -137,11 +137,15 @@ def test_config_context_works():
 def test_host_backend_target_offload(target):
     from sklearnex.neighbors import NearestNeighbors
 
-    err_msg = r"DPC"
+    err_msg = r"DPC|SYCL"
 
     est = NearestNeighbors()
-    if target != "auto":
-        with pytest.raises(ValueError, match=err_msg):
+    if not isinstance(target, str):
+        with pytest.raises((ValueError, RuntimeError, TypeError)):
+            with sklearnex.config_context(target_offload=target):
+                est.fit(np.eye(5, 8))
+    elif target != "auto":
+        with pytest.raises((ValueError, RuntimeError), match=err_msg):
             with sklearnex.config_context(target_offload=target):
                 est.fit(np.eye(5, 8))
     else:
--- a/sklearnex/tests/test_config.py
+++ b/sklearnex/tests/test_config.py
@@ -137,9 +137,7 @@ def test_config_context_works():
 def test_host_backend_target_offload(target):
     from sklearnex.neighbors import NearestNeighbors
 
-    err_msg = (
-        r"device use via \`target_offload\` is only supported with the DPC\+\+ backend"
-    )
+    err_msg = r"only supported with the DPC\+\+ backend"
 
     est = NearestNeighbors()
     if target != "auto":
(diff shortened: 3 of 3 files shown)


sklearnex/tests/test_config.py
```diff
@@ -131,7 +131,7 @@ def test_config_context_works():
 
 
 @pytest.mark.skipif(
-    onedal._default_backend.is_dpc, reason="requires host default backend"
+    not onedal._default_backend.is_dpc, reason="requires host default backend"
```
david-cortes-intel:
```suggestion
    onedal._default_backend.is_dpc, reason="requires host default backend"
```

sklearnex/tests/test_config.py
```diff
@@ -137,13 +137,15 @@ def test_config_context_works():
 def test_host_backend_target_offload(target):
     from sklearnex.neighbors import NearestNeighbors
 
-    err_msg = (
-        r"device use via \`target_offload\` is only supported with the DPC\+\+ backend"
-    )
+    err_msg = r"DPC|SYCL"
```
Vika-F: Isn't the regex too ambiguous (comparing to the previous version)? Is it possible to match something irrelevant with this pattern?

david-cortes-intel: We have 3 possible error types and messages. I don't think there's many possible error messages with "DPC" or "SYCL" in the name that could be thrown from this library.

Vika-F: I am just worried that some other messages could potentially contain those sub-strings.
Isn't this the only message that have to be matched:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/__init__.py#L181 ?

david-cortes-intel: > I am just worried that some other messages could potentially contain those sub-strings. Isn't this the only message that have to be matched: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/__init__.py#L181 ?

No, it can also throw errors from DPCTL when trying to create the queue, plus there's still the older error before that one.

onedal/tests/utils/_device_selection.py
```diff
@@ -19,6 +19,7 @@
 
 import pytest
 
+from onedal import _dpc_backend
```
avolkov-intel, line 23: Is it None in case _dpc_backend is not available? Can it throw an error in some case?

onedal/tests/utils/_device_selection.py
```diff
@@ -54,6 +55,8 @@ def get_queues(filter_: str = "cpu,gpu") -> list[SyclQueue]:
         or `pytest.xfail` instead.
     """
     queues = [None] if "cpu" in filter_ else []
+    if _dpc_backend is None:
+        return queues
```
avolkov-intel, line 64: I think it worth showing a warning here if "gpu" filter was set but we ignored it.

sklearnex/tests/test_config.py
avolkov-intel: I agree with Vika err_msg is too ambigious it may lead to confusions if new errors are added in the future. If there are only 3 types of error just specify them here with | .

onedal/tests/utils/_device_selection.py
david-cortes-intel: > Is it None in case _dpc_backend is not available?

Yes.

> Can it throw an error in some case?

No:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/edaa06916b26960ea50ec9a4bbb2ef59f420a850/onedal/__init__.py#L121

onedal/datatypes/tests/test_data.py
```diff
@@ -648,6 +648,7 @@ def test_table_convert_to_host_dlpack(dataframe, queue, order, data_shape, dtype
     assert X_out.flags.writeable
 
 
+@pytest.mark.skipif(_dpc_backend is None, reason="Functionality requires DPC module")
```
avolkov-intel, line 651: Is it really the only test that requires such skip if?

onedal/tests/utils/_device_selection.py
david-cortes-intel: Added a warning.

onedal/datatypes/tests/test_data.py
david-cortes-intel: It should be,  given the CI results. The rest typically handle this through `get_dataframes_and_queues`.

sklearnex/tests/test_config.py
david-cortes-intel: That's a regular expression. It wouldn't work like that. It'd need to have a separate if-else for each parametrization if done like that. Plus in many cases it doesn't really matter if the error is thrown from DPCTL or from onedal_py_dpc first.

sklearnex/tests/test_config.py
```diff
@@ -137,13 +137,15 @@ def test_config_context_works():
...
 
     est = NearestNeighbors()
-    if target != "auto":
-        with pytest.raises(ValueError, match=err_msg):
+    if not isinstance(target, str):
+        with pytest.raises((ValueError, RuntimeError, TypeError)):
```
avolkov-intel, line 151: Why did we need to extend previous code? What happens if we provide 3? Won't it be just a ValueError?
Do we need Exception instead of ValueError for the rest of cases because of errors from DPCTL?

david-cortes-intel: > Do we need Exception instead of ValueError for the rest of cases because of errors from DPCTL?

Yes.

avolkov-intel:
I see that we don't have this error in main for the same CI job: https://github.com/uxlfoundation/scikit-learn-intelex/actions/runs/24883083680/job/72856205666
Do you know why it happens? Is it because it was 3 days ago or because there's some difference in these jobs?

sklearnex/tests/test_config.py
avolkov-intel: Why it wouldn't work? I thought with pipes if any of the pattern matches the string regexp will catch the string. Why it's not possible to write down more specific regular expressions for each of the 3 error types?

david-cortes-intel: @avolkov-intel Are you able to come up with a regular expression that would catch all possible error messages? We have the following:
* dpctl._sycl_queue.SyclQueueCreationError: SYCL Device 'gpu' could not be created.
    (where 'gpu' varies according to the argument)
* TypeError: Positional argument 3 is not a filter string or a SyclDevice
    (where '3' varies according to the argument)
* RuntimeError: oneDAL GPU/DPC++ support is not available in the current installation.
    (plus the rest of the message)

Plus the older message in case it somehow gets thrown (e.g. lack of dpctl):
device use via `target_offload` is only supported with the DPC++ backend

ethanglaser:
> I see that we don't have this error in main for the same CI job: https://github.com/uxlfoundation/scikit-learn-intelex/actions/runs/24883083680/job/72856205666 Do you know why it happens? Is it because it was 3 days ago or because there's some difference in these jobs?

The first instance of these issues was CI run 8122, this linked one is 8120

sklearnex/tests/test_config.py
ethanglaser: I think what Anatoly is suggesting is to write all out specifically and add pipe between each, like

```
"|".join([
    r"SYCL Device '.+' could not be created",
    r"Positional argument .+ is not a filter string or a SyclDevice",
    r"oneDAL GPU/DPC\+\+ support is not available in the current installation",
    r"device use via `target_offload` is only supported with the DPC\+\+ backend",
])
```

david-cortes-intel: Updated.

ethanglaser (approved):
As long as its expected to have hundreds of `RuntimeError: Requested device type is not supported`

david-cortes-intel:
> As long as its expected to have hundreds of `RuntimeError: Requested device type is not supported`

That's only for CI jobs that are executed under target_offload for CPU.

PR #3132, opened 2026-04-27
FIX: Fix cases of BufferError when calling DLPack capsule

david-cortes-intel:
## Description

ref https://github.com/uxlfoundation/scikit-learn-intelex/pull/3117#discussion_r3128645700

Fixes an issue in which calls to DLPack capsules from non-writeable arrays throw errors while trying to convert them to oneDAL tables.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


icfaust (changes requested):
this is it adding a branch to a very often called function for a very rare circumstance (numpy backed dlpack frameworks which have nonwritable arrays, if you can give an example outside of the array api strict test datatype let me know). it is best to instead place this in pybind11 in dlpack's data_conversion.cpp.

secondly the tests for data conversion should be placex in test_data.py for similarity. 

please refactor.

david-cortes-intel:
> this is it adding a branch to a very often called function for a very rare circumstance (numpy backed dlpack frameworks which have nonwritable arrays, if you can give an example outside of the array api strict test datatype let me know). it is best to instead place this in pybind11 in dlpack's data_conversion.cpp.
> 
> secondly the tests for data conversion should be placex in test_data.py for similarity.
> 
> please refactor.

* There would be no speed advantage in doing this in pybind11, because it would just call functions from modules.
* This could in theory happen for other classes - see array [API design](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html):
    >  Implementations should raise BufferError when the data cannot be exported as DLPack (e.g., incompatible dtype or strides).

icfaust:
the reason for the movement to pybind11 is not speed but to colocate with similar copy functionality that exists (see the recursed boolean) and reduce branching in common pathways which directly impact maintainability. i still want it moved.

david-cortes-intel:
> the reason for the movement to pybind11 is not speed but to colocate with similar copy functionality that exists (see the recursed boolean) and reduce branching in common pathways which directly impact maintainability. i still want it moved.

Moved the logic to pybind11. But that will become really hard to debug if it starts erroring out later on for other reasons.

avolkov-intel:
Is there are some real scenario where this error came from (where the user didn't explicitly specified x.flags.writeable = False)? What was the trigger to fix this?

onedal/datatypes/tests/test_data.py
```diff
@@ -673,3 +674,15 @@ def test_table_writable_dlpack(queue):
         X_out = xp.from_dlpack(X_table, copy=copy_bool)
         # verify that table immutability is gone and copy behavior has been followed
         assert X_out.flags["W"] is copy_bool
+
+
+def test_nonwriteable_arrays():
```
avolkov-intel, line 684: Does it only happen with numpy? Is there a case where it happens with array located on GPU? If yes, I think we need to extend test functionality

david-cortes-intel: > Is there a case where it happens with array located on GPU? If yes, I think we need to extend test functionality

I don't know yet, but the array API specification would allow for such errors - see https://github.com/uxlfoundation/scikit-learn-intelex/pull/3132#issuecomment-4327491506

> Does it only happen with numpy?

No, it can happen with other libraries that keep a numpy capsule, such as array-api-strict.

david-cortes-intel:
> Is there are some real scenario where this error came from (where the user didn't explicitly specified x.flags.writeable = False)? What was the trigger to fix this?

Yes: one can end up with non-writeable arrays like that when pandas is involved.

onedal/datatypes/tests/test_data.py
icfaust: its also worth checking dpnp which has a writable flag. 

can be tested by extending test with https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/datatypes/tests/test_data.py#L88

david-cortes-intel: It's the same with DPNP, since it also forwards the call to numpy, just like array-api-strict:
```python
import dpnp
import numpy as np
x = np.arange(2)
x.flags.writeable = False
x = dpnp.from_dlpack(x)
x.__dlpack__()
```
```
BufferError: Cannot export readonly array since signalling readonly is unsupported by DLPack (supported by newer DLPack version).
```

icfaust:
so far i like the new implementation but the new test is failing in some CI jobs

onedal/datatypes/tests/test_data.py
icfaust: no i mean a dpnp has a writeable flag mimicing numpy. worth playing with it

david-cortes-intel:
> so far i like the new implementation but the new test is failing in some CI jobs

I didn't find any way to solve this for older versions of numpy, because the functions there do not support the same arguments. I don't think it's a priority to fix either, since scikit-learn has this notice in [their documentation](https://scikit-learn.org/dev/modules/array_api.html#enabling-array-api-support):
image

So I'm marking this test to be skipped on older numpy.

onedal/datatypes/tests/test_data.py
david-cortes-intel: It doesn't allow setting it in arbitrary arrays the way numpy does.

icfaust: hrm good to know. I'm a bit surprised, may be worth pinging the dpnp people about this.

icfaust (approved):
Pending the green CI's, and verification by another user on private CI status.

PR #3162, opened 2026-05-14
FIX: Fix creation of oneDAL objects in linear models after fallbacks

david-cortes-intel:
## Description

Fixes cases in which creation of oneDAL objects from fitted coefficients errors out due to incompatibilities with array API objects.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/onedal/linear_model/linear_model.py
+++ b/onedal/linear_model/linear_model.py
@@ -62,7 +62,7 @@ class BaseLinearRegression(metaclass=ABCMeta):
 
         return params
 
-    def _create_model(self, coef_, intercept_, xp):
+    def _create_model(self, coef_, intercept_, xp) -> None:
         model = self.model()
 
         dim0 = coef_.shape[0] if len(coef_.shape) > 1 else 1
@@ -93,8 +93,6 @@ class BaseLinearRegression(metaclass=ABCMeta):
         self.coef_ = coef_
         self.intercept_ = intercept_
         self._onedal_model = model
-
-        return model
 
     @supports_queue
     def fit(self, X, y, queue=None):
--- a/onedal/linear_model/linear_model.py
+++ b/onedal/linear_model/linear_model.py
@@ -62,31 +62,32 @@ class BaseLinearRegression(metaclass=ABCMeta):
 
         return params
 
-    def _create_model(self):
+    def _create_model(self, coef_, intercept_, xp):
         model = self.model()
 
-        # force dtype and shape for all supported estimators to numpy
-
-        if np.isscalar(self.coef_):
-            coef = np.asarray(self.coef_).reshape(1, 1)
+        dim0 = coef_.shape[0] if len(coef_.shape) > 1 else 1
+        if dim0 == 1:
+            coef_ = xp.reshape(coef_, (1, -1))
+        if self.fit_intercept:
+            intercept_ = xp.reshape(intercept_, (dim0, -1))
+            packed_coefficients = xp.concat([intercept_, coef_], axis=1)
         else:
-            # generalized atleast_2d for numpy and array_api inputs
-            # if an empty array, will fail for a multitude of reasons
-            coef = from_table(
-                to_table(self.coef_[None] if self.coef_.ndim == 1 else self.coef_)
-            )
-        if np.isscalar(self.intercept_):
-            intercept = np.asarray(self.intercept_).reshape(1, 1)
-        else:
-            intercept = from_table(to_table(self.intercept_))
+            if hasattr(coef_, "device"):
+                packed_coefficients = xp.zeros(
+                    (dim0, coef_.shape[1] + 1), device=coef_.device
+                )
+            else:
+                # Note: on numpy<2, these objects and functions do not have
+                # the 'device' attribute/argument.
+                packed_coefficients = xp.zeros((dim0, coef_.shape[1] + 1))
+            packed_coefficients[:, 1:] = coef_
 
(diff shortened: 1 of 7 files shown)


onedal/linear_model/tests/test_incremental_linear_regression.py
```diff
@@ -140,6 +140,7 @@ def test_reconstruct_model(queue, dtype):
     model = IncrementalLinearRegression(fit_intercept=True)
     model.coef_ = coef.T
     model.intercept_ = intercept
+    model._onedal_model = model._create_model(model.coef_, model.intercept_, np)
```
Vika-F: The proposed method's API look strange. Why do you have to pass `model.coef_`, `model.intercept_` into the method instead of using `self.coef_` and `self.intercept_` inside the method?
Same to `model._onedal_model`.

```suggestion
    model._create_model(np)
```

david-cortes-intel: This is a test for the internal onedal module. Users are not supposed to do this.

> Why do you have to pass model.coef_, model.intercept_ into the method instead of using self.coef_ and self.intercept_ inside the method?

The method can also be called in cases where those will not exist. For example, if the coefficients are obtained through scikit-learn without involvement of a onedal_estimator object, and then such object is created when calling `.predict()`.

sklearnex/linear_model/tests/test_linear_common.py
```diff
@@ -0,0 +1,103 @@
...
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
+        model.fit(X, y, w)
+    assert not hasattr(model, "_onedal_estimator")
+
+    pred = model.predict(X)
+    assert pred.__class__ == X.__class__
+    assert pred.dtype == array_api_strict.float32
+
+    assert hasattr(model, "_onedal_estimator")
```
Vika-F, line 93: It is not clear why _onedal_estimator is present after the prediction stage in this case comparing to the previous test case?

david-cortes-intel: The call to `.predict()` should generate the object if it wasn't already there.

onedal/linear_model/tests/test_incremental_linear_regression.py
Vika-F: I see.
But anyway, at this line it does the second assignment to `model._onedal_model` for some reason.

In my understanding it should be
- either `model._onedal_model = model._create_model(model.coef_, model.intercept_, np)` everywhere. Then remove the assignment `self._onedal_model = model` within the `_create_model` method.
- or `model._create_model(model.coef_, model.intercept_, np)`. Then remove `return model` within the method.

sklearnex/linear_model/tests/test_linear_common.py
Vika-F: But why this does not happen without the Array API?

david-cortes-intel: It can happen without array API too. For example, if passing weights to `.fit()`.

onedal/linear_model/tests/test_incremental_linear_regression.py
david-cortes-intel:
```suggestion
    model._create_model(model.coef_, model.intercept_, np)
```

onedal/linear_model/tests/test_linear_regression.py
```diff
@@ -143,6 +143,7 @@ def test_reconstruct_model(queue, dtype):
     model = LinearRegression(fit_intercept=True)
     model.coef_ = coef.T
     model.intercept_ = intp
+    model._onedal_model = model._create_model(model.coef_, model.intercept_, np)
```
david-cortes-intel:
```suggestion
    model._create_model(model.coef_, model.intercept_, np)
```

onedal/linear_model/tests/test_incremental_linear_regression.py
david-cortes-intel: Changed to not use the return value to assign. But it looks like every other estimator is also doing the same thing, with both in-place assignment during the call and then assignment of the result.

sklearnex/linear_model/tests/test_linear_common.py
Vika-F: Ah, sorry, never mind.
From the first glance the asserts in array_api and no-array_api tests looked differently, but now I see they are the same.

Thanks for the clarification!

onedal/linear_model/linear_model.py
```diff
@@ -62,35 +62,38 @@ def _get_onedal_params(self, dtype):
...
+                )
+            else:
+                # Note: on numpy<2, these objects and functions do not have
+                # the 'device' attribute/argument.
+                packed_coefficients = xp.zeros((dim0, coef_.shape[1] + 1))
+            packed_coefficients[:, 1:] = coef_
+
+        # packed_coefficients = xp.reshape(packed_coefficients, (1, -1))
         model.packed_coefficients = to_table(
             packed_coefficients, queue=QM.get_global_queue()
```
avolkov-intel: I don't like the fact that global queue is used here, I think we should define queue based on data or provide it to create_model as parameter

david-cortes-intel: Fixed.

Vika-F (approved):

PR #3164, opened 2026-05-14
FIX, MAINT: Implement namespace and device checks for logistic regression

david-cortes-intel:
## Description

Implements device and namespace compatibility checks for methods in logistic regression. Fixes some cases where data was incorrectly being moved in sklearn1.9 along the way.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


sklearnex/linear_model/logistic_regression.py
```diff
@@ -478,21 +496,20 @@ def _onedal_predict(self, X, queue=None):
             res = self._onedal_estimator.predict(X, queue=queue, classes=self.classes_)
```
avolkov-intel, line 516: Since you are adding move_to to the result of predict, you may also remove like=classes from here: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/linear_model/logistic_regression.py#L113 and remove classes attribute from _onedal_estimator.predict fucntion

sklearnex/linear_model/logistic_regression.py
```diff
@@ -508,17 +525,27 @@ def _onedal_predict_proba(self, X, queue=None):
             y = xp.reshape(res, (-1,))
             return xp.stack([1 - y, y], axis=1)
 
+        # Note: it might at first glance appear that defining these methods is not
```
avolkov-intel, line 549: I think the issue is mostly with the fact that we would like to have dispatch between daal4py and onedal on our side, also for daal4py we have an optimized branch for predictions, maybe simplify this comment a little bit

sklearnex/linear_model/logistic_regression.py
david-cortes-intel: The problem is that oneDAL will not always succeed in converting the table to whatever you pass under `like`.

sklearnex/linear_model/logistic_regression.py
david-cortes-intel: Do you have some suggestion for the full comment?

sklearnex/linear_model/logistic_regression.py
```diff
@@ -449,19 +451,35 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
...
+        # This should only be called when 'X' is on CPU
+        def _error_out_on_incompatible_devices(self, X, method_name: str) -> None:
+            # We don't use onedal backend for CPU, so we need an additional check here
+            if "spmd" in self._onedal_LogisticRegression.__module__:
+                raise RuntimeError(
+                    "Executing functions from SPMD backend requires a queue"
+                )
+
+            # This can happen when fitting on GPU and then passing a CPU array to predict
+            if not isinstance(self.coef_, np.ndarray) and not issparse(self.coef_):
```
avolkov-intel, line 481: I think that self.coef_ can not be sparse. Even though we can take sparse matrix as input, the coef_ should be dense. Or did something change with array api?

david-cortes-intel: It can if you call `.sparsify()`:
https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#sklearn.linear_model.LogisticRegression.sparsify

And can come like that after a fallback.

sklearnex/linear_model/logistic_regression.py
```diff
@@ -449,19 +451,35 @@ def _onedal_fit(self, X, y, sample_weight=None, queue=None):
...
-                    raise RuntimeError(
-                        "Executing functions from SPMD backend requires a queue"
-                    )
+                self._error_out_on_incompatible_devices(X, "predict")
 
                 # TODO add array-api support for CPU
                 return daal4py_predict(self, X, "computeClassLabels")
 
+            if sklearn_check_version("1.9"):
+                check_same_namespace(X, self, attribute="coef_", method="predict")
```
avolkov-intel: Can you, please check that it works fine if you pass numpy during fit and pandas during predict? I think there might be an issue if you put this check before data validation

david-cortes-intel: It works correctly as far as I can tell, but moved the checks to happen after the validations just in case.

sklearnex/linear_model/logistic_regression.py
```diff
@@ -327,19 +336,8 @@ def _onedal_gpu_predict_supported(self, method_name, *data):
             patching_status = PatchingConditionsChain(
                 f"sklearn.linear_model.{class_name}.{method_name}"
             )
-            dal_ready = patching_status.and_conditions(
-                [
-                    (
-                        method_name in ["predict_proba", "predict"]
-                        or not sklearn_check_version("1.9"),
```
avolkov-intel, line 334: We do need this check, since we use accuracy_score from sklearn to compute the score for gpu. If this function doesn't support array api, score function for gpu branch will not work.

sklearnex/linear_model/logistic_regression.py
avolkov-intel: We might write something like this, but honestly I think we can just remove this comment
"Note, that even though the logic of this functions is similar to what is done is sklearn, we still need them implemented on sklearnex side to dispatch between cpu and gpu implementation"

sklearnex/linear_model/logistic_regression.py
david-cortes-intel: accuracy_score from scikit-learn supports array API.

sklearnex/linear_model/logistic_regression.py
david-cortes-intel: I would say that doesn't really explain the justification. It was not obvious for me to figure out why exactly they are needed. One could as well dispatch the scikit-learn version with array API in both branches, for example.

sklearnex/linear_model/logistic_regression.py
avolkov-intel: Not for sklearn < 1.4 but seems that in this case it indeed works without errors

avolkov-intel (approved):

PR #3182, opened 2026-05-20
FIX, MAINT: Implement namespace checks, 'everything follows X', and fix oneDAL model re-creation in SVMs

david-cortes-intel:
## Description

Implements changes for array API support in sklearn1.9 for SVMs, and along the way fixes several issues in codepaths that were untested.

It includes some detailed comments about the reasons for why things are being done the way they are and why there are multiple attributes with similar names.

It also adds a few more tests, such as checking that dense models work with sparse data, or checking that oneDAL estimator objects created from fitted attributes produce correct results, among others.

Additionally, updates error messages thrown under different scenarios to match with scikit-learn's.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


david-cortes-intel:
Comment for @Vika-F and @Alexandr-Solovev : for some reason, it appears that creating a oneDAL SVM estimator from scratch without calling `train()` with parameter `class_count` > 2 and setting attributes like `biases` to arrays with length larger 2 will yield an object that's still set as binary classifier and produces a vector from `.decision_function()`.

Large parts of this PR are about working around this issue.

onedal/svm/svm.py
```diff
@@ -131,13 +131,23 @@ def fit(self, X, y, sample_weight=None, class_count=0, queue=None):
...
-        m.coeffs = to_table(self.dual_coef_.T)
-        m.biases = to_table(self.intercept_)
-        return m
+        self.support_vectors_ = support_vectors_
+        self.dual_coef_ = dual_coef_
+        self.intercept_ = intercept_
+        self._sparse = is_sparse(support_vectors_)
+        self.class_count_ = dual_coef_.shape[0] + 1
+
+        m.support_vectors = to_table(support_vectors_)
```
avolkov-intel, line 146: Do we need to specify queue here? Does it work fine with both cpu and gpu and different array api formats?

david-cortes-intel: As far as I know, it should extract the device from the DLpack capsule of the object that's passed to `to_table`. We currently do not have tests for it, but as far as I can tell, it works correctly when reassigning GPU attributes:
image

sklearnex/svm/_base.py
```diff
@@ -125,35 +130,200 @@ class BaseSVM(oneDALEstimator):
 
     _onedal_factory = None
 
+    # A few comments about the attributes:
```
ethanglaser, line 133: I'd be in favor of migrating this massive comment block to docs with a reference here instead. Either that or compressing - for instance lines 133-162 could be written as something along the lines of "Not possible to run scikit-learn inference on a oneDAL-trained model due to some nuances with public/private attributes"

david-cortes-intel: This is about internal implementation details of both sklearn and sklearnex. The docs already cover the user-facing limitations for the public API.

> "Not possible to run scikit-learn inference on a oneDAL-trained model due to some nuances with public/private attributes"

That, for example, is an internal detail, since a user cannot explicitly trigger a branch with `.predict()` taken from sklearn using the public methods of sklearnex objects.

avolkov-intel: Can we create a subsection under Development Guides for that? I agree with Ethan that we shouldn't keep such long comments here

david-cortes-intel: I disagree about that part. This PR was very hard for me to make up precisely because those sorts of details are not obvious, not explained in the file, and not centralized to connect how different things interact together.

I think it'd make it easier to edit this file if the explanations for why things are done in a certain way are right there. Otherwise they aren't easy to find or to notice during reviews for changes in the same file.

sklearnex/svm/_classes.py
```diff
@@ -122,52 +125,63 @@ def _onedal_gpu_supported(self, method_name, *data):
...
         elif method_name in self._n_jobs_supported_onedal_methods:
-            conditions.append(
-                (hasattr(self, "_onedal_estimator"), "oneDAL model was not trained")
+            patching_status.and_conditions(
+                [
+                    (
+                        not issparse(self.support_vectors_),
+                        "Predictions on sparse support vectors are not supported",
+                    ),
+                    (hasattr(self, "_onedal_estimator"), "oneDAL model was not trained"),
```
avolkov-intel: Do we need this check if we can manually create onedal_estimator?

david-cortes-intel: Not strictly (depending on other conditions), but (a) creating the onedal_estimator is not always a zero-copy operation, (b) it cannot always be created, (c) it's not possible to fit an SVM model on GPU through scikit-learn.

This is precisely why I wanted to put a very big block of comment explaining the logic BTW.

david-cortes-intel: But since we'll need to change this to support `move_estimator_to`, I replaced this check. Now it should create the object when possible.

david-cortes-intel: And the proof that it works in case you wanted to ask for it:
image

sklearnex/svm/_classes.py
```diff
@@ -122,52 +125,67 @@ def _onedal_gpu_supported(self, method_name, *data):
...
         )
         X = data[0]
-        conditions = [
-            (
-                self.kernel in ["linear", "rbf"],
-                f'Kernel is "{self.kernel}" while '
-                'only "linear" and "rbf" are supported on GPU.',
-            ),
-            (not is_sparse(X), "Sparse input is not supported on GPU."),
-            (self.class_weight is None, "Class weight is not supported on GPU."),
```
avolkov-intel, line 132: Why did we remove class_weight check and multiclassification check?

david-cortes-intel: They are not removed, just moved further below and executed only if the first check passes.

sklearnex/svm/_base.py
avolkov-intel: Ok, I agree with you

avolkov-intel:
Could you, please extend the description of this PR, so that it's clear what issues were fixed

sklearnex/svm/_base.py
```diff
@@ -125,35 +130,200 @@ class BaseSVM(oneDALEstimator):
...
             self._onedal_estimator.dual_coef_ = value
             self._onedal_estimator._onedal_model = None
+        self._dualcoef_ = value
 
+    @dual_coef_.deleter
+    def dual_coef_(self):
+        if hasattr(self, "_onedal_estimator") and self._is_multi_class_classifier():
+            self._raise_immutable_error()
+        if hasattr(self, "_dualcoef_"):
+            del self._dualcoef_
```
avolkov-intel: Should we delete _onedal_model in this case?

sklearnex/svm/_base.py
```diff
@@ -125,35 +130,200 @@ class BaseSVM(oneDALEstimator):
...
+
+    # Do NOT flip signs here, see earlier comments
     @_dual_coef_.deleter
     def _dual_coef_(self):
-        del self._dualcoef_
+        if hasattr(self, "_onedal_estimator") and self._is_multi_class_classifier():
+            self._raise_immutable_error()
+        if hasattr(self, "_dualcoef_"):
```
avolkov-intel: Same comment here and in few deleters below

sklearnex/svm/_base.py
david-cortes-intel: Fixed. But it's not necessary to delete the oneDAL object, because it can re-create its underlying model object from attributes assigned to it.

sklearnex/svm/_base.py
```diff
@@ -125,35 +130,200 @@ class BaseSVM(oneDALEstimator):
...
+        return self._dualcoef_
+
     @property
     def _dual_coef_(self):
+        if self._is_binary_classifier():
+            return -self._dualcoef_
         return self._dualcoef_
 
+    @dual_coef_.setter
+    def dual_coef_(self, value):
```
avolkov-intel, line 225: General comment: I don't like that names of the properties and actual attributes are mixed and there's no logic in how they are named. It's not clear when some variable is referenced in the code is it a property for which we defined setter/getter/deleter or is it an internal variable that we can reference. Could we please rename all variables that are not intended for external access with specific name, like _dual_coef_internal_. That would improve readability significantly

sklearnex/svm/_base.py
david-cortes-intel: This is a private attribute. It's only used by sklearn, not by oneDAL.

sklearnex/svm/_base.py
avolkov-intel: Especially for _dual_coef_ and _dualcoef_, it's impossible to understand what is a variable and what is a property, we should rename _dualcoef_

sklearnex/svm/_base.py
```diff
@@ -125,35 +130,209 @@ class BaseSVM(oneDALEstimator):
...
+    #   is why these setters and deleters error out.
+    # - And since they error out when trying to assign to public attributes with
+    #   an already existing oneDAL object, which scikit-learn would do during
+    #   fallbacks when there are repeated calls to '.fit()', this means that the
+    #   oneDAL object needs to be deleted before a fallback in order not to
+    #   trigger these errors from scikit-learn.
+    # - Note also that oneDAL uses the transpose of 'dual_coef_', so if this
+    #   array is modified in-place instead of re-assigned, changes would not
+    #   propagate to the oneDAL estimator, and upon deserialization, there
+    #   will be more than one copy in memory.
```
avolkov-intel, line 195: Let's also add the description of meaning of each attribute/property as a list here

sklearnex/svm/_base.py
david-cortes-intel: Renamed and added it to the comment block.

sklearnex/svm/_base.py
david-cortes-intel: That's described in scikit-learn's docs, since they are public attributes:
https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html

david-cortes-intel:
> Could you, please extend the description of this PR, so that it's clear what issues were fixed

Updated.

sklearnex/svm/_base.py
david-cortes-intel: Actually on a second look it should, updated.

sklearnex/svm/_base.py
```diff
@@ -125,35 +130,216 @@ class BaseSVM(oneDALEstimator):
...
+    def dual_coef_(self):
+        if hasattr(self, "_onedal_estimator") and self._is_multi_class_classifier():
+            self._raise_immutable_error()
+        if hasattr(self, "_dual_coef_internal"):
+            del self._dual_coef_internal
+        if hasattr(self, "_onedal_estimator"):
+            self._onedal_estimator.dual_coef_ = None
+            self._onedal_estimator._onedal_model = None
 
+    # Do NOT flip signs here, see earlier comments
```
avolkov-intel: Is this comment located correctly? Why do we need it before deleter?

sklearnex/svm/_base.py
```diff
@@ -213,11 +404,31 @@ def _onedal_cpu_supported(self, method_name, *data):
...
             patching_status.and_conditions(
-                [(hasattr(self, "_onedal_estimator"), "oneDAL model was not trained.")]
+                [
+                    (
+                        isinstance(self.kernel, str)
+                        and self.kernel in ["linear", "rbf", "poly", "sigmoid"],
+                        "Predictions on pre-computed and callable kernels are not supported.",
+                    ),
+                    (
+                        hasattr(self, "_onedal_estimator")
```
avolkov-intel, line 420: I think this logic should be described in the docs

sklearnex/svm/_base.py
```diff
@@ -213,11 +404,31 @@ def _onedal_cpu_supported(self, method_name, *data):
...
+                        or (
+                            not self._is_multi_class_classifier()
+                            and (
+                                not sp.issparse(self.support_vectors_)
+                                or (hasattr(X, "shape") and X.shape[0] > 1)
+                            )
+                        ),
+                        "Single-row prediction from a scikit-learn model has"
```
avolkov-intel, line 428: I think the comment isn't aligned with the whole condition

sklearnex/svm/_base.py
```diff
@@ -307,44 +539,63 @@ def _onedal_cpu_supported(self, method_name, *data):
...
-            patching_status.and_conditions(
-                [
-                    (
-                        (xp.any(y_nonzero != y_nonzero[0])),
-                        "Invalid input - all samples with positive weights belong to the same class.",
-                    ),
-                ]
-            )
+
+            if data[2] is not None:
```
avolkov-intel, line 545: Minor comment: can we add X, y, sample_weight = data in the beginning of the function and don't use data argument after that anywhere in this function

sklearnex/svm/_base.py
david-cortes-intel: Fixed.

sklearnex/svm/_base.py
david-cortes-intel: Why would you think that?

sklearnex/svm/_base.py
david-cortes-intel: Added documentation.

sklearnex/svm/_base.py
david-cortes-intel: No, sample_weight will not always be there.

sklearnex/svm/_classes.py
```diff
@@ -122,52 +125,76 @@ def _onedal_gpu_supported(self, method_name, *data):
...
+                    (
+                        not issparse(self.support_vectors_)
+                        or hasattr(self, "_onedal_estimator"),
+                        "Predictions on sparse support vectors are not supported",
+                    ),
+                    # This can be remove once oneDAL model creation from multi-class
+                    # arrays is fixed.
+                    (
+                        hasattr(self, "_onedal_estimator")
+                        or self._is_binary_classifier(),
```
avolkov-intel, line 190: Why is it different from this condition? https://github.com/uxlfoundation/scikit-learn-intelex/pull/3182/changes#diff-3b73c349245083f66689ea4f9585f3c92d2451fe6c24604f687f3cb1fdd9366bR420

david-cortes-intel: That link doesn't show any particular line number.

sklearnex/svm/_base.py
avolkov-intel: We have conditions related to sparsity in the same block, not only to single row inference

sklearnex/svm/_base.py
avolkov-intel: Yes, but it will be None in this case, no?

doc/sources/algorithms.rst
```diff
@@ -48,10 +48,15 @@ Classification
...
      - Negative weights are not supported.
      - ``probability=True`` is not supported with array API classes other than NumPy.
+       It's not possible to get accelerated predictions out of a multi-class classifier
+       that was fitted by falling back to scikit-learn.
    * - :obj:`sklearn.svm.NuSVC`
      - ``kernel`` must be one of [``"linear"``, ``"rbf"``, ``"poly"``, ``"sigmoid"``]
      - Negative weights are not supported.
-     - 
+     - ``probability=True`` is not supported with array API classes other than NumPy.
+       It's not possible to get accelerated predictions out of a multi-class classifier
```
avolkov-intel, line 57: Should we also add this description to GPU section?

sklearnex/svm/_base.py
david-cortes-intel: No, it would be a tuple with 2 elements instead of 3.

doc/sources/algorithms.rst
david-cortes-intel: It already says that multi-class classification is not supported on GPU.

sklearnex/svm/_base.py
david-cortes-intel: There's only one condition about sparsity in this block. Where do you see the other?

sklearnex/svm/tests/test_svm.py
```diff
@@ -223,3 +239,360 @@ def test_class_weight(queue):
...
+    if n_classes > 0:
+        decision_fn = model.decision_function(X)
+        decision_fn_skl = model_skl.decision_function(X)
+        np.testing.assert_allclose(decision_fn, decision_fn_skl)
+
+
+# This one has no oneDAL involvement, but need to verify that it works as expected.
+@pytest.mark.allow_sklearn_fallback
+@pytest.mark.parametrize("decision_function_shape", ["ovr", "ovo"])
+def test_multi_class_prediction_after_fallback(decision_function_shape):
```
avolkov-intel, line 329: Is it true that both fit and predict methods will run with stock sklearn? Should we add checks that no onedal estimator is created?

avolkov-intel:
I see that some svm tests are failing: FAILED s/svm/tests/test_svm.py::test_error_on_sparse_predict_with_dense_fit[numpy-False-SVC]

sklearnex/svm/tests/test_svm.py
david-cortes-intel: Added a check.

david-cortes-intel:
> I see that some svm tests are failing: FAILED s/svm/tests/test_svm.py::test_error_on_sparse_predict_with_dense_fit[numpy-False-SVC]

Fixed.

sklearnex/svm/_classes.py
avolkov-intel: Nevermind

ethanglaser:
No major objections from my end at this point, but please address docs fail before merge

david-cortes-intel:
> No major objections from my end at this point, but please address docs fail before merge

Fixed.

onedal/svm/svm.py
icfaust: Some info on this issue can be found here: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/utils/_sycl_queue_manager.py#L104 and https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/datatypes/dlpack/data_conversion.cpp#L77. It will yield a generic queue for the device if a dlpack device unless its pytorch (which we have special hooks developed to access it via a pointer). Ideally if a queue is given, we should prioritize it.

avolkov-intel (approved):

ethanglaser (changes requested):
Errors in private CI for py3.10, should be quick fix hopefully: `test_error_on_sparse_predict_with_dense_fit`

ethanglaser (approved):

PR #3184, opened 2026-05-20
FIX, DOC: Fix different signature of `train_test_split` and document additional arguments

david-cortes-intel:
## Description

Function `train_test_split` for some reason has args/kwargs only as signature, which is not in line with scikit-learn.

This PR modifies the signature to named arguments, and along the way documents an additional parameter from this function which scikit-learn doesn't have.

Since there's also another case of an additional parameter that is currently under a section where one would not expect to find them, it moves both of them into a new doc section.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


doc/sources/additional_parameters.rst
```diff
@@ -0,0 +1,65 @@
...
+..
+.. Unless required by applicable law or agreed to in writing, software
+.. distributed under the License is distributed on an "AS IS" BASIS,
+.. WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+.. include:: substitutions.rst
+
+#################################################
+Additional parameters in estimators and functions
```
avolkov-intel, line 17: I think it worth mentioning here that n_jobs parameter is also supported for all estimators in sklearnex even if it's missing in some of the sklearn estimators (e.g. KMeans) with a reference to n_jobs section in the docs

doc/sources/additional_parameters.rst
```diff
@@ -0,0 +1,65 @@
...
+   * - ``max_bins``
+     - `[2, inf)`
+     - ``256``
+     - Number of bins in the histogram with the discretized training data.
+   * - ``min_bin_size``
+     - `[1, inf)`
+     - ``5``
+     - Minimum number of training data points in each bin after discretization.
+
+Note that using discretized training data can greatly accelerate model training times, especially for larger data sets. However, due to the reduced fidelity of the data, the resulting model can present worse performance metrics compared to a model trained on the original data. In such cases, the number of bins can be increased with the ``max_bins`` parameter.
```
avolkov-intel, line 52: How should inf parameter value be provided? I didn't see such examples in other libraries, so I think it worth being more specific. Should user provide math.inf or somethin else?

david-cortes-intel: Added a paragraph about it.

david-cortes-intel: Sorry, replied to wrong comment. It shows a half-open interval `[2, inf)`. That means infinites are excluded.

avolkov-intel: So how should user force the defualt behavior?

david-cortes-intel: I don't get what you mean. If you look at the rendered docs, it would put that interval `[2, inf)` under the column that says 'Possible values', and would put `256` under the column that says `Default value`.

david-cortes-intel: image

avolkov-intel: I meant that it's not possible to get the behavior equivalent to sklearn

avolkov-intel (approved):

doc/sources/additional_parameters.rst
david-cortes-intel: No, it's currently not possible. There is a parameter:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/18c4c4649172b941447823e091fdc7b0de11685b/onedal/ensemble/forest.py#L116
https://uxlfoundation.github.io/oneDAL/onedal/algorithms/ensembles/decision-forest.html#training-method-dense

But it's not possible to modify it from python:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/18c4c4649172b941447823e091fdc7b0de11685b/sklearnex/ensemble/_forest.py#L217

PR #2707, opened 2025-10-01
DOC: Describe behaviors from array API

david-cortes-intel:
## Description

This PR updates the doc section about array API support to reflect the behaviors of the library as of today.

Since it looks like there won't be major changes in the dispatching logic for a while, the PR tries to describe all of the current behaviors, even if they might not sound very logical.

I am not 100% certain that I'm  getting all the details right so @icfaust please give it a thorough read. In particular, I am not sure when sklearnex models will have attributes that are array API classes and when not.

During the changes here, I noticed that the doc page for array API was linking to examples that pass dpnp/dpctl data, but those weren't using array API so data was being moved back and forth regardless. Since those do not have 'array_api' in their names I simply removed them and added a more descriptive example.

---

<details>
<summary>Checklist:</summary>

**Completeness and readability**

- [x] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [x] I have run it locally and tested the changes extensively.
- [x] All CI jobs are green or I have provided justification why they aren't.

</details>

PR #3186, opened 2026-05-22
DOC: Document lack of support for sparse torch tensors

david-cortes-intel:
## Description

From some experiments, it appears that some cases with array API support in scikit-learn work fine with sparse torch arrays, even though these aren't strictly part of array API.

In sklearnex, the table convertors do not have any special handling of sparse torch arrays, so they usually error out. This PR documents the lack of support for them.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/doc/sources/array_api.rst
+++ b/doc/sources/array_api.rst
@@ -178,6 +178,11 @@ also extends to cases where :doc:`patching <patching>` is applied, but the ``.fi
 through |sklearn| as a fallback (see :doc:`algorithms` for details) - meaning: ``move_estimator_to`` cannot
 be used with classes from the |sklearnex| regardless of which backend was used to fit the estimator.
 
+Sparse arrays
+~~~~~~~~~~~~~
+
+Currently, sparse arrays from libraries other than SciPy (e.g. ``torch.sparse``) are not supported.
+
 Example usage
 =============
 


doc/sources/array_api.rst
```diff
@@ -178,6 +178,11 @@ also extends to cases where :doc:`patching <patching>` is applied, but the ``.fi
 through |sklearn| as a fallback (see :doc:`algorithms` for details) - meaning: ``move_estimator_to`` cannot
 be used with classes from the |sklearnex| regardless of which backend was used to fit the estimator.
 
+Sparse arrays
+~~~~~~~~~~~~~
+
+Currently, sparse arrays from libraries other than SciPy (e.g. ``torch.sparse``) are not supported.
```
Vika-F, line 184: Are those supported in stock scikit?
I mean, I do not think we should specify in the docs everyting that might be supported, but is not supported for now in sklearnex.

david-cortes-intel: Their docs do not mention anything about them, but they work in some cases:
https://github.com/scikit-learn/scikit-learn/issues/34087

Technically, sparse classes from pytorch are torch tensors and inherit from the class used by dense tensors, which according to the docs should be supported:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/103c7a5331f07f0b0dfd38c6add196e79d60217d/doc/sources/array_api.rst?plain=1#L26

So this PR clarifies that they aren't actually supported.

Vika-F: Form the conversation in https://github.com/scikit-learn/scikit-learn/issues/34087 I still do not see that it is necessary to add this kind of note on our side.

david-cortes-intel: But our docs do mention that array API works on `torch.Tensor`:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/be964838bc3e7a6873bbe213b77c879da521e0cc/doc/sources/array_api.rst?plain=1#L26

.. and sparse tensor technically fall into that category.

icfaust: > But our docs do mention that array API works on `torch.Tensor`:
> 
> https://github.com/uxlfoundation/scikit-learn-intelex/blob/be964838bc3e7a6873bbe213b77c879da521e0cc/doc/sources/array_api.rst?plain=1#L26
> 
> .. and sparse tensor technically fall into that category.

@david-cortes-intel you are referencing your own work here in #2707, I consider that method of justification to be underhanded. I have to agree with @Vika-F on this.

EDIT: While I did approve that PR, decisions and forms of how the codebase operates should not be dictated by the docs, but the docs should instead be a reflection of the operation.  How things should operate should be dictated by the team and not by a single reviewer and PR writer. A reasonable user would not get a torch sparse tensor to work with sklearn, I do not feel it noteworthy to reinforce a consistency with that codebase.

david-cortes-intel: @icfaust Counterexample:
```python
import os

import numpy as np

from scipy.sparse import random as sp_random
X = sp_random(100, 10, .5, random_state=rng, format="csr")
y = rng.integers(2, size=X.shape[0])

import torch
Xt = torch.tensor(X.toarray()).to_sparse()
yt = torch.tensor(y)

from sklearn import set_config
set_config(array_api_dispatch=True)

from sklearn.linear_model import LogisticRegression
LogisticRegression().fit(Xt, yt)
```

PR #3191, opened 2026-05-25
FIX: Use default compiler when building sklearn from source

david-cortes-intel:
## Description

For some reason, under the latest meson, when it builds scikit-learn from source in jobs that pull it from their GH main branch, it selects ICX as the default compiler, which then results in failures due to Cython not being able to work with it.

This PR fixes the issue by unsetting environment variables related to compilation before attempting to install scikit-learn from source.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/.ci/scripts/setup_sklearn.sh
+++ b/.ci/scripts/setup_sklearn.sh
@@ -29,6 +29,7 @@ if [ "$sklearn_version" == "main" ]; then
     # install sklearn build dependencies
     pip install threadpoolctl joblib scipy
     # install sklearn from main branch of git repo
+    unset CC CXX CFLAGS CXXFLAGS LDFLAGS MAKEFLAGS 
     pip install git+https://github.com/scikit-learn/scikit-learn.git@main
 else
     sed -i.bak -E "s/scikit-learn==[0-9a-zA-Z.]*/scikit-learn==${sklearn_version}.*/" requirements-test.txt


.ci/scripts/setup_sklearn.sh
```diff
@@ -29,6 +29,7 @@ if [ "$sklearn_version" == "main" ]; then
     # install sklearn build dependencies
     pip install threadpoolctl joblib scipy
     # install sklearn from main branch of git repo
+    unset CC CXX CFLAGS CXXFLAGS LDFLAGS MAKEFLAGS
```
avolkov-intel, line 32: Shouldn't we set them back after installation?

avolkov-intel:
The logs of azp nightly job are not visible and the job is still failing

.ci/scripts/setup_sklearn.sh
david-cortes-intel: No. The script finishes after. They have no effect on the calling process.

avolkov-intel: But the environment is left modified even though the process is finished

david-cortes-intel:
> The logs of azp nightly job are not visible and the job is still failing

Yes, this doesn't get it to pass, but it gets it past the step of installing test dependencies.

.ci/scripts/setup_sklearn.sh
david-cortes-intel: The environment is not inherited by a process executing a script through `bash`.

avolkov-intel (approved):

PR #3198, opened 2026-05-27
FIX: Fix KMeans prediction on sparse data after dense fit

david-cortes-intel:
## Description

Fixes an issue in which namespace checks error out if a KMeans model is fitted to dense data but then used to predict on sparse data.

The error is actually coming from scikit-learn:
https://github.com/scikit-learn/scikit-learn/issues/34132

It's not possible to fix it in other cases where this should be supported like linear models, because the error in those cases comes from scikit-learn's code.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/cluster/tests/test_kmeans.py
+++ b/sklearnex/cluster/tests/test_kmeans.py
@@ -18,9 +18,14 @@ import numpy as np
 import pandas as pd
 import polars as pl
 import pytest
+import scipy.sparse as sp
 from numpy.testing import assert_allclose
-from scipy.sparse import csr_matrix
 from sklearn.datasets import make_blobs
+
+if hasattr(sp, "csr_array"):
+    CSR_CTOR = sp.csr_array
+else:
+    CSR_CTOR = sp.csr_matrix
 
 from daal4py.sklearn._utils import daal_check_version, sklearn_check_version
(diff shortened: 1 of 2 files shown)


sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -261,3 +272,31 @@ def test_cov_error_on_incompatible_devices(with_array_api):
         _ = model.transform(X_gpu)
     with pytest.raises(ValueError, match=err_match):
         _ = model.score(X_gpu)
+
+
+@pytest.mark.parametrize(
+    "array_api", [False] + ([True] if sklearn_check_version("1.5") else [])
+)
+def test_sp_predict_on_dense_fit(array_api):
```
Vika-F: Can you please make the name more descriptive?
```suggestion
def test_sparse_predict_on_dense_fit(array_api):
```

Vika-F:
@david-cortes-intel Should the case of sparse fit() + dense predict() be also checked?

david-cortes-intel:
> @david-cortes-intel Should the case of sparse fit() + dense predict() be also checked?

I don't think that'd be necessary, because sparse fit() would produce dense attributes.

Vika-F (approved):
Let's wait for the CI results and LGTM.

PR #3208, opened 2026-05-28
TST: Ensure unconstrained trees fit perfectly training data

cakedev0:
## Description

Adds a non regression-test for https://github.com/uxlfoundation/oneDAL/issues/3648

The test checks the same invariant for the 4 types of trees. It's supposed to be passing for `ExtraTreesClassifier`, `RandomForestClassifier` and `RandomForestRegressor` and failing for `ExtraTreesRegressor` as long as https://github.com/uxlfoundation/oneDAL/pr/3649 is not merged/released.

<details>

<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/onedal/ensemble/tests/test_random_forest.py
+++ b/onedal/ensemble/tests/test_random_forest.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import itertools
 
 import numpy as np
 import pytest
@@ -20,7 +21,12 @@ from numpy.testing import assert_allclose
 from sklearn.datasets import make_classification, make_regression
 
 from daal4py.sklearn._utils import daal_check_version
-from onedal.ensemble import RandomForestClassifier, RandomForestRegressor
+from onedal.ensemble import (
+    ExtraTreesRegressor,
+    ExtraTreesClassifier,
+    RandomForestClassifier,
+    RandomForestRegressor,
+)
 from onedal.tests.utils._device_selection import get_queues
 
 PREDICT_DATA = np.array([[0, 0, 0, 0]], dtype=np.float64)
@@ -63,6 +69,34 @@ def test_rf_regression(queue):
             assert_allclose([-6.97], rf.predict(PREDICT_DATA, queue=queue), atol=1e-2)
         else:
             assert_allclose([-6.83], rf.predict(PREDICT_DATA, queue=queue), atol=1e-2)
+
+
+@pytest.mark.skipif(
+    not daal_check_version((2023, "P", 200)),
+    reason="ExtraTrees requires OneDAL 2023.2",
+)
+@pytest.mark.parametrize("queue", get_queues())
+@pytest.mark.parametrize("Tree, data", [
+    *itertools.product([
+        [RandomForestRegressor, ExtraTreesRegressor],
+        [make_regression(n_samples=500, n_features=10, random_state=0)]
+    ]),
+    *itertools.product([
+        [RandomForestClassifier, ExtraTreesClassifier],
+        [make_classification(n_samples=500, n_features=10, random_state=0)]
+    ]),
+])
+def test_single_tree_fits_training_data(queue, Tree, data):
+    X, y = data
+    regressor = Tree(
+        n_estimators=1,
+        bootstrap=False,
+        max_bins=X.shape[0],
+        random_state=0,
+    ).fit(X, y, queue=queue)
+
+    prediction = regressor.predict(X, queue=queue)
+    assert_allclose(prediction, y, atol=1e-12)
 
 
 @pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


cakedev0:
In https://github.com/uxlfoundation/oneDAL/pull/3649#issuecomment-4563267070 @Vika-F said:

> In scikit-learn-intelex we are trying to rely mostly on scikit-learn test coverage

So maybe I should close this PR and open one in scikit-learn instead? Honestly I'm surprised we don't have this test already in scikit-learn, it would make sense to add it there :+1:

david-cortes-intel:
> In [uxlfoundation/oneDAL#3649 (comment)](https://github.com/uxlfoundation/oneDAL/pull/3649#issuecomment-4563267070) @Vika-F said:
> 
> > In scikit-learn-intelex we are trying to rely mostly on scikit-learn test coverage
> 
> So maybe I should close this PR and open one in scikit-learn instead? Honestly I'm surprised we don't have this test already in scikit-learn, it would make sense to add it there 👍

This kind of correctness/edge-case test would fit better under the test suite for oneDAL. If it's added to oneDAL, it'd be redundant to then run it again here. Perhaps @Alexandr-Solovev or @Vika-F could add the same test in oneDAL as part of the other PR (https://github.com/uxlfoundation/oneDAL/pull/3649).

Also sounds like it'd be good to have a similar test in scikit-learn itself, considering that you have plans to standardize decision tree backends among classes. Or at least to have it for the Hist variants.

cakedev0:
Yes @Vika-F wrote a test in oneDAL, I will pull it in my PR.

This CI here confirms what I expected: 

> It's supposed to be passing for ExtraTreesClassifier, RandomForestClassifier and RandomForestRegressor and failing for ExtraTreesRegressor 

So I saw what I wanted to see, I'm closing this PR and will open a similar one directly in scikit-learn.

cakedev0:
Hum... But with have this test already in scikit-learn ([this one](https://github.com/scikit-learn/scikit-learn/blob/39d0ad03934e24196899c01c8c5ac465cf15401a/sklearn/tree/tests/test_tree.py#L337) for instance), but it's on single tree classes (`DecisionTree*`/`ExtraTree*`), which are not implemented by `sklearnex`... 

I think the best would be to implement those in `sklearnex` through ensemble models + `n_estimators=1`, at least for the tests (private classes + monkey patching could work maybe?).

Wdyt @david-cortes-intel?

david-cortes-intel:
> Hum... But with have this test already in scikit-learn ([this one](https://github.com/scikit-learn/scikit-learn/blob/39d0ad03934e24196899c01c8c5ac465cf15401a/sklearn/tree/tests/test_tree.py#L337) for instance), but it's on single tree classes (`DecisionTree*`/`ExtraTree*`), which are not implemented by `sklearnex`...
> 
> I think the best would be to implement those in `sklearnex` through ensemble models + `n_estimators=1`, at least for the tests (private classes + monkey patching could work maybe?).
> 
> Wdyt @david-cortes-intel?

I still think it'd be better to have the test in oneDAL instead, since it's not specific to the Python interface, and not exactly related to compatibility with other Python packages. If not possible, then having it here would be fine too, but the pytest condition would have to be on onedal>=2026.2 which is where the fix would likely land.

cakedev0:
> I still think it'd be better to have the test in oneDAL instead

Yes, I agree with that.

But I think you missed my point (which was maybe not explicit enough).

**Statements:**
- `scikit-learn-intelex` relies mostly on `scikit-learn` test coverage
- `scikit-learn-intelex` doesn't implement "single tree" estimators (`DecisionTree*`/`ExtraTree*`)
- In `scikit-learn`, correctness tests of "bagged trees" estimators (`RandomForest*`/`ExtraTrees*`) are week, because we relies on correctness of "single tree" estimators for which tests are very strong.

**Conclusion/Problem:** `scikit-learn-intelex` "bagged trees" estimators are not well covered in term of correctness tests.

**Proposed solution:** `scikit-learn-intelex` could implement test-only classes that wraps "bagged trees" estimators with `n_estimators=1` and `bootstrap=False`, and run `sklearn/tree` tests against those classes. With that, correctness of the implementation will be very thoroughly tested. (some tests will pass only with `max_bins>=n_samples`, but this is also something that can be set by the wrappers).

david-cortes-intel:
It is indeed the case that we rely on scikit-learn's test suite to a large degree, but we still have lots of correctness-related tests in oneDAL, especially when they are not too complicated or involve something that's not easily triggerable from Python.

For example:
https://github.com/uxlfoundation/oneDAL/blob/4681fc938a4b8062d286931589bdadc3398a12c0/cpp/oneapi/dal/algo/linear_regression/test/fixture.hpp#L326
https://github.com/uxlfoundation/oneDAL/blob/4681fc938a4b8062d286931589bdadc3398a12c0/cpp/oneapi/dal/algo/linear_regression/test/fixture.hpp#L68

Some classes are covered better than others in oneDAL's tests, and there is indeed a lot of things that are currently missing and only tested through scikit-learn, but we'd like to improve on those deficiencies if possible.

PR #3224, opened 2026-06-01
MAINT: Skip RF float32 test on GPU that fails due to tolerances

david-cortes-intel:
## Description

Skips a test that will start failing due to small floating point differences after this fix:
https://github.com/uxlfoundation/oneDAL/pull/3649#issuecomment-4590964782

---

<details>
<summary>Checklist:</summary>

</details>

### DIFF
--- a/deselected_tests.yaml
+++ b/deselected_tests.yaml
@@ -636,6 +636,9 @@ gpu:
   - tests/test_common.py::test_estimators[RandomForestClassifier(n_estimators=5)-check_class_weight_classifiers]
   - tests/test_common.py::test_estimators[ExtraTreesClassifier(n_estimators=5)-check_class_weight_classifiers]
 
+  # Fails for float32 due to small numeric tolerances in the order of 1e-6
+  - ensemble/tests/test_forest.py::test_memory_layout[float32
+
   # --------------------------------------------------------
   # The following tests fail at collection time, so they cannot be deselected with --deselect and must be ignored with --ignore.
 ignored_tests:


cakedev0:
Note that I noted a difference of behavior between scikit-learn and scikit-learn-intelex regarding precision in trees:
- scikit-learn-intelex aligns everything on X (including y, and calculation on y like impurity)
- scikit-learn always uses float32 for X and float64 for y...

... I think the behavior I would prefer would be to keep the input dtypes of X and y. i.e.:
- do sorting/splitting/binning using the dtype of X
- do impurity calculations using the dtype of y

 I'll check how hard it would be to have that in scikit-learn, and if it's simple (which I think it is), I'll open an RFC to move towards this behavior.

Edit: sorting in float64 will be expensive I guess, so we don't want to do that, we'll have to do features indexing first, like it's done in onedal => **we will keep the current behavior for now**

david-cortes-intel:
> Note that I noted a difference of behavior between scikit-learn and scikit-learn-intelex regarding precision in trees:
> 
> * scikit-learn-intelex aligns everything on X (including y, and calculation on y like impurity)
> * scikit-learn always uses float32 for X and float64 for y...
> 
> ... I think the behavior I would prefer would be to keep the input dtypes of X and y. i.e.:
> 
> * do sorting/splitting/binning using the dtype of X
> * do impurity calculations using the dtype of y
> 
> I'll check how hard it would be to have that in scikit-learn, and if it's simple (which I think it is), I'll open an RFC to move towards this behavior.
> 
> Edit: sorting in float64 will be expensive I guess, so we don't want to do that, we'll have to do features indexing first, like it's done in onedal => **we will the current behavior for now**

That is currently not the case for the CPU version. Impurity calculations are done in float64 regardless of the input types.

There's a comment about it at the top of the file explaining why it needs to be like that:
https://github.com/uxlfoundation/oneDAL/blob/4681fc938a4b8062d286931589bdadc3398a12c0/cpp/daal/src/algorithms/dtrees/forest/regression/df_regression_train_dense_default_impl.i#L53

cakedev0:
The impurity accumulators themselves are indeed `double`.

However, I think the observed behavior is still impacted by `X.dtype` through the stopping threshold: the oneDAL wrapper sets the forest `fptype` from the data table dtype:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/4af2a966903fb39a9955873eff3999c21a1987ad/onedal/ensemble/forest.py#L114-L120

And in oneDAL training, `_impurityThreshold` is floored at `EpsilonVal<algorithmFPType>`:
https://github.com/uxlfoundation/oneDAL/blob/4681fc938a4b8062d286931589bdadc3398a12c0/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i#L540-L549

So even with `impurity_threshold=0`, the effective threshold is approximately:

```text
max(user_impurity_threshold, eps(algorithmFPType_from_X))
```

In the reproducer bellow, we see that X dtype impacts a lot the number of leaves of the tree, while y dtype not so much.

<details>

```python
import itertools

import numpy as np
from scipy.stats import rankdata
from sklearn.datasets import make_regression
from sklearnex.ensemble import RandomForestRegressor

dtypes = ['float64', 'float32']

for (X_dtype, y_dtype) in itertools.product(dtypes, dtypes):
    X, y = make_regression(
        3 * 10**5,
        n_features=1,
        random_state=0,
    )
    X[:, 0] = rankdata(X[:, 0])
    # ^ just to have only unique values in X, even after conversion to float32
    X = X.astype(X_dtype)
    y = y.astype(y_dtype)
    print('X dtype:', X.dtype, " | n unique values:", np.unique(X.ravel()).size)
    print('y dtype:', y.dtype, " | n unique values:", np.unique(y).size)
    rf = RandomForestRegressor(5, max_bins=len(X), bootstrap=False, n_jobs=1)
    rf.fit(X, y)
    n_leaves = set(int(rf.estimators_[i].tree_.n_leaves) for i in range(len(rf.estimators_)))
    print('n leaves in trees:', n_leaves)
    print()
```

```text
X dtype: float64  | n unique values: 300000
y dtype: float64  | n unique values: 300000
n leaves in trees: {299962}

X dtype: float64  | n unique values: 300000
y dtype: float32  | n unique values: 299402
n leaves in trees: {299406}

X dtype: float32  | n unique values: 300000
y dtype: float64  | n unique values: 300000
n leaves in trees: {123191}

X dtype: float32  | n unique values: 300000
y dtype: float32  | n unique values: 299402
n leaves in trees: {123191}
```

</details>

Note that I don't think it has a real impact on users, it's just a slightly inconsistent behavior I'd say (but scikit-learn has some too, like always converting X to float32).

david-cortes-intel:
Thanks for pointing this out. Should be fixed here: https://github.com/uxlfoundation/oneDAL/pull/3653

cakedev0:
Oh nice! I hadn't realized it was so simple to fix, very neat :ok_hand:

ethanglaser (approved):
PR works as intended: http://intel-ci.intel.com/f15deb43-a12d-f168-a5d2-d4f5ef20c6a0

PR #3228, opened 2026-06-02
[maintenance] Remove onedal/sklearnex functionality mixing in `KNeighborsClassifier.predict` using `from_table`

icfaust:
## Description

`KNeighborsClassifier` uses the onedal `from_table` functionality and internal parameter setting functionality in a way which is different from the rest of the codebase.  This makes it similar to all other uses in the codebase.  This is not a final solution, as a refactoring of the `_get_onedal_params` needs to be done, and therefore includes a temporary (but limited) solution.

Follow up: evaluate how to use a _create_model function in the classifier, as a `model` is available for this specific estimator (allowing for the definition of a `_create_model` function).

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/onedal/neighbors/neighbors.py
+++ b/onedal/neighbors/neighbors.py
@@ -196,7 +196,7 @@ class KNeighborsClassifier(NeighborsBase):
         return self.train(params, X_table, y_table).model
 
     @supports_queue
-    def predict(self, X, params, queue=None):
+    def predict(self, X, queue=None):
         X_table = to_table(X, queue=queue)
         # pass Y as non-None to trigger 'responses'
         params = self._get_onedal_params(X_table, 0)
--- a/onedal/neighbors/neighbors.py
+++ b/onedal/neighbors/neighbors.py
@@ -202,7 +202,7 @@ class KNeighborsClassifier(NeighborsBase):
         params["fptype"] = X.dtype
         result = self.infer(params, model, X)
 
-        return result
+        return from_table(result.responses, like=X)
 
     @supports_queue
     def fit(self, X, y, queue=None):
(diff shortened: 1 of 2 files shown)


onedal/neighbors/neighbors.py
```diff
@@ -195,22 +195,36 @@ def _onedal_fit(self, X, y):
...
-        if "responses" not in params["result_option"]:
-            params["result_option"] += "|responses"
-        params["fptype"] = X.dtype
-        result = self.infer(params, model, X)
+    @supports_queue
+    def predict(self, X, queue=None):
+        X_table = to_table(X, queue=queue)
+        # pass Y as non-None to trigger 'responses'
+        params = self._get_onedal_params(X_table, 0)
+        result = self.infer(params, self._onedal_model, X)
```
yuejiaointel, line 203: should probably pass X_table here

onedal/neighbors/neighbors.py
```diff
@@ -195,22 +195,36 @@ def _onedal_fit(self, X, y):
...
+        _check_is_fitted(self)
+
+        if n_neighbors is None:
+            n_neighbors = self.n_neighbors
+
+        if X is None:
+            X = self._fit_X
+
+        X_table = to_table(X)
+        params = self._get_onedal_params(X, n_neighbors=n_neighbors)
```
yuejiaointel, line 222: This should also be X_table

onedal/neighbors/neighbors.py
```diff
@@ -195,22 +195,36 @@ def _onedal_fit(self, X, y):
...
+        prediction_results = self.infer(self._onedal_model, X_table, params)
```
yuejiaointel, line 223: The parameter ordering is not correct, should be (params, self._onedal_model, X_table)

PR #3248, opened 2026-06-03
[maintenance] Simplify wrap_output_data and fix transform_output guard

yuejiaointel:
The transform_output host-transfer block was duplicated across the array_api and non-array_api branches. Hoist it to a single early-return at the top of the wrapper.

Also drop the redundant `transform_output in ("default", None)` guard around the array-api result conversion. By the time control reached it, the earlier transform/fit_transform block had already returned for the non-default case, so the guard was either always True or incorrectly suppressed namespace conversion on predict/score paths when the user had set a global transform_output=polars/pandas.

## Description

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


sklearnex/_device_offload.py
```diff
@@ -192,52 +192,38 @@ def wrap_output_data(func: Callable) -> Callable:
     @wraps(func)
     def wrapper(self, *args, **kwargs) -> Any:
         result = func(self, *args, **kwargs)
-        # In case ARRAY API is enabled the result is already converted to the required type
+
+        # When transform_output is polars/pandas, sklearn's _set_output wrapper
+        # calls pl.DataFrame(result) which can't handle GPU arrays. Transfer to
```
david-cortes-intel: @yuejiaointel Did you see any kind of errors around it somewhere? As far as I can tell, everything works correctly with polars:
Image

But there are some issues with pandas in specific:
Image

david-cortes-intel: Actually it does fail if the tensor is on XPU.

david-cortes-intel:
@yuejiaointel Please add tests checking the expected results under KMeans, including dpnp and torch on CPU and GPU, plus array-api-strict.

sklearnex/_device_offload.py
yuejiaointel: Right, now this works
image
and on GPU also works before would get type error
image

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -239,6 +240,52 @@ def test_array_api_dispatch_output_type(dataframe, queue):
...
+        km = KMeans(n_clusters=3, random_state=0, n_init=1).fit(X)
+        _check_kmeans_results(km, X, X_np)
+
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="array_api_dispatch requires sklearn >= 1.2"
+)
+@pytest.mark.parametrize("device", ["cpu"] + (["xpu"] if torch_xpu_available else []))
+def test_torch_dispatch_results(device):
+    torch = pytest.importorskip("torch")
```
david-cortes-intel: Please use variable `torch_available`.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -239,6 +240,52 @@ def test_array_api_dispatch_output_type(dataframe, queue):
...
+    with config_context(array_api_dispatch=True):
+        km = KMeans(n_clusters=3, random_state=0, n_init=1).fit(X)
+        _check_kmeans_results(km, X, X_np)
+
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="array_api_dispatch requires sklearn >= 1.2"
+)
+@pytest.mark.parametrize("device", ["cpu"] + (["xpu"] if torch_xpu_available else []))
+def test_torch_dispatch_results(device):
```
david-cortes-intel: Please add tests checking that:
* Outputs from different types of transform match with the default option.
* They work under both pandas and polars.
* They work under target offload.
* They work when passed DPNP without array API.
* They work when passed array API from DPNP, torch, and array-api-strict, under CPU and GPU.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, fixed

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Existing tests only checked type match. Added value-match tests (wrapped output == default):

- `test_transform_output_matches_default`: parametrized over `(xp, device)` (numpy/array-api-strict/dpnp/torch, CPU+GPU) and `method` (transform/fit_transform)
- `test_transform_output_dpnp_no_array_api`: dpnp without array_api_dispatch
- `test_transform_output_target_offload`: under target_offload

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -215,6 +221,75 @@ def test_transform_output_gpu(dataframe, queue, transform_output):
...
+)
+@pytest.mark.parametrize(
+    "dataframe,queue", get_dataframes_and_queues("numpy,dpnp,array_api")
+)
+@pytest.mark.parametrize("transform_output", ["polars", "pandas"])
+def test_transform_output_matches_default(dataframe, queue, transform_output):
+    X_np = generate_dense_dataset(200, 10, 0.5, 3)
+    X = _convert_to_dataframe(X_np, sycl_queue=queue, target_df=dataframe)
+
+    with config_context(array_api_dispatch=True):
```
david-cortes-intel: Please use this common fixture instead:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/ffab32dcf35e539d08ebbef4cd562b7fdbd0d916/sklearnex/conftest.py#L71

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -215,6 +221,75 @@ def test_transform_output_gpu(dataframe, queue, transform_output):
...
+    X = _convert_to_dataframe(X_np, sycl_queue=queue, target_df=dataframe)
+
+    km = KMeans(n_clusters=3, random_state=0, n_init=1).fit(X)
+    _assert_transform_output_matches_default(km, X, transform_output)
+
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="transform_output requires sklearn >= 1.2"
+)
+@pytest.mark.parametrize("queue", get_queues("gpu"))
```
david-cortes-intel: Please add a conditional skip when no GPU device is available.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -239,6 +314,51 @@ def test_array_api_dispatch_output_type(dataframe, queue):
         assert isinstance(sc, float)
 
 
+def _check_kmeans_results(km, X, X_np):
```
david-cortes-intel: This could be merged into the previous tests as a first step.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -215,6 +221,75 @@ def test_transform_output_gpu(dataframe, queue, transform_output):
...
+    _assert_transform_output_matches_default(km, X, transform_output)
+
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="transform_output requires sklearn >= 1.2"
+)
+@pytest.mark.parametrize("queue", get_queues("gpu"))
+@pytest.mark.parametrize("transform_output", ["polars", "pandas"])
+def test_transform_output_target_offload(queue, transform_output):
+    X_np = generate_dense_dataset(200, 10, 0.5, 3)
```
david-cortes-intel: Perhaps it could be made smaller to make the tests faster.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -215,6 +221,75 @@ def test_transform_output_gpu(dataframe, queue, transform_output):
...
+        _assert_transform_output_matches_default(km, X, transform_output)
+
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="array_api_dispatch requires sklearn >= 1.2"
+)
+@pytest.mark.skipif(not torch_available, reason="Functionality to test requires torch.")
+@pytest.mark.parametrize("device", ["cpu"] + (["xpu"] if torch_xpu_available else []))
+@pytest.mark.parametrize("transform_output", ["polars", "pandas"])
+def test_transform_output_matches_default_torch(device, transform_output):
```
david-cortes-intel: Please add a similar test for fit_transform. Or maybe parametrize by it.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -239,6 +314,51 @@ def test_array_api_dispatch_output_type(dataframe, queue):
...
+    assert_allclose(sc, -np.square(trans.min(axis=1)).sum(), rtol=1e-5)
+
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="array_api_dispatch requires sklearn >= 1.2"
+)
+@pytest.mark.parametrize(
+    "dataframe,queue", get_dataframes_and_queues("numpy,dpnp,array_api")
+)
+def test_array_api_dispatch_results(dataframe, queue):
```
david-cortes-intel: Why would it have a different test for torch than for the rest? See for example other array API tests elsewhere:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/ffab32dcf35e539d08ebbef4cd562b7fdbd0d916/sklearnex/linear_model/tests/test_mixed_inputs.py#L193

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -239,6 +314,51 @@ def test_array_api_dispatch_output_type(dataframe, queue):
...
+    assert_allclose(trans, expected_distances, rtol=1e-4, atol=1e-4)
+    assert_allclose(pred, trans.argmin(axis=1))
+    assert_allclose(sc, -np.square(trans.min(axis=1)).sum(), rtol=1e-5)
+
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="array_api_dispatch requires sklearn >= 1.2"
+)
+@pytest.mark.parametrize(
+    "dataframe,queue", get_dataframes_and_queues("numpy,dpnp,array_api")
```
david-cortes-intel: Could also have a test in which 'X' is pandas / polars as input. Maybe combined with/without target offload.

david-cortes-intel:
@yuejiaointel I'm seeing several test failures with torch. Example:
```
_____________________ test_transform_output_matches_default_torch[pandas-xpu] ______________________

device = 'xpu', transform_output = 'pandas'

    @pytest.mark.skipif(
        not sklearn_check_version("1.2"), reason="array_api_dispatch requires sklearn >= 1.2"
    )
    @pytest.mark.skipif(not torch_available, reason="Functionality to test requires torch.")
    @pytest.mark.parametrize("device", ["cpu"] + (["xpu"] if torch_xpu_available else []))
    @pytest.mark.parametrize("transform_output", ["polars", "pandas"])
    def test_transform_output_matches_default_torch(device, transform_output):
        X_np = generate_dense_dataset(200, 10, 0.5, 3)
        X = torch.tensor(X_np, device=device)
    
        with config_context(array_api_dispatch=True):
            km = KMeans(n_clusters=3, random_state=0, n_init=1).fit(X)
>           _assert_transform_output_matches_default(km, X, transform_output)

sklearnex/cluster/tests/test_kmeans.py:263: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
sklearnex/cluster/tests/test_kmeans.py:227: in _assert_transform_output_matches_default
    default = _as_numpy(km.transform(X))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
onedal/tests/utils/_dataframes_support.py:141: in _as_numpy
    return np.asarray(obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = tensor([[10.7700, 12.9257, 19.2683],
        [12.2518, 12.9991, 14.2836],
        [13.7644,  9.1373, 16.6484],
       ....4312],
        [ 8.2847, 13.4288, 21.0877],
        [ 8.3473, 10.9056, 14.8079]], device='xpu:0', dtype=torch.float64)
dtype = None

    def __array__(self, dtype=None):
        if has_torch_function_unary(self):
            return handle_torch_function(Tensor.__array__, (self,), self, dtype=dtype)
        if dtype is None:
>           return self.numpy()
                   ^^^^^^^^^^^^
E           TypeError: can't convert xpu:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.

../../miniforge3/envs/torch19/lib/python3.13/site-packages/torch/_tensor.py:1253: TypeError
```

sklearnex/_device_offload.py
```diff
@@ -192,52 +192,41 @@ def wrap_output_data(func: Callable) -> Callable:
...
+            return result
+
+        data = (*args, *kwargs.values())[0]
+
+        if usm_iface := getattr(data, "__sycl_usm_array_interface__", None):
+            return copy_to_dpnp(usm_iface["syclobj"], result)
+
+        if hasattr(data, "dtype"):
+            xp, is_array_api = get_namespace(data)
+            if is_array_api and not _is_numpy_namespace(xp):
```
david-cortes-intel, line 224: Isn't the second condition redundant?

sklearnex/_device_offload.py
```diff
@@ -192,52 +192,41 @@ def wrap_output_data(func: Callable) -> Callable:
...
+                device = getattr(data, "device", None)
+                if isinstance(result, tuple):
+                    result = tuple(xp.asarray(r, device=device) for r in result)
+                elif not isinstance(result, (int, float)):
```
david-cortes-intel: Can oneDAL return scalar types from numpy?

sklearnex/_device_offload.py
```diff
@@ -192,52 +192,41 @@ def wrap_output_data(func: Callable) -> Callable:
...
-                        device = getattr(data, "device", None)
-                        if isinstance(result, tuple):
-                            result = tuple(xp.asarray(r, device=device) for r in result)
-                        elif not isinstance(result, (int, float)):
-                            result = xp.asarray(result, device=device)
+
+        if not args and not kwargs:
+            return result
+
+        data = (*args, *kwargs.values())[0]
```
david-cortes-intel, line 217: Is there a case where some method would _not_ have `X` under `*args`?

sklearnex/_device_offload.py
```diff
@@ -192,52 +192,41 @@ def wrap_output_data(func: Callable) -> Callable:
...
+
+        # When transform_output is polars/pandas, sklearn's _set_output wrapper
+        # builds a DataFrame from the result. For array-api results whose data
+        # lives on a non-CPU device the DataFrame constructor raises: pandas with
+        # "Implicit conversion to a NumPy array is not allowed" and polars with
+        # "DataFrame constructor called with unsupported type". Transfer to host
+        # first so sklearn can wrap into the requested format.
+        if func.__name__ in ("transform", "fit_transform") and (
+            get_config().get("transform_output") not in ("default", None)
+            or getattr(self, "_sklearn_output_config", {}).get("transform", "default")
```
david-cortes-intel, line 204: Perhaps it could default to 'default' to make it more clear.

sklearnex/_device_offload.py
david-cortes-intel: Actually there is:
https://github.com/yuejiaointel/scikit-learn-intelex/blob/2a24df542c4e7ddc1423b7dfda91ddbb27f6ea90/sklearnex/neighbors/knn_unsupervised.py#L95

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, fixed — switched to the `with_array_api` conftest fixture.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, fixed — added a conditional skip with `is_sycl_device_available("gpu")`.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, fixed — merged into the parametrized (xp, device) tests; the separate output-type test is gone.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, fixed — reduced the dataset from 200x10 to 50x5.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, fixed — parametrized over `method` (`transform` / `fit_transform`).

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, fixed — unified torch/dpnp/array-api-strict/numpy into a single `(xp, device)` parametrization, mirroring test_mixed_inputs.py.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: thx, fixed — added `test_dispatch_results_pandas_polars`, parametrized over pandas/polars input and with/without GPU target_offload.

sklearnex/_device_offload.py
yuejiaointel: thx, fixed — now `get_config().get("transform_output", "default")`.

sklearnex/_device_offload.py
yuejiaointel: Good catch — `kneighbors(X=None)` is exactly that case. When it is called without data, `args`/`kwargs` are empty so the `if not args and not kwargs: return result` guard above returns early; the `(*args, *kwargs.values())[0]` indexing only runs when data was actually passed.

sklearnex/_device_offload.py
yuejiaointel: Not redundant -- `array_api_dispatch=True` makes numpy match both conditions:
```python
>>> from sklearn.utils._array_api import get_namespace
>>> with config_context(array_api_dispatch=True):
...     xp, is_array_api = get_namespace(np.zeros(3))
>>> is_array_api, _is_numpy_namespace(xp)
(True, True)
```
So without `not _is_numpy_namespace(xp)`, numpy-under-dispatch hits `xp.asarray(result, device=...)` -- needless, and fails on older numpy with no `device=` kwarg.

sklearnex/_device_offload.py
yuejiaointel: [example ](https://github.com/uxlfoundation/scikit-learn-intelex/blob/64b661cfa8a8d6d1deb486031e74b0d23b13a7f7/onedal/cluster/kmeans.py#L313) is `score()` returns `-1 * result.objective_function_value`, which is a python `float` (verified, `ndim` 0). So oneDAL doesn't return a numpy scalar in this path, but `ndim` is the robust guard either way -- it leaves any scalar (python or numpy) untouched and only converts real arrays:
```python
>>> getattr(np.float64(1.0), "ndim", 0), getattr(1.0, "ndim", 0), np.zeros(3).ndim
(0, 0, 1)
```
Also avoids the `isinstance(result, (int, float))` inconsistency where `np.int64` would slip through and get wrapped. Kept namespace-agnostic, no numpy import.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -176,57 +177,33 @@ def test_dense_vs_sparse(queue, init, algorithm, dims):
...
-@pytest.mark.skipif(
-    not sklearn_check_version("1.5"),
-    reason="Functionality introduced in later sklearn versions.",
+def _convert(arr, xp, device):
+    """Convert a numpy array to the array-API backend ``xp`` on ``device``."""
+    if xp is np:
+        return arr
+    if xp is array_api_strict:
+        return array_api_strict.asarray(arr)
+    return xp.asarray(arr, device=device)
```
david-cortes-intel, line 300: This works for array-api-stric too.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -333,29 +277,36 @@ def _check_kmeans_results(km, X, X_np):
...
-@pytest.mark.skipif(not torch_available, reason="Functionality to test requires torch.")
-@pytest.mark.parametrize("device", ["cpu"] + (["xpu"] if torch_xpu_available else []))
-def test_torch_dispatch_results(device):
-    X_np = generate_dense_dataset(200, 10, 0.5, 3)
-    X = torch.tensor(X_np, device=device)
+@pytest.mark.parametrize(
+    "queue", [None] + (get_queues("gpu") if is_sycl_device_available("gpu") else [])
+)
+@pytest.mark.parametrize("dataframe", [pd.DataFrame, pl.DataFrame])
+def test_dispatch_results_pandas_polars(dataframe, queue):
```
david-cortes-intel: It should test different values of transform output.

sklearnex/cluster/tests/test_kmeans.py
david-cortes-intel: Looks like it's not testing for different transform outputs.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -169,74 +177,137 @@ def test_dense_vs_sparse(queue, init, algorithm, dims):
...
-            with config_context(transform_output=transform_output):
-                km.fit(X_torch)
-                result = km.transform(X_torch)
 
+
+def _assert_transform_output_matches_default(km, X, transform_output, method):
+    """The polars/pandas transform_output wrapping must preserve the values of
+    the default (un-wrapped) output, independent of input type/device."""
+    default = _as_numpy(getattr(km, method)(X))
+    with config_context(transform_output=transform_output):
```
david-cortes-intel: There's 2 ways to trigger this:
* config_context
* estimator.set_output

Please test both.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -169,74 +177,137 @@ def test_dense_vs_sparse(queue, init, algorithm, dims):
...
-    expected_type = pl.DataFrame if transform_output == "polars" else pd.DataFrame
-    assert isinstance(result, expected_type)
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="transform_output requires sklearn >= 1.2"
+)
+@pytest.mark.skipif(
+    not is_sycl_device_available("gpu"), reason="Test for GPU-specific functionality."
+)
+@pytest.mark.parametrize("queue", get_queues("gpu"))
```
david-cortes-intel: This kind of pattern would make the skip happen after the call to create the queue. It would throw this warning if executed in an environment that hasn't built the DPC module:
```
onedal/tests/utils/_device_selection.py:61
  /localdisk2/mkl/dcortes/repos/scikit-learn-intelex/onedal/tests/utils/_device_selection.py:61: UserWarning: Attempting to get a GPU queue, but DPC backend is not available.
    warnings.warn(
```

Please put the calls to GPU queues inside the tests.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Right, removed the branch — `array_api_strict.asarray(arr, device=None)` works (it gets `device=None` from `_array_api_inputs`), so the generic line covers it.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Added — `_check_kmeans_results` now also asserts `transform_output=polars/pandas` wraps into a DataFrame whose values match the default, so this test exercises it for pandas/polars input (with and without GPU offload).

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Folded into `_check_kmeans_results`, which both dispatch tests call, so `transform_output=polars/pandas` value-match now runs across numpy/array-api-strict/dpnp/torch (CPU+GPU) too.

sklearnex/cluster/tests/test_kmeans.py
david-cortes-intel: @yuejiaointel Forgot to push?

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Fixed — queues are now created inside the tests via `target_offload="gpu"`, so collection no longer touches the GPU.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Yes, sorry — pushed now (888c7f1e).

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Both triggers are now covered — the helper checks `config_context(transform_output=...)` and `estimator.set_output(transform=...)`.

yuejiaointel:
Fixed the failing `test_array_api_dispatch_results[array_api_strict]` — reverted the `_as_numpy` dlpack change; `array_api_strict` 2.5 arrays reject both `__array__` and readonly-buffer DLPack export, and plain `np.asarray` handles them. Pushed in e7c6b8c7.

david-cortes-intel:
@yuejiaointel I'm still getting conversion errors with torch:
```
____________________________ test_array_api_dispatch_results[torch-xpu] ____________________________

xp = <module 'torch' from '/mnt/cached_oses/dcortes/miniforge3/envs/torch19/lib/python3.13/site-packages/torch/__init__.py'>
device = 'xpu', with_array_api = None

    @pytest.mark.skipif(
        not sklearn_check_version("1.2"), reason="array_api_dispatch requires sklearn >= 1.2"
    )
    @pytest.mark.parametrize("xp,device", _array_api_inputs)
    def test_array_api_dispatch_results(xp, device, with_array_api):
        X_np = generate_dense_dataset(50, 5, 0.5, 3)
        X = _convert(X_np, xp, device)
    
        km = KMeans(n_clusters=3, random_state=0, n_init=1).fit(X)
        # predict/transform/cluster_centers_ follow the input type.
        assert type(km.predict(X)) == type(X)
        assert type(km.transform(X)) == type(X)
        assert type(km.cluster_centers_) == type(X)
>       _check_kmeans_results(km, X, X_np)

sklearnex/cluster/tests/test_kmeans.py:309: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
sklearnex/cluster/tests/test_kmeans.py:278: in _check_kmeans_results
    centers = _as_numpy(km.cluster_centers_)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
onedal/tests/utils/_dataframes_support.py:141: in _as_numpy
    return np.asarray(obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = tensor([[-5.3155, -9.0958,  4.6081,  1.5780,  1.9101],
        [-8.4886,  3.2664,  1.6732, -1.2613, -0.2838],
        [-0.6270,  2.3935,  2.5095, -0.6282, -4.0319]], device='xpu:0',
       dtype=torch.float64)
dtype = None

    def __array__(self, dtype=None):
        if has_torch_function_unary(self):
            return handle_torch_function(Tensor.__array__, (self,), self, dtype=dtype)
        if dtype is None:
>           return self.numpy()
                   ^^^^^^^^^^^^
E           TypeError: can't convert xpu:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.

../../miniforge3/envs/torch19/lib/python3.13/site-packages/torch/_tensor.py:1253: TypeError
```

yuejiaointel:
Confirmed passing on a local Intel GPU (Max 1100) with the fix:
```
test_array_api_dispatch_results[torch-xpu] PASSED
test_transform_output_matches_default[{fit_transform,transform}-{pandas,polars}-torch-xpu] 4 passed
```
Reproduced the CI env via dal-infra#986 — the trick is torch **nightly** xpu (oneAPI 2026, matches oneDAL-DPC); the stable wheel ships oneAPI 2025.3 and conflicts.

onedal/tests/utils/_dataframes_support.py
```diff
@@ -138,7 +139,13 @@ def _as_numpy(obj, *args, **kwargs):
         return obj.to_numpy(*args, **kwargs)
     if sp.issparse(obj):
         return obj.toarray(*args, **kwargs)
-    return np.asarray(obj, *args, **kwargs)
+    try:
+        return np.asarray(obj, *args, **kwargs)
+    except (TypeError, RuntimeError, ValueError):
```
david-cortes-intel: This might also throw BufferError. There's also `np.from_dlpack`.

sklearnex/decomposition/tests/test_pca.py
```diff
@@ -27,10 +29,25 @@
...
     get_dataframes_and_queues,
+    torch_available,
+    torch_xpu_available,
 )
 from onedal.tests.utils._device_selection import is_sycl_device_available
+from sklearnex import config_context
 from sklearnex.decomposition import PCA
 
+try:
+    import array_api_strict
```
david-cortes-intel: This is a mandatory dependency, no need to put it under 'try'.

sklearnex/decomposition/tests/test_pca.py
```diff
@@ -195,3 +212,92 @@ def test_pca_error_on_incompatible_devices(with_array_api):
...
+        pca._sklearn_output_config = original_config
+
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="array_api_dispatch requires sklearn >= 1.2"
+)
+@pytest.mark.parametrize("xp,device", _array_api_inputs)
+@pytest.mark.parametrize("transform_output", ["polars", "pandas"])
+@pytest.mark.parametrize("method", ["transform", "fit_transform"])
+def test_transform_output_matches_default(
```
david-cortes-intel, line 259: Please remember to add a case testing when the input is pandas/polars, with and without target_offload.

sklearnex/decomposition/tests/test_pca.py
```diff
@@ -195,3 +212,92 @@ def test_pca_error_on_incompatible_devices(with_array_api):
...
+def test_transform_output_dpnp_no_array_api(dataframe, queue, transform_output, method):
+    X = _convert_to_dataframe(
+        load_iris(return_X_y=True)[0], sycl_queue=queue, target_df=dataframe
+    )
+    pca = PCA(n_components=3).fit(X)
+    _assert_transform_output_matches_default(pca, X, transform_output, method)
+
+
+@pytest.mark.skipif(
+    not sklearn_check_version("1.2"), reason="transform_output requires sklearn >= 1.2"
```
david-cortes-intel: CC @ethanglaser will also need to be removed in the sklearn1.0 PR.

sklearnex/decomposition/tests/test_pca.py
```diff
@@ -195,3 +212,92 @@ def test_pca_error_on_incompatible_devices(with_array_api):
...
+
+    # 1) global config_context(transform_output=...)
+    with config_context(transform_output=transform_output):
+        out = getattr(pca, method)(X)
+    assert isinstance(out, expected_type)
+    assert_allclose(out.to_numpy(), default, rtol=1e-5, atol=1e-5)
+
+    # 2) per-estimator set_output(transform=...). Restore the original config
+    # afterwards -- set_output("default") would *pin* it and override (1).
+    original_config = getattr(pca, "_sklearn_output_config", {}).copy()
```
david-cortes-intel: You could put `.set_output(transform="default")` in the `finally` block instead.

onedal/tests/utils/_dataframes_support.py
yuejiaointel: Added `BufferError`. `dlpack_to_numpy` already uses `np.from_dlpack` under the hood, so the fallback covers that.

sklearnex/decomposition/tests/test_pca.py
yuejiaointel: Done — imported directly.

sklearnex/decomposition/tests/test_pca.py
yuejiaointel: Added `test_transform_output_pandas_polars_input` (pandas/polars input, with and without target_offload).

sklearnex/decomposition/tests/test_pca.py
yuejiaointel: Noted, thanks.

sklearnex/decomposition/tests/test_pca.py
yuejiaointel: `set_output(transform="default")` pins the config and then overrides a later `config_context`, so the test breaks:
```python
>>> pca.set_output(transform="default")
>>> with config_context(transform_output="pandas"):
...     type(pca.transform(X))
<class 'numpy.ndarray'>   # expected polars/pandas DataFrame
```
Restoring the saved `_sklearn_output_config` avoids that.

sklearnex/decomposition/tests/test_pca.py
```diff
@@ -223,9 +218,13 @@ def _convert(arr, xp, device):
 
 # (xp, device) array-API input combinations, CPU and GPU; device-specific entries
 # are dropped at collection time when the hardware/library is unavailable.
+# array_api_strict is intentionally omitted: PCA rebuilds its oneDAL model from the
+# fitted ``components_``, and array_api_strict arrays are readonly, so the
```
david-cortes-intel: This is not the case. There was a bug in older numpy that might have raised a similar error, but it shouldn't show up in newer numpy. Do you see this error in the sklearn1.8-1.9 jobs that use the latest numpy?

yuejiaointel: Good catch, added back

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -169,74 +178,164 @@ def test_dense_vs_sparse(queue, init, algorithm, dims):
     )
 
 
-@pytest.mark.skipif(
-    not sklearn_check_version("1.5"),
-    reason="Functionality introduced in later sklearn versions.",
+def _convert(arr, xp, device):
```
david-cortes-intel: CC @icfaust @ethanglaser This kind of thing can be removed after dropping sklearn1.0.

david-cortes-intel:
@yuejiaointel It appears some conversation threads got deleted. Errors with read-only array conversions are not supposed to happen on newer versions of numpy, regardless of operating system. If they do, it's likely to be an error on the onedal data conversion side.

Do you have a log showing that sort of error on windows with the latest numpy?

sklearnex/decomposition/tests/test_pca.py
yuejiaointel: Yes — WindowsCondaRecipe Python3.12, **numpy 2.5.0 + array-api-strict 2.5** (latest), still fails on 4 PCA cases:
```
onedal/decomposition/pca.py:64: in _create_model
    m.eigenvectors = to_table(self.components_, queue=queue)
onedal/datatypes/_data_conversion.py:33: in _convert_one_to_table
    return backend.to_table(... arg, queue)   # arg = readonly array_api_strict components_
array_api_strict/_array_object.py:632: in __dlpack__
    return self._array.__dlpack__(stream=stream)
E   BufferError: Cannot export readonly array since signalling readonly is unsupported by DLPack.
```
Linux passes (same numpy 2.5.0). As you say, this points at the oneDAL conversion side — `backend.to_table` calls `__dlpack__(stream=...)` on the legacy path (no `max_version`), which can't signal readonly; the newer DLPack protocol supports it. Only Windows hits it. For this PR I'll drop array_api_strict from the PCA test (numpy covers the host path, KMeans covers array_api_strict) and file the oneDAL conversion issue separately — unless you'd prefer to keep it and fix the converter here.

yuejiaointel:
> @yuejiaointel It appears some conversation threads got deleted. Errors with read-only array conversions are not supposed to happen on newer versions of numpy, regardless of operating system. If they do, it's likely to be an error on the onedal data conversion side.
> 
> Do you have a log showing that sort of error on windows with the latest numpy?

added log in the reply above

sklearnex/decomposition/tests/test_pca.py
david-cortes-intel: @yuejiaointel That kind of problem should be solved here. Please investigate what happens with code like this which is supposed to work around this error:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/9c37e177f05f065c76886ad5ad561da862d18eb7/onedal/datatypes/dlpack/data_conversion.cpp#L123

sklearnex/decomposition/tests/test_pca.py
david-cortes-intel: @yuejiaointel Support for sklearn1.0 was removed by now, please remove these.

david-cortes-intel:
@yuejiaointel It appears some messages might have gotten lost in this PR.

Regarding the read-only buffer errors: let's leave those for a different PR, but add a very clear TODO comment to remove those conditional skips here.

yuejiaointel:
> @yuejiaointel It appears some messages might have gotten lost in this PR.
> 
> Regarding the read-only buffer errors: let's leave those for a different PR, but add a very clear TODO comment to remove those conditional skips here.

added some comments, will do it a follow up pr to fix the read only error due to windows env numpy version < 2.1

sklearnex/_device_offload.py
```diff
@@ -192,52 +192,41 @@ def wrap_output_data(func: Callable) -> Callable:
...
+
+        if not args and not kwargs:
+            return result
+
+        data = (*args, *kwargs.values())[0]
+
+        if usm_iface := getattr(data, "__sycl_usm_array_interface__", None):
+            return copy_to_dpnp(usm_iface["syclobj"], result)
+
+        if hasattr(data, "dtype"):
```
icfaust, line 222: Could you add some code comments as to what this is doing to help maintenance in the future? With a walrus operator you may be able to collapse 222-224 into a single line.

I'm thinking that the line 222 is redundant from what I am reading here, but I am not sure. The real check is 224, where 223 should work regardless of input.

Let me know.

sklearnex/_device_offload.py
```diff
@@ -192,52 +192,41 @@ def wrap_output_data(func: Callable) -> Callable:
     @wraps(func)
     def wrapper(self, *args, **kwargs) -> Any:
         result = func(self, *args, **kwargs)
-        # In case ARRAY API is enabled the result is already converted to the required type
+
+        # When transform_output is polars/pandas, sklearn's _set_output wrapper
```
icfaust, line 196: I love this comment. Thanks for adding it.

Is the comment for the implicit conversion coming specifically from dpnp? And if so, what does torch xpu put out in that case?

sklearnex/decomposition/tests/test_pca.py
```diff
@@ -191,3 +209,113 @@ def test_pca_error_on_incompatible_devices(with_array_api):
...
+    + ([(array_api_strict, None)] if _numpy_supports_readonly_dlpack else [])
+    + ([(dpnp, "cpu")] if dpnp_available and _dpc_backend is not None else [])
+    + (
+        [(dpnp, "gpu")]
+        if dpnp_available and _dpc_backend is not None and is_sycl_device_available("gpu")
+        else []
+    )
+    + ([(torch, "cpu")] if torch_available else [])
+    + ([(torch, "xpu")] if torch_xpu_available else [])
+)
```
icfaust: I'd honestly put this in `_dataframes_suppport.py`, I can see this being used elsewhere. (sorry, I know its a bit of a nitpick).

icfaust (commented):
Some code could be centralized for testing, some changes in the logic may simplify things, and some `dict` accesses can be simplified due to the drop of sklearn 1.0 in #3259

icfaust:
Thanks for submitting this @yuejiaointel, its good work.

sklearnex/decomposition/tests/test_pca.py
david-cortes-intel: @icfaust This pattern is present throughout all the tests for estimators with array API. Perhaps you could refactor it in a separate PR.

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -399,3 +288,133 @@ def test_sparse_predict_on_dense_fit(array_api):
         np.testing.assert_allclose(sp_pred, dense_pred)
         np.testing.assert_allclose(sp_transf, dense_transf, rtol=1e-4, atol=1e-4)
         np.testing.assert_allclose(sp_score, dense_score, rtol=1e-6, atol=1e-6)
+
+
+def _convert(arr, xp, device):
```
david-cortes-intel: When pytest collects tests, global variables like these will not be limited to the scope of the file, so please set a name that would not potentially conflict with other tests.

david-cortes-intel:
@yuejiaointel Please add back the test that was failing under numpy2.5 on windows due to issues with read-only arrays in DLpack, with a skip and a TODO comment to remove the skip once the issue is solved in skelarnex.

The issues numpy<2.1 are not on our side to solve, only the ones with numpy>=2.5.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Done — `_assert_transform_output_matches_default` checks both the `transform_output` config and per-estimator `set_output`.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Removed with the sklearn 1.0 purge merge — only the sklearn>=1.9 guard remains, which is still needed.

sklearnex/cluster/tests/test_kmeans.py
yuejiaointel: Done — renamed the module-level inputs/helpers with `_KMEANS_`/`_PCA_` prefixes so they cannot collide during collection.

sklearnex/_device_offload.py
yuejiaointel: Seems like older code fixed after merging in the 1.0 drop

sklearnex/_device_offload.py
yuejiaointel: Both hit it: on a non-CPU device pandas raises "Implicit conversion to a NumPy array is not allowed" and polars raises "unsupported type" (dpnp and torch xpu alike). The host transfer covers both.

sklearnex/decomposition/tests/test_pca.py
yuejiaointel: Agreed it belongs in a shared location — will centralize it as a follow-up PR, since the same pattern is used across the array-API estimator tests.

yuejiaointel:
Added back — array_api_strict runs again, skipped only for numpy < 2.5 with a TODO to remove once fixed.

david-cortes-intel:
> Added back — array_api_strict runs again, skipped only for numpy < 2.5 with a TODO to remove once fixed.

What about the windows job with numpy2.5 that was failing due to read-only arrays here:
https://github.com/uxlfoundation/scikit-learn-intelex/pull/3248#discussion_r3514457596

yuejiaointel:
Bisected locally across numpy 1.26→2.5.1: array_api_strict fails on <=2.2.4 (read-only `components_`) and passes on >=2.2.5. The original Windows failure was numpy 1.26.4 in the test env (the 2.5.0 in that log was the build env). Skip set to numpy<2.2.5; CI is green.

david-cortes-intel:
@yuejiaointel Please add similar tests for IncrementalPCA which is under preview:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/preview/decomposition/tests/test_incremental_pca.py

yuejiaointel:
> @yuejiaointel Please add similar tests for IncrementalPCA which is under preview: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/preview/decomposition/tests/test_incremental_pca.py

thx for suggestion! added

sklearnex/cluster/tests/test_kmeans.py
```diff
@@ -273,3 +288,138 @@ def test_sparse_predict_on_dense_fit(array_api):
...
+        [(dpnp, "gpu")]
+        if dpnp_available and _dpc_backend is not None and is_sycl_device_available("gpu")
+        else []
+    )
+    + ([(torch, "cpu")] if torch_available else [])
+    + ([(torch, "xpu")] if torch_xpu_available else [])
+)
+
+
+def _assert_transform_output_matches_default(km, X, transform_output, method):
```
david-cortes-intel: There's multiple functions with this same name. Perhaps they could be put in a centralized place, otherwise they should have unique names to avoid conflicts.

yuejiaointel: Done — moved it to `sklearnex.tests.utils.assert_transform_output_matches_default`; all three tests import the one shared helper now.

sklearnex/_device_offload.py
```diff
@@ -192,52 +192,41 @@ def wrap_output_data(func: Callable) -> Callable:
...
-                            result = tuple(xp.asarray(r, device=device) for r in result)
-                        elif not isinstance(result, (int, float)):
-                            result = xp.asarray(result, device=device)
+
+        if not args and not kwargs:
+            return result
+
+        data = (*args, *kwargs.values())[0]
+
+        if usm_iface := getattr(data, "__sycl_usm_array_interface__", None):
```
david-cortes-intel, line 219: @ethanglaser Would this need to be undone in your other PR?

david-cortes-intel (approved):
Internal CI is not working correctly, but it doesn't appear to be caused by these changes.

ethanglaser:
> Internal CI is not working correctly, but it doesn't appear to be caused by these changes.

Need to rebase

sklearnex/_device_offload.py
ethanglaser: This path would be removed when combined with #3322 but can be kept until they're merged

Issue #3220, opened 2026-05-29
LogisticRegression GPU fallback ignores allow_sklearn_after_onedal=False


cakedev0:
**Describe the bug**

`sklearnex.config_context(target_offload="gpu", allow_fallback_to_host=False, allow_sklearn_after_onedal=False)` still allows `sklearnex.linear_model.LogisticRegression` to fall back to stock scikit-learn on host for a GPU-targeted run.

The operation completes successfully and only logs:

```text
fallback to original Scikit-learn. All input data transferred to host for further backend computations.
```

This looks inconsistent with the config docs:

- `allow_fallback_to_host=False` should disallow GPU to host fallback.
- `allow_sklearn_after_onedal=False` should disallow fallback to stock scikit-learn when no accelerated version is available.

I also checked `main`, the dispatch path in `sklearnex/_device_offload.py` still appears to call the sklearn branch after writing the fallback log without checking `allow_sklearn_after_onedal` in that path.

**To Reproduce**

```python
from sklearn.datasets import make_classification
from sklearn.model_selection import train_test_split
from sklearnex import config_context
from sklearnex.linear_model import LogisticRegression

X, y = make_classification(
    n_samples=3000,
    n_features=20,
    n_informative=10,
    n_classes=2,
    random_state=42,
)
X_train, X_test, y_train, y_test = train_test_split(
    X.astype("float32"), y, test_size=0.2, random_state=0
)

with config_context(
    target_offload="gpu",
    allow_fallback_to_host=False,
    allow_sklearn_after_onedal=False,
):
    clf = LogisticRegression()
    clf.fit(X_train, y_train)
    clf.predict(X_test)
```

**Observed behavior**

The script completes without an exception. With sklearnex logging displayed, it reports fallback to stock scikit-learn on host:

```text
INFO:sklearnex: sklearn.linear_model.LogisticRegression.fit: fallback to original Scikit-learn. All input data transferred to host for further backend computations.
INFO:sklearnex: sklearn.linear_model.LogisticRegression.predict: fallback to original Scikit-learn. All input data transferred to host for further backend computations.
INFO:sklearnex: sklearn.linear_model.LogisticRegression.decision_function: fallback to original Scikit-learn. All input data transferred to host for further backend computations.
```

**Expected behavior**

With both fallback controls set to `False`, this should fail instead of silently running stock scikit-learn on host. 
**Environment:**

- OS: Linux
- Python: 3.12
[...]

PR #3262, opened 2026-06-11
[FIX] Correct usage of the `allow_fallback_to_host` configuration option

icfaust:
## Description

Address issues brought up in #3220 and by @david-cortes-intel where newer refactor of `sklearnex._device_offload.dispatcher` occurring in the array API support commit (#2096) disabled an error upon falling back to scikit-learn when using GPU data. It was the case that this option was broken since #2168, but was untested and the new tests introduced in #2096 did not encompass the fallback to sklearn. This did not reflect the initial implementation which occurred in #772.  This PR adds an additional check and expands test to cover this scenario, it only occurs after all oneDAL checks have been exhausted so it has no performance impact on oneDAL acceleration.  This should only happen in the case that `array_api_dispatch` is False, and `allow_fallback_to_host` is False.  In the case that `array_api_dispatch` is `True` and `allow_fallback_to_host` is `True`, it will prioritize oneDAL `cpu` acceleration over sklearn's array API support.

Additionally, new functions are added to the sycl_queue_manager which should simplify device checks.

NOTE: This PR will stay in draft as it should wreak havoc on private CI sklearn conformance testing, and changes for that are oncoming.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/.circleci/run_xpu_tests.py
+++ b/.circleci/run_xpu_tests.py
@@ -95,7 +95,7 @@ if __name__ == "__main__":
     with config_context(allow_sklearn_after_onedal=False):
 
         if args.device == "gpu":
-            with config_context(target_offload=args.device, allow_fallback_to_host=False):
+            with config_context(target_offload=args.device, allow_fallback_to_host=True):
                 pytest.main(
                     pytest_params + ["--pyargs", "sklearn"] + yml_deselected_tests
                 )
--- a/onedal/utils/_sycl_queue_manager.py
+++ b/onedal/utils/_sycl_queue_manager.py
@@ -83,6 +83,9 @@ def remove_global_queue():
     __globals.queue = None
 
 
+is_cpu_device = lambda: (get_global_queue() := queue) is None or getattr(queue.sycl_device, "is_cpu", True)
+is_gpu_device = lambda: (get_global_queue() := queue) is not None and getattr(queue.sycl_device, "is_gpu", False)
+
 def update_global_queue(queue):
     """Update the global queue.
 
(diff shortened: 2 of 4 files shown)


ethanglaser:
> In the case that array_api_dispatch is True and allow_fallback_to_host is True, it will prioritize oneDAL cpu acceleration over sklearn's array API support.

So it would rather fallback to oneDAL CPU than sklearn GPU?

ethanglaser:
To me it does not make sense to run GPU conformance tests with `allow_sklearn_after_onedal` set to False and `allow_fallback_to_host` to True - neither of these are the defaults

icfaust:
@ethanglaser you are making a good point. That would make a difference in those algos in sklearn which contain array API support like ridge and PCA. `allow_sklearn_after_onedal` only impacts linear regression in the case of a failure, so its a bit misleading too. Maybe we need to come together again and figure out what is the best way forward on this.

cc @david-cortes-intel

david-cortes-intel:
> > In the case that array_api_dispatch is True and allow_fallback_to_host is True, it will prioritize oneDAL cpu acceleration over sklearn's array API support.
> 
> So it would rather fallback to oneDAL CPU than sklearn GPU?



> To me it does not make sense to run GPU conformance tests with `allow_sklearn_after_onedal` set to False and `allow_fallback_to_host` to True - neither of these are the defaults

In the tests, it will not always be the case that the operation will be supported on GPU on oneDAL or under arrayAPI in sklearn, and it would be quite a lot of effort to narrow down the tests to only things that can run on GPU and then update them as patching changes.

I'd say `allow_fallback_to_host=True` makes sense for conformance tests.

PR #3271, opened 2026-06-12
[maintenance] Drop use of `daal_check_version` in `onedal` module

icfaust:
## Description

This is necessary for removing sklearn dependency from the `onedal` module, which we currently rely on `daal4py` for (which imports sklearn). This will standardize use around python version checking for simplicity and reuse.

Just as a note: using a tuple comparison in Python is faster than lru_cache and faster than converting data to C++ for a similar comparison there (either by string etc.). We also simplify the interface to be different than `daal_check_version` which is easier to understand and use.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/onedal/__init__.py
+++ b/onedal/__init__.py
@@ -200,7 +200,7 @@ def onedal_check_version(
         Minor version to compare against oneDAL's LibraryVersionInfo.minorVersion.
     update : int
         Update version to compare against oneDAL's LibraryVersionInfo.updateVersion.
-    _v: tuple(int, int, int), default=_default_backend.__version_tuple__
+    _v : tuple(int, int, int), default=_default_backend.__version_tuple__
         Version tuple defined by oneDAL (major, minor, update) to compare against.
 
     Returns
--- a/onedal/__init__.py
+++ b/onedal/__init__.py
@@ -189,7 +189,27 @@ def _ensure_dpc_available(require_spmd: bool = False) -> None:
     )
 
 
-# Core modules to export
+def onedal_check_version(major: int, minor: int, update: int, _v=_default_backend.__version_tuple__) -> bool:
+    """Check if runtime onedal library version is greater than or equal to a required version.
+    
+    Parameters
+    ----------
+    major : int
+        major version to compare against oneDAL's LibraryVersionInfo.majorVersion
+    minor : int
+        minor version to compare against oneDAL's LibraryVersionInfo.minorVersion
+    update : int
+        update version to compare against oneDAL's LibraryVersionInfo.updateVersion
+
+    Returns
+    -------
+        bool
+            runtime oneDAL version is greater than or equal to the given required oneDAL version
+    """
+    return _v >= (major, minor, update)
+
+
+# Core modules and functions to export
 __all__ = [
     "_ensure_dpc_available",
     "_host_backend",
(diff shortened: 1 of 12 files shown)


david-cortes-intel:
@icfaust Wouldn't this hard-code the version it sees at compile time? What happens if then the version changes at runtime?

icfaust:
> @icfaust Wouldn't this hard-code the version it sees at compile time? What happens if then the version changes at runtime?

I assume you (@david-cortes-intel ) mean that someone injects a different set of onedal shared objects which are different than those that were used to compile the pybind11 interface. As of now neither the daal4py or onedal extract any info from the lib/so files about version, but instead reference a header. If we want to be careful for this circumstance, we need to make changes in onedal for it first (so far as I know onedal does not expose any of this info via a formal function in the symbol table - @Vika-F @Alexandr-Solovev ). Its not a guaranteed guard for binary compatibility as it is now implemented. This does go down a slippery slope of arguments about backwards compatibility of oneDAL versions, which we really shouldn't be doing but was done to allow for conda testing with the dal package (in my opinion).

i wanted it to be compile time since a header dependency would suggest doing that. let me know if i am misunderstanding anything.

david-cortes-intel:
Looks like there is a utility to get the runtime version, but it's not used (current logic for runtime checks is also to take version from headers), and not sure if it works as intended:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/967dd7f3473dced7dd6edf264ff797352e849f0a/generator/wrapper_gen.py#L1335

icfaust:
I'm glad you pushed on this, there is a DAAL_EXPORT'ed class which can be queried, which should be the true way of doing it. However, if the library is swapped to something with a different ABI, it could just fail from missing symbols etc. (since we control breakage, but not addition to the ABI). I will add an import error if the linked oneDAL is less than the one used for compilation.

icfaust:
Note to reviewers, I used claude to make the requisite swaps to the various files after the definition (out of laziness):

```
All daal_check_version calls in the onedal/ module have been migrated to onedal_check_version.

  Files changed:
  - onedal/ensemble/forest.py:21,141 — import + (2023,"P",101) → (2023, 1, 1)
  - onedal/cluster/__init__.py:17,24 — import + (2023,"P",200) → (2023, 2, 0)
  - onedal/cluster/kmeans.py:22,28,215 — import + two (2023,"P",200) → (2023, 2, 0)
  - onedal/cluster/kmeans_init.py:20,28 — import + (2023,"P",200) → (2023, 2, 0)
  - onedal/linear_model/linear_model.py:21,59 — import + (2024,"P",600) → (2024, 6, 0)
  - onedal/linear_model/logistic_regression.py:21,83 — import + (2024,"P",300) → (2024, 3, 0)
  - onedal/covariance/covariance.py:20,44,46,111 — import + (2024,"P",1) → (2024, 0, 1), (2024,"P",400) → (2024, 4, 0)
  - onedal/covariance/incremental_covariance.py:19,139 — import + (2024,"P",1) → (2024, 0, 1)
  - onedal/common/hyperparameters.py:21,23,103,107 — import + (2024,"P",0) → (2024, 0, 0), (2024,"P",300) → (2024, 3, 0), (2025,"P",700) → (2025, 7, 0)

  Version-tuple conversion uses the MMUU split you already applied in __init__.py (e.g., (2023,"P",200) → minor=2, update=0). One judgment call:
  (2024,"P",600) in linear_model.py maps to (2024, 6, 0) — I read the third field as a 4-digit MMPP, so 0600 = minor 6, update 0. Verify that matches your
  intent (versus (2024, 0, 600) or similar) before merging.
```

david-cortes-intel:
@icfaust Please update the documentation:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/674852c023ca8657b70ac322c4c780b023a7f193/doc/sources/contributor-reference.rst?plain=1#L66

david-cortes-intel (approved):

onedal/__init__.py
```diff
@@ -189,7 +187,31 @@ def _ensure_dpc_available(require_spmd: bool = False) -> None:
...
+    minor : int
+        Minor version to compare against oneDAL's LibraryVersionInfo.minorVersion.
+    update : int
+        Update version to compare against oneDAL's LibraryVersionInfo.updateVersion.
+    _v : tuple(int, int, int), default=_default_backend.__version_tuple__
+        Version tuple defined by oneDAL (major, minor, update) to compare against.
+
+    Returns
+    -------
+        bool
```
david-cortes-intel, line 215: Missing a name.

icfaust: https://numpydoc.readthedocs.io/en/latest/format.html#returns name is optional on return, especially if its anonymous. I'd just name it `result` but I think that wouldn't be informative either...

david-cortes-intel (approved):
CI failures do not look related to the changes here.

david-cortes-intel:
ref https://github.com/uxlfoundation/scikit-learn-intelex/issues/3377

PR #3275, opened 2026-06-17
MAINT: Remove logic of falling back to scikit-learn after linear model fit failures

david-cortes-intel:
## Description

There is some logic in some linear models to fall back to scikit-learn if oneDAL fitting procedures fail (introduced in https://github.com/uxlfoundation/scikit-learn-intelex/pull/1907), but:
* In the case of linear regression, both the CPU and GPU versions by now have C++ fallbacks to try more robust methods when the first one fails (e.g. data with linear dependencies).
* When the input is originally on GPU, it would trigger a fallback to CPU with scikit-learn, which is usually not desirable. Better would be to let it fail.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/linear_model/linear.py
+++ b/sklearnex/linear_model/linear.py
@@ -290,30 +290,16 @@ class LinearRegression(oneDALEstimator, _sklearn_LinearRegression, _BaseLinearMo
 
         self._initialize_onedal_estimator()
 
-        try:
-            self._onedal_estimator.fit(X, y, queue=queue)
+        self._onedal_estimator.fit(X, y, queue=queue)
 
-            self.n_features_in_ = self._onedal_estimator.n_features_in_
-            self._sparse = False
-            self._coef_ = self._onedal_estimator.coef_
-            self._intercept_ = self._onedal_estimator.intercept_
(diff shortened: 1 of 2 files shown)


sklearnex/linear_model/linear.py
```diff
@@ -290,30 +290,16 @@ def _onedal_fit(self, X, y, sample_weight, queue=None):
 
         self._initialize_onedal_estimator()
 
-        try:
-            self._onedal_estimator.fit(X, y, queue=queue)
+        self._onedal_estimator.fit(X, y, queue=queue)
```
avolkov-intel, line 293: Should we still try catch it to provide meaningful python error message? Note that it might be useful to keep oneDAL error message as well

david-cortes-intel: Wouldn't the error message from oneDAL already come as a python exception?

icfaust: agreed. ideally we should mention the PRs which brought this feature in for reference in the description

david-cortes-intel: Added a link to the PR that introduced this mechanism.

icfaust (approved):

PR #3312, opened 2026-07-08
MAINT, DOC: Document and simplify solver parameter for IncrementalPCA

david-cortes-intel:
## Description

IncrementalPCA in sklearnex allows changing the solver, whereas the sklearn equivalent doesn't.

This PR updates the documentation to reflect this additional parameter, and renames the solvers to be in line with sklearn's, similar to https://github.com/uxlfoundation/scikit-learn-intelex/pull/3276 which does the same for PCA.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/doc/sources/additional_parameters.rst
+++ b/doc/sources/additional_parameters.rst
@@ -68,7 +68,7 @@ It can take the following values:
 - ``"covariance_eigh"`` (default).
 - ``"full"``.
 
-Note that under the ``"full"`` solver, further calls to :~meth:`sklearn.decomposition.IncrementalPCA.partial_fit` will make the object's stored memory grow with the data.
+Note that under the ``"full"`` solver, further calls to :meth:`~sklearn.decomposition.IncrementalPCA.partial_fit` will make the object's stored memory grow with the data.
 
 Train-test splitting
 ====================
--- a/doc/sources/additional_parameters.rst
+++ b/doc/sources/additional_parameters.rst
@@ -58,6 +58,18 @@ This parameter is available in the following classes:
 - :obj:`sklearn.ensemble.ExtraTreesRegressor`
 - :obj:`sklearn.ensemble.ExtraTreesClassifier`
 
+IncrementalPCA
+==============
+
+:obj:`sklearn.decomposition.IncrementalPCA` accelerated with the |sklearnex| (currently under preview mode - see :doc:`preview` for more information) allows an additional argument ``svd_solver`` akin to the same argument that is accepted by :obj:`sklearn.decomposition.PCA`.
+
+It can take the following values:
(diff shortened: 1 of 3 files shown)


david-cortes-intel:
In the failing jobs, it's somehow trying to use the DPC backend where it's not built. Oddly, it fails only in sklearn conformance testing, not in sklearnex, and the failures are in multiple places unrelated to this PR.

Not sure why that happens with this PR.

doc/sources/algorithms.rst
```diff
@@ -552,7 +552,7 @@ Dimensionality Reduction
    * - :obj:`sklearn.decomposition.IncrementalPCA`
      - All parameters are supported except:
 
-       - ``svd_solver`` not in [`'full'`, `'covariance_eigh'`]
+       - ``svd_solver`` != ``'full'``
```
ethanglaser: Would it really be considered an exception if sklearn does not support this? I almost feel like its not necessary to mention svd_solver here - and maybe reference additional_parameters.rst instead

icfaust: Good catch. Something is a bit off here in the docs originally as this isn't a supported parameter. We should note to the users that a special parameter is available instead to change the svd solver.

david-cortes-intel: Fixed.

ethanglaser: What about further up in this page? Looks like the same change may be needed there as well

david-cortes-intel: Fixed.

doc/sources/additional_parameters.rst
```diff
@@ -58,6 +58,18 @@ This parameter is available in the following classes:
...
+==============
+
+:obj:`sklearn.decomposition.IncrementalPCA` accelerated with the |sklearnex| (currently under preview mode - see :doc:`preview` for more information) allows an additional argument ``svd_solver`` akin to the same argument that is accepted by :obj:`sklearn.decomposition.PCA`.
+
+It can take the following values:
+
+- ``"covariance_eigh"`` (default).
+- ``"full"``.
+
+Note that under the ``"full"`` solver, further calls to :~meth:`sklearn.decomposition.IncrementalPCA.partial_fit` will make the object's stored memory grow with the data.
```
ethanglaser: Is there confirmation that this ~meth rendered as intended? I don't see any other usage in the codebase so just wondering

david-cortes-intel: Fixed.

ethanglaser (approved):
LGTM. Perhaps @Alexandr-Solovev should confirm this aligns with expectations.

PR #3323, opened 2026-07-14
DOC: Add reminder to inherit sklearn docstrings of methods

david-cortes-intel:
## Description

Adds a small comment in the instructions to inherit the docstrings from sklearn. This detail is already mentioned at the end of the same file, but somehow it got missed by LLMs when crafting the PR for MaxAbsScaler. Perhaps having it more clearly expressed in the instructions might help.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/dummy/_dummy.py
+++ b/sklearnex/dummy/_dummy.py
@@ -89,6 +89,11 @@ from ..utils.validation import validate_data
 # parameters must extend the dictionary for checking.  This validation
 # normally occurs in the ``fit`` method.
 #
+# 11) All methods that are overridden from scikit-learn's base (e.g. 'fit')
+# or otherwise passed through some decorator must take the method docstrings
+# from scikit-learn's class, adding documentation for additional arguments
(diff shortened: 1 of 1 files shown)


sklearnex/dummy/_dummy.py
```diff
@@ -89,6 +89,11 @@
 # parameters must extend the dictionary for checking.  This validation
 # normally occurs in the ``fit`` method.
 #
+# 11) All methods that are overridden from scikit-learn's base (e.g. 'fit')
```
icfaust, line 92: See below (EDIT farther down in the `_dummy.py` file):

```python

    # All public methods (i.e. without leading underscores) should have
    # documentation which conforms to the numpy-doc standard.  Generally
    # if a defined method replaces an inherited Scikit-Learn estimator
    # method, the ``__doc__`` attribute should be re-applied to the new
    # implementation. Any new additional characteristics compared to the
    # equivalent sklearn estimator should be appended to the sklearn doc
    # string.
```

Should adapt it there (if necessary).

PR #3328, opened 2026-07-19
[docs] add comments to explain github nightly CI for improved maintainability

icfaust:
## Description

I ran claude, then I had claude run adversarially for first corrections, and then I went through and corrected/added what info was missing. That means I have reviewed and understood the comments.  I have also removed odd LLM characters which I detest.

This is in order to help others maintain this code. It also highlights how proper coding agents can get 90% of the necessary actionable info for developers.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,32 +19,39 @@
 # ------------------------------------------------------------------------------
 # This workflow builds and tests scikit-learn-intelex against the *nightly*
 # oneDAL binaries produced by the uxlfoundation/oneDAL repository, rather than
-# building oneDAL from source here. The high-level flow is:
+# building oneDAL from source in order to save time while enabling access to
+# recent oneDAL development. The high-level flow is:
 #
-#   onedal_nightly ──▶ sklearn_lnx   (matrix over Python x sklearn on Linux)
-#                  ├─▶ sklearn_win   (matrix over Python x sklearn on Windows)
-#                  └─▶ build_uxl ──▶ test_uxl   (UXL Foundation dedicated flow)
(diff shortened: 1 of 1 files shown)


ethanglaser (changes requested):
IMO, since these comments are intended for human consumption this is one instance where claude should not be used. To me the most confusing part is how the oneDAL nightly builds are grabbed, so https://github.com/uxlfoundation/oneDAL/actions/workflows/nightly-build.yml should be referenced in a comment. Outside of that just illustrating the overall workflow with a few concise comments should do the trick - try to think back about how you designed this originally. Just my 2 cents.

icfaust:
> IMO, since these comments are intended for human consumption this is one instance where claude should not be used. To me the most confusing part is how the oneDAL nightly builds are grabbed, so https://github.com/uxlfoundation/oneDAL/actions/workflows/nightly-build.yml should be referenced in a comment. Outside of that just illustrating the overall workflow with a few concise comments should do the trick - try to think back about how you designed this originally. Just my 2 cents.

Gotcha, thanks for the clarification as I wasn't sure what you all wanted. Doing that will cost more time, I will need to reconsider when I can deliver that. The problem is that I need to lay bare some underlying info about how github actions work and the consequences. While this can be divined from github's docs, bring it together in a concise manner is not trivial. Unfortunately this whole machination was a constraint by management that the build needed to occur in oneDAL and not here.

Should we also document some of the more difficult pitfalls of our public CI's (as this one is unfortunately a simpler case), like the reuse of the azure pipelines from sklearnex in oneDAL, how I do code coverage generation, or how and why I do recall of builds for ABI checking in github on oneDAL side? 

Let me know.

PR #3322, opened 2026-07-13
Remove non-array api dpnp support

ethanglaser:
## Description

Removes direct support for non-array API dpnp support. Migrates all examples and spmd tests to set `array_api_dispatch=True` for fully on-device handling without conversion to host. Using dpnp inputs without setting `array_api_dispatch` now returns numpy arrays.

---

<details>
<summary>Checklist:</summary>

**Completeness and readability**

- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.
- [ ] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/doc/sources/contribute.rst) for details)_.
- [ ] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [ ] I have run it locally and tested the changes extensively.
- [ ] All CI jobs are green or I have provided justification why they aren't.
- [ ] I have extended testing suite if new functionality was introduced in this PR.

**Performance**

- [ ] I have measured performance for affected algorithms using [scikit-learn_bench](https://github.com/IntelPython/scikit-learn_bench) and provided at least a summary table with measured data, if performance change is expected.
- [ ] I have provided justification why performance and/or quality metrics have changed or why changes are not expected.
- [ ] I have extended the benchmarking suite and provided a corresponding scikit-learn_bench PR if new measurable functionality was introduced in this PR.

</details>

PR #3330, opened 2026-07-20
isolate dpnp sua offset fix

ethanglaser:
## Description

Isolated from https://github.com/uxlfoundation/scikit-learn-intelex/pull/3322 where it was initially discovered. Basically any dpnp non-host offloaded array slicing operation would unsuccessfully apply an offset (ie X_dpnp_device[5:10] would actually pull X_dpnp_device[0:5]). Per Claude:

### Root cause

The C++ `convert_to_homogen_impl` in `onedal/datatypes/sycl_usm/data_conversion.cpp` built the oneDAL table from `__sycl_usm_array_interface__['data'][0]` (the base allocation pointer) but **never applied the `offset` field**. A dpnp slice view (e.g. `X[a:b]`) keeps the *same* `data[0]` base pointer and encodes its start position via `offset` (in **elements**, not bytes). Without applying it, oneDAL read from row 0 of the allocation for every view rather than from the view's actual start.

### Symptom

Under `array_api_dispatch=True`, `IncrementalEmpiricalCovariance.fit(dpnp_X)` produced a wrong `location_`/score — surfaced as `test_whitened_toy_score[dpnp-*]` returning ≈ -13.58 vs expected ≈ -14.18 (identical on CPU and GPU). `IncrementalEmpiricalCovariance.fit` batches its input via slice views (`X[batch_start:batch_end]`), so every batch after the first read the wrong rows. Single-batch fits (`batch_size ≥ n_samples`) and explicitly copied batches were unaffected.

Minimal repro, independent of covariance: a dpnp `X[5:10]` view round-tripped through `onedal.datatypes.to_table`/`from_table` returns row-0 data instead of row-5.

### Why it wasn't caught before

On `main`, dpnp inputs without `array_api_dispatch` are routed through `_transfer_to_host` (→ `asnumpy`) before reaching the SUA→table path, so slice views never hit this conversion. This is a **latent backend bug**; #3322 (removing non-array-API dpnp support) only *exposes* it by passing dpnp slice views directly to `to_table`.

### Fix

Added `get_sua_offset()` in `sycl_usm_utils.cpp/.hpp` (reads the optional `offset`, defaults to 0) and applied it: `ptr = get_sua_ptr(...) + offset`.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>


icfaust (approved):
@ethanglaser quick question, wouldn't this have thrown off all `use_raw_input` results?

ethanglaser:
> @ethanglaser quick question, wouldn't this have thrown off all `use_raw_input` results?

Yes. The reason it was not caught is there are minimal slicing operations in the spmd test suite. But any slicing operation in these device-only dpnp arrays would not account for the indicated offset.

ethanglaser:
Pretty big bug and equally large test gap

onedal/datatypes/tests/test_data.py
```diff
@@ -228,6 +229,32 @@ def test_input_zero_copy_sycl_usm(dataframe, queue, order, dtype):
...
+def test_sliced_view_offset_sycl_usm(dataframe, queue, start, dtype):
+    """A sliced view (e.g. X[start:]) keeps the base allocation pointer in
+    `__sycl_usm_array_interface__['data'][0]` and encodes its start via the
+    `offset` field. The table conversion must apply that offset, otherwise the
+    view reads from row 0 (regression test for the SUA offset bug).
+    """
+    X_np = np.arange(50, dtype=dtype).reshape(10, 5)
+    X_dp = _convert_to_dataframe(X_np, sycl_queue=queue, target_df=dataframe)
+
+    X_view = X_dp[start:]
```
david-cortes-intel: Could also test other types - for example, non-unit slices, selecting a subset of columns, etc.

ethanglaser: Added more cases

david-cortes-intel (approved):

PR #3003, opened 2026-03-06
Support torch validation and add to spmd tests

ethanglaser:
## Description

Enables torch tensor validation alongside dpnp arrays in spmd test scope, as well as any tests using default `ONEDAL_PYTEST_FRAMEWORKS` scope, with support added in _dataframes_support.py. Also adds additional pytest fixtures, `mixed_device_params`, `pd_to_torch_working` which are outlined in contributor reference documentation and used in various mixed device tests.

---

<details>
<summary>Checklist:</summary>

**Completeness and readability**

- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.
- [ ] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/doc/sources/contribute.rst) for details)_.
- [ ] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [ ] I have run it locally and tested the changes extensively.
- [ ] All CI jobs are green or I have provided justification why they aren't.
- [ ] I have extended testing suite if new functionality was introduced in this PR.

**Performance**

- [ ] I have measured performance for affected algorithms using [scikit-learn_bench](https://github.com/IntelPython/scikit-learn_bench) and provided at least a summary table with measured data, if performance change is expected.
- [ ] I have provided justification why performance and/or quality metrics have changed or why changes are not expected.
- [ ] I have extended the benchmarking suite and provided a corresponding scikit-learn_bench PR if new measurable functionality was introduced in this PR.

</details>

PR #3332, opened 2026-07-20
force array_api_dispatch for non-numpy array API frameworks

ethanglaser:
## Description

Initial draft of adding pytest fixture to force `array_api_dispatch=True` for dpnp pytest scope. In similar scope as https://github.com/uxlfoundation/scikit-learn-intelex/pull/3322 and depends on https://github.com/uxlfoundation/scikit-learn-intelex/pull/3330

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>


sklearnex/conftest.py
```diff
@@ -59,6 +59,24 @@ def pytest_runtest_call(item):
         yield
 
 
+@pytest.fixture(autouse=True)
+def _array_api_dispatch_for_device_frameworks(request):
+    # Non-numpy array API inputs (dpnp, array_api) are only supported on-device
+    # under array_api_dispatch; without it they silently host-transfer, leaving the
+    # on-device paths untested. Enable dispatch whenever a test is parametrized with
+    # such a dataframe so these inputs exercise the real array API path. Tests marked
+    # allow_sklearn_fallback intentionally cover host paths and are left untouched.
```
david-cortes-intel: Please document this very clearly in the contributor reference.

ethanglaser: Documented

doc/sources/contributor-reference.rst
```diff
@@ -126,8 +126,14 @@ Test helpers
...
+
+  - Under dispatch, **fitted attributes and method outputs stay in the input's namespace** (a dpnp input yields dpnp results). Comparisons against a NumPy expected value must therefore convert the result first (see ``_as_numpy`` / ``_as_numpy_checked`` below) — ``numpy.testing.assert_allclose`` cannot consume a dpnp array on either side.
+  - Tests marked ``@pytest.mark.allow_sklearn_fallback`` are **left untouched** by this fixture (dispatch is not forced), since those tests intentionally exercise the host/fallback path. Use this mark for any dpnp test whose configuration is expected to fall back to |sklearn| rather than run on device.
```
icfaust: That could get dicey when the sklearn fallback supports `array_api_dispatch`, therefore not exercising logic in `sklearnex._device_offload`.

doc/sources/contributor-reference.rst
```diff
@@ -126,8 +126,14 @@ Test helpers
...
+
+- ``@pytest.mark.allow_sklearn_fallback``: will avoid having tests fail when they end up calling procedures from |sklearn| instead of from the |onedal|. This can be helpful for example when testing that some corner case falls back correctly when it should. It also disables the automatic array API dispatch described above.
 - ``onedal.tests.utils._dataframes_support._as_numpy``: this function can be used to convert an input array or data frame to NumPy, regardless of whether it lives on host or on device, and regardless of array API support.
+- ``onedal.tests.utils._dataframes_support._as_numpy_checked``: like ``_as_numpy``, but first asserts that the object belongs to the array namespace implied by the test's ``dataframe`` parameter (e.g. a ``"dpnp"`` test must produce a dpnp result) before converting. Use it on result-side values to verify that outputs stay on device (dpnp-in → dpnp-out) rather than being silently host-transferred, while still allowing the value comparison against a NumPy expected result.
```
icfaust: This is good. I haven't looked yet in detail, but this may help in some centralized patching tests?

sklearnex/basic_statistics/tests/test_incremental_basic_statistics.py
```diff
@@ -382,7 +396,9 @@ def test_fit_all_option_on_random_data(
         assert_allclose(gtr, res, atol=tol)
 
 
-@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+# dpnp/array_api excluded: fitted state stays device-bound under array_api_dispatch
```
icfaust, line 406: This may be a bug (function loss) in the current code given the first pickleable implementation (#2211), and the array API swap. Previously we would drop the queue and the store things, but we may need to make a ticket and address this.

onedal/tests/utils/_dataframes_support.py
```diff
@@ -149,6 +150,52 @@ def _as_numpy(obj, *args, **kwargs):
...
+    Estimators that rebuild a oneDAL model from fitted arrays (PCA/IncrementalPCA
+    components_, DummyRegressor constant_) route them back through ``to_table``.
+    numpy < 2.2.5 returns those arrays read-only, which ``to_table`` cannot export
+    through DLPack, so array_api_strict inputs raise a BufferError / read-only
+    assignment error under forced array_api_dispatch. numpy >= 2.2.5 returns
+    writeable arrays.
+    TODO: remove once the oneDAL data conversion handles read-only arrays.
+    """
+    if dataframe == "array_api" and not _package_check_version("2.2.5", np.__version__):
+        pytest.skip("TODO: sklearnex read-only DLPack conversion fails on numpy<2.2.5")
```
icfaust: Ideally we should also modify the logic in the function `get_dataframes_and_queues` to set the following:

```python
    if (
        "array_api" in dataframe_filter_
        and "array_api" in array_api_modules
        or array_api_enabled()
    ):
        if not _package_check_version("2.2.5", np.__version__):
            dataframes_and_queues.append(pytest.param("array_api", None, id="array_api"), marks=pytest.mark.xfail(reason="DLPack conversion fails on numpy<2.2.5"))
        else:
            dataframes_and_queues.append(pytest.param("array_api", None, id="array_api"))

```
In order to limit the necessary use of this function to only circumstances which do not use `get_dataframes_and_queues` to reduce maintenance.

sklearnex/linear_model/tests/test_linear.py
icfaust, line 1: Why is the diff so weird on this file?

sklearnex/linear_model/tests/test_logreg.py
```diff
@@ -60,7 +60,8 @@ def check_preview_is_enabled() -> bool:
 
 
 @pytest.mark.parametrize(
-    "dataframe,queue", get_dataframes_and_queues(device_filter_="cpu")
+    "dataframe,queue",
```
icfaust: This drops array api testing. Was this intentional?

sklearnex/neighbors/common.py
```diff
@@ -382,20 +383,15 @@ def _kneighbors_postprocess(
...
             except RuntimeError:
                 # Fallback for array API < 2024.12 (e.g. array-api-strict
-                # with api_version='2023.12' has the function but raises)
+                # with api_version='2023.12' has the function but raises).
+                # dlpack_to_numpy avoids np.from_dlpack(device=), whose device
+                # kwarg only exists on numpy >= 2.1.
+                seq_np = dlpack_to_numpy(seq)
                 indices = xp.from_dlpack(
-                    np.take_along_axis(
-                        np.from_dlpack(indices, device="cpu"),
```
icfaust, line 388: This looks to be a bugfix?

onedal/tests/utils/_dataframes_support.py
```diff
@@ -149,6 +150,52 @@ def _as_numpy(obj, *args, **kwargs):
         return dlpack_to_numpy(obj)
 
 
+def _assert_in_namespace(obj, dataframe):
+    """Assert obj belongs to the array namespace implied by ``dataframe``.
+
+    Under array_api_dispatch, sklearnex outputs stay in the input namespace,
+    so an on-device dpnp/array_api input should yield an on-device result
+    (dpnp-in -> dpnp-out). Scalars are namespace-agnostic and are ignored.
+    """
```
icfaust: If you are using claude, force it to do typing and a full numpydoc on the new functions. If you are doing it yourself, not worth the time.

icfaust (commented):
Don't get too annoyed by this suggestion (as it would require some refactoring and testing of the PR):

In all but a handful of cases the new function `_as_numpy_checked` is always tied to `assert_allclose`. Couldn't we just spin our own `assert_allclose` at this point which always forces things to numpy, and then add an intentional `_assert_in_namespace` in the various functions? That way we don't have to constantly bookkeep the inputs into `assert_allclose`?

Probably a borderline nitpick, but its worth thinking about how to keep this as simple for development and maintenance as possible.

sklearnex/linear_model/tests/test_linear.py
ethanglaser: Resolved. Long story short is that windows and linux based editors have different line endings (crlf vs lf) and some edits (ie if claude rewrites a file to make an edit, or idk how claude goes about making an edit) can trigger the line endings to change for the entire file, leading to diff of entire file instead of the specific lines where code changes were made.

[dpnp](https://github.com/IntelPython/dpnp/blob/master/.gitattributes), [numpy](https://github.com/numpy/numpy/blob/main/.gitattributes), etc. have their gitattributes set up to normalize this across all files by default. We don't have this currently but could add it if we care about. I drafted a PR of what the normalization would look like but its very optional: https://github.com/uxlfoundation/scikit-learn-intelex/pull/3343

onedal/tests/utils/_dataframes_support.py
ethanglaser: Done

sklearnex/linear_model/tests/test_linear.py
icfaust: amazing catch.

ethanglaser:
> Don't get too annoyed by this suggestion (as it would require some refactoring and testing of the PR):
> 
> In all but a handful of cases the new function `_as_numpy_checked` is always tied to `assert_allclose`. Couldn't we just spin our own `assert_allclose` at this point which always forces things to numpy, and then add an intentional `_assert_in_namespace` in the various functions? That way we don't have to constantly bookkeep the inputs into `assert_allclose`?
> 
> Probably a borderline nitpick, but its worth thinking about how to keep this as simple for development and maintenance as possible.

I am not a fan of all of the _as_numpy conversions, and we already have some of our own versions of assert_allclose in the spmd test utils so I don't see any reason why we shouldn't do this here. Check latest commit (https://github.com/uxlfoundation/scikit-learn-intelex/pull/3332/commits/718dc888423fb3bd4ab69e2e4f67a4f01894bf72) to see what that looks like and if there are any changes you'd want from that prototype.

sklearnex/linear_model/tests/test_logreg.py
ethanglaser: Intentional because test_sklearnex_multiclass_classification is CPU-only and logreg _onedal_cpu_supported rejects array API inputs

onedal/tests/utils/_dataframes_support.py
```diff
@@ -149,6 +152,104 @@ def _as_numpy(obj, *args, **kwargs):
...
+    if np.isscalar(obj):
+        return
+    if dataframe == "dpnp":
+        assert dpnp_available and isinstance(
+            obj, dpnp.ndarray
+        ), f"expected dpnp output, got {type(obj)}"
+    elif dataframe in array_api_modules:
+        xp = array_api_modules[dataframe]
+        assert (
+            hasattr(obj, "__array_namespace__") and obj.__array_namespace__() is xp
```
david-cortes-intel: Wouldn't this same condition also work for DPNP? And shouldn't it check device too?

sklearnex/covariance/tests/test_incremental_covariance.py
```diff
@@ -213,11 +218,14 @@ def test_whitened_toy_score(dataframe, queue):
...
         est.fit(X_df)
-        result = _as_numpy(est.score(X_df))
-    assert_allclose(expected_result, result, atol=1e-6)
+        score = est.score(X_df)
+        _assert_in_namespace(score, dataframe)
+    assert_allclose_numpy(expected_result, score, atol=1e-6)
 
 
-@pytest.mark.parametrize("dataframe,queue", get_dataframes_and_queues())
+# dpnp/array_api excluded: fitted state stays device-bound under array_api_dispatch
```
david-cortes-intel: What about array_api_strict?

sklearnex/neighbors/common.py
ethanglaser: yeah, a bug exposed by these changes

doc/sources/contributor-reference.rst
ethanglaser: Updated to reflect that assert_allclose_numpy has replaced _as_numpy_checked. but yeah I like it, and I tried to replace usage where possible throughout the repo. possibly missed some cases.

doc/sources/contributor-reference.rst
ethanglaser: This is for intentional host-fallback cases and doesn't guarantee _device_offload coverage. But updated for clarification

sklearnex/covariance/tests/test_incremental_covariance.py
ethanglaser: Updated here and in related cases

onedal/tests/utils/_dataframes_support.py
ethanglaser: Good point, I've added some logic in https://github.com/uxlfoundation/scikit-learn-intelex/pull/3332/commits/9e7fa873a790c0170bc62003a0b0b0e78744c1e6 to address - let me know if that works

onedal/tests/utils/_dataframes_support.py
```diff
@@ -167,27 +175,32 @@ def _assert_in_namespace(obj: Any, dataframe: str) -> None:
...
-        assert (
-            hasattr(obj, "__array_namespace__") and obj.__array_namespace__() is xp
-        ), f"expected {dataframe} output, got {type(obj)}"
+    xp = _expected_namespaces[dataframe]
+    assert (
+        hasattr(obj, "__array_namespace__") and obj.__array_namespace__() is xp
+    ), f"expected {dataframe} output, got {type(obj)}"
+    if device is not None:
+        # a SyclQueue is not itself a device; compare the device it wraps
+        expected = getattr(device, "sycl_device", device)
```
david-cortes-intel: I think this would still pass if passing DPNP on CPU as input and the outputs have different queues. Maybe it could use `check_same_namespace` from sklearn.

david-cortes-intel:
@ethanglaser Should this PR be merged before https://github.com/uxlfoundation/scikit-learn-intelex/pull/3003 , or the other way around?

ethanglaser:
> @ethanglaser Should this PR be merged before #3003 , or the other way around?

Good question, I think either order should be fine, and then whichever gets merged last can be updated to accommodate the other. Most important thing was getting https://github.com/uxlfoundation/scikit-learn-intelex/pull/3322 in first.

onedal/tests/utils/_dataframes_support.py
ethanglaser: Good catch, I confirmed that there were some gaps with that check, though I ended up with a different fix than check_same_namespace, see https://github.com/uxlfoundation/scikit-learn-intelex/pull/3332/commits/f994499a023c7451273192db5692b0a472bac76c

david-cortes-intel:
@ethanglaser Looks like it needs line ending renormalization.

sklearnex/conftest.py
```diff
@@ -59,6 +59,24 @@ def pytest_runtest_call(item):
...
+    dataframe = getattr(request.node, "callspec", None)
+    dataframe = dataframe.params.get("dataframe") if dataframe else None
+    if dataframe in ("dpnp", "array_api") and not request.node.get_closest_marker(
```
david-cortes-intel: I think this will need adjustment after merging the torch PR.

ethanglaser: Updated

david-cortes-intel (approved):

david-cortes-intel:
@ethanglaser Any blockers for this PR?

ethanglaser:
No, but would like to confirm with @icfaust whether previous comments were addressed properly before merge and if there are any other comments on recent changes

onedal/dummy/dummy_onedal.hpp
icfaust, line 1: Changes to dummy onedal should be a separate PR.

onedal/dummy/dummy_onedal.hpp
```diff
@@ -270,33 +287,10 @@ struct infer_ops {
...
-            }
-            case dal::data_type::int64: {
-                inp = static_cast<float_t>(*reinterpret_cast<const std::int64_t*>(ptr));
-                break;
-                default: throw std::runtime_error("incompatible input type");
-            }
-        }
+        // The accessor converts from the table's own data type, so no dispatch
+        // over 'get_metadata().get_data_type(0)' is needed here.
+        const float_t inp = read_first_element<float_t>(ctx, input.constant);
```
icfaust, line 293: OneDAL accepts int32 data, why was it dropped?

sklearnex/conftest.py
```diff
@@ -59,6 +59,29 @@ def pytest_runtest_call(item):
         yield
 
 
+@pytest.fixture(autouse=True)
+def _array_api_dispatch_for_device_frameworks(request):
+    # Non-numpy array API inputs (dpnp, array_api, torch) are only supported
+    # on-device under array_api_dispatch; without it they silently host-transfer,
+    # leaving the on-device paths untested. torch additionally cannot be restored to
+    # its own namespace on the host path, since ``Tensor`` has no
+    # ``__array_namespace__``, so undispatched torch inputs come back as numpy.
```
icfaust: Never thought about it, but true w.r.t. `support_input_format` and similar functions. Yikes, tbh it may be worth mentioning the 2-3 functions which are impacting this, since I had to think about it a bit to understand, and someone with less experience would take some time to parse this.

sklearnex/conftest.py
```diff
@@ -59,6 +59,29 @@ def pytest_runtest_call(item):
...
+    # Enable dispatch whenever a test is parametrized with such a dataframe so these
+    # inputs exercise the real array API path. Tests marked allow_sklearn_fallback
+    # intentionally cover host paths and are left untouched.
+    dataframe = getattr(request.node, "callspec", None)
+    dataframe = dataframe.params.get("dataframe") if dataframe else None
+    if dataframe in (
+        "dpnp",
+        "array_api",
+        "torch",
```
icfaust: I think this is a bit fragile tbh, a whitelist hidden in `conftest` would be a bit obscure. It really should be yanking this from `onedal.tests.utils._dataframes_support` since that tends to have all the dataframes handling there. We should be able to have a list of array API supported frameworks there that should be referenced here, ease of maintenance, centralization of purpose, etc.

sklearnex/neighbors/_lof.py
```diff
@@ -150,14 +150,10 @@ def _predict(self, X=None):
...
-            if sycl_queue is not None:
-                is_inlier = xp.ones(
-                    self.n_samples_fit_, dtype=xp.int64, sycl_queue=sycl_queue
-                )
-            else:
-                is_inlier = xp.ones(self.n_samples_fit_, dtype=xp.int64)
             mask = self.negative_outlier_factor_ < self.offset_
+            # ``ones_like`` inherits the queue/device of the fitted scores, which a
+            # bare ``ones`` drops for namespaces without a ``sycl_queue`` (e.g. torch).
+            is_inlier = xp.ones_like(mask, dtype=xp.int64)
```
icfaust, line 156: Good change. I really need to look at the LocalOutlierFactor changes w.r.t array API. Ideally should've been a separate PR.

icfaust: Just leave it to accelerate the PR's inclusion.

icfaust (commented):
This almost ready for primetime, something I want transferred to `_dataframes_support.py`, but otherwise good to go.

onedal/dummy/dummy_onedal.hpp
ethanglaser: It wasn't dropped — int32 still works, the handling just moved into oneDAL.

The switch was there to serve the raw byte-pointer read: `get_original_data(...).get_data()` returns a `byte_t*`, so you need the table's dtype to pick the right `reinterpret_cast` before dereferencing. That's what the switch was doing — it was plumbing for the pointer read, not the int32 feature itself.

`row_accessor<const float_t>` isn't a typed view that requires the table to already hold `float_t`; it's a converting accessor. Per the header: *"the new memory block is allocated, the data from the table rows is converted and copied into this block."* So pulling `float_t` from an int32 table performs exactly the `static_cast<float_t>(*reinterpret_cast<const std::int32_t*>(ptr))` that was written out by hand.

Confirmed against a local build by forcing `constant_` to each dtype:

```
int32:   OK -> [3. 3. 3.]
int64:   OK -> [3. 3. 3.]
float32: OK -> [3. 3. 3.]
float64: OK -> [3. 3. 3.]
```

One real behavior change though: the old `default: throw std::runtime_error("incompatible input type")` is gone, so an unsupported type now surfaces as whatever `pull` raises rather than that message. Happy to add an explicit guard back if you'd prefer to keep the specific error.

sklearnex/conftest.py
ethanglaser: let me know if changes in https://github.com/uxlfoundation/scikit-learn-intelex/pull/3332/commits/68ab0ab75199cdc9bd998fdb2251fded711bef71 adequately address concerns here

sklearnex/conftest.py
ethanglaser: Updated

david-cortes-intel:
@ethanglaser Please solve merge conflicts.

ethanglaser:
Merge conflicts addressed, moved dummy stuff into https://github.com/uxlfoundation/scikit-learn-intelex/pull/3396 which should be merged first

david-cortes-intel:
CI issue happens also in other PRs so unlikely to be related to the changes here.

PR #3337, opened 2026-07-23
Address coverity + cleanup pairwise logic

ethanglaser:
## Description

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/daal4py/sklearn/metrics/_pairwise.py
+++ b/daal4py/sklearn/metrics/_pairwise.py
@@ -14,33 +14,9 @@
 # limitations under the License.
 # ===============================================================================
 
-import warnings
-from functools import partial
-
 import numpy as np
-from joblib import effective_n_jobs
-from sklearn.exceptions import DataConversionWarning
 from sklearn.metrics import pairwise_distances as pairwise_distances_original
-from sklearn.metrics.pairwise import (
-    _VALID_METRICS,
-    PAIRWISE_BOOLEAN_FUNCTIONS,
-    PAIRWISE_DISTANCE_FUNCTIONS,
-    _pairwise_callable,
-    _parallel_pairwise,
-    check_pairwise_arrays,
-)
-from sklearn.utils.validation import check_non_negative
-
-try:
-    from sklearn.metrics.pairwise import _precompute_metric_params
-except ImportError:
-
-    def _precompute_metric_params(*args, **kwrds):
-        return dict()
-
-
-from scipy.sparse import issparse
-from scipy.spatial import distance
+from sklearn.metrics.pairwise import _VALID_METRICS
 from sklearn.utils._param_validation import (
     Hidden,
     Integral,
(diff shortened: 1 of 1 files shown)


daal4py/sklearn/metrics/_pairwise.py
```diff
@@ -132,54 +102,13 @@ def _pairwise_distances(
...
-        X, _ = check_pairwise_arrays(
-            X, Y, precomputed=True, ensure_all_finite=force_all_finite
-        )
-        whom = (
-            "`pairwise_distances`. Precomputed distance "
-            " need to have non-negative values."
-        )
-        check_non_negative(X, whom=whom)
-        return X
-    if metric in PAIRWISE_DISTANCE_FUNCTIONS:
```
david-cortes-intel, line 151: I think this would disable oneDAL for the cases where a user passes a supported metric from scikit-learn as a function object.

ethanglaser: It's impossible to reach anything below line 140 with the logic in this function

icfaust: Just so that I understand, wouldn't @david-cortes-intel 's concern be a gap in the current code (as on line 108 we dont go to daal when its a callable)?

david-cortes-intel: Yes, looks like it's also a problem in the current main branch. Could nevertheless be fixed here.

ethanglaser: Just to clarify, these deletions are copied directly from the sklearn functionality: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/metrics/pairwise.py#L2454-L2496 and we have [this condition](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3337/changes#diff-ca2695b03ff78155a0e201c2d28c8132982422fcc481f62f0d1ecb597c09cdd3R64) that uses dal if metric is supported. What more would this help with?

david-cortes-intel: The second link doesn't show line number, only a whole-file diff.

ethanglaser: This: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/daal4py/sklearn/metrics/_pairwise.py#L88 which is unaffected by this PR

david-cortes-intel: @ethanglaser Yes, my point is that those conditions could be changed in this same PR using the dict logic that's currently not reachable, so that they'd work when passing recognized functions.

ethanglaser: The scope of the dict doesn't really overlap with the scope of our dal ready metrics though - we have cosine and correlation, the sklearn PAIRWISE_DISTANCE_FUNCTION dict does have cosine but does not have correlation, and has a bunch that we do not support. Additionally, the dict's values are 2-D matrix functions like cosine_distances(X, Y), but a callable metric is invoked per row pair as a scalar (out[i, j] = metric(x, y)), so a user can't pass those functions in the first place — a reverse lookup would never match.

david-cortes-intel: Looks like scikit-learn indeed doesn't work like their docs describe.

david-cortes-intel (approved):

PR #3346, opened 2026-07-30
address sklearn nightly conformance fails

ethanglaser:
## Description

Addresses fails in latest SEV (private CI run against sklearn main branch): http://intel-ci.intel.com/f188fa4b-ad83-f11f-857d-d4f5ef20c6a0

Changes include:

- Removal of bracket from some logreg array api deselected tests to accommodate both original case and updated pytest scope with additional parameter: https://github.com/scikit-learn/scikit-learn/commit/add0e835b7d01bcd14f6d585393d0b8154795bc3
- Alignment of forest `estimators_` reconstruction with categorical split support: `Tree.__cinit__` now requires `n_categories`, `NODE_DTYPE` gained a `left_cat_bitset` field, and `_validate_X_predict` requires `is_categorical_`: https://github.com/scikit-learn/scikit-learn/commit/b8e52fc543b0d7f18cef7098f52f9817794f4a51
- Alignment of logreg path with `class_weight` cleanup: it is no longer a path parameter (folded into `sample_weight` by the caller) and `y` now arrives label-encoded while `classes` keeps the original labels: https://github.com/scikit-learn/scikit-learn/commit/28ba5c0c22f79ea84e743d3684a87ebe568c7800
- Windows-only deselection of a logreg array api test (convergence issue), version-gated as the test itself was added in sklearn 1.9: https://github.com/scikit-learn/scikit-learn/commit/9ca766bafa756ad714d14a98f270cd3723e6dc22
- Re-adds kmeans vacuous cluster deselection that was removed in #1770 but unclear if actually resolved - it pops up in nightly azure pipelines

Latest run from this branch: http://intel-ci.intel.com/f190fdcc-1240-f1f9-8ac4-d4f5ef20c6a0

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


onedal/datatypes/dlpack/data_conversion.cpp
```diff
@@ -252,7 +252,11 @@ managed_t* construct_dlpack_tensor(const dal::array<byte_t>& array,
 }
 
 static inline void move_dlpack_data(dal::array<byte_t>& array, py::tuple dl_device, bool copy) {
-    DLDevice requested{ dl_device[0].cast<DLDeviceType>(), dl_device[1].cast<std::int32_t>() };
+    // The array API standard specifies dl_device as a tuple of plain ints, and
+    // pybind11 enums are not implicitly constructible from int, so the device
+    // type must be cast as an integer before being narrowed to the enum.
+    DLDevice requested{ static_cast<DLDeviceType>(dl_device[0].cast<std::int32_t>()),
```
david-cortes-intel: Why would this change be needed?

daal4py/sklearn/linear_model/logistic_path.py
```diff
@@ -107,6 +107,17 @@ def _get_additional_lbfgs_options_dict(k, v):
 from sklearn.linear_model._logistic import _logistic_regression_path as lr_path_original
 from sklearn.preprocessing import LabelBinarizer, LabelEncoder
 
+if sklearn_check_version("1.10"):
```
david-cortes-intel: It'd be easier for future maintenance (and easier to read) if it could add a conditional function signature, the way it's currently done for other versions.

david-cortes-intel:
The logistic regression failures are due to small numeric differences. Will also need to be skipped.

daal4py/sklearn/linear_model/logistic_path.py
ethanglaser: Good point, aligning it with other files/versions

onedal/datatypes/dlpack/data_conversion.cpp
ethanglaser: I can drop for now and see how SEV looks without it

ethanglaser: Migrated to here: https://github.com/uxlfoundation/scikit-learn-intelex/pull/3348
I do think the change is beneficial

david-cortes-intel:
Some of the remaining failures for logistic regression are due to small numeric differences. Those can be deselected.

But the ones for LogisticRegressionCV will need some investigation.

sklearnex/ensemble/_forest.py
```diff
@@ -80,6 +80,42 @@
...
+        # The same change widened NODE_DTYPE with a 'left_cat_bitset' field, which
+        # oneDAL's node struct does not have. Copy field-wise into the wider dtype
+        # so the absent categorical bitset stays zeroed (i.e. no categories).
+        nodes = np.zeros(node_ar.shape, dtype=NODE_DTYPE)
+        for name in node_ar.dtype.names:
+            nodes[name] = node_ar[name]
+        return nodes
+
+    def _set_tree_fitted_attrs(est):
+        # '_validate_X_predict' requires this attribute, normally set by
```
icfaust: I'd like for us to follow the approach going forward where we try to get sklearnex support to act like the latest sklearn as default and then modify to match older versions (like in #1313). At some point when we get to 1.10+ support only we will have to come back through and modify everything to make it the default anyway. This saves future work and maintenance.

deselected_tests.yaml
```diff
@@ -354,13 +354,14 @@ deselected_tests:
   - linear_model/tests/test_logistic.py::test_logistic_cv_sparse[42-csr_matrix]
   - linear_model/tests/test_logistic.py::test_logistic_cv_sparse[42-csr_array]
   - tests/test_common.py::test_estimators[LogisticRegressionCV(cv=3,max_iter=5,use_legacy_attributes=False)-check_sample_weight_equivalence_on_dense_data]
-  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[array_api_strict-CPU_DEVICE-float64-None-False-False-False]
-  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[array_api_strict-CPU_DEVICE-float64-None-False-True-False]
-  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[dpnp-cpu-float64-None-False-False-False]
-  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[dpnp-cpu-float64-None-False-True-False]
+  # Unterminated ids: sklearn 1.10 appends the solver, so match on the id prefix
```
icfaust: Is this addressing why there isn't a `]` at the end?

deselected_tests.yaml
```diff
@@ -354,13 +354,14 @@ deselected_tests:
...
+  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[array_api_strict-CPU_DEVICE-float64-None-False-False-False
+  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[array_api_strict-CPU_DEVICE-float64-None-False-True-False
+  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[dpnp-cpu-float64-None-False-False-False
+  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[dpnp-cpu-float64-None-False-True-False
 
   # Logistic regression array api compliance test fails on Windows only due to convergence issue
-  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[array_api_strict-device1-float32-balanced-True-True-True]
+  - linear_model/tests/test_logistic.py::test_logistic_regression_array_api_compliance[array_api_strict-device1-float32-balanced-True-True-True
```
icfaust: This is claiming to be a windows only deselection but does not have `win32` at the end.

icfaust (commented):
Since there are already review comments here, I want to add weigh in.

icfaust:
Please when you fill in the description, link to this: https://github.com/scikit-learn/scikit-learn/pull/33354 because this will help people who want to see what caused this change.

onedal/datatypes/dlpack/data_conversion.cpp
icfaust: @david-cortes-intel @ethanglaser the cause is not from the first line but the later ones. Realistically I did not add enough guards to the dl_device tuple and we do not check for type which is implicitly assumed to be able to convert to a `DLDeviceType` which cannot be done in `pybind11` directly from a python Int type. If we really wanted to do it right we would need to add additional checks here (https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/datatypes/table.cpp#96) to guarantee the cast to `int32_t` can work.

deselected_tests.yaml
ethanglaser: Right, reason for this is that sklearn 1.10 will be adding an additional parameter to these tests, so the updated line handles both the original deselection case and the new one - test is still deselected for the same reasons. Since the comment isn't really related to why the test fails, I can remove if preferred

deselected_tests.yaml
ethanglaser: Updated

sklearnex/ensemble/_forest.py
ethanglaser: Check if https://github.com/uxlfoundation/scikit-learn-intelex/pull/3346/changes/2ce494b3367daa92445fe115018c0762630410d7 adequately addresses this

ethanglaser:
> Please when you fill in the description, link to this: [scikit-learn/scikit-learn#33354](https://github.com/scikit-learn/scikit-learn/pull/33354) because this will help people who want to see what caused this change.

Description added with details on origin of changes

onedal/datatypes/dlpack/data_conversion.cpp
ethanglaser: Addressed on that PR

david-cortes-intel:
@ethanglaser I see from the logs that similar changes are needed for LogisticRegressionCV and for the newton-cg solver, which run under preview mode:
https://dev.azure.com/daal/daal4py/_build/results?buildId=62103&view=logs&j=9170ab64-83bd-5a6a-4a87-20bcd33aae99&t=42d710fe-fdf0-5901-4ef3-1b49c35ff4b5&l=56234

deselected_tests.yaml
```diff
@@ -182,6 +190,7 @@ deselected_tests:
   # test_non_uniform_strategies fails due to differences in handling of vacuous clusters after update
   # See https://github.com/uxlfoundation/scikit-learn-intelex/issues/69
   - cluster/tests/test_k_means.py::test_kmeans_relocated_clusters >=0.24
+  - preprocessing/tests/test_discretization.py::test_nonuniform_strategies[kmeans
```
david-cortes-intel, line 193: This is a real bug that needs to be fixed, currently assigned to @Alexandr-Solovev . Please remove it from the PR.

ethanglaser: Why not deselect while a fix is in progress?

ethanglaser: With azp nightly green we'll know when a regression occurs

david-cortes-intel: Because it's hard to reproduce this issue and the logs with failures are not kept forever. Also it has happened multiple times in the past that these deselects are added and then no one remembers to remove them, or they get added back in merge conflicts and similar.

ethanglaser: The issue is 7 years old at this point - if we're worried about forgetting it then im not sure what to say. There is already a deselection associated with it. If we want to reproduce then run it locally or draft a PR fix that removes the deselection. We have deselected tests to maintain green CI.

david-cortes-intel: I'll leave to @Alexandr-Solovev to comment about whether fresh logs are needed or not.

Alexandr-Solovev: I believe that it should not block merging process, especially in case that we dont really change Kmeans algo in cpp side for a long time. Can you please create a JIRA ticker for this deselected test and add TODO mark here.

david-cortes-intel: @ethanglaser Please add a TODO comment for this deselection reminding to remove it after the bug has been fixed in oneDAL.

david-cortes-intel (approved):

PR #3357, opened 2026-08-05
FIX: Fall back to sklearn for forest predictions on NaN data

david-cortes-intel:
## Description

ref https://github.com/uxlfoundation/scikit-learn-intelex/issues/3356

Solves the first point in the issue, in which passing data with missing values to prediction-related methods generates an error, by falling back to scikit-learn in those cases.

Note that this still doesn't produce correct results due to the second problem in the issue (incorrect assignment of missing branches).

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklearnex/ensemble/_forest.py
+++ b/sklearnex/ensemble/_forest.py
@@ -358,9 +358,11 @@ class BaseForest(oneDALEstimator, ABC):
                 return patching_status
             xp, _ = get_namespace(X)
             has_nan = xp.any(xp.isnan(X))
-            patching_status.and_conditions([
-                (not has_nan, "Missing values are not supported."),
-            ])
+            patching_status.and_conditions(
+                [
+                    (not has_nan, "Missing values are not supported."),
+                ]
+            )
 
             if method_name == "predict_proba":
                 patching_status.and_conditions(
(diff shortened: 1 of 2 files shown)


ethanglaser (approved):
LGTM for now. Maybe in future we could come up with a creative solution to still allow non-nan rows to be inferenced in oneDAL.

sklearnex/ensemble/_forest.py
```diff
@@ -354,6 +354,15 @@ def _onedal_cpu_supported(self, method_name, *data):
                     ),
                 ]
             )
+            if not dal_ready:
+                return patching_status
+            xp, _ = get_namespace(X)
+            has_nan = xp.any(xp.isnan(X))
```
icfaust: unlike previous changes recently like this (adding nan checks) this is on the hot path. it is guaranteed to reduce performance. please move the finiteness check here and disable its use in validate_data in onedal_predict in order to prevent an additional load/eval.

icfaust (changes requested):
sorry need to throw this on because of the approval

sklearnex/ensemble/_forest.py
david-cortes-intel: Fixed. But looks like the finite checker was also checking for infinites, which should be supported for predictions.

icfaust: We have an optimized finite checker to improve perf, this unfortunately is still not as performant. What trick the inf+nan check does is that if it has an Inf, it will offload to sklearn and fail on sklearn-side (where performance loss isn't a big deal: I'd replicate similar behavior already in this file for fit, unless I am missing something).

david-cortes-intel: Updated. Looks like scikit-learn indeed doesn't support infinites, even though oneDAL does.

icfaust (approved):
Just make sure to also modify the gpu version before merging. Thanks for doing the hard work.

PR #3358, opened 2026-08-05
FIX: Set branch for missing values in sklearn tree structures

david-cortes-intel:
## Description

ref https://github.com/uxlfoundation/scikit-learn-intelex/issues/3356

Sets the branch for missing values in decision tree objects according to the same logic that scikit-learn uses.

Note that the branch for missing values is set in the trees according to a majority logic when the input doesn't contain missing values; and this piece of information is still used by scikit-learn when falling back on predictions.

I am not sure if this is all that's needed as I see that the generated trees forwarded to sklearn on .predict() with missing values would still not match against predictions from treelite, for instance.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/onedal/primitives/tree_visitor.cpp
+++ b/onedal/primitives/tree_visitor.cpp
@@ -299,7 +299,7 @@ void init_get_tree_state(py::module_& m) {
 
             // Sets the 'missing_go_to_left' values on non-terminal nodes.
             // Note that this requires having the values in the arrays
-            // already filled-in for the children of node, so it cannot be
+            // already filled-in for the children of each node, so it cannot be
             // done in the same pass that sets these initial values.
             py::array_t<skl_tree_node>& node_ar = tsv.node_ar;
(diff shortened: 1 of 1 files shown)


onedal/primitives/tree_visitor.cpp
```diff
@@ -296,6 +296,21 @@ void init_get_tree_state(py::module_& m) {
                                                      n_classes);
             node_visitor<Task, decltype(tsv)> tsv_decorator{ &tsv };
             model.traverse_depth_first(iTree, std::move(tsv_decorator));
+
+            // Sets the 'missing_go_to_left' values on non-terminal nodes.
+            // Note that this requires having the values in the arrays
+            // already filled-in for the children of node, so it cannot be
```
david-cortes-intel:
```suggestion
            // already filled-in for the children of each node, so it cannot be
```

avolkov-intel (approved):

PR #3325, opened 2026-07-16
[enhancement] Add CPython free-threading support

napetrov:
## Summary

Adds the first supported scikit-learn-intelex configuration for free-threaded CPython: **CPython 3.14t on Linux x86-64, CPU/host execution**. Native imports keep the GIL disabled.

## Changes

- Mark the generated daal4py and tested host oneDAL extensions as free-threading compatible.
- Protect mutable native algorithm/model state and Python-owned dense/CSR data used across native threads.
- Harden capsule, transceiver, and MPI initialization/finalization lifetimes; require `MPI_THREAD_MULTIPLE` where MPI is used.
- Bind native builds to the packaging interpreter, validate SOABI, and handle supported oneDAL layouts consistently.
- Add focused build-discovery, ownership, lifecycle, and concurrency tests, plus opt-in sanitizer build modes.

## Validation

Exact-head GitHub Actions and Azure Pipelines are green at `29020567`, including the Linux CPython 3.14t build, wheel ABI smoke test, and free-threading tests. The native extensions import without re-enabling the GIL.

## Current limitations / follow-ups

- **TSan is not a CI gate in this PR.** The PR only provides the opt-in `SKLEARNEX_SANITIZER=thread` build mode. A manually instrumented run fails on reported sanitizer errors (`-fno-sanitize-recover=all`), but no TSan-clean claim is made here. CI enablement and feasibility with the external native runtimes need separate evaluation.
- **CatBoost and SHAP are excluded from the CPython 3.14t test environment.** They are optional integrations whose upstream packages are not currently usable in this configuration; their integration tests skip when the packages are absent.
- **Windows free-threaded, DPC/SYCL, SPMD, and distributed configurations are outside this PR's support scope.** They should be considered, enabled, and validated explicitly in follow-up PRs.
- The wheel produced in CI is only a build/ABI validation artifact. A standalone or publication-ready wheel is **not** a goal of this PR.

This PR therefore claims Linux x86-64 CPU/host support only; it does not claim sanitizer-clean or cross-backend/platform readiness.

---

<details>
<summary>Checklist</summary>

- [x] Code and non-obvious synchronization/lifetime invariants are documented.
- [x] User-facing support limitations are documented.
- [x] Commits contain the required sign-off.
- [x] Exact-head GitHub Actions and Azure Pipelines are green.
- [x] Focused tests cover the new build and free-threading behavior.
- [x] No performance/quality change is expected: synchronization is scoped to mutable manager/model objects; independent estimators are not globally serialized.

</details>

PR #3360, opened 2026-08-07
[bug] Make the CMake backend build reproducible and its oneDAL discovery explicit

napetrov:
Split out of #3325, which reviewers asked to break up. This part carries no free-threading changes at all and stands on its own.

### oneDAL library discovery

`custom_build_cmake_clib` assumed a single `$DALROOT/lib/<arch>` layout. Conda packages and some source installs put the libraries in `$DALROOT/lib` or, on Windows, `$DALROOT/Library/lib`, and the build then failed at link time with a missing-library error that pointed at the wrong directory.

Discovery now walks the known layouts and picks the first one that contains the *complete* set of libraries the selected backend needs, rather than the first one that merely exists. When none match, the error names every directory it looked in and exactly which SONAMEs were missing from each.

### One source for oneDAL library names

`setup.py:get_libs()` and the discovery code each spelled out the oneDAL library names per platform, so the two build paths could disagree about what oneDAL is called and only find out at link time. Naming now lives in `scripts/build_backend.py` alone:

* `get_onedal_libraries()` returns linker names, and `get_libs()` delegates to it
* `get_onedal_library_filenames()` returns the on-disk names discovery looks for

Both derive from one stem list keyed on the interface (`daal` / `host` / `dpc` / `spmd_dpc`), the major binary version and the platform, so the check is exact rather than a glob. Output is identical to the previous `get_libs()` for every reachable interface × platform × parameters-lib × version combination.

One divergence had to be resolved to do this: on Windows with the parameters library enabled, `get_libs()` emitted `onedal_parameters*` while `scripts/CMakeLists.txt` links `onedal_core_parameters_*_dll`. The unified version follows CMakeLists, since `use_parameters_lib` is unconditionally false on Windows (`setup.py`) and the `get_libs()` branch was therefore dead code.

### Build reproducibility

* `CMAKE_BUILD_TYPE` is now passed explicitly on the command line and `scripts/CMakeLists.txt` only defaults it when unset, so `--debug` actually produces a debug build. Previously the CMakeLists hard-set `Release` unconditionally, silently discarding the caller's choice. The default is kept, without `FORCE`, so a direct `cmake` invocation with no `-D` still gets optimization flags.
* The empty generator argument is no longer appended on non-Windows. An empty string in `argv` is not portable across CMake versions and shows up as a confusing `""` in the build log.
* The Windows-only `$ENV{PYTHON}` hint that overrode `PYTHON_EXECUTABLE` is removed - it silently shadowed the interpreter the build was actually invoked with. Note `PYTHON_EXECUTABLE` is the deprecated FindPythonInterp spelling, which nothing in this project reads; the modern `Python_EXECUTABLE` is set by #3325, the only PR that adds a `find_package(Python)` consumer.

`-j<n>` stays as one argument and installation stays on `cmake --install`. An earlier revision changed both for CMake 3.13/3.15 compatibility; that was dropped after review pointed out `dependencies-dev` pins `cmake==4.4.2`, so the floor is set by the pin, not by the oldest CMake the invocation would parse.

### MPI library name

`MPI_LIBNAME` can now be set in the environment. The previous logic probed for `mpi.lib` / `impi.lib` and assumed `mpi` elsewhere, which does not cover MPICH or a non-standard Intel MPI layout.

### Other

* `onedal/dummy/dummy_onedal.hpp` asserts the kind of the constant it reads instead of assuming it.
* `tests/test_build_backend.py` covers only what CI cannot reach: the debug configuration (no CI job sets `DEBUG_BUILD` or passes `--debug`), oneDAL layouts other than the single one each CI job installs, and the naming agreement itself - which is structurally out of CI's reach, since each job builds one platform and none compares `setup.py` against `CMakeLists.txt`. The Windows import-library names are checked by grepping `scripts/CMakeLists.txt` rather than by restating them. 22 cases, no oneDAL installation required.

Depended on by #3325.


scripts/build_backend.py
```diff
@@ -102,28 +180,31 @@ def custom_build_cmake_clib(
...
+        cmake_args.append(cmake_generator)
+    cmake_args += [
         "-S" + builder_directory,
         "-B" + abs_build_temp_path,
         "-DCMAKE_INSTALL_PREFIX=" + install_directory,
         "-DCMAKE_PREFIX_PATH=" + install_directory,
         "-DIFACE=" + iface,
         "-DONEDAL_MAJOR_BINARY=" + str(onedal_major_binary_version),
+        "-DPYTHON_EXECUTABLE=" + sys.executable,
+        "-DPython_EXECUTABLE=" + sys.executable,
```
david-cortes-intel: Why would it need both?

scripts/build_backend.py
```diff
@@ -140,12 +221,18 @@ def custom_build_cmake_clib(
     # the number of parallel processes is dictated by MAKEFLAGS (see setup.py)
     # using make conventions (i.e. -j flag) but is set as a cmake argument to
     # support Windows and Linux simultaneously
-    make_args = ["cmake", "--build", abs_build_temp_path, "-j" + str(n_threads)]
+    # Keep the job count as a separate argument: CMake 3.13 accepts ``-j 2``
+    # but not the concatenated ``-j2`` spelling.
+    make_args = ["cmake", "--build", abs_build_temp_path, "-j", str(n_threads)]
 
+    # ``cmake --install`` was added in CMake 3.15. Use the generated install
```
david-cortes-intel: CMake's install and build do different things. If some post-processing is needed, it could be done in the setup file, like it currently does for .lib files.

tests/test_build_backend.py
```diff
@@ -0,0 +1,221 @@
...
+        is_win=is_win,
+        is_lin=not is_win,
+        debug_build=debug_build,
+    )
+
+    build_type_args = [arg for arg in calls[0] if arg.startswith("-DCMAKE_BUILD_TYPE=")]
+    assert build_type_args == [f"-DCMAKE_BUILD_TYPE={expected_build_type}"]
+    assert "" not in calls[0]
+    assert ("-GNinja" in calls[0]) is is_win
+    assert not any(arg.startswith("-DSKLEARNEX_SANITIZER=") for arg in calls[0])
```
david-cortes-intel: `DSKLEARNEX_SANITIZER` is only referenced in this test file. Looks like it could be left out.

scripts/build_backend.py
napetrov: It doesn't — removed. Nothing in `scripts/CMakeLists.txt` reads either variable; the build gets its Python paths from `PYTHON_INCLUDE_DIR`, `PYTHON_LIBRARY_DIR` and `pybind11_DIR`, all of which were already being passed.

I had added them on the assumption that pybind11's `FindPython` needed steering toward the interpreter running the build. That was worse than unnecessary: the lowercase `PYTHON_EXECUTABLE` is the deprecated `FindPythonInterp` spelling, and passing both risks pointing at *different* interpreters rather than pinning one.

scripts/build_backend.py
napetrov: Agreed on both counts — reverted to `cmake --install`.

You are right that they are not equivalent: `--install` runs only the install rules, while `--build --target install` goes through the generator and can rebuild before installing. And my justification for the switch does not survive checking: I claimed CMake 3.15 compatibility, but `dependencies-dev` pins `cmake==4.4.2` and this PR itself raises the conda recipe to `cmake >=3.18`. There is no supported configuration where `cmake --install` is unavailable, so the change bought nothing and lost the narrower semantics.

The `-j` change went with it for the same reason — `-j<n>` concatenated is back, since the split `-j <n>` form was only justified by that same phantom CMake 3.13 support.

Noted on post-processing belonging in `setup.py` alongside the existing `.lib` handling; nothing here needs it today.

tests/test_build_backend.py
napetrov: Correct — `SKLEARNEX_SANITIZER` appears nowhere else in the repository. Assertion removed.

It was left over from an earlier iteration that had a sanitizer build flag. What remained was a test pinning a variable name instead of a behavior, which is worse than no test: it would have failed if someone ever legitimately added that flag.

tests/test_build_backend.py
```diff
@@ -0,0 +1,220 @@
...
+    build_type_args = [arg for arg in calls[0] if arg.startswith("-DCMAKE_BUILD_TYPE=")]
+    assert build_type_args == [f"-DCMAKE_BUILD_TYPE={expected_build_type}"]
+    assert "" not in calls[0]
+    assert ("-GNinja" in calls[0]) is is_win
+    assert calls[1][:2] == ["cmake", "--build"]
+    assert Path(calls[1][2]).parts[-2:] == ("build", "backend_host")
+    assert calls[1][3:] == ["-j1"]
+    assert calls[2][:2] == ["cmake", "--install"]
+    assert Path(calls[2][2]).parts[-2:] == ("build", "backend_host")
+    assert calls[2][3:] == []
```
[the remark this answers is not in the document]
napetrov: Correct, and the PR description was the stale half — I've updated it.

Both of those changes existed in an earlier revision of this PR and were reverted after @david-cortes-intel pointed out that `dependencies-dev` pins `cmake==4.4.2`. My original rationale — "support CMake 3.13" — was simply false for this project: the build cannot run on a CMake that old, so `-j2` vs `-j 2` and `cmake --install` vs `cmake --build --target install` are both unreachable compatibility concerns. The tests assert the current shape, which is the reverted-to-upstream shape, so implementation and tests already agreed; only the description didn't.

Worth noting the reverts are load-bearing in the other direction too: the third item in that group, dropping the `$ENV{PYTHON}` → `PYTHON_EXECUTABLE` hint, is *kept*. That is the deprecated FindPythonInterp spelling and nothing in this project reads it. #3325 sets the modern `Python_EXECUTABLE` instead, because it is the only PR that adds a `find_package(Python)` consumer — and `tests/test_build_backend.py` now asserts the deprecated spelling is absent, so a future re-add fails the test rather than silently steering the interpreter search.

scripts/CMakeLists.txt
```diff
@@ -21,7 +21,9 @@ project("oneDAL python interface"
     HOMEPAGE_URL https://github.com/uxlfoundation/oneDAL
 )
 
-set(CMAKE_BUILD_TYPE Release)
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type" FORCE)
```
david-cortes-intel: Wouldn't this be set automatically if passing `-DCMAKE_BUILD_TYPE`?

scripts/build_backend.py
```diff
@@ -28,6 +28,73 @@
 logger = logging.getLogger("sklearnex")
 
 
+def _get_required_onedal_libraries(
```
david-cortes-intel: This would need to be unified with the setup.py file:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/7b4847a3e938cc941b72633fe2602f00ac9b2b51/setup.py#L197

tests/test_build_backend.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+    monkeypatch.setattr(build_backend.np, "get_include", lambda: "numpy-include")
+    monkeypatch.setattr(build_backend, "get_paths", lambda: {"include": "python-include"})
+    monkeypatch.setattr(
+        build_backend,
+        "get_config_var",
+        lambda name: {
+            "LIBDEST": str(tmp_path / "python" / "Lib"),
+            "LIBDIR": str(tmp_path / "python" / "lib"),
+            "SOABI": "cp312-win_amd64" if is_win else "cpython-312-x86_64-linux-gnu",
```
david-cortes-intel: This is hard-coding a python version.

tests/test_build_backend.py
```diff
@@ -0,0 +1,220 @@
...
+@pytest.mark.parametrize(
+    "is_win,debug_build,expected_build_type",
+    [
+        (False, False, "Release"),
+        (False, True, "Debug"),
+        (True, False, "Release"),
+        (True, True, "Debug"),
+    ],
+)
+def test_cmake_build_type_is_explicit(
```
david-cortes-intel: These tests feel a bit redundant considering that we have CI configurations that test these as non-mocked scenarios.

scripts/build_backend.py
napetrov: Done in ba2ad006 — naming now lives in one place and `setup.py:get_libs()` delegates to it.

`build_backend.py` grew two functions over the same stem list:

* `get_onedal_libraries()` — linker names (`:libonedal_core.so.4`, `onedal_core_dll.4`, `onedal_core.4`)
* `get_onedal_library_filenames()` — on-disk names discovery looks for (`libonedal_core.so.4`, `onedal_core_dll.4.lib`, `libonedal_core.4.dylib`)

`get_libs()` keeps its `onedal`/`onedal_dpc` spelling at the call site and maps it to the backend's `host`/`dpc`, so no caller changed.

I checked the replacement is output-identical rather than assuming it: reproduced upstream `get_libs()` verbatim as a reference and compared against the new path across every interface × platform × parameters-lib × major-version combination — 30 cases, no differences.

One real divergence the unification had to resolve, worth flagging explicitly. On Windows with `use_parameters_lib` on, upstream `get_libs()` emitted `onedal_parameters.N` and `onedal_parameters_dll.N`, while `scripts/CMakeLists.txt` links `onedal_core_parameters_dll.N`. Those cannot both be right. The unified version follows CMakeLists, because that spelling is the one actually exercised — `setup.py:113` sets `use_parameters_lib = (not IS_WIN) and ...`, so the Windows branch of that `if` was unreachable dead code. That is also why the comparison above shows no differences: the combination where they disagree is not reachable.

`tests/test_build_backend.py` now checks the agreement rather than restating it — one test greps the Windows import-library names straight out of `CMakeLists.txt` and compares against what the Python side produces, so a future edit to either one fails the test instead of failing at link time.

scripts/CMakeLists.txt
napetrov: Not automatically, no — but you are right that the `FORCE` was wrong, and it is gone in ba2ad006.

Passing `-DCMAKE_BUILD_TYPE=X` sets the cache variable, so `if(NOT CMAKE_BUILD_TYPE)` is false and the block does not run. What it does *not* do is make this block unnecessary, because the block covers the case where nothing passed the flag: a direct `cmake -S scripts -B build` with no `-D`, which is what a developer poking at the backend by hand does. Upstream's unconditional `set(CMAKE_BUILD_TYPE Release)` gave that invocation optimization flags; deleting it outright would silently turn hand-run builds into no-optimization builds, since CMake's own default for `CMAKE_BUILD_TYPE` is empty.

The `FORCE` was a leftover and actively counterproductive: `FORCE` overwrites a cached value, which is exactly what the surrounding guard exists to prevent. It would only ever matter in the situation the guard already rules out. Removed, and the comment now says why the block is there at all:

```cmake
# build_backend.py always passes -DCMAKE_BUILD_TYPE, so this only applies to a
# direct cmake invocation, which would otherwise get no optimization flags at
# all. Guarded rather than unconditional so it cannot discard a caller's choice.
if(NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type")
endif()
```

tests/test_build_backend.py
napetrov: Right — that was `SOABI` in the fake `get_config_var`, and it was pointless as well as hard-coded. Nothing in `build_backend.py` reads `SOABI`; it was left over from an earlier revision. Gone in ba2ad006, along with the fake `LIBDEST` pretending to be a Windows layout.

The rewritten test only stubs what the code under test actually calls (`LIBDEST`, `LIBDIR`), and asserts only the build type and the absence of the empty generator argument — no Python version, no ABI tag, no interpreter paths.

tests/test_build_backend.py
napetrov: Fair, and I cut the file down in ba2ad006 to only what CI cannot reach. The old version did contain tests that just re-ran what CI already does for real; those are gone.

What is left, and why CI does not cover it:

* **Debug configuration.** No CI job sets `DEBUG_BUILD` or passes `--debug` — I grepped `.ci/` and `.github/` for both and there are no hits; the only references are in `setup.py` itself. Since the bug this PR fixes is precisely that `--debug` silently produced a Release build, CI would have stayed green through it. That is the one behaviour here I actually want pinned.
* **oneDAL layouts other than the one CI installs.** Each CI job has exactly one oneDAL installation, so it exercises exactly one layout. It cannot tell you that the flat `$DALROOT/lib` conda layout works, nor that a stale arch directory from a different major version is correctly skipped rather than preferred — which is the actual bug in the old `os.path.isdir` check, and it fails *silently* in the sense that it fails at link time with an error pointing at the wrong directory.
* **The naming agreement itself.** This is the part CI structurally cannot check: each job builds one platform, so no job ever compares the Windows names against the Linux ones, and none compares `setup.py` against `CMakeLists.txt`. The test that greps CMakeLists for the import-library names is cheap and catches a class of error that otherwise only shows up as a link failure on a platform the author was not building.

I also dropped the assertions on the exact `cmake --build` / `cmake --install` argv, which is what @Copilot's comment above was reacting to. Pinning the command shape was testing the mock, not the code — CI runs those commands for real.

Net: 22 test cases (mostly parametrization over platform and index dtype), no oneDAL installation needed, and the file now has a docstring stating this scoping rule so the next person does not re-add CI-duplicating tests.

scripts/CMakeLists.txt
```diff
@@ -21,7 +21,12 @@ project("oneDAL python interface"
     HOMEPAGE_URL https://github.com/uxlfoundation/oneDAL
 )
 
-set(CMAKE_BUILD_TYPE Release)
+# build_backend.py always passes -DCMAKE_BUILD_TYPE, so this only applies to a
+# direct cmake invocation, which would otherwise get no optimization flags at
+# all. Guarded rather than unconditional so it cannot discard a caller's choice.
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type")
+endif()
```
david-cortes-intel: I don't think it matters that optimization flags aren't passed here on a custom call, because this only contains simple bindings where there isn't much to optimize. That plus a custom call could as well pass the flags it wants as a different option.

tests/test_build_backend.py
```diff
@@ -0,0 +1,216 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""Tests for the parts of the build driver CI cannot reach.
+
+CI builds the release configuration on Linux, Windows and macOS, so the happy
+path is covered there for real. What it never exercises is a debug build, a
+oneDAL layout other than the one its own installation happens to use, or the
```
david-cortes-intel: This actually is covered in CI configs - for instance, the one that builds the docs installs oneDAL from pip.

tests/test_build_backend.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def test_get_onedal_library_dir_names_every_directory_and_miss(tmp_path):
+    with pytest.raises(FileNotFoundError) as exc_info:
+        _get_onedal_library_dir(tmp_path, "intel64", major_version=4)
+
+    message = str(exc_info.value)
+    assert str(tmp_path / "lib" / "intel64") in message
+    assert str(tmp_path / "lib") in message
+    assert "libonedal_thread.so.4" in message
```
david-cortes-intel: This .4 could be defined as a global variable at the beginning of the file.

tests/test_build_backend.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def test_get_onedal_library_dir_skips_incomplete_directory(tmp_path):
+    # An arch directory left over from a different oneDAL major version must not
+    # win over a complete one, which is what a plain isdir() check would do.
+    flat = tmp_path / "lib"
+    _create_libraries(flat, LINUX_HOST_LIBRARIES)
+    _create_libraries(flat / "intel64", ("libonedal.so.3", "libonedal_core.so.3"))
+
+    assert _get_onedal_library_dir(tmp_path, "intel64", major_version=4) == str(flat)
```
david-cortes-intel: Is sklearnex meant to work with aarch64 setups at the moment?

tests/test_build_backend.py
```diff
@@ -0,0 +1,216 @@
...
+
+def test_unsupported_backend_is_rejected():
+    with pytest.raises(ValueError, match="Unsupported oneDAL backend"):
+        get_onedal_libraries("host_dpc", 4)
+
+
+@pytest.mark.parametrize(
+    "debug_build,expected_build_type", [(False, "Release"), (True, "Debug")]
+)
+def test_cmake_build_type_follows_the_debug_flag(
```
david-cortes-intel: I think this is not necessary. The condition it tests is pretty straightforward to debug.

scripts/CMakeLists.txt
napetrov: Agreed — dropped the default rather than keeping a guarded one. `build_backend.py` always passes `-DCMAKE_BUILD_TYPE`, and a direct `cmake` call can pass whatever flags it wants, so this file no longer has an opinion.

Worth recording why the line could not simply stay as it was, since that is the bug the PR was fixing: `set(CMAKE_BUILD_TYPE Release)` creates a *normal* variable, and a normal variable shadows the cache entry that `-DCMAKE_BUILD_TYPE=Debug` writes. So `setup.py --debug` configured a Debug cache value and then compiled with Release flags. Removing the line fixes that with no defaulting logic at all.

In 8b67874d.

tests/test_build_backend.py
napetrov: Correct — the docstring was wrong and is fixed in 8b67874d. The documentation job installs oneDAL from pip, which gives the flattened `lib/` tree, so the second layout is exercised for real in CI and a mistake there shows up as a build failure.

The docstring now only claims what the build cannot report: which candidate directory was rejected and why, and that `setup.py` and `scripts/CMakeLists.txt` still agree on the Windows import-library names — a disagreement there surfaces as a link error a long way from its cause.

tests/test_build_backend.py
napetrov: Done in 8b67874d — `MAJOR_VERSION = 4` at the top of the file, and every name is derived from it, including the deliberately-different `MAJOR_VERSION - 1` in the stale-directory test.

tests/test_build_backend.py
napetrov: No — there is no aarch64 support to speak of today: no published build, and no CI job on that architecture.

Nothing in this PR claims otherwise. The `aarch64 -> arm` entry in `plt_dict` is unchanged from `main` (`scripts/build_backend.py:93`) and only records what oneDAL calls that directory; the tests here use `intel64` exclusively and assert nothing about aarch64.

If you would rather the mapping not imply support at all, I am happy to drop the `aarch64` entry in this PR — say the word. I left it because removing it is a behaviour change unrelated to oneDAL discovery.

tests/test_build_backend.py
napetrov: Removed in 8b67874d.

It is redundant now for a concrete reason rather than just on taste: with the `set(CMAKE_BUILD_TYPE Release)` line gone from `scripts/CMakeLists.txt` (see the other thread), there is nothing left that could shadow the caller`s `-DCMAKE_BUILD_TYPE`, so the failure mode the test guarded is structural rather than behavioural. It also had to monkeypatch `pybind11`, `numpy`, `sysconfig` and `subprocess.check_call` to observe one argv entry, which is a lot of machinery for something a build log shows directly.

scripts/CMakeLists.txt
```diff
@@ -21,12 +21,9 @@ project("oneDAL python interface"
...
 
-# build_backend.py always passes -DCMAKE_BUILD_TYPE, so this only applies to a
-# direct cmake invocation, which would otherwise get no optimization flags at
-# all. Guarded rather than unconditional so it cannot discard a caller's choice.
-if(NOT CMAKE_BUILD_TYPE)
-    set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type")
-endif()
+# CMAKE_BUILD_TYPE is left to the caller: build_backend.py always passes
+# -DCMAKE_BUILD_TYPE, and a direct cmake invocation can pass whatever it wants.
+# Setting it here as a normal variable would silently discard that choice.
```
david-cortes-intel: I think this comment is not helpful since it's quite obvious. Could be removed.

scripts/build_backend.py
```diff
@@ -28,6 +28,122 @@
...
+    """Return linker names for the oneDAL libraries the given backend needs."""
+    library_names = _get_onedal_library_names(iface, use_parameters_lib, is_win)
+    if is_win:
+        return [
+            f"{_to_windows_import_name(name)}.{major_version}" for name in library_names
+        ]
+    if is_mac:
+        return [f"{name}.{major_version}" for name in library_names]
+    # The ':' prefix asks the linker for this exact SONAME rather than for a
+    # development symlink, which the runtime packages do not ship.
```
david-cortes-intel: Second part of the statement is not correct.

tests/test_build_backend.py
david-cortes-intel: I think it'd be more helpful to add support for a potential aarch64 structure in this PR than to remove it, since it will have to be done at some point and this PR already touches the directory logic.

scripts/CMakeLists.txt
napetrov: Removed in 78e8775a.

scripts/build_backend.py
napetrov: Corrected in 78e8775a: the comment now only says that `-l:` selects the exact SONAME; it no longer makes a claim about runtime-package symlinks.

tests/test_build_backend.py
napetrov: Added in 78e8775a. `get_onedal_arch_dir()` maps `aarch64` to oneDAL’s `arm` directory, and the discovery tests now create and select a complete `lib/arm` layout in addition to asserting all three architecture spellings.

scripts/build_backend.py
```diff
@@ -28,6 +28,127 @@
...
+        library_names.append("onedal_thread")
+    return library_names
+
+
+def _to_windows_import_name(name):
+    return name if name.endswith("_dll") else f"{name}_dll"
+
+
+def get_onedal_arch_dir(machine):
+    return {"x86_64": "intel64", "AMD64": "intel64", "aarch64": "arm"}.get(
```
david-cortes-intel: What kind of function would return AMD64 with uppercase letters?

napetrov: `platform.machine()` on Windows. It goes through `_get_machine_win32()`, which reads the WMI CPU architecture - code 9 is `AMD64` - and falls back to `PROCESSOR_ARCHITEW6432` / `PROCESSOR_ARCHITECTURE`; the same call answers `x86_64` on Linux and macOS.

Both spellings reach this function: at build time from `plt.machine()` a few lines below, and at import time from the same call in `onedal/__init__.py` and `daal4py/__init__.py`, inside their `if "Windows" in platform.system():` branches. So both have to map to `intel64` or the Windows wheel cannot find its own libraries. Added a comment naming the source in [`26a7afe2`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3360/commits/26a7afe2071a43ce72e8d62f1aaae18a4214cc52); `tests/test_build_backend.py` already parametrizes all three keys.

scripts/build_backend.py
```diff
@@ -28,6 +28,133 @@
...
+    "host": "onedal",
+    "dpc": "onedal_dpc",
+    "spmd_dpc": "onedal_dpc",
+}
+
+
+def _get_onedal_library_names(iface, use_parameters_lib=True, is_win=False):
+    """Return the oneDAL libraries a backend needs, in link order.
+
+    Names are the platform-independent stems; ``get_onedal_libraries`` and
```
david-cortes-intel:
```suggestion
    Names are the platform-independent stems, ``get_onedal_libraries`` and
```

scripts/build_backend.py
```diff
@@ -28,6 +28,133 @@
...
+        raise ValueError(f"Unsupported oneDAL backend: {iface}")
+
+    iface_library = _ONEDAL_IFACE_LIBRARY[iface]
+    library_names = []
+    if iface_library is not None:
+        library_names.append(iface_library)
+        if use_parameters_lib:
+            is_dpc = iface_library == "onedal_dpc"
+            if is_win:
+                # On Windows the parameters entry points ship in the core
```
david-cortes-intel:
```suggestion
                # On Windows, the parameters library entry points come from the core
```

scripts/build_backend.py
```diff
@@ -28,6 +28,133 @@
...
+
+
+def _to_windows_import_name(name):
+    return name if name.endswith("_dll") else f"{name}_dll"
+
+
+def get_onedal_arch_dir(machine):
+    # Keyed by what ``platform.machine()`` reports, which is not the same string
+    # for the same hardware on every platform: on Windows it goes through
+    # ``_get_machine_win32()`` and answers ``AMD64``, while on Linux and macOS it
```
david-cortes-intel: Would this return ARM64 on windows-aarch64?

scripts/build_backend.py
```diff
@@ -28,6 +28,133 @@
...
+    # answers ``x86_64``. Both have to select ``intel64``, and both spellings
+    # reach here - at build time from ``plt.machine()`` below, and at import time
+    # from the same call in ``onedal/__init__.py`` and ``daal4py/__init__.py``.
+    return {"x86_64": "intel64", "AMD64": "intel64", "aarch64": "arm"}.get(
```
david-cortes-intel: This is also defined here btw:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/7a3faf7adc9a2ec5428c20395002adcc73e7fe6a/tests/run_examples.py#L64

scripts/build_backend.py
napetrov: Yes - good catch, that is a real gap. `platform.machine()` answers `ARM64` on Windows-on-ARM: `_get_machine_win32()` reads the WMI CPU architecture and its table spells code 12 `ARM64`, the same way it spells x86-64 `AMD64` rather than `x86_64`. Only the Linux/macOS spelling was in the dict, so such a build would have fallen through to `ARM64` as the directory name and looked for oneDAL somewhere that does not exist.

Added the key and a test case in [`aaac9ca0`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3360/commits/aaac9ca0).

scripts/build_backend.py
napetrov: It was, and with the same `ARM64` gap - which is exactly the drift worth removing: the build would look in one oneDAL directory and the example runner in another. `tests/run_examples.py` now loads `get_onedal_arch_dir` from the build driver instead of keeping its own copy, the same way `tests/test_build_backend.py` already does ([`aaac9ca0`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3360/commits/aaac9ca0)).

Two more copies of the mapping exist at runtime, in `onedal/__init__.py` and `daal4py/__init__.py`. Those cannot share this helper - it lives in `scripts/`, which is build-time only and is not installed - so deduplicating them means adding a small runtime helper to the installed package. Happy to do that here if you want it, or leave it for its own PR since it touches the import path of both packages.

tests/run_examples.py
```diff
@@ -60,9 +62,17 @@
...
+_SPEC = importlib.util.spec_from_file_location(
+    "sklearnex_build_backend", _BUILD_BACKEND_PATH
+)
+_build_backend = importlib.util.module_from_spec(_SPEC)
+_SPEC.loader.exec_module(_build_backend)
+
+arch_dir = _build_backend.get_onedal_arch_dir(plt.machine())
 assert 8 * struct.calcsize("P") in [32, 64]
 
 if 8 * struct.calcsize("P") == 32:
```
david-cortes-intel: Are these actually needed?

napetrov: No - and neither is the loader above them. In this file `arch_dir` names exactly one thing, `tests/_results/<arch>`: a log directory nothing reads. It is gitignored, and no CI pipeline invokes this script - `doc/sources/tests.rst` documents it as a manual helper. The `struct.calcsize` assert and the `ia32` branch were the last 32-bit remnants left in the repository, kept alive only by that path.

So the better answer to your earlier note about the mapping being duplicated here is to delete the second copy rather than share it: `logdir = jp(runner_dir, "_results")`, and the `importlib` block, the `plt`/`struct`/`Path` imports and the 32-bit branch go with it ([`8dfa8177`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3360/commits/8dfa8177)). `get_onedal_arch_dir()` keeps its two real users, `scripts/build_backend.py` and `tests/test_build_backend.py`.

The two runtime copies of the mapping in `onedal/__init__.py` and `daal4py/__init__.py` are untouched - those need a helper inside the installed package, so still my preference to do them in their own PR unless you want them here.

david-cortes-intel (approved):

PR #3361, opened 2026-08-07
[bug] Fix reference counting and buffer ownership in the NumPy/NumericTable conversions

napetrov:
Split out of #3325, which reviewers asked to break up. These are reference-counting and lifetime bugs that are present on ordinary GIL builds today; nothing here is free-threading-specific and the ABI is unchanged.

### The `__2daalnt__` protocol converted only the first object per process

`to_daal_nt` held its result in a function-local `static NumericTablePtr ntptr` and guarded it with `if (true || !ntptr)`. The `true ||` short-circuit made the condition constant, so the static was rewritten on every call - but it was still a single shared object, so the pointer handed back to one caller could be replaced underneath it by the next. Two different inputs could not be converted independently in one process. The static is gone.

### The capsule was untyped and its destructor never ran

The capsule carrying a `NumericTablePtr` out of `__2daalnt__` had no name and no destructor. Any object whose `__2daalnt__` returned a capsule that was not subsequently consumed leaked the `shared_ptr` it held, and a capsule from an unrelated source was accepted and its pointer dereferenced. It is now a named `"daal4py.NumericTablePtr"` capsule with a destructor. Unnamed capsules are still accepted for backwards compatibility.

### `NpyNumericTable` decremented a reference it never incremented

The constructor stored the borrowed array without an incref while the destructor called `Py_DECREF` on it. Every table built over a non-contiguous array therefore consumed a reference it did not own, so the caller's array could be freed while the table still pointed into its buffer. The constructor now increfs, with the decref on the failure path, and the destructor is guarded so it does not touch the interpreter during finalization.

### The CSR conversion leaked four references per call

`data_conversion.cpp` fetched `data`, `indices`, `indptr` and the array from `PyArray_FROMANY` and wrapped each in `reinterpret_borrow`. All of those return *new* references, so each conversion leaked them. They now use `reinterpret_steal`.

### Borrowed buffers handed to oneDAL without an owner

Tables built over a NumPy buffer were given a raw pointer with no tie to the object owning the memory. The owner is now carried by a `std::shared_ptr` with a no-op deleter, so the table keeps the array alive for as long as oneDAL holds it.

### Unchecked C-API calls

`PyArray_SimpleNewFromData` and `PyArray_SetBaseObject` failures were ignored, and a failed `PyArray_SetBaseObject` leaked the base it was handed. `PyList_GetItemRef` is used on 3.13+. A small `PyOwnedRef` RAII wrapper replaces the manual incref/decref pairs on the paths that had them.

### Tests

`tests/test_numeric_table_references.py` asserts the ownership rules directly rather than numerical results: the protocol consumes its capsule exactly once over 64 iterations, unnamed capsules still work, non-capsules are rejected, two different inputs convert independently (the regression above), input reference counts are balanced for C-contiguous / F-contiguous / non-contiguous arrays, and the table's data survives dropping the input array and churning the allocator. Two small `_make_nt_capsule_for_testing` / `_roundtrip_nt_for_testing` helpers are exported to make the capsule path reachable from Python.

Marked as draft until CI has run on it.

Depended on by #3325.


src/npy4daal.h
```diff
@@ -35,6 +35,15 @@
     #define PyDataType_FIELDS(descr) ((descr)->fields)
 #endif
 
+inline bool can_decref_python_object()
+{
+    if (!Py_IsInitialized()) return false;
+#if PY_VERSION_HEX >= 0x030D0000
```
icfaust: Add a comment behind for what the hex translates to in ASCII (for maintenance and understanding)
```suggestion
#if PY_VERSION_HEX >= 0x030D0000 // >= Python 3.13
```

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -196,32 +205,37 @@ dal::table convert_to_table(py::object inp_obj,
...
-        if (!(is_array(py_data) && is_array(py_column_indices) && is_array(py_row_indices) &&
-              array_numdims(py_data) == 1 && array_numdims(py_column_indices) == 1 &&
-              array_numdims(py_row_indices) == 1)) {
+        py::object py_data = py::reinterpret_steal<py::object>(PyObject_GetAttrString(obj, "data"));
+        py::object py_column_indices =
+            py::reinterpret_steal<py::object>(PyObject_GetAttrString(obj, "indices"));
+        py::object py_row_indices =
+            py::reinterpret_steal<py::object>(PyObject_GetAttrString(obj, "indptr"));
+        py::object py_shape =
+            py::reinterpret_steal<py::object>(PyObject_GetAttrString(obj, "shape"));
```
icfaust:
```suggestion
        py::object py_data = py::reinterpret_steal<py::object>(obj.attr("data"));
        py::object py_column_indices =
            py::reinterpret_steal<py::object>(obj.attr("indices"));
        py::object py_row_indices =
            py::reinterpret_steal<py::object>(obj.attr("indptr"));
        py::object py_shape =
            py::reinterpret_steal<py::object>(obj.attr("shape"));
```

Can you clarify why `PyObject_GetAttrString` is used rather than pybind11's `attr`? (see line 205 before this for similar use.

src/daal4py.cpp
```diff
@@ -48,25 +49,30 @@
...
 private:
-    PyArrayObject * _ndarray;
+    static std::shared_ptr<PyArrayObject> make_owner(PyArrayObject * a)
+    {
+        Py_INCREF(a);
+        return std::shared_ptr<PyArrayObject>(a, [](PyArrayObject * array) {
+            if (!can_decref_python_object()) return;
+            PyGILState_STATE state = PyGILState_Ensure();
+            Py_DECREF(array);
+            PyGILState_Release(state);
```
david-cortes-intel, line 78: If DECREF sets the GIL, shouldn't INCREF also set it?

src/npy4daal.h
napetrov: Done in 721aa41 — the annotation is on both occurrences of the guard now:

```cpp
#if PY_VERSION_HEX >= 0x030D0000 // >= Python 3.13
```

onedal/datatypes/numpy/data_conversion.cpp
napetrov: Switched to pybind11 in 721aa41, thanks — there was no reason for the raw C API here.

One adjustment to the literal suggestion: `obj.attr("data")` already returns an owning `py::object` (via `object_api::attr` → `getattr`, which does the `PyObject_GetAttr` and adopts the result). Wrapping that in `reinterpret_steal` would take ownership of an already-owned reference and decref it twice on scope exit. `reinterpret_steal` is for adopting a raw `PyObject*` that you got a new reference to from the C API — which is what the *old* code would have needed, and is exactly the bookkeeping that made the raw calls worth removing.

So the committed form is a plain `py::getattr`:

```cpp
// py::getattr returns an owning py::object and raises Python's own
// AttributeError if the lookup fails, so no reference bookkeeping or
// null check is needed here.
py::object py_data = py::getattr(obj, "data");
py::object py_column_indices = py::getattr(obj, "indices");
py::object py_row_indices = py::getattr(obj, "indptr");
py::object py_shape = py::getattr(obj, "shape");
```

This also drops the manual null checks the raw version needed: `py::getattr` raises `AttributeError` itself rather than returning null.

src/daal4py.cpp
napetrov: Yes — good catch, fixed in 721aa41. The two sides are symmetric now:

```cpp
static std::shared_ptr<PyArrayObject> make_owner(PyArrayObject * a)
{
    // Symmetric with the deleter below. Every current caller reaches this from
    // a Python-facing conversion and so already holds the GIL, which makes the
    // Ensure a cheap no-op recursive acquisition; the deleter genuinely needs
    // it, because oneDAL can drop the last table reference on a worker thread.
    // Guarding both sides means neither depends on the other's caller.
    PyGILState_STATE state = PyGILState_Ensure();
    Py_INCREF(a);
    PyGILState_Release(state);
    ...
```

To be explicit about the asymmetry that prompted the original code: the INCREF side is reached only from Python-facing conversions, so the GIL is in fact already held and `PyGILState_Ensure` is a no-op recursive acquisition. The DECREF side is the one that genuinely needs it — oneDAL can drop the last reference to a table on a worker thread with no Python state at all. But relying on that argument means the INCREF's correctness depends on a property of its callers that nothing enforces, so guarding both is the cheaper invariant.

src/daal4py.cpp
```diff
@@ -357,24 +408,26 @@ daal::data_management::NumericTablePtr make_nt(PyObject * obj)
...
+            const char * capsule_name = PyCapsule_GetName(capsule.get());
+            py_err_check();
+            if (capsule_name && std::strcmp(capsule_name, numeric_table_capsule_name) != 0) throw std::invalid_argument("__2daalnt__ returned a capsule with an unsupported name");
+            // Keep accepting the legacy unnamed capsule while producers migrate
+            // to the typed protocol.
+            auto nt = reinterpret_cast<daal::data_management::NumericTablePtr *>(PyCapsule_GetPointer(capsule.get(), capsule_name));
+            py_err_check();
+            auto ntptr = *nt;
+            if (PyCapsule_SetDestructor(capsule.get(), nullptr) < 0) py_err_check();
+            delete nt;
```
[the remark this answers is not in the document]
napetrov: The re-use hazard is real but it predates this PR and is not something the capsule can defend against on its own, so I'd rather not change it here.

`__2daalnt__` is specified to transfer ownership of a heap-allocated `NumericTablePtr`. Upstream already does `delete nt` on it (`src/daal4py.cpp:374` on main), so a second conversion of the same capsule is already a use-after-free today. This PR narrows that window rather than widening it: it validates the capsule name before touching the pointer, and it clears the destructor with `PyCapsule_SetDestructor(capsule.get(), nullptr)` so the pointer is deleted exactly once instead of racing the capsule's own destructor.

On the specific suggestion — `PyCapsule_SetPointer(capsule, nullptr)` is not available as a safe fix, because CPython rejects a NULL pointer with `ValueError` rather than storing it. Storing a sentinel non-NULL value instead would only convert the dangling read into a wild read at a different address, which is not an improvement.

The actual guarantee has to come from the producer: a `__2daalnt__` implementation must return a fresh capsule per call, which is what the protocol requires and what every in-tree implementation does. `sklearnex/tests/test_free_threading.py::test_daal4py_numeric_table_protocol_is_thread_local` pins that down — it round-trips through the protocol 1000 times across 8 threads with a fresh capsule per call and checks each result matches the array it came from, so a producer that returned a shared capsule would fail loudly there.

If you'd like the capsule to be defensively self-invalidating, that's a protocol change — the capsule would need to carry a small owned struct with a consumed flag rather than a bare `NumericTablePtr*` — and it should be its own PR with the producers updated in the same change.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -196,32 +205,36 @@ dal::table convert_to_table(py::object inp_obj,
...
+            PyArray_FROMANY(py_data.ptr(), array_type(py_data.ptr()), 0, 0, NPY_ARRAY_CARRAY));
+        py::object np_column_indices = py::reinterpret_steal<py::object>(
+            PyArray_FROMANY(py_column_indices.ptr(),
                             NPY_UINT64,
                             0,
                             0,
-                            NPY_ARRAY_CARRAY | NPY_ARRAY_ENSURECOPY | NPY_ARRAY_FORCECAST);
-        PyObject *np_row_indices =
-            PyArray_FROMANY(py_row_indices,
+                            NPY_ARRAY_CARRAY | NPY_ARRAY_ENSURECOPY | NPY_ARRAY_FORCECAST));
```
david-cortes-intel: I don't think `NPY_ARRAY_ENSURECOPY` should be required. This array (and the indptr too) is only used as an input to make a 1-base copy later on:
https://github.com/napetrov/scikit-learn-intelex/blob/13724cce8ddef8f22ef57a2bcfc7691e69a022eb/onedal/datatypes/numpy/data_conversion.cpp#L112

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -61,18 +82,13 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
...
-    auto res_table = dal::homogen_table(
+    auto owner = make_python_owner(reinterpret_cast<PyObject *>(np_data));
+    return dal::homogen_table(
         data_pointer,
         row_count,
         column_count,
-        [np_data](const T *data) {
-            Py_DECREF(np_data);
-        },
+        [owner = std::move(owner)](const T *) {},
```
david-cortes-intel: What difference does `std::move` make here?

onedal/datatypes/numpy/data_conversion.cpp
napetrov: Agreed — removed in 13353817. You are right that this array and `indptr` are only read, to build the one-based `dal::array`s that `convert_to_csr_impl` allocates itself (`data_conversion.cpp:109-124`), so nothing ever writes through the caller's buffers and a shared view is safe.

`FORCECAST` stays and does most of the copying anyway: it forces a copy whenever the input dtype is not `uint64`, which is scipy's usual `int32`/`int64` indices. So in practice `ENSURECOPY` only changed behaviour for input that was *already* `uint64` — the one case where the copy was pure overhead rather than an unavoidable cast.

Added a test (`test_csr_conversion_does_not_modify_input`) that the caller's `indices`, `indptr` and `data` come back unchanged, parametrized over `index_dtype` including `uint64` — that is the case where nothing forces a copy on the way in, so a stray write would actually be observable there and nowhere else.

Note the `data` array keeps `NPY_ARRAY_CARRAY` without `ENSURECOPY`, unchanged — that buffer *is* borrowed by the table rather than copied, which is what the owner `shared_ptr` in this PR exists to keep alive.

onedal/datatypes/numpy/data_conversion.cpp
napetrov: It saved one atomic refcount bump plus the matching decrement, and nothing else — but the named local it was moving from existed only in order to be moved from, so the cleaner fix is to remove the local. Done in 13353817:

```cpp
return dal::homogen_table(
    data_pointer,
    row_count,
    column_count,
    [owner = make_python_owner(reinterpret_cast<PyObject *>(np_data))](const T *) {},
    layout);
```

To be precise about what the `std::move` did and did not do. `owner` is a `std::shared_ptr<PyObject>`; the init-capture copy-constructs from it, which is an atomic increment of the control block's refcount, and then `owner` goes out of scope and atomically decrements it. `std::move` turned that into a pointer steal. Real but trivial, and once per conversion — not the reason to write it.

What made it worth changing is that a named-then-moved local invites the question your comment asks, and the answer is only ever "nothing you need to care about". Initializing the capture directly from the temporary gets the same code with no move to justify: the capture is initialized from an rvalue, so there is no copy to elide in the first place. Same in `convert_to_csr_impl`.

While there I added a line on each saying *why* the capture exists — the table borrows the NumPy buffer, so the owner has to outlive every copy of the table — since that is the non-obvious part, unlike the move.

david-cortes-intel (approved):
LGTM, but it could be improved by calculating the base1 arrays for CSR matrices without making an extra copy that would get immediately discarded. Could have templated versions for dtypes and unit-vs-nonunit stride where the base1 arrays would be calculated from the base0 input as it comes.

PR #3362, opened 2026-08-07
[bug] Fix transceiver and MPI lifecycle in the daal4py distributed backend

napetrov:
Split out of #3325, which reviewers asked to break up. These are lifetime and error-handling bugs in the distributed backend that exist on ordinary GIL builds; the only threading-related change is requesting `MPI_THREAD_MULTIPLE`.

### `get_transceiver()` returned a pointer to an object that `daalfini()` could free

`get_transceiver()` returned a raw `transceiver *` aliasing an object owned by a file-scope `shared_ptr`, and `del_transceiver()` reset that `shared_ptr`. Every caller still using the returned pointer - which includes any algorithm in the middle of a collective - was left with a dangling pointer. It now returns a `shared_ptr`, so a caller keeps the transceiver alive for as long as it uses it. All existing call sites bind with `auto`, so this is source-compatible. `del_transceiver()` swaps the global out and destroys the old value *outside* the lifecycle mutex, since destruction calls into MPI.

### `daalfini()` finalized MPI that daal4py did not initialize

`init()` called `MPI_Init()` whenever MPI was not yet initialized and `fini()` called `MPI_Finalize()` unconditionally. In a process where MPI was initialized by someone else - mpi4py being the common case - the first `daalfini()` finalized it, and every subsequent MPI call in the process failed. Init and fini are now reference counted, `MPI_Finalize()` is only called when daal4py owns the initialization, and repeated or concurrent `daalfini()` is idempotent.

### MPI thread support level

Initialization goes through `MPI_Init_thread()` requesting `MPI_THREAD_MULTIPLE`, and queries the level with `MPI_Query_thread()` when someone else initialized. A level below `MPI_THREAD_MULTIPLE` is reported as a `RuntimeWarning` rather than an error, since single-threaded distributed use stays correct - callers who care can escalate it with `-W error`.

### Exceptions escaping destructors

`transceiver::~transceiver()` called `fini()`, which could throw, from a destructor. Both are now `noexcept`; teardown failures are best-effort because there is no caller to report them to.

### Every MPI return code was discarded

`MPI_Send`, `MPI_Recv`, `MPI_Gather(v)`, `MPI_Bcast`, `MPI_Allreduce`, `MPI_Exscan`, `MPI_Comm_size` and `MPI_Comm_rank` all had their return codes dropped, so a failed collective was indistinguishable from a successful one and produced silently wrong results. They are now checked and translated into exceptions carrying `MPI_Error_string`.

### `gather()`

The buffers from `daal_malloc` leaked on every overflow and allocation-failure path, the varying-size offsets were accumulated in `int` and overflowed for large gathers, and an allocation failure on the root left every other rank blocked in `MPI_Gather` until the job was killed. Sizes are now computed in `size_t` with explicit range checks, the buffer is freed on all failure paths, and the root's outcome is agreed across ranks before entering the collective. The hot-path collectives (`bcast`, `reduce_all`, `reduce_exscan`) deliberately keep their validation local, since it is rank-symmetric and paying an extra `MPI_Allreduce` per call inside the k-means convergence loop is not worth it.

### `to_mpi()` mapped unsigned types to signed datatypes

`UINT32` and `UINT64` mapped to `MPI_INT32_T` and `MPI_INT64_T`, which gives wrong answers for `OP_MAX` / `OP_MIN` on values above the signed range.

### Module init

`PyInit_mpi_transceiver` ignored the result of `PyObject_SetAttrString` and leaked the object it created when it failed.

### Tests

* `tests/test_mpi_lifecycle.py` (pytest-mpi) covers the mpi4py-owns-MPI direction: repeated concurrent init/fini cycles must not finalize MPI, and a GIL-holding `num_procs()` waiter must not block distributed lazy initialization.
* `tests/mpi_lifecycle_smoke.py` covers the daal4py-owns-MPI direction. It has to be a standalone script under `mpiexec -n 2`, because it requires that nothing has touched MPI beforehand and MPI cannot be reinitialized after `MPI_Finalize`, so the whole lifecycle has to fit in one process. It runs in the Azure Linux pipeline on the Python 3.13 configuration.

Marked as draft until CI has run on it.

Depended on by #3325.


src/mpi/mpi_transceiver.cpp
```diff
@@ -18,103 +18,224 @@
...
+    {
+        local_error = std::current_exception();
+    }
+
+    int local_failed = local_error ? 1 : 0;
+    int any_failed   = 0;
+    mpi_check(MPI_Allreduce(&local_failed, &any_failed, 1, MPI_INT, MPI_MAX, MPI_COMM_WORLD), "MPI_Allreduce(collective preflight)");
+    if (!any_failed) return;
+    if (local_error) std::rethrow_exception(local_error);
+    throw std::runtime_error("MPI collective validation or allocation failed on another rank");
```
david-cortes-intel, line 118: Python calls that result in errors should throw Python exceptions.

src/mpi/mpi_transceiver.cpp
```diff
@@ -18,103 +18,224 @@
...
+    try
+    {
+        std::lock_guard<std::mutex> lock(m_lifecycle_mutex);
+        if (!m_users || --m_users) return;
+        const bool owns_mpi = m_owns_mpi;
+        m_owns_mpi          = false;
+        m_initialized       = false;
+        if (owns_mpi) mpi_finalize_noexcept();
+    }
+    catch (...)
```
david-cortes-intel, line 195: Shouldn't it report the errors to the user?

david-cortes-intel (commented):
Shouldn't it also test a scenario in which MPI is initialized through a different library?

src/mpi/mpi_transceiver.cpp
napetrov: They do become Python exceptions. Everything in this file is reached from Python only through `src/transceiver.h`, which daal4py declares to Cython with `except +`, so Cython emits a `try`/`catch(...)` around every call and routes it into `__Pyx_CppExn2PyErr()`. I generated the C++ to confirm the mapping rather than take it on trust:

| C++ exception | Python exception |
|---|---|
| `std::bad_alloc` | `MemoryError` |
| `std::bad_cast`, `std::bad_typeid` | `TypeError` |
| `std::domain_error`, `std::invalid_argument` | `ValueError` |
| `std::ios_base::failure` | `IOError` |
| `std::out_of_range` | `IndexError` |
| `std::overflow_error` | `OverflowError` |
| `std::range_error`, `std::underflow_error` | `ArithmeticError` |
| `std::exception` (anything else) | `RuntimeError` |

The `what()` string is preserved as the message in every case, which is why the errors here are constructed with a description of the failing MPI operation plus `MPI_Error_string`'s text.

Checking that table did surface a real bug, though: the two root-rank range checks were throwing `std::out_of_range`, which lands on `IndexError`. An out-of-range root rank is a bad argument, not a bad index, so ed7874d changes both to `std::invalid_argument` → `ValueError`. `mpi_count`'s `std::overflow_error` → `OverflowError` and `DAAL4PY_CHECK_MALLOC`'s `std::bad_alloc` → `MemoryError` were already landing where they should.

I added a comment on the throwing helper so the next reader does not have to re-derive this:

```cpp
// Raises std::runtime_error, which Cython's ``except +`` translates into a
// Python RuntimeError at the daal4py boundary, with this message preserved.
void mpi_check(int code, const char * operation)
```

src/mpi/mpi_transceiver.cpp
napetrov: Yes — fixed in ed7874d. `fini()` still has to be `noexcept` (it runs from a destructor and from interpreter shutdown, where there is no caller left to catch anything), but "cannot throw" is not the same as "must be silent". Failures are now reported:

```cpp
// Report a teardown failure without throwing. fini() runs from a destructor and
// from interpreter shutdown, so there is no caller left to catch an exception -
// but staying silent hides a real fault, so surface it as a Python warning when
// the interpreter can still take one, and on stderr when it cannot.
void report_teardown_failure(const std::string & reason) noexcept
{
    const std::string message = "daal4py could not shut MPI down cleanly: " + reason;

    if (Py_IsInitialized()
#if PY_VERSION_HEX >= 0x030D0000 // >= Python 3.13
        && !Py_IsFinalizing()
#endif
    )
    {
        const PyGILState_STATE state = PyGILState_Ensure();
        // A warning raised here cannot propagate, so clear it rather than
        // leaving an exception set for an unrelated later call to trip over.
        if (PyErr_WarnEx(PyExc_RuntimeWarning, message.c_str(), 1) < 0) PyErr_Clear();
        PyGILState_Release(state);
        return;
    }
    std::fputs((message + "\n").c_str(), stderr);
}
```

`RuntimeWarning` rather than a hard error, because by the time `fini()` runs the computation has already completed and its results are valid; a dirty MPI shutdown is worth surfacing but does not invalidate anything the caller already has. Using the warnings machinery also means it can be escalated with `-W error::RuntimeWarning` by anyone who does want it fatal.

The `PyErr_Clear()` matters: a warning raised from here cannot propagate, so leaving the exception set would hand it to some unrelated later call to trip over.

`MPI_Finalize` and `MPI_Finalized` failures now go through the same path, and the remaining `catch` in `fini()` reports rather than swallows:

```cpp
catch (const std::exception & e)
{
    // Only the lock_guard can get here; MPI failures are already reported by
    // mpi_finalize_noexcept. Reported rather than swallowed so a failure to
    // take the lifecycle mutex is not invisible.
    report_teardown_failure(e.what());
}
```

I compiled the helper against an MPI stub and exercised all four paths (interpreter live, finalizing, never initialized, warning-raise-fails) to confirm none of them re-enter Python at a point where that is unsafe and none leave an exception set.

src/mpi/mpi_transceiver.cpp
```diff
@@ -153,36 +322,45 @@ static MPI_Op to_mpi(transceiver_iface::operation_type o)
...
-    static std::shared_ptr<mpi_transceiver> s_smt;
-    PyObject * m;
+    static std::shared_ptr<mpi_transceiver> transceiver_instance;
     static struct PyModuleDef moduledef = {
-        PyModuleDef_HEAD_INIT, "daal4py.mpi_transceiver", "No docs", -1, NULL,
+        PyModuleDef_HEAD_INIT, "daal4py.mpi_transceiver", "No docs", -1, nullptr,
     };
-    m = PyModule_Create(&moduledef);
-    if (m == NULL) return NULL;
+    PyObject * module = PyModule_Create(&moduledef);
```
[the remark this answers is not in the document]
napetrov: The cast is pre-existing and, in this specific case, well-defined — though only because of a detail worth making explicit, so thanks for flagging it.

Upstream does exactly the same thing: `PyInit_mpi_transceiver` exports `&s_smt` where `s_smt` is a `static std::shared_ptr<mpi_transceiver>` (`src/mpi/mpi_transceiver.cpp:177,186` on main), and `create_transceiver` reinterprets it as `std::shared_ptr<transceiver_iface>*` (`src/transceiver.cpp:62` on main). This PR did not introduce or change either side; it added the null and error checks around the import that upstream was missing.

On the correctness question: you're right that `shared_ptr<Derived>` and `shared_ptr<Base>` are distinct types and the reinterpret_cast is not blessed in general. What makes the dereference work here is that `mpi_transceiver` derives from `transceiver_impl` which derives from `transceiver_iface`, all single non-virtual public inheritance with `transceiver_iface` as the primary base — so the `Derived*` and `Base*` values are numerically identical, and `shared_ptr`'s layout (stored pointer + control block pointer) is identical between the two instantiations. `*iface` then copy-constructs a `shared_ptr<transceiver_iface>` from that reinterpreted object and the control block is shared correctly, so refcounts are not corrupted.

That is a real constraint on the class hierarchy that nothing currently states or checks, which is the part I agree is a defect. Rather than change the ABI across the module boundary in a PR about MPI lifecycle, I'd fix it properly by exporting `std::shared_ptr<transceiver_iface>` — the base type the consumer expects — so no cast is needed at all. That touches the module's public contract and any out-of-tree `D4P_TRANSCEIVER` implementation, so it belongs in its own PR rather than this one.

In the meantime I'd rather not leave the constraint undocumented, so I can add a `static_assert` next to the export pinning what the cast actually relies on — that `transceiver_iface` is a non-virtual base of `mpi_transceiver` — which turns a future hierarchy change from silent UB into a compile error. Say the word and I'll add it here; otherwise I'll fold it into the follow-up.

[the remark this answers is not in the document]
napetrov: Added it — 24ced208. Pinning the constraint here rather than folding it into the follow-up, since the follow-up changes the module's public contract and should not be a prerequisite for documenting what today's code already relies on.

Two `static_assert`s for the parts the type system can state, and a runtime check for the part it cannot:

```cpp
static_assert(std::is_base_of<transceiver_iface, mpi_transceiver>::value,
              "mpi_transceiver must derive from transceiver_iface for the exported pointer cast");
static_assert(sizeof(std::shared_ptr<mpi_transceiver>) == sizeof(std::shared_ptr<transceiver_iface>),
              "shared_ptr layout must match between mpi_transceiver and transceiver_iface");
```

Neither of those can express "the `transceiver_iface` subobject sits at offset zero", which is the assumption the reinterpret_cast actually depends on — `std::is_base_of` is satisfied by a multiple-inheritance hierarchy where the base is *not* first, and that is the case that would silently corrupt the control block. So `PyInit_mpi_transceiver` compares the two pointer values once at import and fails with an `ImportError` if they differ, rather than letting the first collective run on a misaligned object.

I checked the guard both ways rather than assuming it fires, by reproducing the hierarchy shape standalone (single non-virtual inheritance chain `mpi_transceiver` → `transceiver_impl` → `transceiver_iface` → `transceiver_base_iface`) and running the consumer's exact cast against it:

* as it is today: offsets match, and the `shared_ptr<transceiver_iface>` copied out of the reinterpreted object shares the control block correctly (`use_count` tracks, stored pointer matches)
* with an extra polymorphic base inserted before `transceiver_impl`: offsets differ and the check fails — i.e. the guard catches exactly the hierarchy change that would break the cast

The proper fix is still to export `std::shared_ptr<transceiver_iface>` so no cast is needed; that touches any out-of-tree `D4P_TRANSCEIVER` implementation and stays a separate PR.

doc/sources/daal4py.rst
```diff
@@ -35,8 +35,11 @@ Documentation for functions that control the global thread settings in ``daal4py
 MPI helpers
 -----------
 
-Documentation for helper functions that can be used in distributed mode, particularly when using MPI without |mpi4py|.
-See :ref:`distributed_daal4py` for examples.
+Documentation for helper functions that can be used in distributed mode.
+See :ref:`distributed_daal4py` for examples. ``daalinit`` configures daal4py
+threads only; it does not initialize MPI. Distributed computations initialize
+their transceiver lazily. ``daalfini`` releases that transceiver after the
+computation, and finalizes MPI only when daal4py initialized MPI itself.
```
david-cortes-intel: I don't think these kinds of details are relevant for this page.

doc/sources/distributed_daal4py.rst
```diff
@@ -95,6 +95,11 @@ From the ``daal4py`` side, in order to use distributed mode, algorithm construct
 be passed argument ``distributed=True``, method ``.compute()`` should be passed the right
 subset of the data for each node, and after the distributed computations are finalized,
 function :obj:`daal4py.daalfini` must be called before accessing the results object.
+``daalinit`` only configures daal4py thread settings: distributed computations
+initialize their MPI transceiver lazily, so it is not needed to initialize MPI.
+``daalfini`` releases that transceiver and finalizes MPI only when daal4py
```
david-cortes-intel: This is ambiguous, especially considering the paragraphs above.

generator/wrapper_gen.py
```diff
@@ -183,19 +186,17 @@ def daalinit(nthreads: int = -1) -> None:
...
+    objects. It releases daal4py's distributed transceiver. When daal4py
+    initialized MPI itself, it also finalizes MPI; when MPI was already
+    initialized (for example by ``mpi4py``), it leaves MPI ownership and
+    finalization to that library.
 
-    Note that software ``mpi4py`` calls this function automatically if it is imported, but
-    it only does so upon process exit, so this still needs to be called before accessing
-    the result objects in the process/rank that will use them.
+    This function is a no-op when no daal4py distributed transceiver is active.
+    ``daalinit`` only configures daal4py thread settings; it is not required to
```
david-cortes-intel: This sounds illogical, perhaps could be reworded.

generator/wrapper_gen.py
```diff
@@ -166,7 +166,10 @@ def __dealloc__(self):
 
 def daalinit(nthreads: int = -1) -> None:
     '''
-    Set number of threads for daal4py
+    Set number of threads for daal4py.
+
+    This function does not initialize MPI or distributed mode. The distributed
```
david-cortes-intel: Could be put as a note at the end of this description instead.

generator/wrapper_gen.py
```diff
@@ -183,19 +186,17 @@ def daalinit(nthreads: int = -1) -> None:
...
-    run through MPI (used for distributed mode).
+    Release daal4py distributed resources.
 
-    This is a wrapper over ``MPI_Finalize``. It does not need to be called if ``mpi4py``
-    was imported before, as ``mpi4py`` calls this function upon process exit.
+    Call this after distributed computations and before accessing their result
+    objects. It releases daal4py's distributed transceiver. When daal4py
+    initialized MPI itself, it also finalizes MPI; when MPI was already
+    initialized (for example by ``mpi4py``), it leaves MPI ownership and
+    finalization to that library.
```
david-cortes-intel: Can the transceiver be initialized more than once?

doc/sources/distributed_daal4py.rst
```diff
@@ -95,6 +95,12 @@ From the ``daal4py`` side, in order to use distributed mode, algorithm construct
 be passed argument ``distributed=True``, method ``.compute()`` should be passed the right
 subset of the data for each node, and after the distributed computations are finalized,
 function :obj:`daal4py.daalfini` must be called before accessing the results object.
+For this multi-process execution, user code does not need to call ``MPI_Init``
+or :obj:`daal4py.daalinit`: the first distributed operation initializes its
+transceiver lazily. If an MPI runtime was already initialized, for example by
+``mpi4py``, ``daal4py`` uses it and :obj:`daal4py.daalfini` leaves finalization
```
david-cortes-intel:
```suggestion
|mpi4py|, ``daal4py`` uses it and :obj:`daal4py.daalfini` leaves finalization
```

doc/sources/distributed_daal4py.rst
```diff
@@ -95,6 +95,12 @@ From the ``daal4py`` side, in order to use distributed mode, algorithm construct
...
+to its initializer. Otherwise, ``daalfini`` finalizes the MPI runtime that
```
david-cortes-intel:
```suggestion
to its initializer. Otherwise, :obj:`daal4py.daalfini` finalizes the MPI runtime that
```

doc/sources/distributed_daal4py.rst
```diff
@@ -84,8 +84,8 @@ same algorithms to much larger problem sizes.
 Using distributed mode
 ----------------------
 
-In order to use distributed mode, the code to execute must be saved in a Python file,
-and this Python file must be executed through an MPI runner (``mpiexec`` / ``mpirun``).
+To run distributed code across multiple processes, save it in a Python file and
```
david-cortes-intel:
```suggestion
To run distributed code across multiple processes, store the code in a Python file and
```

doc/sources/distributed_daal4py.rst
```diff
@@ -186,7 +194,8 @@ Same example calling MPI functionalities from ``mpi4py`` instead:
     qr_algo = daal4py.qr(distributed=True)
     qr_result = qr_algo.compute(X_node)
 
-    daal4py.daalfini() # call before accessing results
+    # Releases daal4py's transceiver; mpi4py retains ownership of MPI.
```
david-cortes-intel:
```suggestion
    # Releases daal4py's transceiver, mpi4py retains ownership of MPI.
```

generator/wrapper_gen.py
```diff
@@ -177,25 +177,25 @@ def daalinit(nthreads: int = -1) -> None:
 
     :param int nthreads: [default: -1] Number of threads to use for further computations in daal4py. If this number is less or equal than zero, then settings will not be changed.
 
+    .. note:: This function configures thread settings only. It does not
+              initialize MPI or distributed mode; the transceiver is
```
david-cortes-intel:
```suggestion
              initialize MPI or distributed mode, the transceiver is
```

generator/wrapper_gen.py
```diff
@@ -177,25 +177,25 @@ def daalinit(nthreads: int = -1) -> None:
...
-    Note that software ``mpi4py`` calls this function automatically if it is imported, but
-    it only does so upon process exit, so this still needs to be called before accessing
-    the result objects in the process/rank that will use them.
+    Release daal4py distributed resources.
+
+    Call this after the last distributed computation and before accessing its
+    result objects. It releases daal4py's distributed transceiver. If daal4py
+    initialized MPI, releasing the last transceiver finalizes MPI and another
+    distributed operation cannot be started in the same process. If MPI was
+    initialized by another library, for example ``mpi4py``, it remains owned by
```
david-cortes-intel:
```suggestion
    initialized by another library, for example |mpi4py|, it remains owned by
```

generator/wrapper_gen.py
```diff
@@ -177,25 +177,25 @@ def daalinit(nthreads: int = -1) -> None:
...
+    that library; later distributed operations can create a new transceiver.
```
david-cortes-intel: It doesn't explain how to create a new transceiver.

doc/sources/distributed_daal4py.rst
```diff
@@ -95,6 +95,12 @@ From the ``daal4py`` side, in order to use distributed mode, algorithm construct
 be passed argument ``distributed=True``, method ``.compute()`` should be passed the right
 subset of the data for each node, and after the distributed computations are finalized,
 function :obj:`daal4py.daalfini` must be called before accessing the results object.
+For this multi-process execution, user code does not need to call ``MPI_Init``
+or :obj:`daal4py.daalinit`: the first distributed operation initializes its
+transceiver lazily. If an MPI runtime was already initialized, for example by
```
david-cortes-intel, line 100: Does it need to initialize an MPI communicator in particular?

generator/wrapper_gen.py
napetrov: Yes, but only when `daal4py` does not own MPI.

`get_transceiver()` (`src/transceiver.cpp`) creates it lazily and `del_transceiver()` drops it, so a later distributed operation builds a new one. The `mpi_transceiver` itself outlives that wrapper though - it is a function-local `static std::shared_ptr<mpi_transceiver>` in `PyInit_mpi_transceiver` - so the second wrapper re-runs the full `init()` path with `m_users` back at 0 and `m_owns_mpi` cleared. From there the two cases diverge: if `daal4py` was the one that called `MPI_Init_thread`, `fini()` has already finalized MPI and `init()` raises `"MPI cannot be reinitialized after MPI_Finalize"`; if `mpi4py` owns MPI, nothing was finalized and re-creation succeeds.

That asymmetry is what the `daalfini` docstring now describes. Leaving this thread open in case you want the wording to say more than it does.

generator/wrapper_gen.py
napetrov: Right - there is no function that creates one. Added that to the docstring in [`d8e55b9a`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/d8e55b9a660aa1e64e8d6ed826a255e1676b38a0): the next `.compute()` of an algorithm constructed with `distributed=True` creates it lazily, the same way the first one was created.

doc/sources/distributed_daal4py.rst
napetrov: No. `daal4py` only checks whether MPI has been initialized at all - `MPI_Initialized()` in `mpi_transceiver::init()` - and runs every collective on `MPI_COMM_WORLD`, so any library that has called `MPI_Init` or `MPI_Init_thread` is enough and no communicator has to be set up. Said so explicitly in [`d8e55b9a`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/d8e55b9a660aa1e64e8d6ed826a255e1676b38a0).

doc/sources/distributed_daal4py.rst
napetrov: Your suggestions on the lines below rewrite this sentence; accepted verbatim in [`d8e55b9a`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/d8e55b9a660aa1e64e8d6ed826a255e1676b38a0).

src/mpi/mpi_transceiver.cpp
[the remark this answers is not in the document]
napetrov: Tracked as #3375 for the proper fix - exporting `std::shared_ptr<transceiver_iface>` so there is no cast at all - since that changes the module's public contract and any out-of-tree `D4P_TRANSCEIVER` implementation along with it.

What `24ced208` added here is the guard for today's code: two `static_assert`s for the parts the type system can state, plus a runtime comparison of the two pointer values at import, because `std::is_base_of` is also satisfied by a hierarchy where the base is not first - which is precisely the case that would corrupt the control block silently.

generator/wrapper_gen.py
```diff
@@ -193,8 +193,11 @@ def daalfini() -> None:
     result objects. It releases daal4py's distributed transceiver. If daal4py
     initialized MPI, releasing the last transceiver finalizes MPI and another
     distributed operation cannot be started in the same process. If MPI was
-    initialized by another library, for example ``mpi4py``, it remains owned by
-    that library; later distributed operations can create a new transceiver.
+    initialized by another library, for example |mpi4py|, it remains owned by
+    that library, and later distributed operations can create a new transceiver.
+    There is no function that creates one: as before, the next call to a
+    ``.compute()`` of an algorithm constructed with ``distributed=True`` creates
+    it lazily.
```
david-cortes-intel, line 202: Please add a test in which there'd be two calls to distributed `.compute()`, with the second one happening after a `daalfini()` call. You'll need to mark them with `@pytest.mark.mpi`.

tests/test_mpi_lifecycle.py
```diff
@@ -0,0 +1,127 @@
...
+
+# Must precede the daal4py import: this is what makes mpi4py the owner of MPI.
+MPI = pytest.importorskip("mpi4py.MPI", exc_type=ImportError)
+
+import daal4py
+
+pytest.importorskip("daal4py.mpi_transceiver", exc_type=ImportError)
+
+
+def test_mpi_is_externally_owned():
```
david-cortes-intel, line 50: Please mark these with `@pytest.mark.mpi`.

.ci/pipeline/build-and-test-lnx.yml
```diff
@@ -73,6 +73,16 @@ steps:
       DPCTL_VERSION: 0.22.1
       DPNP_VERSION: 0.20.0
     displayName: "Install testing requirements"
+  - script: |
+      . /usr/share/miniconda/etc/profile.d/conda.sh
+      conda activate CB
+      conda install -q -y -c conda-forge mpi4py
```
david-cortes-intel: @ethanglaser @icfaust Is IMPI guaranteed to be installed for these CI jobs?

src/mpi/mpi_transceiver.cpp
```diff
@@ -18,103 +18,274 @@
...
+    )
+    {
+        const PyGILState_STATE state = PyGILState_Ensure();
+        // A warning raised here cannot propagate, so clear it rather than
+        // leaving an exception set for an unrelated later call to trip over.
+        if (PyErr_WarnEx(PyExc_RuntimeWarning, message.c_str(), 1) < 0) PyErr_Clear();
+        PyGILState_Release(state);
+        return;
+    }
+    std::fputs((message + "\n").c_str(), stderr);
```
david-cortes-intel, line 76: This `stderr` is not the same as the Python error stream.

src/transceiver.cpp
```diff
@@ -15,66 +15,85 @@
...
+
+[[noreturn]] void throw_python_error(const char * message)
+{
+    if (PyErr_Occurred()) PyErr_Print();
+    throw std::runtime_error(message);
+}
+
+std::shared_ptr<transceiver> create_transceiver()
+{
+    const char * modname = std::getenv("D4P_TRANSCEIVER");
```
david-cortes-intel, line 65: * Please document this.
* What's the use-case for this?

src/transceiver.cpp
```diff
@@ -15,66 +15,85 @@
...
+    if (!modname) modname = "daal4py.mpi_transceiver";
+
+    pyobject_ptr mod(PyImport_ImportModule(modname));
```
david-cortes-intel, line 68: Shouldn't this require the GIL?

david-cortes-intel (changes requested):

generator/wrapper_gen.py
napetrov: Added in [`9b8fccf8`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/9b8fccf8) as `test_distributed_compute_works_again_after_daalfini`: two `daal4py.qr(distributed=True).compute()` calls with a `daalfini()` between them, comparing the two R factors so it holds for any rank count. The whole module is now marked `mpi` (see below), so it picks that up.

The daal4py-owned mirror of this case - where the second attempt *must* fail, because MPI cannot be reinitialized after `MPI_Finalize` - is already asserted at the end of `tests/mpi_lifecycle_smoke.py`. That one has to stay a standalone `mpiexec` script rather than a pytest module, since it needs a process where nothing has touched MPI beforehand.

tests/test_mpi_lifecycle.py
napetrov: Done in [`9b8fccf8`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/9b8fccf8), as a module-level `pytestmark = pytest.mark.mpi` rather than a decorator per test - everything in the file drives MPI, so the mark belongs to the module. That makes it select and skip the same way the SPMD suites do.

src/mpi/mpi_transceiver.cpp
napetrov: Agreed, and that is why it is the fallback rather than the primary path. The branch above it prefers `PyErr_WarnEx`, which does go through the Python warning machinery, and `stderr` is reached only when the interpreter cannot take a warning at all - `!Py_IsInitialized()`, or `Py_IsFinalizing()` on 3.13+.

In that state there is no Python stream to write to: `PySys_WriteStderr` needs both the GIL and a live `sys.stderr`, and this function is `noexcept` and runs from a destructor and from interpreter shutdown. So the choice there is C `stderr` or silence, and silence would hide a genuine MPI teardown fault. Happy to reword the comment if it reads as though `stderr` were meant to be equivalent to the Python stream - it is meant as the last resort when the Python one is gone.

src/transceiver.cpp
napetrov: Documented in [`9b8fccf8`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/9b8fccf8). It is pre-existing - `main` reads the same variable in `get_transceiver()` - and the use case is swapping in an out-of-tree communication backend without rebuilding daal4py: any module exporting a `transceiver` attribute that holds the address of a `std::shared_ptr<transceiver_iface>` can be named through it. Nothing in this repository sets it, which is why it needed saying out loud; it is honored so that such a build keeps working.

It is also the reason the export type in #3375 is a contract change rather than an internal detail.

src/transceiver.cpp
napetrov: It does, and it is held: `create_transceiver()` is in an anonymous namespace with exactly one caller, `get_transceiver()`, which takes `gil_state_guard gil` before calling in - so the import, the attribute lookup and `PyLong_AsVoidPtr` all run under the GIL. `main` did the same thing with a `PyGILState_Ensure` inline in that function; this PR moved it out so the guard also covers publishing the initialization state, which is what keeps the lock order consistent on free-threaded builds.

Stated as a precondition in the comment now ([`9b8fccf8`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/9b8fccf8)), since it was only implied by the call site before.

.ci/pipeline/build-and-test-lnx.yml
napetrov: Not answering for you two, but the facts about what that job does today, since it is this PR's job:

```yaml
conda install -q -y -c conda-forge mpi4py
mpiexec -n 2 python tests/mpi_lifecycle_smoke.py
```

So it does not rely on IMPI being present - it pulls `mpi4py` and whatever MPI runtime conda-forge selects for it, and the step is gated on `RUN_MPI_TESTS == '1'`.

That is worth a second look though, and it is the same trap this PR's own docs warn about: `conda install -c conda-forge mpi4py` without an `mpi=*=impi` pin can resolve to MPICH or OpenMPI, while the library under test is built against the MPI backend from `DALROOT`. `doc/sources/distributed_daal4py.rst` explicitly tells users to pin `mpi=*=impi` for exactly this reason, so the CI step arguably ought to do what the docs ask:

```yaml
conda install -q -y -c conda-forge mpi4py 'mpi=*=impi'
```

I have not changed it, because if these jobs are meant to run against a non-Intel MPI on purpose then pinning would be wrong, and you two know which it is. Say which and I will make the change here.

tests/test_mpi_lifecycle.py
```diff
@@ -41,6 +41,10 @@
 
 pytest.importorskip("daal4py.mpi_transceiver", exc_type=ImportError)
 
+# Everything in this module drives MPI, so it is selectable and skippable as a
+# unit the same way the SPMD suites are.
+pytestmark = pytest.mark.mpi
```
david-cortes-intel: Please add `@pytest.mark.mpi` manually to each test so that it'd be greppable.

tests/test_mpi_lifecycle.py
```diff
@@ -41,6 +41,10 @@
...
+
 
 def test_mpi_is_externally_owned():
```
david-cortes-intel, line 50: You'll need to add these test files here:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/7a3faf7adc9a2ec5428c20395002adcc73e7fe6a/conda-recipe/run_test.sh#L86

And similarly for the windows counterpart.

src/transceiver.cpp
david-cortes-intel: @napetrov Please document it in the online documentation for daal4py:
https://uxlfoundation.github.io/scikit-learn-intelex/2026.1/about_daal4py.html#distributed-mode

tests/test_mpi_lifecycle.py
napetrov: Done in [`bcdd98ff`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/bcdd98ff) - the module-level `pytestmark` is gone and each of the six tests carries `@pytest.mark.mpi` itself, so `grep -n "mark.mpi"` lists them all.

tests/test_mpi_lifecycle.py
napetrov: Added in [`bcdd98ff`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/bcdd98ff), to `conda-recipe/run_test.sh` and `run_test.bat` both. Worth knowing that `test_mpi_lifecycle.py` was already being collected by the plain `pytest -s "${sklex_root}/tests"` run at the top of those scripts, where pytest-mpi skips it for lack of `--with-mpi`; the new step under the MPI launcher is what actually runs it. `mpi_lifecycle_smoke.py` is launched directly rather than through `helper_mpi_tests.py`, because that helper imports `mpi4py` and would take MPI ownership away from `daal4py` - which is the exact path that script exists to cover.

Two things surfaced while wiring it up, both fixed in the same commit:

* `tests/mpi_lifecycle_smoke.py` asserted exactly two ranks, so it would have failed under this script's `-n 4` Intel MPI path. It now works at any rank count above one.
* `tests/helper_mpi_tests.py` called `pytest.main(...)` and discarded the result, so it always exited zero: every MPI pytest step in `run_test.sh` and `run_test.bat`, including the two pre-existing spmd ones, reported success no matter what failed. It now exits with pytest's status, per your and my exchange on gating - flagging it explicitly because it can surface pre-existing failures in those two older steps, and if you would rather keep this PR narrow I will split it back out.

On Windows this stays inert either way: `conda-recipe/meta.yaml` sets `NO_DIST=1  # [win]`, and `mpi4py`, `pytest-mpi` and `mpi * impi` are all `# [not win]` in the test requirements, so the block never executes in the recipe. Added for symmetry rather than for coverage.

Incidentally, that same `meta.yaml` does pin `mpi * impi` for the test environment, which is the pin the Azure step in the other thread lacks.

src/transceiver.cpp
napetrov: Documented in [`bcdd98ff`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/bcdd98ff). The section you linked is the stub in `about_daal4py.rst` that `:ref:`s to `distributed_daal4py`, so the text went on the linked page as "Selecting the communication backend" - the default module, what `D4P_TRANSCEIVER` overrides it with, and the contract a module has to satisfy (an attribute `transceiver` holding the address of a `std::shared_ptr<transceiver_iface>`, read once per transceiver creation). The stub now names the variable and points there instead of just saying "see the section".

It carries a warning that this is an advanced, unsupported extension point and that the exported type is expected to change, with a link to #3375 - documenting the variable makes it discoverable, and I would rather not have it read as a stable interface right as that issue changes the type. Say if you would rather have the whole thing inline in `about_daal4py` instead of on the page it links to.

conda-recipe/run_test.sh
```diff
@@ -99,6 +99,14 @@ if [[ ! $NO_DIST ]]; then
...
         pytest ${PYTEST_CONFIG} -s "${sklex_root}/tests/test_daal4py_spmd_examples.py" $@ $(generate_pytest_args mpi_legacy)
     return_code=$(($return_code + $?))
+    mpirun ${EXTRA_MPI_ARGS} python "${sklex_root}/tests/helper_mpi_tests.py" \
+        pytest --with-mpi ${PYTEST_CONFIG} -s "${sklex_root}/tests/test_mpi_lifecycle.py" $@ $(generate_pytest_args mpi_lifecycle)
+    return_code=$(($return_code + $?))
+    # Launched directly rather than through the helper above: this one covers the
+    # daal4py-owned MPI path, which requires that nothing - mpi4py included - has
+    # initialized MPI before daal4py does.
+    mpirun ${EXTRA_MPI_ARGS} python "${sklex_root}/tests/mpi_lifecycle_smoke.py"
+    return_code=$(($return_code + $?))
```
david-cortes-intel: These should go through the helper file. See comments about what it does:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/44f9e332e599d46fe4888b9cb7b15e9cc26d8414/tests/helper_mpi_tests.py#L17-L22

david-cortes-intel: Actually I see the issue now. Perhaps you could find some other way of achieving the same without mpi4py. Daal4py should also have functions to query the rank.

napetrov: Done - `tests/mpi_lifecycle_smoke.py` no longer imports `mpi4py` at all ([`2cd89a14`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/2cd89a14)). You are right that it was self-defeating: the import initializes MPI, which is the one thing this script needs *not* to have happened before `daal4py` does, so using it as the observer worked against the case under test.

Each use is replaced by something that cannot take ownership:

* rank and size from `daal4py.num_procs()` / `my_procid()`, cross-checked against what the launcher publishes in the environment (`PMI_RANK`/`PMI_SIZE`, or Open MPI's own names) rather than against `MPI.COMM_WORLD` - an oracle independent of daal4py that initializes nothing;
* the thread support level by promoting to an error the `RuntimeWarning` the transceiver already raises when the library provides below `MPI_THREAD_MULTIPLE`, instead of `MPI_Query_thread` through mpi4py;
* the pre-teardown barrier and the "MPI is still usable" check by one real distributed `covariance` computation, whose collectives synchronize the ranks;
* `MPI.Is_finalized()` by the reinitialization failure already asserted at the end - `"MPI cannot be reinitialized after MPI_Finalize"` can only come from `MPI_Finalized()` being true, so it proves the same thing.

On routing it through `helper_mpi_tests.py`: that helper is a pytest launcher - it strips the `pytest` token and the json-report arguments and calls `pytest.main(...)` - so a plain script has nothing to hand it, and there is no json report to gate on rank zero.

And it cannot become a pytest module either, which is the part I should have stated in the first place instead of only naming the helper's own mpi4py import. With `--with-mpi`, `pytest-mpi` does `from mpi4py import MPI; comm = MPI.COMM_WORLD` in [`pytest_runtest_setup`](https://github.com/aragilar/pytest-mpi/blob/master/src/pytest_mpi/__init__.py#L127-L145), before the body of any `mpi`-marked test runs, and that import initializes MPI. So under `--with-mpi` the daal4py-owned path is unreachable by construction, independently of what this file imports.

Happy to switch the helper's rank detection from mpi4py to those same launcher environment variables if you want it to stop initializing MPI - it only needs the rank to decide who writes the json report. That is a separate change though, and it would not move this script, since it still is not pytest.

.ci/pipeline/build-and-test-lnx.yml
napetrov: Answering my own question with data, since the run settled it: **no, IMPI is not installed in these jobs, and it does not need to be.**

`conda list` in "Install testing requirements" on `LinuxCondaEnv Python3.13_Sklearn1.9` shows `mpi 1.0.1 mpich` / `mpich 5.0.1`, so `mpiexec` there is MPICH's and `conda install -c conda-forge mpi4py` resolves against it. The two-rank lifecycle step ran under exactly that combination and passed ([build 62787](https://dev.azure.com/daal/scikit-learn-intelex/_build/results?buildId=62787), task "MPI two-rank lifecycle smoke", `succeeded`), which is the expected outcome: daal4py's Intel MPI build is ABI compatible with MPICH, and `doc/sources/distributed_daal4py.rst` already says only non-MPICH-compatible backends such as OpenMPI need a source build.

So adding `mpi=*=impi` would not fix anything here - it would *change* the backend these jobs currently exercise. I left it unpinned and said why in the step's comment.

While there, [`f740be52`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/f740be52) makes that step run both halves rather than only the smoke script: `tests/test_mpi_lifecycle.py` through `helper_mpi_tests.py` with `--with-mpi` for the mpi4py-owned case, then the smoke script in its own process for the daal4py-owned one. That is worth doing because the recipe jobs you asked me to wire these into are red repo-wide right now on a channel-solve failure - identically on #3386, #3383, #3381 and #3379 - so the recipe wiring proves nothing until that clears, while this pipeline actually executes them. The step also got `set -e`, since an Azure `script:` block otherwise reports only its last command's status, and the smoke script now prints one line per rank - it asserts silently, so a skipped step and a passing one used to produce identical logs.

tests/mpi_lifecycle_smoke.py
```diff
@@ -104,3 +104,11 @@ def query_topology(_):
...
 else:
     raise AssertionError("distributed use unexpectedly reinitialized finalized MPI")
+
+# Everything above asserts silently, which in CI is indistinguishable from a step
+# that never ran the script. Say what was covered.
+print(
+    f"rank {rank} of {size}: daal4py-owned MPI initialized, used from threads, "
+    "torn down once, and confirmed unusable afterwards",
+    flush=True,
+)
```
david-cortes-intel: There will be a race condition when printing this. Better to print only from one rank.

napetrov: Done in [`96254261`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/96254261) — the summary now prints from rank zero only. It can speak for the whole run because the distributed `covariance` computation above it is a collective, so no rank reaches that line unless every rank took part, and a rank that failed anywhere fails the launcher whether or not it printed.

Keeping a print at all is what distinguishes a passing step from a skipped one in the log, since everything before it asserts silently.

tests/test_mpi_lifecycle.py
napetrov: Follow-up on the Windows counterpart: `WindowsCondaRecipe` went red on the last run, and it is worth reporting what was underneath it, because the block you pointed me at was already running there — as a no-op.

Fixed in [`d876ac47`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/d876ac47).

<details>
<summary>What was happening</summary>

Per `doc/sources/tests.rst`, distributed tests on Windows expect the launcher to come from `$PYTHON` (`set "PYTHON=mpiexec -n 2 python"`), and `run_test.bat` guards the no-launcher case with `if "%PYTHON%"=="python"`. That only matches the literal string. conda-build's test phase defines `PYTHON` as `%PREFIX%\python.exe`, which slips through, so `NO_DIST` arrived empty — the recipe log shows `"NO_DIST="` — and the distributed block ran with a bare interpreter and no MPI stack at all, since `meta.yaml` marks `mpi`, `mpi4py` and `pytest-mpi` as `# [not win]`.

That reported success rather than failure: `pytest` rejects `--with-mpi` as an unknown argument when `pytest-mpi` is not installed, and `helper_mpi_tests.py` discarded `pytest.main()`'s return value. On current main's Windows recipe leg you can see the consequence directly — the helper banner prints, no pytest session follows, and the script still exits zero. So the pre-existing `sklearnex_spmd` and `mpi_legacy` steps have never run on Windows.

This surfaced now for two reasons: `helper_mpi_tests.py` propagating pytest's status in this PR turns those silent no-ops into real failures, and `LinuxCondaRecipe` only just started passing after the conda channel state healed, which is what got the recipe legs exercised at all.

Two changes:

- `NO_DIST` is set unless `%PYTHON%` names an MPI runner, which is the contract the documentation already describes. `run_test.sh` needs nothing equivalent: on Linux `mpirun` is invoked explicitly and the recipe does install the MPI dependencies.
- The smoke test now runs as `%PYTHON% tests\mpi_lifecycle_smoke.py`, like the steps beside it, instead of the `mpiexec -n 2 %PYTHON%` I had written. Hardcoding the launcher would have expanded to `mpiexec -n 2 mpiexec -n 2 python` under the documented configuration; taking the rank count from `$PYTHON` is both correct and consistent.

The Windows pipeline jobs are unaffected either way — `build-and-test-win.yml` passes `NO_DIST: 1` explicitly. The nightly workflow sets `NO_DIST=YES`, which is not `"1"` and so was also entering the block; it now skips for the same reason.
</details>

I have left the `sklearnex_spmd` and `mpi_legacy` gap on Windows as a skip rather than trying to make it work, since installing an MPI stack into the Windows recipe is a separate change from this PR. Happy to file it as an issue if you would like it tracked.

tests/mpi_lifecycle_smoke.py
```diff
@@ -106,9 +106,14 @@ def query_topology(_):
...
-print(
-    f"rank {rank} of {size}: daal4py-owned MPI initialized, used from threads, "
-    "torn down once, and confirmed unusable afterwards",
-    flush=True,
-)
+# that never ran the script. Say what was covered, from a single rank: the ranks
+# write to the same captured stream in no defined order, and one line is enough.
+# It can speak for all of them because the distributed computation above is a
+# collective - no rank reaches this point unless every rank took part - and a rank
+# that failed anywhere would fail the launcher regardless of what is printed here.
```
david-cortes-intel: Most of this comment block is redundant.

conda-recipe/run_test.bat
```diff
@@ -46,6 +46,17 @@ if "%~2"=="--json-report" (
...
 )
 
+rem Distributed tests on Windows require PYTHON to carry an MPI launcher, as in
+rem set "PYTHON=mpiexec -n 2 python" (see doc/sources/tests.rst). The check above
+rem only recognizes the literal string "python", so any other bare interpreter -
+rem conda-build passes %PREFIX%\python.exe - reached the block below and ran the
+rem MPI steps single-rank with no MPI stack installed at all, since meta.yaml
+rem marks mpi, mpi4py and pytest-mpi as "# [not win]". That looked like a pass
+rem rather than an error: pytest rejects --with-mpi as an unknown argument when
+rem pytest-mpi is missing, and helper_mpi_tests.py used to discard its status.
```
david-cortes-intel: Please write the comments as if they refer to the codebase as it will be once the PR is merged, without referencing previous discarded work.

tests/mpi_lifecycle_smoke.py
napetrov: Cut to the two things a reader cannot get from the code in [`d150d21f`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/d150d21f) - that the summary exists because everything above it asserts silently, and that one rank prints it because the ranks share the stream in no defined order:

```python
# The assertions above are silent, so say what was covered. One rank does it,
# since the ranks share the captured stream in no defined order.
```

The rest was an argument for why one rank is enough, which belongs in this thread rather than in the file.

conda-recipe/run_test.bat
napetrov: Fair point, and it applied in more places than this one - taken as a general rule and swept across the PR in [`d150d21f`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/d150d21f). This block now states the requirement and the skip, with no narration of the check beside it or of the failure it avoids:

```bat
rem Distributed tests on Windows need PYTHON to carry an MPI launcher, as in
rem set "PYTHON=mpiexec -n 2 python" (see doc/sources/tests.rst). Skip them when it
rem does not, which is also the case under conda-build: PYTHON is a plain
rem interpreter there and the recipe installs no MPI stack on Windows, since
rem meta.yaml marks mpi, mpi4py and pytest-mpi as "# [not win]".
```

<details>
<summary>The other four the same rule caught</summary>

- `tests/helper_mpi_tests.py` said "without this the script always exits zero", describing the behavior being replaced. Now: "Exit with pytest's status, so that a rank whose tests fail fails the launcher and the caller along with it."
- `conda-recipe/run_test.sh` and `run_test.bat` both explained at length why the smoke script is launched directly instead of through the helper. Reduced to the one reason that matters - the helper imports `mpi4py` and would take MPI ownership away - since the script's own module docstring covers the rest.
- `src/transceiver.cpp` called `D4P_TRANSCEIVER` "a pre-existing hook"; after merge it is simply a hook.
- `tests/mpi_lifecycle_smoke.py`, per your other comment.

Net effect is 30 comment lines removed for 18 added. No code changed, so the green run on `d876ac47` still covers it.
</details>

src/mpi/mpi_transceiver.cpp
napetrov: Reworded in [`5e13c8c2`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3362/commits/5e13c8c2), since I had offered that and not done it. The comment no longer reads as though the two were interchangeable, and it now says why there is nothing else left to write to:

```cpp
// Report a teardown failure without throwing. fini() runs from a destructor and
// from interpreter shutdown, so there is no caller left to catch an exception -
// but staying silent hides a real fault. Warn through Python while the interpreter
// can still take one; past that point sys.stderr is gone too, so fall back to the
// process's own stderr, which is a different stream and the only one left.
```

The behavior is unchanged: `PyErr_WarnEx` remains the path taken whenever the interpreter is live, and the `fputs` is reached only under `!Py_IsInitialized()` or `Py_IsFinalizing()`, where `PySys_WriteStderr` would need a GIL and a `sys.stderr` that no longer exist.

david-cortes-intel (approved):
I see the tests aren't actually executing in private CI, but I ran them manually and they all passed.

PR #3363, opened 2026-08-07
[enhancement] Replace the per-call std::regex result-option scan with a shared parser

napetrov:
Split out of #3325, which reviewers asked to break up. No free-threading changes here.

Five estimator wrappers - `basic_statistics`, `dbscan`, `linear_model`, `logistic_regression`, `neighbors` - open-coded the same scan of the result-option string with `std::regex("\\w+")`. Each call, i.e. each fit or predict, constructed and compiled that regex and then walked it with `std::sregex_iterator` to split a short string into identifier tokens.

Three problems beyond the duplication:

* Regex construction and compilation sit on a path that runs once per estimator call.
* `std::regex` is imbued with the global locale, so the meaning of `\w` depends on process-wide state that Python code can change underneath it.
* It pulls `<regex>` into five translation units for a scan that a dozen lines of straight-line code express exactly.

`onedal/common/result_options.hpp` adds `for_each_result_option()`, which invokes a callback for each `[A-Za-z0-9_]` run in the input. That token definition is exactly ECMAScript `\w`, so the set of accepted and rejected option strings is unchanged - a differential test against `std::regex("\\w+")` over the fixed option vocabulary plus 200k random byte strings finds no difference. `std::isalnum()` is deliberately not used: it is locale-sensitive and does not accept `_`.

Note the `catch (std::regex_error&)` arms being removed were unreachable, since the pattern was a fixed valid literal. The `ONEDAL_PARAM_DISPATCH_THROW_INVALID_VALUE` throw for an unrecognized option was never caught by them and still propagates to the caller unchanged.

`onedal/common/tests/test_result_options.py` covers the tokenizer: multiple and single tokens, idempotent repeats, rejection of unknown tokens and of prefixes of real ones, every non-word separator byte including embedded NULs and non-ASCII, and concurrent parsing from eight threads.

An earlier revision of this PR did that with a standalone `tests/result_options_threading.cpp` binary compiled by the Linux CI job, on the grounds that the parser is header-only and unexposed to Python. That was dropped after review: a native test binary is a new build artifact and a new CI step to maintain, and driving the tokenizer through `BasicStatistics(result_options=...)` proves strictly more - that the parsed tokens actually arrive as the requested results, not merely that a string was split. One case (`test_any_non_word_byte_separates_tokens`) reads the raw backend result rather than going through `fit()`, because `fit()` unpacks results by iterating options split on `"|"` and so can only round-trip that one separator; the docstring on `_compute_raw` explains it.

Depended on by #3325.


.github/workflows/ci.yml
```diff
@@ -141,6 +141,13 @@ jobs:
           pip install --upgrade -r requirements-test.txt
           if [ "${{ steps.set-env.outputs.DPCFLAG }}" == "" ]; then pip install dpctl==${{ env.DPCTL_VERSION }} dpnp==${{ env.DPNP_VERSION }}; fi
           pip list
+      - name: Result-option parser test
```
david-cortes-intel: A better way to test this would be to make a python wrapper and put it under the tests folder.

napetrov: Agreed, and done in 784ae11 — the native binary and its bespoke `g++` CI step are gone, replaced by `onedal/common/tests/test_result_options.py`.

It turned out not to need a wrapper at all. The tokenizer is already reachable from Python through every estimator that takes a `result_options` argument: `BasicStatistics` joins the list with `|` and hands the string to the backend, where `get_onedal_result_options` splits it and dispatches each token. So the test drives the real public path instead of a binding that exists only for testing:

```python
def _compute(options, queue=None):
    data = np.array([[0.0, 1.0], [2.0, 3.0], [4.0, 5.0]], dtype=np.float64)
    estimator = BasicStatistics(result_options=options)
    return estimator.fit(data, queue=queue)
```

It covers everything the native test asserted, plus one thing that test structurally could not:

- multi-token, single-token and repeated-token strings
- every separator class, including the two the native test existed for — embedded `NUL` and bytes above `0x7f` (`"\xff"`, and `"é"` to cover a multi-byte UTF-8 character being split on its lead byte)
- `_` *not* separating, since it is a word character — `["min_max"]` must be rejected as one unknown token rather than accepted as two known ones
- unknown and prefix-of-known tokens rejected with `RuntimeError` (`ONEDAL_PARAM_DISPATCH_THROW_INVALID_VALUE` throws `std::runtime_error`, which pybind11 maps to `RuntimeError`)
- concurrent parsing from 8 threads, which is the property that motivated replacing the per-call `std::regex` in the first place

The thing the native test could not do: because these assertions go through `fit`, they check that the parsed tokens actually *arrive* as the requested results (`result.min_`, `result.max_`, `result.mean_`), not merely that a string was split into the expected pieces. It also now runs on every CI leg and over the queue parametrization, rather than only in the one hand-written Linux step.

icfaust (commented):
I always thought it was overkill to be doing regexes on such limited numbers of options. Will give a more thorough review next week.

ethanglaser (approved):
Thanks for separating this, much easier to review. Only suggestion is to also remove `include <regex>` from the kmeans and kmeans_init cpp files as they're unused.

onedal/common/tests/test_result_options.py
```diff
@@ -0,0 +1,154 @@
...
+
+@pytest.mark.parametrize("separator", ["|", ",", " ", "\0", "\xff", "é"])
+@pytest.mark.parametrize("queue", get_queues())
+def test_any_non_word_byte_separates_tokens(separator, queue):
+    """Token boundaries are every byte outside [A-Za-z0-9_], matching ECMAScript \\w.
+
+    This covers the cases the standalone native test used to assert - embedded
+    NULs and bytes above 0x7f - through the public path. ``min`` and ``max`` are
+    both real options, so a correctly split string yields both results.
+
```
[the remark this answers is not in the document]
napetrov: Correct — the description was stale and I've rewritten that paragraph. There is no `result_options_threading.cpp` in the repo, and the pytest module's reference to "the standalone native test" was pointing at a file that no longer exists.

History: an earlier revision did add that native binary plus a Linux CI step to compile and run it. It was dropped after review, for two reasons. A native test binary is a new build artifact and a new CI step to maintain for a header-only helper. And driving the tokenizer through `BasicStatistics(result_options=...)` proves strictly more than the native test did — it shows the parsed tokens actually arrive as the requested *results*, not merely that a string was split into the expected pieces.

Docstrings reworded in c8cb51b0 so they describe what each test asserts rather than referring to a deleted file. No test logic changed.

One wrinkle the description now records: `test_any_non_word_byte_separates_tokens` reads the raw backend result instead of going through `fit()`. `fit()` unpacks results by iterating `self.options` and doing `getattr(result, option)`, and the `options` setter splits on `"|"` only — so `"min,max"` survives as a single Python-level option and `getattr` fails on it. `"|"` is the only separator that round-trips through Python, which is precisely why testing *any* non-word byte has to read the backend result directly.

david-cortes-intel (approved):

PR #3368, opened 2026-08-12
[enhancement] Build the base-1 CSR index arrays without an intermediate cast array

napetrov:
Follow-up to the review comment on #3361, which approved that PR with:

> LGTM, but it could be improved by calculating the base1 arrays for CSR matrices without making an extra copy that would get immediately discarded. Could have templated versions for dtypes and unit-vs-nonunit stride where the base1 arrays would be calculated from the base0 input as it comes.

That is what this does. It is kept out of #3361 because that one was a refcount and ownership bugfix, already approved and clean.

## The extra copy

`convert_to_table()` cast a `csr_matrix`s `indices` and `indptr` to uint64 with `PyArray_FROMANY`, and `convert_to_csr_impl()` then allocated a `dal::array` and walked that cast array adding one. Two allocations and two passes per index array, where the first existed only to be read once and thrown away.

scipy stores `indices` and `indptr` as int32 whenever the matrix fits in 32 bits, so that was the *common* path, not a corner case. The sharp edge: the previous code was zero-copy exactly in the rare case - uint64 indices, where `FORCECAST` had nothing to do - and copied in the default one.

For nnz non-zeros, with int32 indices:

| | peak for the index array | bytes moved |
|---|---|---|
| before | 16·nnz (uint64 temp + dal::array) | ~24·nnz |
| after | 8·nnz | ~12·nnz |

At nnz = 10⁸ that is 800 MB of peak allocation and ~1.6 GB of traffic. `indptr` adds 8·(rows+1), minor.

## What replaces it

`make_one_based_indices()` reads the callers buffer as it comes:

- `rebase_indices_to_one<Src>` widens and rebases in one pass, templated on source width and signedness, reading the source as `Src`. `make_one_based_indices` requires `PyArray_ISALIGNED` before taking that path, which is what makes the typed dereference well defined with respect to alignment for every element; the stride test alongside it is what makes `stride / itemsize` exact, because NumPy folds only the strides of dimensions holding more than one element into its alignment flag. Three loops follow from that: broadcast, unit stride, and any other step including negative.
- `rebase_indices_to_one_unaligned<Src>` is the cold path, and the only remaining `memcpy` in the file. NumPy allows a unit-stride view to start at an odd byte, where a typed dereference would be undefined, so such a buffer is walked byte by byte instead. scipy does not produce one.
- `classify_index_type()` dispatches on itemsize and signedness rather than on NumPy type number, because `NPY_INT64` is `NPY_LONG` on Linux and `NPY_LONGLONG` on Windows and a caller may hand over either spelling on either platform.
- Anything the loops cannot read directly - a byte-swapped array, or a width scipy does not produce, such as int16 - falls back to numpys own cast and then takes the fast path on the result, which is native contiguous int64 by construction. Generality is unchanged; recursion is one level deep.

Two smaller things follow from it:

- The fallback casts to `NPY_INT64` rather than `NPY_UINT64`, which removes the reinterpretation of a uint64 buffer through an `int64_t *`.
- The flags lose `NPY_ARRAY_WRITEABLE`, which was forcing a copy of any read-only index array, for buffers that are only ever read.

## Behaviour is unchanged, including at the edges

Deliberately so - no new validation, so this stays a pure performance change. A base-0 index of `-1` still rebases to `0` rather than to a huge positive, and a uint64 index above 2⁶³ still wraps exactly as before.

The alignment guard does not change that either. A differential harness compared the current dispatch against the previous one over 277 index buffers - every accepted dtype plus byte-swapped and unsupported widths, contiguous, reversed, strided, negative-strided, broadcast, read-only, zero- and one-length, and deliberately misaligned at each byte offset - with no divergence in any value, no extra casts, and one level of recursion at most. All four paths were exercised (unit stride 150 cases, general stride 32, broadcast 15, byte-wise fallback 80). Under `-fsanitize=alignment` the guarded dispatch is clean across all 277; bypassing the guard reports `load of misaligned address ... which requires 4 byte alignment` for each of the four index dtypes, which is the control showing the harness detects what it is looking for.

I checked that rather than asserting it: a standalone differential program compared the new loops against the old cast-then-widen pipeline (`(int64_t)(uint64_t)value + 1`) over both stride variants for int32, uint32, int64 and uint64, including `INT32_MIN`, `INT64_MIN`, `UINT32_MAX` and `UINT64_MAX - 1`, with the strided case offset by one byte so the elements are unaligned as well. All cases agree.

```
int32: ok (10 values)
uint32: ok (6 values)
int64: ok (9 values)
uint64: ok (6 values)
all index-rebasing cases match the previous pipeline
```

Rejecting negative indices outright would be a real improvement on top of this, but it is a behaviour change and belongs in its own PR.

## Tests

`onedal/datatypes/tests/test_data.py` gains `test_csr_index_representations_are_equivalent`, which converts the same matrix through every index dtype (int32, uint32, int64, uint64) crossed with every buffer layout (contiguous, strided, read-only, byte-swapped) and checks the per-column sums against scipy. Per-column sums catch a misread of `indices` directly, and a misread of `indptr` too, because the row boundaries decide which entries are visited at all.

`test_csr_conversion_does_not_modify_input` picks up uint32 in its dtype list for the same reason.

One thing the new test has to work around, which is pre-existing and not touched here: the conversion asks scipy whether the indices are sorted, and scipy answers by calling into its own C routines, which reject some index dtypes outright - `ValueError: unsupported data types in input` for uint64. The test asserts `has_sorted_indices` instead of letting scipy compute it. Verified against scipy 1.18.

## Validation

No oneDAL is available in the environment I wrote this in, so the Python tests have not been run locally and CI is the gate for them. What was run locally: the standalone differential program above, `black`/`isort`/`clang-format` on the changed files, and a scipy-only check that all sixteen dtype/layout combinations actually reach the conversion in the representation the test intends - scipy preserves the exact dtype, strides and byte order of an assigned index array, so none of the parametrizations is silently normalized before it gets there.

<details>
<summary>Checklist</summary>

- [x] The non-obvious invariants - platform-dependent NumPy type numbers, the fallback path, the unaligned strided case - are documented in comments.
- [x] Behaviour parity with the code being replaced is verified rather than assumed.
- [x] Tests cover the dtype and layout matrix the dispatch introduces.
- [x] No user-facing behaviour change, so no documentation change.
- [x] Performance: no scikit-learn_bench table, because the gain is in allocation and traffic rather than in a kernel's throughput, and it is quantified in the table above (16·nnz → 8·nnz peak for the index array, ~24·nnz → ~12·nnz moved). The per-element loop cost is deliberately *not* claimed as an improvement: measured in a separate translation unit with the stride supplied at run time, as the real build compiles and calls it, the typed loops are at parity with the `memcpy` ones on gcc and about 1.1x better on clang, and no size, stride or dtype shows more.

</details>

### DIFF
--- a/onedal/datatypes/numpy/data_conversion.cpp
+++ b/onedal/datatypes/numpy/data_conversion.cpp
@@ -96,28 +96,19 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
 
 // Widen `count` base-0 indices of type Src into base-1 std::int64_t.
 //
-// The unit-stride case gets its own loop rather than folding into the general
-// one: it is what every scipy matrix hits, and a compile-time-known step is
-// what lets the compiler vectorize the widening.
+// Keep the source byte-addressed: NumPy permits both arbitrary strides and
+// unit-stride views whose first element is unaligned for Src. Fixed-size memcpy
+// avoids an unaligned or aliasing-unsafe Src* dereference; compilers recognize
+// and optimize this load for the common unit-stride case.
 template <typename Src>
-inline void rebase_indices_to_one(const char *source,
+inline void rebase_indices_to_one(const char *source_bytes,
                                   std::int64_t stride,
                                   std::int64_t *destination,
                                   std::int64_t count) {
-    if (stride == static_cast<std::int64_t>(sizeof(Src))) {
-        const Src *typed_source = reinterpret_cast<const Src *>(source);
-        for (std::int64_t i = 0; i < count; ++i) {
-            destination[i] = static_cast<std::int64_t>(typed_source[i]) + 1;
-        }
-    }
-    else {
-        // memcpy rather than a cast through Src *, because a strided array can
-        // also be unaligned for its element type.
-        for (std::int64_t i = 0; i < count; ++i) {
-            Src value;
-            std::memcpy(&value, source + i * stride, sizeof(Src));
-            destination[i] = static_cast<std::int64_t>(value) + 1;
-        }
+    for (std::int64_t i = 0; i < count; ++i) {
+        Src value;
+        std::memcpy(&value, source_bytes + i * stride, sizeof(Src));
+        destination[i] = static_cast<std::int64_t>(value) + 1;
     }
 }
 
--- a/onedal/datatypes/numpy/data_conversion.cpp
+++ b/onedal/datatypes/numpy/data_conversion.cpp
@@ -99,12 +99,44 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
 // Keep the source byte-addressed: NumPy permits both arbitrary strides and
 // unit-stride views whose first element is unaligned for Src. Fixed-size memcpy
 // avoids an unaligned or aliasing-unsafe Src* dereference; compilers recognize
-// and optimize this load for the common unit-stride case.
+// and optimize this load for the common unit-stride case. The special cases
+// below still use memcpy: a contiguous byte walk avoids repeated stride
+// multiplication, while a zero-stride broadcast only needs one load.
 template <typename Src>
 inline void rebase_indices_to_one(const char *source_bytes,
                                   std::int64_t stride,
                                   std::int64_t *destination,
                                   std::int64_t count) {
+    if (stride == 0 && count > 0) {
+        Src value;
+        std::memcpy(&value, source_bytes, sizeof(Src));
+        const std::int64_t rebased = static_cast<std::int64_t>(value) + 1;
+        for (std::int64_t i = 0; i < count; ++i) {
+            destination[i] = rebased;
+        }
+        return;
+    }
+
+    if (stride == static_cast<std::int64_t>(sizeof(Src))) {
+        const char *source = source_bytes;
+        for (std::int64_t i = 0; i < count; ++i, source += sizeof(Src)) {
+            Src value;
+            std::memcpy(&value, source, sizeof(Src));
+            destination[i] = static_cast<std::int64_t>(value) + 1;
+        }
+        return;
+    }
+
+    if (stride > 0 && stride % static_cast<std::int64_t>(sizeof(Src)) == 0) {
+        const char *source = source_bytes;
+        for (std::int64_t i = 0; i < count; ++i, source += stride) {
+            Src value;
+            std::memcpy(&value, source, sizeof(Src));
+            destination[i] = static_cast<std::int64_t>(value) + 1;
+        }
+        return;
+    }
+
     for (std::int64_t i = 0; i < count; ++i) {
         Src value;
         std::memcpy(&value, source_bytes + i * stride, sizeof(Src));
--- a/onedal/datatypes/numpy/data_conversion.cpp
+++ b/onedal/datatypes/numpy/data_conversion.cpp
@@ -16,6 +16,8 @@
 
 #define NO_IMPORT_ARRAY
 
+#include <cstdint>
+#include <cstring>
 #include <memory>
 #include <stdexcept>
 #include <string>
@@ -92,6 +94,115 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
         layout);
 }
 
+// Widen `count` base-0 indices of type Src into base-1 std::int64_t.
+//
+// The unit-stride case gets its own loop rather than folding into the general
+// one: it is what every scipy matrix hits, and a compile-time-known step is
+// what lets the compiler vectorize the widening.
+template <typename Src>
+inline void rebase_indices_to_one(const char *source,
+                                  std::int64_t stride,
+                                  std::int64_t *destination,
+                                  std::int64_t count) {
+    if (stride == static_cast<std::int64_t>(sizeof(Src))) {
+        const Src *typed_source = reinterpret_cast<const Src *>(source);
+        for (std::int64_t i = 0; i < count; ++i) {
+            destination[i] = static_cast<std::int64_t>(typed_source[i]) + 1;
+        }
+    }
+    else {
+        // memcpy rather than a cast through Src *, because a strided array can
+        // also be unaligned for its element type.
+        for (std::int64_t i = 0; i < count; ++i) {
+            Src value;
+            std::memcpy(&value, source + i * stride, sizeof(Src));
+            destination[i] = static_cast<std::int64_t>(value) + 1;
+        }
+    }
+}
+
+enum class index_kind { unsupported, int32, uint32, int64, uint64 };
+
+// Classify by width and signedness rather than by NumPy type number, because
+// NPY_INT32 and NPY_INT64 are aliases whose targets differ per platform: an
+// int64 array is NPY_LONG on Linux and NPY_LONGLONG on Windows, and a caller
+// can hand over either spelling on either platform.
+static index_kind classify_index_type(int npy_type, std::int64_t itemsize) {
+    const bool is_signed = npy_type == NPY_INT || npy_type == NPY_LONG || npy_type == NPY_LONGLONG;
+    const bool is_unsigned =
+        npy_type == NPY_UINT || npy_type == NPY_ULONG || npy_type == NPY_ULONGLONG;
+    if (itemsize == 4) {
+        return is_signed ? index_kind::int32
+                         : (is_unsigned ? index_kind::uint32 : index_kind::unsupported);
+    }
+    if (itemsize == 8) {
+        return is_signed ? index_kind::int64
+                         : (is_unsigned ? index_kind::uint64 : index_kind::unsupported);
+    }
+    return index_kind::unsupported;
+}
+
+// Build the base-1 index array oneDAL's csr_table expects, reading the caller's
+// buffer in whatever integer dtype and stride it arrives in.
+//
+// Casting to a fixed dtype first - which is what this used to do - allocates
+// twice per index array: once for the cast array, whose only purpose is to be
+// read once and discarded, and once for the dal::array that outlives the call.
+// scipy stores `indices` and `indptr` as int32 whenever the matrix fits in 32
+// bits, so that was the common path, not a corner case: for nnz non-zeros it
+// cost an extra 8*nnz bytes of peak allocation and two extra passes over them.
+static dal::array<std::int64_t> make_one_based_indices(PyObject *py_indices) {
+    PyArrayObject *np_indices = reinterpret_cast<PyArrayObject *>(py_indices);
+
+    // The loops below read the source dtype directly, so they need native byte
+    // order and a width they can widen. Anything else - a byte-swapped array,
+    // or an index dtype scipy does not produce, such as int16 - goes through
+    // numpy's own cast and then takes the fast path on the result, which is
+    // native, contiguous int64 by construction. Recursion is therefore one
+    // level deep.
+    const index_kind kind =
+        classify_index_type(static_cast<int>(array_type(np_indices)),
+                            static_cast<std::int64_t>(array_type_sizeof(np_indices)));
+    if (!array_is_native(np_indices) || kind == index_kind::unsupported) {
+        // No WRITEABLE in the flags: these buffers are only ever read, and
+        // asking for writeability copies a read-only input for nothing.
+        py::object cast_indices = py::reinterpret_steal<py::object>(
+            PyArray_FROMANY(py_indices,
+                            NPY_INT64,
+                            0,
+                            0,
+                            NPY_ARRAY_C_CONTIGUOUS | NPY_ARRAY_ALIGNED | NPY_ARRAY_FORCECAST));
+        if (!cast_indices) {
+            throw std::invalid_argument(
+                "[convert_to_table] Could not convert csr_matrix indices to 64-bit integers.");
+        }
+        return make_one_based_indices(cast_indices.ptr());
+    }
+
+    const std::int64_t count = static_cast<std::int64_t>(array_size(np_indices, 0));
+    auto one_based = dal::array<std::int64_t>::empty(count);
+    std::int64_t *const destination = one_based.get_mutable_data();
+    const char *const source = static_cast<const char *>(array_data(np_indices));
+    const std::int64_t stride = static_cast<std::int64_t>(array_stride(np_indices, 0));
+
+    switch (kind) {
+        case index_kind::int32:
+            rebase_indices_to_one<std::int32_t>(source, stride, destination, count);
+            break;
+        case index_kind::uint32:
+            rebase_indices_to_one<std::uint32_t>(source, stride, destination, count);
+            break;
+        case index_kind::int64:
+            rebase_indices_to_one<std::int64_t>(source, stride, destination, count);
+            break;
+        default:
+            // classify_index_type() admits nothing else past the guard above.
+            rebase_indices_to_one<std::uint64_t>(source, stride, destination, count);
+            break;
+    }
+    return one_based;
+}
+
 template <typename T>
 inline csr_table_t convert_to_csr_impl(PyObject *py_data,
                                        PyObject *py_column_indices,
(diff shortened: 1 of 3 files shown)


onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -92,42 +94,131 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
         layout);
 }
 
+// Widen `count` base-0 indices of type Src into base-1 std::int64_t.
+//
+// The unit-stride case gets its own loop rather than folding into the general
+// one: it is what every scipy matrix hits, and a compile-time-known step is
+// what lets the compiler vectorize the widening.
+template <typename Src>
+inline void rebase_indices_to_one(const char *source,
```
david-cortes-intel: Why would it need to be `char` if the template already has Src as parameter?

onedal/datatypes/tests/test_data.py
```diff
@@ -710,7 +711,62 @@ def test_nonwriteable_arrays():
...
+    reinterpreted.indptr = _reinterpret_index_array(X.indptr, index_dtype, layout)
+    # The conversion asks scipy whether the indices are sorted, and scipy answers
+    # that by calling into its own C routines, which reject some of the dtypes
+    # under test here ("unsupported data types in input" for uint64). Assert the
+    # answer instead - the indices came from a sorted matrix above.
+    reinterpreted.has_sorted_indices = True
+
+    result = BasicStatistics(result_options="sum").fit(reinterpreted)
+
+    assert_allclose(result.sum_, expected, rtol=1e-12)
```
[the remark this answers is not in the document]
napetrov: Fixed in fd620417. The test now uses `rtol=0, atol=1e-9`, consistent with the BasicStatistics float64 tests.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -92,42 +94,131 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
...
+    }
+    else {
+        // memcpy rather than a cast through Src *, because a strided array can
+        // also be unaligned for its element type.
+        for (std::int64_t i = 0; i < count; ++i) {
+            Src value;
+            std::memcpy(&value, source + i * stride, sizeof(Src));
+            destination[i] = static_cast<std::int64_t>(value) + 1;
+        }
+    }
```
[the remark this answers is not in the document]
napetrov: Fixed in fd620417, with a stronger approach than an alignment branch: `rebase_indices_to_one` now uses fixed-size `memcpy` from byte-addressed source storage for every stride. This avoids both unaligned and aliasing/lifetime-unsafe `Src*` dereferences. Added deterministic regressions for deliberately offset-by-one, unit-stride int32/uint32/int64/uint64 CSR index buffers; the test asserts both unit stride and misalignment.

onedal/datatypes/numpy/data_conversion.cpp
napetrov: It remains `char*` intentionally, renamed `source_bytes` in fd620417. The function accepts arbitrary NumPy views, including negative/non-unit stride and offset-by-one unit-stride storage, so it must do byte-address arithmetic. Each element is loaded into the templated `Src` with fixed-size `memcpy`; this keeps the typed conversion while avoiding an invalid `Src*` dereference.

david-cortes-intel: @napetrov In that case, it would make sense to have a specialized version for the case where strides are equal to sizeof(Src), and another for the case where strides are non-negative and a multiple of sizeof(Src) (or zero), since those kinds of operations could be optimized compared to something with arbitrary strides.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -99,12 +99,44 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
...
+            destination[i] = rebased;
+        }
+        return;
+    }
+
+    if (stride == static_cast<std::int64_t>(sizeof(Src))) {
+        const char *source = source_bytes;
+        for (std::int64_t i = 0; i < count; ++i, source += sizeof(Src)) {
+            Src value;
+            std::memcpy(&value, source, sizeof(Src));
```
david-cortes-intel: This and the one below don't need memcpy. Could be put into a loop with an ivdep pragma to allow vectorization.

onedal/datatypes/tests/test_data.py
```diff
@@ -725,6 +725,11 @@ def _reinterpret_index_array(indices, index_dtype, layout):
         view = np.repeat(typed, 2)[::2]
         assert view.strides[0] == 2 * typed.itemsize
         return view
+    if layout == "negative":
```
david-cortes-intel, line 736: Why would it need a special handling if the convertor uses a signed int type for the strides?

onedal/datatypes/numpy/data_conversion.cpp
napetrov: Agreed on the memcpy, and both are gone in [`11341035`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3368/commits/11341035f7253245784c547f827468c7938fee24): requiring `PyArray_ISALIGNED` in the guard in `make_one_based_indices()` lets the loops take `const Src *` directly, so the byte addressing disappears from both of these branches and one byte-wise loop is left as the cold path for buffers that are genuinely unaligned - which is what Copilot suggested on the 12th and I should have taken then.

On the pragma specifically: I tried it, and on this loop it is a no-op. `#pragma GCC ivdep` emits zero SIMD instructions and gcc's refusal does not change, because it is not refusing on a dependence - which is all `ivdep` asserts - but because it has no vector type for the byte-wise load. Removing the `memcpy` is what actually unblocks the vectorizer, so the change you asked for is the right one; it is just not the pragma that gets you there.

<details>
<summary>What I measured: the pragma, the codegen, and what vectorizing is actually worth</summary>

`-O3 -std=c++17` plus the UNIX `SDL_FLAGS` from `scripts/CMakeLists.txt:47`, no `-march`. gcc 13.3.0, clang 18.1.3.

**The pragma is inert here.** `#pragma GCC ivdep` on the strided `memcpy` loop, recompiled at three ISA levels - baseline, `-march=x86-64-v3`, `-march=sapphirerapids` (AVX-512, gather available). Identical result every time, and `objdump` finds 0 SIMD instructions:

```
missed: couldn't vectorize loop
missed: not vectorized: no vectype for stmt: *_4 = _5;
```

**A `-fopt-info-vec` line on the typed general-stride loop is misleading on its own.** gcc does report it as vectorized, but the SIMD block is behind a runtime check:

```
cmp    $0x1,%rsi
jne    <scalar>          # anything but unit stride goes here
...
movdqu (%rdi,%rax,1),%xmm0
pcmpgtd %xmm0,%xmm2
punpckldq / punpckhdq ; paddq ; movups
```

Since the dispatch handles `element_stride == 1` in its own branch above, the general branch never reaches that block. That is also why swapping `memcpy` for a typed load measures the same at non-unit stride - both end up running scalar loops. Kernels in their own translation unit with the stride supplied at run time, medians of 301, pinned core: gcc is at parity (0.98-1.03x) across sizes, steps and dtypes; clang is 1.10-1.18x in favour of the typed loop at L2-resident sizes, from unrolling rather than vectorization.

**Vectorizing the non-unit-stride case is worth real money, and no pragma reaches it.** A hand-written SSE2 stride-2 deinterleave (`movdqu` x2, `shufps 0x88`, sign-extend, `movdqu` x2), verified bit-exact against the scalar loop, ns/element:

| n | working set | scalar | hand SSE2 | ratio |
|---|---|---|---|---|
| 1e4 | 0.2 MiB | 0.6942 | 0.2624 | 2.65x |
| 1e5 | 1.5 MiB | 0.6673 | 0.3496 | 1.91x |
| 1e6 | 15.3 MiB | 0.6672 | 0.5956 | 1.12x |
| 1e7 | 152.6 MiB | 0.9952 | 1.0859 | 0.92x |

Real while the working set fits L2, gone past it. It is out of reach of any pragma because `array_stride()` is a runtime value - the compiler will only vectorize this shape by specialising for a stride it can see. If you want that path I would rather add it explicitly, for `element_stride == 2` and whichever other small constants are worth peeling, as a separate change on top of this one. Say the word and I will open it.

Footnote on portability, since it would decide the spelling anyway: bare `ivdep` is the Intel-compiler pragma, gcc wants `#pragma GCC ivdep` and MSVC wants `#pragma loop(ivdep)`. `scripts/CMakeLists.txt:28` sets `CMAKE_COMPILE_WARNING_AS_ERROR ON` and the glob at `:77` pulls this file into that target, and the `WindowsNightly` jobs compile it with MSVC, so an unguarded pragma would need a three-way `#ifdef`.
</details>

onedal/datatypes/numpy/data_conversion.cpp
napetrov: The `char *` is gone in [`11341035`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3368/commits/11341035f7253245784c547f827468c7938fee24) - requiring `PyArray_ISALIGNED` in the guard means the loops can take `const Src *` directly, so the byte-addressed spelling I defended here is no longer needed on any path except the unaligned fallback.

onedal/datatypes/tests/test_data.py
napetrov: The signed stride type is what makes a negative stride representable; it is not what makes it covered, and a negative stride still lands on a different loop from the unit-stride one, so it needs its own parameter here.

The `typed[::-1].copy()[::-1]` is only there to keep the oracle valid - a bare `typed[::-1]` would reverse the index *values* too, and the per-column sums this test asserts would no longer match. Both of those are now comments on the helper in [`11341035`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3368/commits/11341035f7253245784c547f827468c7938fee24), and I corrected the neighbouring one while I was there: it claimed the strided view is "not necessarily aligned", which is wrong - `np.repeat(t, 2)[::2]` is always aligned, and after this change that is exactly what puts it on the typed general-stride loop rather than the byte-wise one. Both layouts now assert `flags.aligned` so that stays true.

onedal/datatypes/numpy/data_conversion.cpp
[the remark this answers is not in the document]
napetrov: Revisiting this: your original suggestion was the right one and my "stronger approach" was not. [`11341035`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3368/commits/11341035f7253245784c547f827468c7938fee24) does what you asked - the fast path now requires `PyArray_ISALIGNED`, and unaligned buffers fall back to the `memcpy` loop rather than every path paying for it.

One addition beyond the flag: the dispatch also tests `stride % itemsize == 0`. NumPy folds only the strides of dimensions holding more than one element into its alignment flag, so a one-element view can report itself aligned while its byte stride is not a whole number of elements, and `stride / itemsize` would silently truncate. There is a regression test for that shape now.

onedal/datatypes/tests/test_data.py
david-cortes-intel: I don't follow. Why would you reverse the array if the C++ code with the iterator can accurately loop over negative strides?

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -92,42 +94,185 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
...
+    }
+}
+
+// The same widening for a buffer whose elements are not aligned for Src, which
+// NumPy permits: a view can be unit-stride and still start at an odd byte. A
+// typed dereference would be undefined there, so this walks bytes and copies
+// each element out with a fixed-size memcpy. It is the cold path - scipy does
+// not produce such a buffer - so it is kept simple rather than specialized.
+template <typename Src>
+inline void rebase_indices_to_one_unaligned(const char *source_bytes,
```
david-cortes-intel, line 139: What would be the issue here with using pybind11 accessors?

onedal/datatypes/tests/test_data.py
napetrov: It can, and that is not what the reversal is for - my earlier answer buried the point.

The conversion walks a negative stride fine. The two reversals are there so this layout hands it the *same* index sequence as every other layout in the matrix. A plain `typed[::-1]` is equally a negative-stride view, but of the values in reverse order, which is a different matrix: the per-column sums this test compares against would no longer match, and a reversed `indptr` is not a valid CSR structure at all. Storing the values backwards and then stepping backwards over them cancels out, leaving the original order behind a negative stride.

Rewrote it in [`c22442c3`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3368/commits/c22442c3) to say exactly that - the intermediate is named, the comment explains the intent rather than the mechanics, and the invariant is asserted with `assert_array_equal(view, typed)` instead of argued.

onedal/datatypes/numpy/data_conversion.cpp
napetrov: That line is the byte-wise fallback for buffers that are *not* aligned for their element type, and the accessors cannot serve it: every pybind11 element accessor ends in a typed dereference at a byte offset, which is precisely the undefined behaviour this fallback exists to avoid.

```cpp
// unchecked_reference::operator()   - numpy.h:687
return *reinterpret_cast<const T *>(data_ + byte_offset_unsafe(strides_, ssize_t(index)...));

// array_t::at()                     - numpy.h:1452
return *(static_cast<const T *>(array::data()) + byte_offset(ssize_t(index)...) / itemsize());
```

(pybind11 3.1.0, the version `dependencies-dev` pins.)

On the aligned paths they would work, but those loops are already plain `const Src *` after the guard, so an accessor would only add a layer - and `unchecked<T, N>()` requires an exact dtype match, so the per-width dispatch would stay either way.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -92,42 +94,185 @@ inline dal::homogen_table convert_to_homogen_impl(PyArrayObject *np_data) {
...
+        npy_type == NPY_UINT || npy_type == NPY_ULONG || npy_type == NPY_ULONGLONG;
+    if (itemsize == 4) {
+        return is_signed ? index_kind::int32
+                         : (is_unsigned ? index_kind::uint32 : index_kind::unsupported);
+    }
+    if (itemsize == 8) {
+        return is_signed ? index_kind::int64
+                         : (is_unsigned ? index_kind::uint64 : index_kind::unsupported);
+    }
+    return index_kind::unsupported;
```
david-cortes-intel, line 188: While SciPy doesn't construct sparse matrices with other types, it does allow changing the types after construction:
```python
import numpy as np

from scipy.sparse import random as sp_random
X = sp_random(100, 50, .1, rng=rng, format="csr")
v = rng.standard_normal(size=X.shape[1])
res = X @ v

X.indices = X.indices.astype(np.uint16)
X.indptr = X.indptr.astype(np.uint16)
res2 = X @ v
np.testing.assert_almost_equal(res, res2)

assert X.indices.dtype == np.uint16
```

napetrov: It already converts correctly, and that case is what the cast fallback is for. `classify_index_type` accepts itemsize 4 and 8 only, so `uint16` (and `int16`, `uint8`, and any byte-swapped array) falls to `index_kind::unsupported` and goes through one `PyArray_FROMANY(..., NPY_INT64, ..., FORCECAST)` followed by a single level of recursion on the resulting native, contiguous, aligned int64 array. Values widen, nothing truncates - `main` took that same cast for *every* index dtype, so this is no worse anywhere and cheaper for the int32/int64 that `scipy.sparse` actually constructs.

What was missing was a test saying so, added in [`c398af98`](https://github.com/uxlfoundation/scikit-learn-intelex/pull/3368/commits/c398af98): `int16` and `uint16` now run through the whole dtype-by-layout matrix, plus `test_csr_indices_retyped_after_construction` in the shape you wrote - build a float64 CSR, retype `indices` and `indptr`, and compare the conversion's result against scipy on the same matrix.

I deliberately did not give the narrow widths a widening loop of their own: scipy cannot produce them on construction, so a loop per width would serve a path no caller reaches through `scipy.sparse` alone.

One adjacent hole while we are here, pre-existing and not addressed by this PR: a `uint64` index above `INT64_MAX` wraps in the `static_cast<std::int64_t>`, and did the same on `main` via `FORCECAST`. Happy to file it or guard it here, whichever you prefer.

david-cortes-intel (approved):

PR #3371, opened 2026-08-17
MAINT: Remove deprecated daal4py sklearn patching

david-cortes-intel:
## Description

Follow up from https://github.com/uxlfoundation/scikit-learn-intelex/pull/3314

Removes the functionality to patch the sklearn module with daal4py classes, which has been deprecated and is not tested nor does it have any use case anymore.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


daal4py/sklearn/monkeypatch/dispatcher.py
```diff
@@ -1,255 +0,0 @@
...
-            ElasticNet_daal4py,
-            ElasticNet_sklearn,
-        ),
-        "sklearn.linear_model.Lasso": (
-            linear_model_module,
-            "Lasso",
-            Lasso_daal4py,
-            Lasso_sklearn,
-        ),
-        "sklearn.linear_model._logistic._logistic_regression_path": (
```
avolkov-intel, line 99: This class patching is not present in https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/sklearnex/dispatcher.py but it's still used for CPU LogisticRegression

david-cortes-intel: It gets temporarily patched during .fit() calls:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/1950db13c929b3fe260ba5922676c6c1ef50ee2f/daal4py/sklearn/linear_model/logistic_path.py#L480

This daal4py patching would not have been applied if calling `sklearnex.patch_sklearn()`.

sklearnex/dispatcher.py
```diff
@@ -398,6 +406,48 @@ def get_patch_names() -> list[str]:
...
+    which, what, _, replacer = descriptor
+    if replacer is not None:
+        setattr(which, what, replacer)
+    elif hasattr(which, what):
+        delattr(which, what)
+
+
+def enable(
+    name: Optional[str] = None,
+    map: Optional[PatchMap] = None,
```
avolkov-intel: I don't see a usecase where map can be None. Probably it worth making it a required argument and swap it with the name, although I understand that in this case we would need to switch the order of name and map in many functions

daal4py/sklearn/__init__.py
```diff
@@ -14,11 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
-from .monkeypatch.dispatcher import _get_map_of_algorithms as sklearn_patch_map
-from .monkeypatch.dispatcher import _patch_names as sklearn_patch_names
-from .monkeypatch.dispatcher import disable as unpatch_sklearn
-from .monkeypatch.dispatcher import enable as patch_sklearn
-from .monkeypatch.dispatcher import patch_is_enabled as sklearn_is_patched
+from . import ensemble, linear_model, manifold, metrics, model_selection, neighbors, utils
```
avolkov-intel, line 17: Why do we need this line if we didn't have it previously?

daal4py/sklearn/test/test_common.py
```diff
@@ -1,79 +0,0 @@
...
-    "KNeighborsClassifier": KNeighborsClassifier(n_neighbors=10),
-    "DaalRandomForestClassifier": DaalRandomForestClassifier(
-        n_estimators=10, random_state=777
-    ),
-    "DaalRandomForestRegressor": DaalRandomForestRegressor(
-        n_estimators=10, random_state=777
-    ),
-}
-
-ORDERS = ["C", "F"]
```
avolkov-intel, line 53: Not sure if numpy + F-order is tested somewhere else

avolkov-intel: Found some tests in test_memory_usage.py please check that the tests from there are sufficient

sklearnex/dispatcher.py
avolkov-intel: Up to you to decide whether you want to fix it now

david-cortes-intel: Fixed.

daal4py/sklearn/__init__.py
david-cortes-intel: Previously, they were being imported in the dispatcher file. After removing it, they aren't imported automatically, so this line ensures they get imported.

daal4py/sklearn/test/test_common.py
david-cortes-intel: This was testing the classes under daal4py.sklearn. Those do not go through the same data validators as sklearnex. For sklearnex, even if there aren't specific tests with different memory layouts, those sorts of situations are still executed in conformance tests.

sklearnex/dispatcher.py
```diff
@@ -530,19 +576,45 @@ def unpatch_sklearn(
...
+    current = getattr(which, what, None)
+    if current is None:
+        return False
+    return current == replacer
+
+
+def check_patch_is_enabled(
+    name: Optional[str] = None, return_map: bool = False, *, map: PatchMap
+) -> Union[bool, dict[str, bool]]:
+    if name is not None:
```
ethanglaser, line 605: Would name ever be None? Could be an opportunity for further simplification

david-cortes-intel: Yes, it can be None when called from this line:
https://github.com/david-cortes-intel/scikit-learn-intelex/blob/bea0f8bf07bbe8c3e588f578636c226a9658f6c8/sklearnex/dispatcher.py#L659

ethanglaser (approved):

PR #3372, opened 2026-08-17
Add SKLEARNEX_NO_ABS_RPATH flag for conda build

ethanglaser:
## Description

Published PyPI wheels carry absolute build-machine paths in their `.so` RPATHs. From the current
nightly of `main`:

```
_daal4py...so   RPATH  [<ws>/<hash>/.pixi/envs/build/lib:<ws>/build/.../daal/latest/lib/intel64:$ORIGIN/../../../]
```

Harmless at runtime — `$ORIGIN/../../../` is what resolves oneDAL, from the `daal` wheel — but it leaks
internal build layout into a public artifact.

The `daal/latest/lib/intel64` entry originates here. When `$DALROOT` points outside the active conda
environment, `conda-recipe/build.sh` passes `--abs-rpath`, which makes `setup.py:326` prepend
`$DALROOT/lib/intel64` to the setuptools-built extensions and `scripts/CMakeLists.txt:57` prepend
`${oneDAL_LIBRARY_DIR}` to the CMake-built ones (plus `$MKLROOT/lib` for the two DPC++ ones). That is the
right default for a local build against an out-of-env oneDAL, and wrong only when the artifacts get
shipped to another machine. `SKLEARNEX_NO_ABS_RPATH=1` opts out. Default behavior is unchanged, and
`conda-build` / the conda-forge feedstock never took this branch anyway (`DALROOT` is unset there).

**Scope:** this clears the `$DALROOT`- and `$MKLROOT`-derived entries — 35 of the 43 absolute entries in
the current nightly, and all three CMake-built extensions completely. The other 8 are the first entry in
the example above: conda-forge python reports `-Wl,-rpath,<env>/lib` in `LDSHARED` and setuptools links
with it verbatim, so it reaches the setuptools-built extensions no matter what the recipe passes. No
recipe-level flag can suppress that one, and it is out of scope here.

**Heads-up:** `setup.py` adds the `-s` link flag only when `--abs-rpath` is absent, so setting this also
strips `_daal4py.so` (9,957,160 → 7,433,280 bytes) and `mpi_transceiver.so` (29,504 → 22,904). That
matches what the `conda-build` path has always produced; decoupling the two in `setup.py` would be a
separate change if unwanted.

**Verified** in an internal CI build of this branch. The flag observably takes effect — the `-s` coupling
above is visible in the binaries — and Linux tests, examples and conformance are green on Python
3.10–3.14. The 35-of-43 split is derived from the code paths above measured against a nightly baseline;
that CI run also had a post-link RPATH rewrite active, so it does not isolate this change's effect on its
own.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/conda-recipe/build.sh
+++ b/conda-recipe/build.sh
@@ -28,7 +28,11 @@ if [ -z "${DALROOT}" ]; then
 elif [ "${DALROOT}" != "${CONDA_PREFIX}" ] && [ ! -z "${CONDA_PREFIX}" ]; then
     # source oneDAL if DALROOT is set outside of conda-build
     source ${DALROOT}/env/vars.sh
-    export DALRPATH=--abs-rpath
+    # builds meant for redistribution set this: $DALROOT is a build-machine path there,
+    # which must not end up in the RPATH of the objects that get published
+    if [ -z "${SKLEARNEX_NO_ABS_RPATH}" ]; then
+        export DALRPATH=--abs-rpath
+    fi
 fi
 
 if [ -z "${MPIROOT}" ] && [ -z "${NO_DIST}" ]; then
--- a/conda-recipe/build.sh
+++ b/conda-recipe/build.sh
@@ -28,8 +28,6 @@ if [ -z "${DALROOT}" ]; then
 elif [ "${DALROOT}" != "${CONDA_PREFIX}" ] && [ ! -z "${CONDA_PREFIX}" ]; then
     # source oneDAL if DALROOT is set outside of conda-build
     source ${DALROOT}/env/vars.sh
-    # builds meant for redistribution set this: $DALROOT is a build-machine path there,
-    # which must not end up in the RPATH of the objects that get published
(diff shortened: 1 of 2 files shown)


conda-recipe/build.sh
```diff
@@ -28,7 +28,11 @@ if [ -z "${DALROOT}" ]; then
 elif [ "${DALROOT}" != "${CONDA_PREFIX}" ] && [ ! -z "${CONDA_PREFIX}" ]; then
     # source oneDAL if DALROOT is set outside of conda-build
     source ${DALROOT}/env/vars.sh
-    export DALRPATH=--abs-rpath
+    # builds meant for redistribution set this: $DALROOT is a build-machine path there,
+    # which must not end up in the RPATH of the objects that get published
```
david-cortes-intel: I think this comment is redundant.

david-cortes-intel (approved):

PR #3374, opened 2026-08-20
RPATH cleansing follow-up

ethanglaser:
## Description

Follow-up to #3372 (DAALL-9602). That PR wired `SKLEARNEX_NO_ABS_RPATH` only into `conda-recipe/build.sh`, which left build-machine paths in the shipped `.so` rpaths. Measured against the `main` nightly, which carries 43 absolute entries across the five Python packages:

**1. Opt-out moved into `setup.py`** (`NO_ABS_RPATH`, beside the other env toggles), so it applies to `bld.bat`, `conda-build`, and direct `setup.py` / `bdist_wheel` calls rather than only to `build.sh`. `build.sh` now always exports `--abs-rpath` and `setup.py` cancels it, so the opt-out cannot be bypassed by using a different entry point. Clears the 35 `$DALROOT` / `$MKLROOT` entries — every CMake-built extension goes to zero.

**2. `setup.py:324` no longer adds `-Wl,-rpath,$MPIROOT/lib`** when opting out (5 entries, `mpi_transceiver` on every leg). `build.sh` sets `MPIROOT=$PREFIX`, so this was the build environment's own path; the MPI runtime is a dependency of the built package, so `$ORIGIN/../../../` resolves it the same way oneDAL is.

**Verification.** Private CI: `main` = 43 absolute entries; the opt-out flag alone (#3372) = 8; with this PR = 3. The 8 residuals were 5 + 3 exactly — the 5 MPI entries above, plus `-Wl,-rpath,<prefix>/lib`, which conda interpreters carry in `LDCXXSHARED` (Python >= 3.12) and which conda's compiler activation script also puts in `$LDFLAGS`. Those 3 are not reachable from any build argument, so they are removed by a post-build `patchelf` pass in the internal build pipeline instead: a job running that pass reported **0** absolute entries across all 25 `.so` files, with `$ORIGIN/../../../` intact and the tag still `RPATH` rather than `RUNPATH`. Both jobs are green on Python 3.10-3.14, including the MPI, conformance and examples legs, so resolving the MPI runtime through `$ORIGIN` instead of `$MPIROOT` holds at runtime.

## Limitations

- Opt-in only — gated on the variable rather than on `not USE_ABS_RPATH`, so builds that pass no `--abs-rpath` (conda-forge, GitHub CI) are unaffected.
- Linux only.
- Does not touch the rpath entries contributed by conda's own link commands (`LDSHARED` / `LDCXXSHARED`, `$LDFLAGS`). Those are managed by distutils and cannot be filtered completely from `setup.py`, so they are stripped from the built objects post-build.

## Possible extensions

- Decouple `-s` from `USE_ABS_RPATH` (`setup.py:290`) to keep symbols in `_daal4py` / `mpi_transceiver`.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>


setup.py
```diff
@@ -389,6 +392,20 @@ def getpyexts():
...
+    # LDCXXSHARED is the one that matters here (the extensions are all language="c++"),
+    # but both are filtered since either can be the effective linker.
+    for holder in (os.environ, cfg_vars):
+        for var in ("LDSHARED", "LDCXXSHARED"):
+            if var in holder:
+                holder[var] = " ".join(
+                    flag
+                    for flag in holder[var].split()
+                    if not flag.startswith("-Wl,-rpath,")
+                )
```
david-cortes-intel: For me, if used in a conda environment, this adds an absolute path with the name of the environment on it.

I think it shouldn't be needed either, because in an environment with dependencies installed with pip or conda, libmpi.so will be in the same folder as libtbb.so.

david-cortes-intel: Actually I see this is removing it. But since these are managed by distutils, shouldn't they be left as-is?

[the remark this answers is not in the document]
david-cortes-intel: Bot here is right.

david-cortes-intel (approved):
Issues with conda tooling to be solved by other means in affected distributions.

PR #3380, opened 2026-08-24
FIX: Fix `__version__` attribute in daal4py and onedal modules

david-cortes-intel:
## Description

This attribute is not present in daal4py and is hard-coded in the onedal module, but it can be queried at runtime instead of being set at install time.

ref https://github.com/uxlfoundation/scikit-learn-intelex/issues/3377

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/daal4py/__init__.py
+++ b/daal4py/__init__.py
@@ -68,6 +68,8 @@ except ImportError as e:
 
     raise
 
+__version__ = ".".join([str(v) for v in eval(_get__version__())[:3]])
+
 from . import mb, sklearn
 
-__all__ = ["mb", "sklearn"]
+__all__ = ["__version__", "mb", "sklearn"]
--- a/setup.py
+++ b/setup.py
@@ -439,7 +439,7 @@ class onedal_build:
         self.onedal_post_build()
         if hasattr(self, "build_lib"):
             # swap out __version__ before install
-            for p in ["onedal", "sklearnex"]:
+            for p in ["sklearnex"]:
                 loc = os.sep.join((self.build_lib, p, "__init__.py"))
                 if os.path.isfile(loc):
                     with open(loc, "r+") as f:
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -439,7 +439,7 @@ def run(self):
         self.onedal_post_build()
         if hasattr(self, "build_lib"):
             # swap out __version__ before install
-            for p in ["onedal", "sklearnex"]:
+            for p in ["sklearnex"]:
```
ethanglaser: How is onedal behavior different than sklearnex?

david-cortes-intel: OneDAL is a separate library and the onedal module is a wrapper over it. It's possible to use different versions of oneDAL and sklearnex together.

icfaust: For loop not necessary.

david-cortes-intel: Removed.

icfaust (commented):
I'd like to see a change in the docs notifying users that the `__version__` does not reflect anything about the nature of the onedal python module, but instead the linked oneDAL shared objects. This is a subtle thing that we should keep clear in the case of debugging.

daal4py/__init__.py
```diff
@@ -68,6 +68,8 @@
 
     raise
 
+__version__ = ".".join([str(v) for v in eval(_get__version__())[:3]])
```
avolkov-intel, line 71: For me it's not clear why daal4py.__version__ should follow the oneDAL version of the C++ library and not the sklearnex version. 

For me daal_version, dal_version, onedal_version should be the versions of the linked C++ library

sklearnex_version, daal4py_version should be the versions of the python library

Not sure what @icfaust @ethanglaser think about it

icfaust: Because daal4py does a rather un-recommended ```from _daal4py import *```, its not so clear. For ```daal4py.sklearn``` or model_builders I definitely see the sklearnex version being important, but for ```daal4py._daal4py``` it should be the oneDAL version.  I'm fine with either, it should be extraordinarily rare for someone to use a different oneDAL version than sklearnex version, so at the end this all is a bit of a wash.

icfaust (approved):

PR #3400, opened 2026-09-01
resolve kmeans spmd example issues:

ethanglaser:
## Description

Addresses infrequent but persistent error in KMeans spmd examples where not all clusters are represented on a given rank by warning for non-spmd but not on spmd as it's a realistic scenario that shouldn't be a cause for concern.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/sklearnex/spmd/cluster/kmeans.py
+++ b/sklearnex/spmd/cluster/kmeans.py
@@ -20,6 +20,11 @@ from ...cluster import KMeans as base_KMeans
 
 
 class KMeans(base_KMeans):
+    def _warn_on_degenerate_clustering(self):
+        # labels_ only covers this rank's shard, so a cluster missing from it means no
+        # point of that cluster landed here -- not that the global clustering degenerated.
+        pass
+
     def _initialize_onedal_estimator(self):
         """Override to use SPMD backend instead of batch backend."""
         onedal_params = {
--- a/sklearnex/cluster/k_means.py
+++ b/sklearnex/cluster/k_means.py
@@ -476,6 +476,10 @@ if daal_check_version((2023, "P", 200)):
             self.n_iter_ = self._onedal_estimator.n_iter_
             self.n_features_in_ = self._onedal_estimator.n_features_in_
 
+            self._warn_on_degenerate_clustering()
+
+        def _warn_on_degenerate_clustering(self):
+            """Warn when fewer than n_clusters clusters received any point."""
             xp, _ = get_namespace(self.labels_)
             unique_func = getattr(xp, "unique_values", np.unique)
             distinct_clusters = unique_func(self.labels_).shape[0]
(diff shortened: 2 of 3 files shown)


sklearnex/spmd/cluster/kmeans.py
```diff
@@ -20,6 +20,11 @@
 
 
 class KMeans(base_KMeans):
+    def _warn_on_degenerate_clustering(self):
+        # labels_ only covers this rank's shard, so a cluster missing from it means no
+        # point of that cluster landed here -- not that the global clustering degenerated.
+        pass
```
david-cortes-intel, line 26: When would this warning be thrown on SPMD?

ethanglaser: It would not

ethanglaser: Because the data is spread across the ranks, there isn't an expectation that every cluster must be represented on every rank

david-cortes-intel: No, but if there's some cluster than doesn't have any points in any rank, then it should throw a warning too.

ethanglaser: I'm not sure I agree that a warning should occur there. Anyone else have opinions?

But if so, we should close this PR, related tickets, and accept occasional CI failures due to the Werror.

david-cortes-intel: Wouldn't the error in the CIs be thrown when any rank has an empty cluster instead of being checked globally across ranks?

And if the latter, wouldn't it be fixed by this PR?
https://github.com/uxlfoundation/oneDAL/pull/3742

ethanglaser: The warning is being raised is:

`ConvergenceWarning: Number of distinct clusters (9) found smaller than n_clusters (10). Possibly due to duplicate points in X.`

This happens occasionally when a certain rank does not have a cluster represented on it, even if the desired number of clusters is present across all ranks collectively. The warning message is misleading and in general I don't think this situation necessitates a warning at all.

david-cortes-intel: It shouldn't throw a warning in that situation, but it should throw a warning when some cluster is globally empty across all ranks.

ethanglaser: It should, but I think enabling this would be significantly more hassle than beneficial. Communication between ranks is only happening in oneDAL, so sklearnex alone could not determine this. Additionally oneDAL already has functionality to address via `fill_empty_clusters`. I think this is in a mergeable state.

david-cortes-intel: Then I guess it shouldn't be a problem to issue the warning later on if a flag about it gets added on the oneDAL side, since that info will be available at fit time:
https://github.com/uxlfoundation/oneDAL/issues/3770

david-cortes-intel (approved):

PR #3403, opened 2026-09-01
FIX: Fix memory leaks and memory safety issues in base1 CSR array conversions

david-cortes-intel:
## Description

Follow up from https://github.com/uxlfoundation/scikit-learn-intelex/pull/3368

Fixes:
* A memory leak caused due to incorrect transfer of ownership of base0 arrays that are created from base1 data.
* A potential use-after-free by getting data pointers out of oneDAL tables to put into numpy arrays where there's no guarantee that the table will outlive the array.
* Accessing signed int types as unsigned and handling arrays as unsigned types when they are meant to be signed.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/onedal/datatypes/numpy/data_conversion.cpp
+++ b/onedal/datatypes/numpy/data_conversion.cpp
@@ -442,6 +442,10 @@ static PyObject *convert_to_numpy_impl(
 
     void *opaque_value = static_cast<void *>(new dal::array<T>(host_array));
     PyObject *cap = PyCapsule_New(opaque_value, NULL, free_capsule<T>);
+    if (!cap) {
+        Py_DECREF(obj);
+        throw std::runtime_error("Python capsule creation failed");
+    }
     PyArray_SetBaseObject(reinterpret_cast<PyArrayObject *>(obj), cap);
     return obj;
 }
@@ -535,18 +539,30 @@ static PyObject *convert_to_py_from_csr_impl(const csr_table &table) {
     const T *data = table.get_data<T>();
     auto data_array = dal::array<T>::wrap(data, non_zero_count);
 
+    // These put the Python objects into temporary pybind11 containers
+    // in order to ensure that if any allocation fails or the conversion
+    // throws, the intermediate objects would get destructed afterwards.
     PyObject *py_data = convert_to_numpy_impl<NpType, T>(data_array, non_zero_count);
+    py::object py_data_holder = py::reinterpret_steal<py::object>(py_data);
 
     PyObject *py_col =
         convert_to_numpy_impl<NPY_UINT64, std::uint64_t>(column_indices_zero_based_array,
                                                          non_zero_count);
+    py::object py_col_holder = py::reinterpret_steal<py::object>(py_col);
+
     PyObject *py_row =
         convert_to_numpy_impl<NPY_UINT64, std::uint64_t>(row_offsets_zero_based_array,
                                                          rows_indices_count);
+    py::object py_row_holder = py::reinterpret_steal<py::object>(py_row);
     PyObject *result = PyTuple_New(3);
+    if (!result)
+        throw std::bad_alloc();
     PyTuple_SetItem(result, 0, py_data);
+    py_data_holder.release();
     PyTuple_SetItem(result, 1, py_col);
+    py_col_holder.release();
     PyTuple_SetItem(result, 2, py_row);
+    py_row_holder.release();
     return result;
 }
 
--- a/onedal/datatypes/numpy/data_conversion.cpp
+++ b/onedal/datatypes/numpy/data_conversion.cpp
@@ -24,6 +24,9 @@
 
 #include "oneapi/dal/table/homogen.hpp"
 #include "oneapi/dal/table/detail/homogen_utils.hpp"
+#if ONEDAL_VERSION >= 20260200
+#include "oneapi/dal/table/detail/csr_utils.hpp"
+#endif
 
 #include "onedal/datatypes/numpy/data_conversion.hpp"
 #include "onedal/datatypes/numpy/numpy_utils.hpp"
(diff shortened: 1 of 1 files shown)


onedal/datatypes/numpy/data_conversion.cpp
ethanglaser, line 1: The oneDAL version <= 2023.1 still contains the leak, should we just remove it entirely? Might be worth a discussion on what oneDAL level of support should be maintained in sklearnex

david-cortes-intel: Technically we support oneDAL>=2021.4:
https://uxlfoundation.github.io/scikit-learn-intelex/2026.1/building-from-source.html

But I don't know if the same functions/methods would be available in that version.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -505,17 +504,26 @@ static PyObject *convert_to_py_from_csr_impl(const csr_table &table) {
...
     if (table.get_indexing() == sparse_indexing::zero_based) {
         column_indices_zero_based_data =
             const_cast<std::uint64_t *>(reinterpret_cast<const std::uint64_t *>(column_indices));
         row_offsets_zero_based_data =
             const_cast<std::uint64_t *>(reinterpret_cast<const std::uint64_t *>(row_offsets));
+
+        column_indices_zero_based_array =
+            dal::array<std::uint64_t>::wrap(column_indices_zero_based_data, non_zero_count);
+        row_offsets_zero_based_array =
+            dal::array<std::uint64_t>::wrap(row_offsets_zero_based_data, rows_indices_count);
```
ethanglaser: The zero-based branch has the opposite problem and it's still here.

The `const_cast` gives `wrap` a mutable pointer, so `wrap` attaches `empty_delete`, `need_mutable_data()` inside `convert_to_numpy_impl` no-ops, and `PyArray_New` hands numpy the table's own buffer. The capsule then holds a non-owning array, so nothing keeps the table alive: dangling read if the table dies before the numpy array (the pybind `result` local in `onedal/svm/svm.py` goes out of scope while `support_vectors_` lives on), and numpy can write straight into oneDAL memory. `data_array` below is safe only because its pointer stays const, which forces `need_mutable_data()` to copy.

Dropping the `const_cast` gets the indices the same treatment, and lets both `*_data` locals move into the one-based branch:

```cpp
    dal::array<std::uint64_t> column_indices_zero_based_array;
    dal::array<std::uint64_t> row_offsets_zero_based_array;

    if (table.get_indexing() == sparse_indexing::zero_based) {
        // Wrapped as const so need_mutable_data() copies rather than aliasing the table.
        column_indices_zero_based_array = dal::array<std::uint64_t>::wrap(
            reinterpret_cast<const std::uint64_t *>(column_indices), non_zero_count);
        row_offsets_zero_based_array = dal::array<std::uint64_t>::wrap(
            reinterpret_cast<const std::uint64_t *>(row_offsets), rows_indices_count);
    }
    else { // sparse_indexing::one_based
        column_indices_zero_based_array = dal::array<std::uint64_t>::empty(non_zero_count);
        row_offsets_zero_based_array = dal::array<std::uint64_t>::empty(rows_indices_count);

        std::uint64_t *column_indices_zero_based_data =
            column_indices_zero_based_array.get_mutable_data();
        std::uint64_t *row_offsets_zero_based_data =
            row_offsets_zero_based_array.get_mutable_data();

        // ...rebasing loops unchanged...
    }
```

That costs one copy of the indices. `csr_accessor::pull(..., sparse_indexing::zero_based)` would avoid it and collapse both branches, but that's more than this PR needs. I also couldn't find a path where oneDAL actually returns zero-based indices here, so this is latent rather than something I've seen crash.

Separately: the `ONEDAL_VERSION <= 20230100` copy of this function above still has the leak you're fixing here.

onedal/datatypes/numpy/data_conversion.cpp
```diff
@@ -528,19 +536,16 @@ static PyObject *convert_to_py_from_csr_impl(const csr_table &table) {
...
         convert_to_numpy_impl<NPY_UINT64, std::uint64_t>(column_indices_zero_based_array,
                                                          non_zero_count);
-    PyTuple_SetItem(result, 1, py_col);
-    auto row_indices_zero_based_array =
-        dal::array<std::uint64_t>::wrap(row_offsets_zero_based_data, rows_indices_count);
     PyObject *py_row =
-        convert_to_numpy_impl<NPY_UINT64, std::uint64_t>(row_indices_zero_based_array,
+        convert_to_numpy_impl<NPY_UINT64, std::uint64_t>(row_offsets_zero_based_array,
                                                          rows_indices_count);
+    PyObject *result = PyTuple_New(3);
```
ethanglaser: Good that the tuple no longer leaks when a conversion throws. One leftover: `convert_to_numpy_impl` can throw (`invalid_argument` if `PyArray_New` fails, `bad_alloc` from `need_mutable_data`), and if `py_col` or `py_row` throws then `py_data` is a leaked reference, since it never reached the tuple. `PyTuple_New` can also return NULL and `PyTuple_SetItem` won't check.

Only reachable under allocation failure, so up to you. pybind11 is already in scope via `data_conversion.hpp`:

```cpp
    py::object py_data = py::reinterpret_steal<py::object>(
        convert_to_numpy_impl<NpType, T>(data_array, non_zero_count));
    // py_col, py_row likewise
    PyObject *result = PyTuple_New(3);
    if (!result)
        throw std::bad_alloc();
    PyTuple_SetItem(result, 0, py_data.release().ptr());   // steals the reference
```

ethanglaser (commented):
(potential improvements via Claude) The leak is real: `wrap` attaches `empty_delete`, so nothing ever freed those two `host_allocator` blocks, and sparse SVM hits it on every fit. Two notes, both about the rest of the function rather than your change.

onedal/datatypes/numpy/data_conversion.cpp
napetrov: Confirmed this experimentally, and there is one extra data point on the "latent" question. ASan on the PR head gives `heap-use-after-free`, READ of size 8, with the free traced through `csr_table_impl::~csr_table_impl()` → `array_impl<long>::~array_impl()` → `host_allocator<long>::deallocate`, and the write direction reproduces too: `indices[0] = 999` from Python rewrites the live table's `column_indices[0]`. Dropping the `const_cast` fixes both — the pointer-identity probe goes from aliased-and-writeable to copied-and-read-only, with output values unchanged.

On reachability: oneDAL appears to *normalise* the output, so the branch may be unreachable rather than merely unexercised. Training a real `dal::svm` on a `zero_based` input table still returns `support_vectors` with `one_based` indexing, so combined with `convert_to_csr_impl` always constructing one-based tables, there is no current path that reaches it. That supports reading the const wrap as cheap insurance rather than a fix for a live bug.

<details>
<summary>probe output</summary>

Real `dal::svm` train (thunder, linear kernel), same problem with both input indexings:

```
########## input base-1 ##########
input table indexing        : one_based
support_vectors kind == csr : yes
support_vectors indexing    : one_based
coeffs          kind == csr : no

########## input base-0 ##########
input table indexing        : zero_based
support_vectors kind == csr : yes
support_vectors indexing    : one_based
coeffs          kind == csr : no
```

Aliasing probe on a zero-based table, PR head vs PR + const wrap:

| | PR head | + const wrap |
|---|---|---|
| `indices` aliases table buffer | yes | no |
| `indptr` aliases table buffer | yes | no |
| `indices.flags.writeable` | True | False |

`data` is unaffected in both, since its pointer stays const and `need_mutable_data()` therefore copies — matching your reading.
</details>

One footnote on the lifetime argument: `result.coeffs` comes back as a *dense* table (`coeffs kind == csr : no` above), so `from_table(result.coeffs)` at `onedal/svm/svm.py:112` goes through the homogen path. A sparse fit puts one CSR table through this function — `support_vectors` — not two.

onedal/datatypes/numpy/data_conversion.cpp
napetrov: Measured this one, and the item leak is a bit worse than "leftover": moving `PyTuple_New` to the end does not reduce it at all. I added counters to `free_capsule` and next to `PyCapsule_New`, so `created - freed` is exactly the number of numpy arrays dropped on the floor, then forced the Nth `convert_to_numpy_impl` call to throw, over 20 attempts:

| variant | throw while building `py_col` | throw while building `py_row` |
|---|---|---|
| `main` (tuple first) | 20 leaked arrays | 40 |
| this PR (tuple last) | 20 | 40 |
| `reinterpret_steal` form | 0 | 0 |

So the tuple move stops the empty tuple itself leaking, but every already-built array still leaks, and each one carries an owning `dal::array` capsule — under a throw in `py_row` on a one-based table that is two index copies plus the data array per attempt, not just three `PyObject` headers. Agree it is allocation-failure-only, but the `reinterpret_steal` form takes it to zero for about three lines.

onedal/datatypes/numpy/data_conversion.cpp
icfaust: @ethanglaser I agree on the 2nd part. The use of onedal version checks is to primarily support the azure pipelines CI system (no user/ customer case). We can do a clean up similar to sklearn version checks up until some recent version (and update the docs based on @david-cortes-intel 's recommendation)

icfaust (commented):
I'm going to leave this up to the reviewers. Do we have any info if this impacts uxlfoundation/oneDAL#3761 's segfault?

onedal/datatypes/numpy/data_conversion.cpp
david-cortes-intel: This other issue needs a solution from the oneDAL side. PR has not been merged yet:
https://github.com/uxlfoundation/oneDAL/pull/3775

onedal/datatypes/numpy/data_conversion.cpp
david-cortes-intel: Fixed.

onedal/datatypes/numpy/data_conversion.cpp
david-cortes-intel: Modified now to use the new utilities.

Vika-F (approved):
Thank you for the prompt update. Let's wait for the [nightly] CI results and LGTM.

PR #3416, opened 2026-09-03
FIX: Ensure sparse support vectors retain full shape

david-cortes-intel:
## Description

When assigning sparse support vectors in CSR format in SVMs, they are constructed from CSR arrays, but if there are empty columns, the resulting shape might be different than what SciPy would deduce from those arrays.

This PR sets the shape explicitly in order to avoid discarding empty columns.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/onedal/svm/svm.py
+++ b/onedal/svm/svm.py
@@ -17,6 +17,7 @@
 from abc import ABC, abstractmethod
 from math import sqrt
 
+import numpy as np
 from scipy import sparse as sp
 
 from daal4py.sklearn._utils import is_sparse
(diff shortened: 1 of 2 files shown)


onedal/svm/svm.py
```diff
@@ -109,8 +110,23 @@ def fit(self, X, y, sample_weight=None, class_count=0, queue=None):
         result = self.train(params, *data_t)
 
         if self._sparse:
+            # Note: the coefficients from oneDAL come as dense, but
+            # scikit-learn stores them as sparse, hence the conversion
```
icfaust, line 114: This looks like this is sklearn conformance. The sparse logic should then be moved to the sklearnex estimators at the copy over.

david-cortes-intel: In this case, `from_table` would return an empty tuple, and if that gets passed to the CSR constructor together with shape arguments, it would produce an error.

david-cortes-intel: If you mean about the coefficients, then yes, that part could be handled in sklearnex, but that's unrelated to this PR.

